PR #3, opened 2019-08-22
History is persisted in local storage

AndrewBastin:
Pretty self explanatory.

History is maintained across multiple sessions using the Local Storage API


liyasthomas:
@all-contributors please add AndrewBastin for code

PR #6, opened 2019-08-22
History is sorted from latest to oldest

AndrewBastin:
A small UX stuff, usually we need the latest history entries on top. So we can avoid the high amounts of scroll


liyasthomas (approved):
You always amaze me 👌

Issue #19, opened 2019-08-22
PWA not installable


NileshSP:
**Describe the bug**
"Install PWA" button doesn't work 

**To Reproduce**
2. Click on 'Install PWA' button
4. Effect - Not installed

**Expected behavior**
Doesn't provide "Install App?" dialog to proceed

**Screenshots**
Lighthouse suggestions through 'Audit' tab in devtools for troubleshooting
![image](https://user-images.githubusercontent.com/13110493/63553172-a67a8100-c507-11e9-9456-694b581b087c.png)

**Additional context for resolution**
Try updating the "start_url" in manifest.json to 
"start_url": "https://liyasthomas.github.io/postwoman" or "." instead

liyasthomas:
#22 fixed this.
Tested and approved.

PR #25, opened 2019-08-23
add button to install PWA

steviebeenz:
Makes a banner, native and designed by your browser display, which let's the user install the PWA. This allows the user to install the PWA easier, and also prompts them to.


liyasthomas:
I afraid this feature has already been added fron PR #19

PR #27, opened 2019-08-23
Auto Update Dependencies 

steviebeenz:
Never have an out of date dependency.
Auto updates dependencies into PRs


larouxn:
Sheer curiosity, why use this over something like [Dependabot](https://dependabot.com)? Both are free and update dependencies via pull requests but Dependabot is, as far as I know, a little more well established and actually owned by GitHub these days.

liyasthomas:
+1 for Depandabot

liyasthomas:
#31 will will possibly come with a fix for this issue too.

PR #28, opened 2019-08-23
Added token based authentication type

reefqi037:
User can choose Bearer Token authentication type to access apis protected with access/id token.


reefqi037:
Yes. I've tested it on my own AWS Api Gateway endpoint configured with Cognito Authorizer.

Bearer Token test

liyasthomas:
Thanks for the quick reply :purple_heart:

Issue #30, opened 2019-08-23
Add copy response to clipboard button


LeoVie:
Add a button with a caption like "Copy to clipboard" or some fancy icon above the response text field to directly copy the response to the users clipboard.

liyasthomas:
#64 fixes this issue and hence closing

PR #34, opened 2019-08-23
`Copy Response` Button 

athul:
This solves #30 And Thank You for this. :heart_decoration: From fellow mallu

### DIFF
--- a/index.html
+++ b/index.html
@@ -224,7 +224,8 @@
         <ul>
           <li>
             <label for="body">response</label>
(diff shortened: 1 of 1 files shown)


AndrewBastin:
> This solves #30 And Thank You for this. heart_decoration From fellow mallu

mallus ftw :rofl:

athul:
Do merge the best one.Since that PR will benefit a whole lot more people

athul:
After Checking I personally Think #35 is far better than mine. Thus closing it

PR #35, opened 2019-08-23
Added copy to clipboard button

edisonaugusthy:


liyasthomas:
#34 and #35 seems to be possible duplicates.

liyasthomas:
This would solve #30

Issue #29, opened 2019-08-23
Ability to store/share/create collections


curtchan:
I like the idea of your project, but first thing that i noticed that it lacks comparing to postman - shareable/saveable collections

It would require:
* ability to group requests into collections (naming requests, naming collections, ordering them)
* ability to save collection/read collection (json export/import?, storing in browser storage?)

keithl8041:
Being able to share collection is the killer feature for Postman and is why they can charge a premium for their team membership. Being able to export/import is a step down from that but still gets you most of the same functionality, without the advantages of being able to sync automatically between team members. 

So I'd second the import/export functionality. JSON seems an obvious choice but need to be careful around how it gets serialised to make sure it can work nicely with git (ie: if two or more people edit a file, can git patch it together nicely). 

I'd like to help out with this if we can agree a good approach. I think the logical separation of creating the collections and the import/export makes sense. How best to proceed?

liyasthomas:
Thumbs up! This feature would make Postwoman top of its class. Looking forward on collaborating with you.

keithl8041:
The sync operation is possibly covered by #26 where using Firebase to sync data is mentioned, so potentially this ticket should just be about creating the collection organisation and management concept.

andrewyoung918:
I like the idea of a work space tab filing the need of collections.

I think being able to import / export openAPI docs as a tab, and persisting tabs in browser storage would be great, not require auth or db server.

An alternative to having cross team synced connections would be a "Share tab on slack" that posted a link that includes a serialized export.

Would make sharing super straight forward. Would serve many teams needs better than postman. It won't solve for late Enterprise, but I don't think this tool needs to.

liyasthomas:
This approach could be implemented straight away. Can you take the initiative and consider making a PR. All supporters will have your back.

jgroom33:
Related to #5 
I'd like to see yaml used for the format.
Perhaps using fetch terminology would be good: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#Supplying_request_options
Another option is to use har (har is a format provided by browser exports)

jgroom33:
Example:
```yaml
name: nocall
  var:
    foo: 1
name: call1
  request:
    url: http://foo.bar
    method: POST
    body:
      foo: {{ nocall.foo }}
name: call2
  request:
    url: http://foo.bar/{{ call1.response.id }}
    method: GET
  until: {{ call2.response.status }} == 200
[...]

PR #42, opened 2019-08-23
Add Share Buttons 

steviebeenz:


NBTX:
I think we can close this and update #29 and #72 to note that we could use this as a mechanism for users sharing their 'shareable' requests. Especially seeing as this PR is from even before the Nuxt migration.

liyasthomas:
Closing absolute PRs

Issue #46, opened 2019-08-24
Colored response codes based on status code


liyasthomas:
**Is your feature request related to a problem? Please describe.**
It would've been great to analyse request responses if they are colored based on their status codes. Ex: green for 200, yellow for 301, red for 404 etc.

**Describe the solution you'd like**
Show colored responses based on its status code to easily identify the response status. Add a class name based on status code to color the status code. Preserve the same for history section.

**Describe alternatives you've considered**
The only way to identify status codes are manually checking them. Colored codes would help it 100x times.

**Additional context**
History section should also follow these colored responses for future reference. Add a string for history array to contain color class.

liyasthomas:
Closing the issue since #52  solved it

PR #52, opened 2019-08-24
Response status colors added.

izerozlu:
Issue #46 's request is fullfilled by adding a computed property to check after `response` changes to determine which category is the `response`.

Categories are implemented from [Http Status Code](https://www.tutorialrepublic.com/html-reference/http-status-codes.php).

### DIFF
--- a/assets/css/styles.scss
+++ b/assets/css/styles.scss
@@ -1,7 +1,7 @@
 :root {
   --bg-color: #121212;
-  --fg-color: #FFF;
-  --ac-color: #51FF0D;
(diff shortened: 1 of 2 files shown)


NBTX:
I like that you used regexes for each of the status code groups!

liyasthomas:
Looks cool, exactly it should've been. 🔥 Would you mind creating a PR to bind the colored class to history entries status codes too..? That would be great!

izerozlu:
Sure. Will create a PR for entries as well soon.

PR #53, opened 2019-08-24
History fieldset minimized.

izerozlu:
History fieldset is minimized by setting a `max-height` style to `fieldset.history .collapsible` to minimize the application view.

This minimization doesn't occur until five `entry` items are added to history.

Also, button to clear history has been moved to the parent to keep its position upon history scroll.

### DIFF
--- a/assets/css/styles.scss
+++ b/assets/css/styles.scss
@@ -130,6 +130,10 @@ fieldset.history legend {
   color: #9B9B9B;
 }
 
+fieldset.history .collapsible {
+  overflow: auto;
+}
+
 fieldset.authentication {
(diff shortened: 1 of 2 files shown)


izerozlu:
#53 and #54 are direct conflicts and two seperate implementations for the history fieldset. #53 doesn't include a component dependency so a bit more lighter than #54. But #54 brings in a performance boost for the big history lists. If both of PRs evaluate to be merge, one of them has to be chosen.

liyasthomas:
Consider rebase and pull again

izerozlu:
Will create a virtual scroll version of this PR. Thus closing.

PR #54, opened 2019-08-24
Virtualscroll added for history fieldset.

izerozlu:
Virtual scrolling added for the `history` fieldset. 

History fieldset doesn't get into scrolling mode until five `entry` items are inserted into `history`.

Also `Clear History` button has moved to the parent.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,7 @@
   "dependencies": {
     "@nuxtjs/pwa": "^3.0.0-0",
     "nuxt": "^2.0.0",
+    "vue-virtual-scroll-list": "^1.4.2",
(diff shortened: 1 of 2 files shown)


izerozlu:
#53 and #54 are direct conflicts and two seperate implementations for the history fieldset. #53 doesn't include a component dependency so a bit more lighter than #54. But #54 brings in a performance boost for the big history lists. If both of PRs evaluate to be merge, one of them has to be chosen.

liyasthomas:
Consider rebase and pull again

izerozlu:
Will create another PR for this subject, thus closing.

PR #63, opened 2019-08-25
Feature/raw input

nickpalenchar:


nickpalenchar:
Hey @liyasthomas thanks for putting this together.

I saw this issue and thought I'd run with it--The contributing guidelines look like a template to me so feel free to ping me if there's something that needs corrections.

Rather than having a separate content type dropdown, I decided to make a 'raw input' a selection checkbox, which replaces the key/value form with free form text input. Since "raw application/json" isn't a new "Content-Type", but a way to enter data to send, it didn't feel right to make it another dropdown option. Raw input is also a lower level option that applies across application types, i.e. one might also want to send raw data with "www-form/url-encoded" or with custom headers in the future; this change should be able to handle those as well.

Thanks!

nickpalenchar:
Regarding text input for Content-Type: I definitely agree though I think that should be a separate issue (one which I'd be keen to pick up next 😎). Those options were there before my changes (which might not be apparent in this PR's diff) and would be better off testing separately.

I also thought of having a "custom" option in the Content-Type dropdown which would give the user a text field, but the auto-complete version you suggested is a great idea.

PR #72, opened 2019-08-27
Enable shareable requests with parameters in url queries

yubathom:
I think this is maybe related to #29. 

It enable shareable requests using query parameters in the url. Examples:

[/url=https://api.thecatapi.com&path=/v1/images/search](https://yubathom.github.io/postwoman/?url=https://api.thecatapi.com&path=/v1/images/search)

[/method=POST](https://yubathom.github.io/postwoman/?method=POST
)

[/method=PUT&contentType=/?method=PUT&contentType=www-form/urlencoded](https://yubathom.github.io/postwoman/?method=PUT&contentType=www-form/urlencoded)

it can be applied to the other items of `data` of the `pages/index` component. 

Maybe would be simpler to store theese settings (that could be stored in the router) in a collection than a encoding/decodign static files.

### DIFF
--- a/pages/index.vue
+++ b/pages/index.vue
@@ -528,7 +528,15 @@
             this.$refs.previewFrame.setAttribute('data-previewing-url', this.url);
           }
         }
+      },
+      setRouteQueries(queries) {
+        for (const key in queries) {
+          if (this[key]) this[key] = queries[key];
+        }
       }
+    },
+    created() {
+      if (Object.keys(this.$route.query).length) this.setRouteQueries(this.$route.query);
     }
   }
 


liyasthomas:
This is quite the easiest and hackiest way to share requests.
> Maybe would be simpler to store theese settings (that could be stored in the router) in a collection than a encoding/decodign static files.
@NBTX @AndrewBastin

NBTX:
Is the end goal to do a serverless implementation of shareable requests? @liyasthomas

liyasthomas:
Using formats such as JSON and YAML for export/import functionalities have their pros and cons, but if any use cases such as sharable requests or collections could be implemented without using any other parties, then isn't that a great thing? This also make scope for publically sharable URLS for testing and all.

liyasthomas:
That's exactly what I meant. Export/import can only be implemented by JSOM/YAML considering the request body, number of requests and all. I specified the use case as sharable (public) URLs which could be implemented with parameter queries.

NBTX:
Yeah exactly, it'd basically just be an ID referring to the JSON/YAML file somewhere?

We definitely want people to be able to share URLs - even better would be if they could embed it like GitHub Gists, but like you say there's just far too much to cram into the query string.

liyasthomas:
Closing absolute PR

liyasthomas:
I closed this by accident. My bad. Reopened it.

liyasthomas:
Rebase and update this one.

yubathom:
Thanks @liyasthomas. I'm still working and I'll update this.

PR #73, opened 2019-08-28
Adds HTML preview.

### LINKED ISSUE #41: Add html preview to response section
If you send an unexpected request to an api (e.g. POST to https://reqres.in/), you'll get a 404 or something back. In the response section you get a html response.
It would be nice if there would be a preview tab in which you could see this html response rendered.

NBTX:
Resolves #41 - adds an HTML preview toggle-able with the 'Preview HTML' / 'Hide Preview' button in the bottom right of the response area.
When 'Preview HTML' is clicked, the HTML source is covered by a preview of the rendered HTML.

This button only shows when content type is HTML and a response has been submitted.
As always, you can test the changes here: https://nbtx.github.io/postwoman-vue/

image

### DIFF
--- a/assets/css/styles.scss
+++ b/assets/css/styles.scss
@@ -171,7 +171,7 @@ fieldset.orange legend {
   color: #F5A623;
 }
 
-.collapsible.hidden {
+.hidden {
   display: none;
 }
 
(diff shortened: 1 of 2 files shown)


NBTX:
Oop - just found a bug, let me fix it real quick.
If you hit preview, and then load another link it doesn't hide the preview and it actually covers the response of the other link.

NBTX:
Okay, good to go!

PR #75, opened 2019-08-28
fixes #74 fallback for findEntryStatus when match is missing or other…

### LINKED ISSUE #74: App is broken with old history in localStorage
**Describe the bug**
Starting after [`271bd06`](https://github.com/liyasthomas/postwoman/commit/271bd061765f1d2d065aac8cf2b82a36d8e68117) (or perhaps around then), old localStorage from a user agent causes the app to break when navigating to the app.

**To Reproduce**
1. Open the browser and open devTools
2. Go to the tab for storage, click on localstorage
3. Edit the history section to be `[{"time":"2:57:54 PM","method":"POST","url":"https://reqres.in","path":"/api/users"},{"time":"2:56:30 PM","method":"GET","url":"https://reqres.in","path":"/api/users"},{"time":"10:33:28 AM","method":"GET","url":"https://reqres.in","path":"/api/users"}]`
4. Go to the postwoman page to get the app to break

**Expected behavior**
The app should fully load, any old history that can't be deciphered should be displayed with as much information as possible.

 - OS: Mac OS Sierra
 - Browser: Firefox
 - Version 68.0 (64-bit)

nickpalenchar:
…wise invalid

Small change with big implication!

As mentoned in #74, it would be best for both old history (which has no status number) _and_ new history to be preserved. 

`findEntryStatus` which in turn calls `findStatusGroup`, calls the `Array.prototype.find` which will return undefined if the callback function provided does not return true on any item. Since the callback function uses a RegEx.prototype.test on the `statusCategories`'s `.statusCodeRegex` to determine a match, I added a final object with a `.statusCodeRegex` of `/.*/` to match anything. Think of it as a default in a switch statement, since all other objects before it have to not match before hitting it.

### DIFF
--- a/assets/css/styles.scss
+++ b/assets/css/styles.scss
@@ -278,6 +278,10 @@ ol li {
   background-color: #B71C1C;
 }
 
+.missing-data-response {
+  background-color: #676767;
+}
+
 fieldset#history {
   .method-list-item {
     position: relative;
--- a/pages/index.vue
+++ b/pages/index.vue
@@ -241,6 +241,13 @@
       statusCodeRegex: new RegExp(/[5][0-9]+/),
       className: 'sv-error-response'
     },
+    {
+      // this object is a catch-all for when no other objects match and should always be last
+      name: 'unknown',
+      statusCodeRegex: new RegExp(/.*/),
+      className: 'missing-data-response'
+    }
+
   ];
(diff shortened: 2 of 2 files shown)


NBTX:
LGTM! Also kudos for the only fix write-up that I've understood without checking the code! Very clear!

PR #78, opened 2019-08-28
Syntax Highlighting for HTML, CSS, JS, XML, JSON and SVG

AndrewBastin:
Sorry for the delays, I am kinda busy with travelling these days...

So yeah, this PR adds syntax highlighting to the responses powered by [Prism](https://prismjs.com/).
It looks at the returned content-type and applies the equivalent highlighting...

Well, I had to change a bit in the way how the responses are rendered, it relies on a property (responseHtml) now to render the raw html into the response box. Also, since textarea doesn't allow formatting, when an unsupported MIME type is returned, it defaults to the old textarea based layout but when a supported mime is returned, instead the content is rendered within a code tag inside a pre tag (ie, `<pre><code>{{ content }}</code></pre>`).

There are some stuff we will lose like the stretching textarea and stuff, and if there are no line breaks, the content box keeps expanding even out of the viewport. Someone should like update this PR with the ability to have proper overflow handling (I kinda don't want to work with the CSS on that one).

### DIFF
--- a/assets/css/styles.scss
+++ b/assets/css/styles.scss
@@ -166,8 +166,7 @@ fieldset.orange legend {
 select,
 input,
 option,
-textarea,
-pre {
+textarea {
   margin: 4px;
   padding: 8px 16px;
   border-radius: 4px;
@@ -176,6 +175,16 @@ pre {
   color: var(--fg-color);
   font-weight: 700;
   font-size: 18px;
+  font-family: monospace;
+}
+
+pre {
+  margin: 4px;
+  padding: 8px 16px;
+  border-radius: 4px;
+  width: calc(100% - 8px);
+  background-color: var(--bg-dark-color);
+  color: var(--fg-color);
   font-family: monospace;
 }
 
@@ -278,6 +287,10 @@ ol li {
 
 #installPWA {
   display: none;
+}
+
+.token {
+  background-color: transparent !important;
 }
 
 .info-response {
--- a/package-lock.json
+++ b/package-lock.json
@@ -2966,6 +2966,17 @@
       "resolved": "https://registry.npmjs.org/cli-boxes/-/cli-boxes-2.2.0.tgz",
       "integrity": "sha512-gpaBrMAizVEANOpfZp/EEUixTXDyGt7DFzdK5hU+UbWt/J0lB0w20ncZj59Z9a93xHb9u12zF5BS6i9RKbtg4w=="
     },
+    "clipboard": {
+      "version": "2.0.4",
+      "resolved": "https://registry.npmjs.org/clipboard/-/clipboard-2.0.4.tgz",
+      "integrity": "sha512-Vw26VSLRpJfBofiVaFb/I8PVfdI1OxKcYShe6fm0sP/DtmiWQNCjhM/okTvdCo0G+lMMm1rMYbk4IK4x1X+kgQ==",
+      "optional": true,
+      "requires": {
+        "good-listener": "^1.2.2",
+        "select": "^1.1.2",
+        "tiny-emitter": "^2.0.0"
+      }
+    },
     "cliui": {
       "version": "3.2.0",
       "resolved": "https://registry.npmjs.org/cliui/-/cliui-3.2.0.tgz",
@@ -3729,6 +3740,12 @@
       "resolved": "https://registry.npmjs.org/delayed-stream/-/delayed-stream-1.0.0.tgz",
       "integrity": "sha1-3zrhmayt+31ECqrgsp4icrJOxhk=",
       "dev": true
+    },
+    "delegate": {
+      "version": "3.2.0",
+      "resolved": "https://registry.npmjs.org/delegate/-/delegate-3.2.0.tgz",
+      "integrity": "sha512-IofjkYBZaZivn0V8nnsMJGBr4jVLxHDheKSW88PyxS5QC4Vo9ZbZVvhzlSxY87fVq3STR6r+4cGepyHkcWOQSw==",
+      "optional": true
     },
     "delegates": {
       "version": "1.0.0",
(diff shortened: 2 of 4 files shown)


liyasthomas:
Report
what works:
- HTML syntax highlighting
- JSON syntax highlighting
- JS syntax highlighting
- CSS syntax highlighting
- (prolly xml,css,svg too)

what does't works:
- responsive UI is broken
- copyResponse() is broken
- preview HTML is broken

screely-1566988602071

:point_up_2: even though syntax highlighting works, HTML tags are trimmed at unusual points such as closing tags and all.

AndrewBastin:
> Report
> what works:
> 
>     * HTML syntax highlighting
> 
>     * JSON syntax highlighting
> 
>     * JS syntax highlighting
> 
>     * CSS syntax highlighting
> 
>     * (prolly xml,css,svg too)
> 
> 
> what does't works:
> 
>     * responsive UI is broken
> 
>     * copyResponse() is broken
> 
>     * preview HTML is broken
> 
> 
> screely-1566988602071

I will work on the copy response and the preview HTML bugs, I am not really fluent in CSS stuff for doing proper overflow handling... So peeps with the knowledge please send a PR to my [repo](https://github.com/AndrewBastin/postwoman) and help me out.

NBTX:
I'll take a look at the CSS when I get home, that's no problem! I can also add draggable resizing ability (or maybe we should do show more/show less instead?).

I think rather than juggling two layouts (which adds more points of potential failure) we should just change it over to the `pre` unilaterally.

I've already done that for the websocket page as I had to add message formatting to make it readable.

liyasthomas:
+1 for <pre>

AndrewBastin:
> I'll take a look at the CSS when I get home, that's no problem! I can also add draggable resizing ability (or maybe we should do show more/show less instead?).
> 
> I think rather than juggling two layouts (which adds more points of potential failure) we should just change it over to the `pre` unilaterally.
> 
> I've already done that for the websocket page as I had to add message formatting to make it readable.

I think so too, I just kept it there because I felt like the behaviour shouldn't affect unsupported MIME types, mostly just for the resizable stuff... That's mostly it.... I too support the move to <pre> tag completely...

By the way, can someone report whether the copy response works on IE10 ? I don't use Windows so I can't check if so... Else, I have to move from the hack and use a hidden textarea with the content and run copy from there...

AndrewBastin:
> I'll take a look at the CSS when I get home, that's no problem! I can also add draggable resizing ability (or maybe we should do show more/show less instead?).
> I think rather than juggling two layouts (which adds more points of potential failure) we should just change it over to the `pre` unilaterally.
> I've already done that for the websocket page as I had to add message formatting to make it readable.



I think so too, I just kept it there because I felt like the behaviour shouldn't affect unsupported MIME types, mostly just for the resizable stuff... That's mostly it.... I too support the move to tag completely... 
 
By the way, can someone report whether the copy response works on IE10 ? I don't use Windows so I can't check if so... Else, I have to move from the hack and use a hidden textarea with the content and run copy from there...

AndrewBastin:
> Report
> what works:
> 
>     * HTML syntax highlighting
> 
>     * JSON syntax highlighting
> 
>     * JS syntax highlighting
> 
>     * CSS syntax highlighting
> 
>     * (prolly xml,css,svg too)
> 
> 
> what does't works:
> 
>     * responsive UI is broken
> 
>     * copyResponse() is broken
> 
>     * preview HTML is broken
> 
> 
> screely-1566988602071

I don't think the Preview HTML problem is related to my code as far as I can see, the code doesn't depend on the syntax highlighter div or anything and I have made sure that `response.body` property is never touched by the syntax highlighting code, the `togglePreview()` function doesn't even refer to any of my code.

NBTX:
Preview HTML covers up the textarea, so it may be a change to the HTML structure that caused it?

NBTX:
Oh I just checked the code - if I understand it correctly - a fix for the responsive UI issue is changing the style for the `pre` tags to include `white-space: pre-wrap`.

AndrewBastin:
> Oh I just checked the code, a fix for the responsive UI issue is changing the style for the `pre` tags to include `white-space: pre-wrap;`.
> 
> Alternatively, I guess you can just make it scrollable by changing it to `overflow: auto;`?

The white-space stuff didn't work, the overflow: auto option I tried before, but it didn't work, we have to give a fixed width for that to work....

I am kinda busy with some travel stuff right now, so I will be a bit slow to respond to stuff, sorry...

AndrewBastin:
> Report
> what works:
> 
>     * HTML syntax highlighting
> 
>     * JSON syntax highlighting
> 
>     * JS syntax highlighting
> 
>     * CSS syntax highlighting
> 
>     * (prolly xml,css,svg too)
> 
> 
> what does't works:
> 
>     * responsive UI is broken
> 
>     * copyResponse() is broken
> 
>     * preview HTML is broken
> 
> 
> screely-1566988602071

Btw, I just checked, preview HTML isn't broken, well it might have been some bug with `dev.to` which didn't make it render properly with our weird stretched out viewport :man_shrugging: 

It is rendering fine now....

liyasthomas:
Let me run some tests once again

NBTX:
If you need a fixed width, off the top of my head I think you can try wrapping the response in a div that's `display: block; position: relative;` and then making it `width: 100%;`, but if that doesn't work I'll play around with it.

starvsion:
> If you need a fixed width, off the top of my head I think you can try wrapping the response in a div that's `display: block; position: relative;` and then making it `width: 100%;`, but if that doesn't work I'll play around with it.

I tried it, by adding a max-width (75vw, or 85 rem), max-height and scroll to pre , but doesn't work very well with view port width 720 - 1200 ... cannot use percentage, cause the HTML thing is wildly long, and none of the parents has a fixed width ...

PR #80, opened 2019-08-28
Readable GitHub logo

pmankar:
For light theme, the logo was not having much contrast.
Now, it just updates the image style to make it more consistent with the theme color.


layouts/default.vue
```diff
@@ -120,7 +120,10 @@
         // Once the PWA code is initialized, this holds a method
         // that can be called to show the user the installation
         // prompt.
-        showInstallPrompt: null
+        showInstallPrompt: null,
+        logoStyle() {
+          return "margin-right: 16px;" + (this.$store.state.postwoman.settings.THEME_CLASS ? " filter: invert(100%); -webkit-filter: invert(100%);" : '')
```
NBTX: THEME_CLASS being set does not necessarily imply light theme. Can you perhaps change it to `.includes('light')`?

pmankar: Yes. Agree.
Updated the code.

PR #89, opened 2019-08-28
Websocket log auto scroll to bottom

pmankar:
With increasing number of messages, the new messages are appended to the bottom and thus needs to be scrolled automatically. Also added timestamp in the log.


pmankar:
also added code for GitHub logo contrast on theme switch.
no storage of theme class also handled.

PR #95, opened 2019-08-29
Add TipTap Editor for code highlighting

starvsion:
This one was inspired from : https://github.com/liyasthomas/postwoman/pull/78 . 

Basically this one is done with TipTap editor (readonly mode) and highlight.js. Given the fact that tiptap is a text editor (but displays as divs), it solves the problem with responsiveness, copying html etc., and this one is able to handle any type of input (but highlighting is only available for js, css,  xml, json or other language if we wanna add.).

Only Problem with this: HTML is always parsed into DOM and display as text, so there is no HTML highting here ...

### DIFF
--- a/assets/css/styles.scss
+++ b/assets/css/styles.scss
@@ -1,166 +1,165 @@
 $responsiveWidth: 720px;
-
 ::selection {
-  background-color: var(--ac-color);
-  color: var(--act-color);
+    background-color: var(--ac-color);
+    color: var(--act-color);
 }
 
 ::-webkit-scrollbar {
-  width: 8px;
+    width: 8px;
 }
 
 ::-webkit-scrollbar-thumb {
-  background-color: #4a4a4a;
+    background-color: #4a4a4a;
 }
 
 * {
-  box-sizing: border-box;
-  outline: 0;
-  border: 0;
-  font-family: "Poppins", "Roboto", "Noto", sans-serif;
+    box-sizing: border-box;
+    outline: 0;
+    border: 0;
+    font-family: "Poppins", "Roboto", "Noto", sans-serif;
 }
 
 a {
-  display: inline-flex;
-  color: inherit;
-  text-decoration: none;
-  font-weight: 700;
+    display: inline-flex;
+    color: inherit;
+    text-decoration: none;
+    font-weight: 700;
 }
 
 body {
-  background-color: var(--bg-color);
-  color: var(--fg-color);
-  font-weight: 500;
-  line-height: 1.5;
-  -webkit-tap-highlight-color: transparent;
-  -webkit-touch-callout: none;
-  -webkit-user-select: none;
-  user-select: none;
+    background-color: var(--bg-color);
+    color: var(--fg-color);
+    font-weight: 500;
+    line-height: 1.5;
+    -webkit-tap-highlight-color: transparent;
+    -webkit-touch-callout: none;
+    -webkit-user-select: none;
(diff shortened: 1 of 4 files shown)


NBTX:
Is this not a direct conflict with #78?

starvsion:
> Is this not a direct conflict with #78?

Just a proposal for an alternative solution to that one ...

NBTX:
Fair enough, I just feel like using a WYSIWYG editor library is a bit overkill. It also abstracts away a lot of the control from us in terms of style. So I prefer fixing #78 over a new implementation. Either way, having an alternative doesn't hurt!

liyasthomas:
Better alternatives are always welcomed but as @NBTX said, this seems like an overkill.

AndrewBastin:
> This one was inspired from : #78 .
> 
> Basically this one is done with TipTap editor (readonly mode) and highlight.js. Given the fact that tiptap is a text editor (but displays as divs), it solves the problem with responsiveness, copying html etc., and this one is able to handle any type of input (but highlighting is only available for js, css, xml, json or other language if we wanna add.).
> 
> Only Problem with this: HTML is always parsed into DOM and display as text, so there is no HTML highting here ...

Thank you for your PR, I have thought about this approach as well, but I kinda had the thought of keeping Postwoman as lean as possible, I mean, that is one sole thing this project focuses on right ? To be a quick and lean alternative to other API Requesters out there. That is even one of the reasons I preferred Prism over highlight.js, we don't need full on parsers, for our use case, a regex based highlighter is more than enough... Prism is lean, and really modular so with the build time tree shaking and stuff, we just have minimal code in the final version.

AndrewBastin:
> > This one was inspired from : #78 .
> > Basically this one is done with TipTap editor (readonly mode) and highlight.js. Given the fact that tiptap is a text editor (but displays as divs), it solves the problem with responsiveness, copying html etc., and this one is able to handle any type of input (but highlighting is only available for js, css, xml, json or other language if we wanna add.).
> > Only Problem with this: HTML is always parsed into DOM and display as text, so there is no HTML highting here ...
> 
> Thank you for your PR, I have thought about this approach as well, but I kinda had the thought of keeping Postwoman as lean as possible, I mean, that is one sole thing this project focuses on right ? To be a quick and lean alternative to other API Requesters out there. That is even one of the reasons I preferred Prism over highlight.js, we don't need full on parsers, for our use case, a regex based highlighter is more than enough... Prism is lean and really modular so with the build time tree shaking and stuff, we just have minimal code in the final version...

starvsion:
> > This one was inspired from : #78 .
> > Basically this one is done with TipTap editor (readonly mode) and highlight.js. Given the fact that tiptap is a text editor (but displays as divs), it solves the problem with responsiveness, copying html etc., and this one is able to handle any type of input (but highlighting is only available for js, css, xml, json or other language if we wanna add.).
> > Only Problem with this: HTML is always parsed into DOM and display as text, so there is no HTML highting here ...
> 
> Thank you for your PR, I have thought about this approach as well, but I kinda had the thought of keeping Postwoman as lean as possible, I mean, that is one sole thing this project focuses on right ? To be a quick and lean alternative to other API Requesters out there. That is even one of the reasons I preferred Prism over highlight.js, we don't need full on parsers, for our use case, a regex based highlighter is more than enough... Prism is lean and really modular so with the build time tree shaking and stuff, we just have minimal code in the final version...

I agree, maybe having an editor is an overkill, I will close this PR now.

PR #96, opened 2019-08-30
Use axios for requests, UI tweaks, and more...

### LINKED ISSUE #61: HTTP request with different library
**Is your feature request related to a problem?**
No

Why the need to use xhr instead, for example, Axios.
One factor should be the need to use interceptors before or after request

Any

**Additional context**
No

NBTX:
Test URL: https://nbtx.github.io/postwoman-vue/

---

image
- Resolves #61 by using the [Nuxt.js axios module](https://axios.nuxtjs.org/) to handle requests instead of browser XHR library.
- Adds progress bar when requests are made.

image
- Adds metro-style slider component and changes checkboxes to sliders.

image
- Shows response body on error, thus allowing you to preview error pages, etc.

---

- Fixes the disabled state on key events and button presses.
- Fixes the disabled *style* on key events and button presses.

### DIFF
--- a/assets/css/styles.scss
+++ b/assets/css/styles.scss
@@ -84,6 +84,11 @@ button {
   font-weight: 700;
   font-size: 16px;
   cursor: pointer;
+
+  &[disabled], &.disabled {
+    opacity: 0.7;
+    cursor: default;
+  }
 }
 
 fieldset {
@@ -220,7 +225,7 @@ input[type="checkbox"] {
   background-color: var(--err-color);
 }
 
-.disabled {
+.disabled, input[disabled] {
   background-color: var(--err-color);
   color: #b2b2b2;
 }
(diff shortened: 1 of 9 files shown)


liyasthomas:
@NBTX I'm getting error on console:

> `Refused to set unsafe header "Content-Length"`

On methods: `PUT`, `POST`, `PATCH`

NBTX:
Weird... weren't they set befor

nickpalenchar:
I'm running an older version and can confirm the same thing happens before this merge (it was set [here](https://github.com/liyasthomas/postwoman/blob/4796fe402446ee01f881b5901cbc18f2fc554f14/pages/index.vue#L420))

It looks like this is the behavior of most browsers rather than the code.

liyasthomas:
So that' probably a browser specific bug then.

NBTX:
Perhaps we just weren't testing POST requests because I read up that browsers block Content-Length as a security feature and it's handled for us by the browser.

liyasthomas:
Think it can be solved by proxy middleware?

NBTX:
Yes

Issue #94, opened 2019-08-29
Show "Send" button all over the page or enable hotkeys


tiagoa:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when configuring the request and then having to scroll up the page just to hit the "Send" button.

**Describe the solution you'd like**
Set a fixed position for the button "Send" or/and enable hotkeys.

NBTX:
What if we made the request section sticky to the top of the window as you scroll?
Though, either way, hotkeys does seem like a good idea.

izerozlu:
I'm working on an implementation with the [IntersectionObserver](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API).
After that implementation i'll look into enabling some hotkeys as well.

liyasthomas:
I did a UI with sticky request section, but it broke responsiveness. Looking forward for suggestiongs.

PR #109, opened 2019-08-30
Fix intersection button send

guastallaigor:
In my opinion something is wrong right now with the button fixed, using the Intersection API (ref #107 #94). At my environment, it only works on the first scroll and every two full page scrolls after that.

This PR changes the logic a bit to make it work all the time.

**Right now:**

screencast-liyasthomas github io-2019 08 30-09-31-02

**After this PR:**

screencast-localhost-3000-2019 08 30-09-35-06

### DIFF
--- a/pages/index.vue
+++ b/pages/index.vue
@@ -533,9 +533,10 @@
       },
       observeRequestButton() {
         const requestElement = this.$refs.request.$el;
+        const sendButton = this.$refs.sendButton;
         const observer = new IntersectionObserver((entries, observer) => {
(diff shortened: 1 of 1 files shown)


izerozlu:
@guastallaigor Your changes look a better way to to implement the feature.

But i couldn't get it worked on the firefox. That should be the problem. I was working on firefox when i was implenting it and it was working fine. I suppose we should find the way for it to work in both in your environment and my environment. But for now your way looks better and also works in chrome. So i'll merge in yours.

PR #114, opened 2019-08-30
😬 Actually send the request body

### LINKED ISSUE #113: Request Body Is Not Sent
**Describe the bug**
During a POST/CREATE/PUT/PATCH request the request body is not actually being sent.

**To Reproduce**
1. Go to postwoman.io
2. Switch to POST
3. Enter: `https://jsonplaceholder.typicode.com` as URL and `/users` as path
4. Add a field "name" with value "foo"
5. Send the request
6. Check the response and see missing field "name" with value "foo"

**Expected behavior**
The response should be `{ "name":"foo", "id": 11}`

 - OS: Ubuntu 18.04
 - Browser: Chrome (Incognito Tab)
 - Version: 76

**Additional context**
I’ve tested this both with a local development project as well as the JSONPlaceholder API and the results are both the same. This problem did not yet occur 20 hours ago, so perhaps it’s a regression from one of the more recent commits.

NBTX:
Fixes #113 

(whoops)

### DIFF
--- a/pages/index.vue
+++ b/pages/index.vue
@@ -378,8 +378,9 @@
 
         // If the request has a request body, we want to ensure Content-Length and
         // Content-Type are sent.
+        let requestBody;
         if (this.hasRequestBody) {
-          const requestBody = this.rawInput ? this.rawParams : this.rawRequestBody;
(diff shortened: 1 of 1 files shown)


NBTX:
Tested with `GET` and `POST`

PR #121, opened 2019-08-31
⚠️ Ensure build code doesn't fail if version can't be obtained from GitHub

NBTX:
See title.

### DIFF
--- a/build.js
+++ b/build.js
@@ -17,17 +17,23 @@ try {
 
 		let version = {};
 		// Get the current version name as the tag from Git.
-		version.name = runCommand('git', ['tag']);
+		version.name = (process.env.TRAVIS_TAG || runCommand('git', ['tag']));
 
 		// FALLBACK: If version.name was unset, let's grab it from GitHub.
 		if(!version.name){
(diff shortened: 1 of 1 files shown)


NBTX:
@liyasthomas okay to merge this on account of the build passing being proof it works?

larouxn:
Hmm, I like this idea but, unless I'm mistaken how this all works, the latest GitHub release is [v0.1.0](https://github.com/liyasthomas/postwoman/releases/tag/v0.1.0), meanwhile the `package.json` states we're on [v1.0.0]. Will we need to synchronize these?
https://github.com/liyasthomas/postwoman/blob/c1abf5f07c6a58cb39c25b5990d13d1a9de941aa/package.json#L3

liyasthomas:
Lets continue on 0.1.0
Better update package.json

NBTX:
It depends. Those numbers shouldn't need to be used but occasionally - like when we're merging a lot of PRs - we hit the rate limit on GitHub's API and can't get the last release.

Thinking about it, I guess the better alternative is just to hide the version name if it can't be defined.

NBTX:
Okay as I was about to submit another PR, I'm gonna update the `package.json`, merge this, submit my PR and then open an issue to continue this discussion.

Issue #100, opened 2019-08-30
default menu option - 'Http' is not highlighted when launched from installed pwa app (UI bug)


NileshSP:
**Describe the bug**
After launching pwa app from desktop, default 'Http' menu is not highlighted

![image](https://user-images.githubusercontent.com/13110493/63991977-ca196a80-cab7-11e9-9a2e-0d1e50533e39.png)

**To Reproduce**
1. Go to 'postwoman.io'
2. Click on 'Install PWA'
3. Open Postwoman app from desktop
4. 'Http' is not highlighted

**Expected behavior**

![image](https://user-images.githubusercontent.com/13110493/63992157-9a1e9700-cab8-11e9-8c1e-7c30cbc217b1.png)

**Desktop :**
 - OS: Win 10
 - Browser chrome
 - Version latest

liyasthomas:
Issue has been identified. Most probably its an issue with start URL of PWA. Will look into it ASAP.

NileshSP:
Have found the issue after a bit of troubleshooting and have a solution, let me know if I can contribute by submitting a PR

NBTX:
@NileshSP go for it!

NileshSP:
@liyasthomas , @NBTX  - submitted PR #125

liyasthomas:
Fixed on #125

PR #125, opened 2019-08-31
update with PR for issue/bug #100

NileshSP:
Issue - ideally the default url expected is '/' to highlight the 'Http' menu option but in PWA's case the url passed is '/?standalone=true' to indicate and load it as an app where 'standalone=true' is automatically appended by the rendering browser engine 

Solution - newly added middleware 'parsedefaulturl.js' would intercept the request before loading the default page and redirect appropriately as desired


izerozlu:
Can't do a review today. I'll be looking into that tonight or tomorrow. Sorry about that.

PR #118, opened 2019-08-30
Bind history state and query params of requests

yubathom:
* Enable uri with preseted headers, params and body. Examples:

- <a href="https://yubathom.github.io/postwoman/?headers=%5B%7B%22key%22%3A%22foo%22,%20%22value%22%3A%22bar%22%7D%5D">/?headers=[{"key":"foo", "value":"bar"}]
 - <a href="https://yubathom.github.io/postwoman/?params=%5B%7B%22key%22%3A%22foo%22,%20%22value%22%3A%22bar%22%7D%5D">/?params=[{"key":"foo", "value":"bar"}]
 - <a href="https://yubathom.github.io/postwoman/?method=POST&bodyParams=%5B%7B%22key%22%3A%22foo%22,%20%22value%22%3A%22bar%22%7D%5D">/?method=POST&bodyParams=[{"key":"foo", "value":"bar"}]</a>

or combined with other params like:
[/?method=POST&auth=Basic&httpUser=someUser&httpPassword=somePass](https://yubathom.github.io/postwoman/?method=POST&url=https%3A%2F%2Freqres.in&path=%2Fapi%2Fusers&auth=Basic&httpUser=someUser&httpPassword=somePass)
* The url and router state are now binded. So the requests can be shared (just copy it from the url bar)

PR #131, opened 2019-09-02
Bind raw body input to url of requests

yubathom:
Enable raw input binding to the url. Raw inputs are now shareable. Example:
[/?method=PUT&rawParams={"foo":"bar","col":[1,2,3]}](https://yubathom.github.io/postwoman/?method=PUT&rawParams=%7B%0A%22foo%22%3A%22bar%22,%0A%22col%22%3A%5B1,2,3%5D%0A%7D)


yubathom:
Related to #29. 
Update of  #72 and #118.

Issue #93, opened 2019-08-29
Import request from cURL


tiagoa:
**Is your feature request related to a problem? Please describe.**
I would like to import a request by providing a cURL command.

**Describe the solution you'd like**
Like in Postman we have the option to import: https://dev.to/stephencweiss/how-to-test-your-api-using-curl-postman-382c

liyasthomas:
#76

nityanandagohain:
@liyasthomas Can I work on this.

liyasthomas:
Yeah sure! Go ahead.

liyasthomas:
@nityanandagohain any updates on the issue? I'm hoping to include this feature in the next version release ✨

nityanandagohain:
@liyasthomas I will send a PR today.

Rohittater255:
Is this issue resolved??  Can we Import API Request as curl In Hoppscoth now??

Anubarak:
I searched for this as well - any way to do that?

arumoy:
> For anyone with the same issue:
> 
> You can click into the URL bar and just paste your request there

This needs to be in the UI. Or, maybe, just a button 😂 — doesn't have to be an easter egg at all.

PR #138, opened 2019-09-06
curl import added

nityanandagohain:
Fixes issue #93 

Hey @liyasthomas  I have tried to add functionality to import curl commands. Please let me know if any changes are required.

Screenshot from 2019-09-06 16-17-42
Screenshot from 2019-09-06 16-19-49
Screenshot from 2019-09-06 16-23-00


pages/index.vue
```diff
@@ -1,6 +1,19 @@
 <template>
   <div class="page">
-    <pw-section class="cyan" label="Request" ref="request">
+    <pw-section class="blue" label="Request" ref="request">
+      <button id="show-modal" @click="showModal = true">IMPORT</button>
+      <import-modal v-if="showModal" @close="showModal = false">
+        <div slot="header">
+          <h2>IMPORT CURL</h2>
```
NBTX: Can you please make this title case?
'Import cURL'

pages/index.vue
```diff
@@ -1,6 +1,19 @@
 <template>
   <div class="page">
-    <pw-section class="cyan" label="Request" ref="request">
+    <pw-section class="blue" label="Request" ref="request">
+      <button id="show-modal" @click="showModal = true">IMPORT</button>
```
NBTX: Can you rename this to 'Import' and move it outside of the `pw-section` for request?

nityanandagohain: @NBTX  I have made the requested changes, please have a look.

pages/index.vue
```diff
@@ -1,10 +1,10 @@
 <template>
   <div class="page">
-    <pw-section class="blue" label="Request" ref="request">
+    <pw-section class="blue" label="Import" ref="import">
       <button id="show-modal" @click="showModal = true">IMPORT</button>
```
NBTX: Can you change this from 'IMPORT' to 'Import'?
Other than that, everything's perfect.

nityanandagohain: Done.

liyasthomas:
@NBTX merge after your approval

liyasthomas:
@nityanandagohain @hosseinnedaee 
Ran into a bug in cURL parser! After importing a cURL, `PATH` is appended to `URL` which is okay. But tje existing `PATH` field stays same. We manually need to empty `PATH` field before sending next request.
To reproduce:
1. Generate a cURL code (⚡ icon under URL field) for the default test request
2. Copy thc cURL and import the same using Import cURL button
3. See URL is appended with PATH as well as the existing PATH stays as same as before
Fix:
Either replace the existing PATH field with the one from the imported cURL (which is the best UX)
Or
Append PATH to URL, but empty the existing PATH field.

I wonder why anybody haven't reported this yet. And need to look at error handler on index.vue regarding importCURL as well since from 2nd import cURL request, function lands me on error handler shooting error toast. 
https://github.com/liyasthomas/postwoman/blob/master/assets/js/curlparser.js

PR #146, opened 2019-09-14
.htaccess file added

realabbas:
**.htaccess file** added for better **Leverage browser caching**, which increases the page speed by a good amount.

htaccess file

Browser Leverage)

_Note:_ The difference in fully page load is differing because of the hosted server.

### DIFF
--- a/.htaccess
+++ b/.htaccess
@@ -1,0 +1,24 @@
+<IfModule mod_expires.c>
+  ExpiresActive On
+
+  # Images
+  ExpiresByType image/jpeg "access plus 1 year"
+  ExpiresByType image/gif "access plus 1 year"
+  ExpiresByType image/png "access plus 1 year"
+  ExpiresByType image/webp "access plus 1 year"
+  ExpiresByType image/svg+xml "access plus 1 year"
+  ExpiresByType image/x-icon "access plus 1 year"
+
+  # Video
+  ExpiresByType video/mp4 "access plus 1 year"
+  ExpiresByType video/mpeg "access plus 1 year"
+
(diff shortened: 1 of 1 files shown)


.htaccess
```diff
@@ -0,0 +1,24 @@
...
+
+  # Images
+  ExpiresByType image/jpeg "access plus 1 year"
+  ExpiresByType image/gif "access plus 1 year"
+  ExpiresByType image/png "access plus 1 year"
+  ExpiresByType image/webp "access plus 1 year"
+  ExpiresByType image/svg+xml "access plus 1 year"
+  ExpiresByType image/x-icon "access plus 1 year"
+
+  # Video
```
NBTX: We can remove this section.

.htaccess
```diff
@@ -0,0 +1,24 @@
...
+  ExpiresByType video/mp4 "access plus 1 year"
+  ExpiresByType video/mpeg "access plus 1 year"
+
+  # CSS, JavaScript
+  ExpiresByType text/css "access plus 1 month"
+  ExpiresByType text/javascript "access plus 1 month"
+  ExpiresByType application/javascript "access plus 1 month"
+
+  # Others
```
NBTX: Also this section.

realabbas:
> I guess it makes sense to remove the sections that don't apply at all.

Okay yes, it does make sense. I have made an amendment to the file, removing the unnecessary sections

realabbas:
> Seems reasonable now that the media types make sense for Postwoman.
> 
> Would certainly like @NBTX's re-review as well though.

Yeah, waiting for @NBTX reviews

Issue #62, opened 2019-08-25
Fast URL entry


NBTX:
**Is your feature request related to a problem? Please describe.**
No.

**Describe the solution you'd like**
Similar to Postman's interface, I think we should make the URL entry a single bar at the top, with the method on the left. Then we should just parse the URL and fill in the relevant information.

We could keep the extra sections but this would make Postwoman a lot more intuitive and faster to use.

**Describe alternatives you've considered**
n/a

**Additional context**
![image](https://user-images.githubusercontent.com/43181178/63654349-20fa0980-c771-11e9-8a7c-4a94e9b199d7.png)
*As shown, Postman parses the URL as it is entered and automatically fills in the relevant sections.*

NBTX:
I'm happy to implement this myself, but I wanted to open a discussion on the matter.

liyasthomas:
This fact might surprise you: other than seeing some screenshots on internet, I haven't used Postman or seen its UI (this is just because, my 7 year old pc can't afford to run another Electron app). Afterall, I didn't intend to create an alternative to Postman, in my opinion this project should outperform any other such tools, since its lightweight, faster & opensource.

If there's any feature we'd like to add, all are free to discuss here. Everything is better when there's freedom and customisation in it.

liyasthomas:
@NBTX see #63 and share your thoughts. Me too think this is an essential feature somehow I forgot at the first place. My bad. Anyway thanks for pulling it up.

NBTX:
I'm actually only drawing the parallel to Postman's interface from experience, rather than because the names are similar, and this is the UI I'm used to - but this, as I said is something I would find useful.

I already very much prefer Postwoman because it's so much lighter. Even on my pretty decent systems Postman takes upwards of 7 seconds to start up, and then usually has several annoying pop ups to deal with.

liyasthomas:
#93 needs similar kinda approach

terranblake:
@liyasthomas and @NBTX i'd like to take this one for hacktoberfest 😄

liyasthomas:
@terranblake you are free to contribute to this 🔥👌 assigning to you.

terranblake:
@liyasthomas for this feature to be implemented, i do have a couple questions about a proposed design. 

1. can the `path` field be removed as a part of the solution?
2. can i move the `permalink` and `show code` buttons below the `url` field on the same plane as `import cURL`?

liyasthomas:
[...]

PR #156, opened 2019-09-22
Sync path and parameters

hosseinnedaee:
I used _url_ and _querystring_ node.js built-in modules to parse the path and query.


liyasthomas:
@hosseinnedaee Import cURL seems to be broken. Can you take a look at that

hosseinnedaee:
> @hosseinnedaee Import cURL seems to be broken. While importing cURL generated from Postwoman itself, `path` is not updated. Default path value is persisted even after cURL import. Can you take a look at that. **Locally it works fine by the way confused**
> To reproduce:
> 
> 1. goto postwoman.io
> 2. send a `POST` request
> 3. copy `cURL` code
> 4. again visit postwoman.io
> 5. import cURL
> 6. see `path` is as same as default placeholder, not updated after cURL import

@liyasthomas  anyway, when I import curl it just shows me this error: "cURL is not formatted properly"
Screenshot from 2019-09-30 22-42-31

liyasthomas:
See, on this screenshot `path` field is appended to `URL` as well as to the path field itself. Guess how did that happen?

hosseinnedaee:
Maybe we should wait until fast URL entry #62 to be closed and then I work around it. because I think they depend on each other and maybe after using just one field for `URL` and `Path` the issue be solved. what do you think?

PR #176, opened 2019-10-01
Collections

### LINKED ISSUE #29: Ability to store/share/create collections

TheHollidayInn:
Hey! I wanted to get this posted to get some feedback and show progress

Closes #29

- Collections
  - [x] Create
  - [x] Read
  - [x] Update
  - [x] Delete
  - [x] Export
- Folders
  - [x] Create
  - [x] Read
  - [x] Update
  - [x] Delete
- Requests
  - [x] Read
  - [x] Update
  - [x] Delete
  - [x] Save Current
- Other
  - [ ] Maybe add as a side panel
  - [ ] Style updates


liyasthomas:
So far, so good. I liked your approach on storing collections.

yubathom:
Just mounted locally. Nice approach with vuex @TheHollidayInn! 

I have two doubts:
1. How would you like to deal the shareability of theese collections? Are you considering an backend api to give you a response so you could populate the store before the creation of the view? 
> obs. Don't keep multiple requests data in urls due to its [characters limit](https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers)
2. Would you consider using the data of history of local storage (`localStorage.history`) to easily populate/update the store of theese collections? Or maybe keeping them synced?

TheHollidayInn:
Hey @yubathom! For this I figured I would not implement a backend, but implement locally (save with local storage). It wouldn't be much to add a backend later or event add an option for users to supply their own.

For sharing, I was thinking of just exporting the JSON like postman does and allow for users to also import JSON.

liyasthomas:
@TheHollidayInn thought I should just ping you to know the status of this PR.

TheHollidayInn:
Hey! Sorry, got a little behind, but I will push my updates tomorrow.

TheHollidayInn:
Alright! I think I have the first round of features (Although there are plenty more I could think of). I'd appreciate any feedback. Sorry it took me so long.

liyasthomas:
MVP seems to be working awesome!

But couldn't find requests in Export modal. Actually it would be great if you could  give a brief intro to using collections. I want to know why requests doesn't appear in while exporting. Or is it only for me! I might have missed something 🤐🤐

liyasthomas:
@TheHollidayInn This is awesome! I've added few stylings.

After some manual testing, I often ran into some errors.
- editCollection often brings up addCollection modal. Especially on first click. From second or third click onwards, modal seems to be correct.
- This same issue happens to editRequest and editFolder oftentimes. But nit always. Any idea what's wrong?

I tried to debug but logic seems to be fine. Even though you use same modal to create and edit, logic seems to be correct. Any idea what went wrong let me know if you could reproduce the bug!

PR #183, opened 2019-10-04
Add Device orientation and Update start_url in manifest

mskian:
- Add Device orientation  -> Site Will Automatically Rotate when the user rotates the device 
- Update `start_url` we can easily track the PWA installation by UTM Source


liyasthomas:
Is there any other contributions you wanna make? Small contributions like this are ignored. You could open an issue which will be fixed in the next minor patch update.

Btw, we already implemented google analytics for usage detection.

Issue #133, opened 2019-09-04
Please add a label for each request. It will be helpful.


MoonSmile1707:
When i post or get a request and then i want to use it again. But i don't know where it is... same issue on postman... 
Please add a label for each request. :D

liyasthomas:
History section has entries with Status code, URL, Path, timestamp. Isn't that enough?

liyasthomas:
This could be related to #29 
A collection of requests with a common label/name would be :100: helpful. Make a collection of such requests and ability to share them.

MoonSmile1707:
when i work with GraphQL. It was enough to find a request quickly. thanks <3

sharath2106:
@liyasthomas I would like to work on this issue. Let me know if no one has claimed to work on this yet.

liyasthomas:
@sharath2106 You're welcome. Assigning to you.

sharath2106:
@liyasthomas You should be getting a PR by today or tomorrow. First timer here.

liyasthomas:
_There is always a first._

liyasthomas:
@sharath2106 just thought I should ping you for PR.

sharath2106:
Done with setting up data and the text box for the request name. Rendering the same in the history section is not happening even after introducing the fields in history.vue component and index.vue.
<img width="1226" alt="Screenshot 2019-10-04 at 7 51 03 AM" src="https://user-images.githubusercontent.com/25933070/66176609-08d4ae80-e67c-11e9-8d75-3f7c2d88bd6b.png">
<img width="1220" alt="Screenshot 2019-10-04 at 7 50 53 AM" src="https://user-images.githubusercontent.com/25933070/66176610-08d4ae80-e67c-11e9-9c41-6bed147615b9.png">

liyasthomas:
History data is fetched from localStorage. And also binded to it. After sanding the request, check `Dev tools > Application tab > Local storage > http://localhost:300 > history` to see entries.

sharath2106:
Have tried to print the history entry in the console

<img width="1680" alt="Screenshot 2019-10-04 at 10 43 29 AM" src="https://user-images.githubusercontent.com/25933070/66182765-fbc3b980-e693-11e9-8b2e-bb6f6d124883.png">

This is the local storage - 
[...]

PR #184, opened 2019-10-04
Add Request name label for every requests

sharath2106:
This pull request is based on issue #133 

I've added a field to Enter the request name for each request. Please find the screenshot below - 

Screenshot 2019-10-04 at 2 47 24 PM

Ideally, the same request name should be listed down in the History section of the request. The changes are also made in the corresponding files.

### DIFF
--- a/pages/index.vue
+++ b/pages/index.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="page">
-      <input :class="{ error: !requestName }" @keyup.enter="requestName ? validRequestName() : null" id="label" name="label" type="label" v-model="label">
+      <input @keyup.enter="requestName" id="label" name="label" type="label" v-model="label">
     <pw-modal v-if="showModal" @close="showModal = false">
       <div slot="header">
         <ul>
@@ -523,7 +523,7 @@
     },
     computed: {
       requestName() {
-        return this.label.match(/^([^?]*)\??/)[1]
+        return this.label
       },
       statusCategory() {
         return findStatusGroup(this.response.status);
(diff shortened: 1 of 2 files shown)


pages/index.vue
```diff
@@ -1,5 +1,6 @@
 <template>
   <div class="page">
+      <input :class="{ error: !requestName }" @keyup.enter="requestName ? validRequestName() : null" id="label" name="label" type="label" v-model="label">
```
liyasthomas: What's the use of `validRequestName()` ?
I meant, why you want to pattern match a `label` string on every keyup?

pages/index.vue
```diff
@@ -520,6 +522,9 @@
       }
     },
     computed: {
+      requestName() {
+        return this.label.match(/^([^?]*)\??/)[1]
+      },
```
liyasthomas: Label will be a string which user decides, we don't want to enforce any pattern for it. Isn't it?

pages/index.vue
```diff
@@ -686,6 +691,13 @@
           behavior: 'smooth'
         });
       },
+      async validRequestName() {
+        if (!this.requestName) {
+          this.$toast.error('Request Name is not valid', {
+            icon: 'error'
+          });
+          return;
+        }},
```
liyasthomas: Can you tell me an use case for this?

pages/index.vue
```diff
@@ -773,6 +785,7 @@
 
             // Addition of an entry to the history component.
             const entry = {
+              label: this.requestName,
```
liyasthomas: Already `label` is binded with input field value isn't it? so we don't need to specify `this.name` for it.

sharath2106:
@liyasthomas Apologies. I just committed and pushed from the wrong repo. I'll make the changes and push again.

PR #185, opened 2019-10-04
Added auto theme

AndrewBastin:
This PR adds an automatic theme which uses the `prefers-color-scheme` media query to automatically go to dark mode or be in light mode respecting the user's system's norms (like Android and iOS dark modes, Windows/macOS dark modes (some change based on time of day)).

Now, I do kinda feel auto to be a good candidate for being the default theme, since it integrates more with the system preference. So if this can be replaced from dark as the default, please feel free to add to my commit.

### DIFF
--- a/assets/css/themes.scss
+++ b/assets/css/themes.scss
@@ -1,10 +1,57 @@
 /**
   Main Themes:
 
+  - auto
   - dark (default)
   - light
   - black
  */
+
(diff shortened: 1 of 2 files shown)


liyasthomas:
I actually had one PR cooking up for this since you add Black mode. And default to auto mode seems reasonable.

PR #199, opened 2019-10-19
Fix CORS and mixed content issue

hosseinnedaee:
In this PR I used  Nuxtjs serverMiddleware to create a proxy server and pass the requests through the proxy and get over CORS and mixed-content issue's

resolving https://github.com/liyasthomas/postwoman/issues/175 and https://github.com/liyasthomas/postwoman/issues/2

### DIFF
--- a/pages/index.vue
+++ b/pages/index.vue
@@ -700,6 +700,28 @@
           behavior: 'smooth'
         });
       },
+      async makeRequest(auth, headers, requestBody) {
+        const config = this.$store.state.postwoman.settings.PROXY_ENABLED ? {
+            method: 'POST',
+            url: '/proxy',
+            data: {
+              method: this.method,
+              url: this.url + this.pathName + this.queryString,
+              auth,
+              headers,
+              data: requestBody ? requestBody.toString() : null
+            }
+          } : {
+            method: this.method,
+            url: this.url + this.pathName + this.queryString,
+            auth,
+            headers,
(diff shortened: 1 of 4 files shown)


pages/index.vue
```diff
@@ -821,13 +838,24 @@
...
             this.$refs.historyComponent.addEntry(entry);
             return;
+          } else {
+            this.response.status = error.message;
+            this.response.body = "See JavaScript console (F12) for details.";
+            this.$toast.error('Something went wrong!', {
+              icon: 'error'
+            });
+            if(!this.$store.state.postwoman.settings.PROXY_ENABLED) {
+              this.$toast.info('Turn on the proxy', {
```
NBTX: This is looking good to me, my only recommendation is that this be renamed to 'Try enabling the proxy' or words to that effect, but it's not that significant.

liyasthomas: How about "Switch to Proxy mode"?

NBTX:
Oh- just one other issue, we need to add `Access-Control-Allow-Headers: *` to the proxy response

pages/index.vue
NBTX: Yes - but perhaps with the question mark. The idea being to imply to the user that it's something that might help with the error.

NBTX:
See #200 which includes these commits with some finishing touches and bug fixes.

PR #215, opened 2019-10-25
Re-order sections and add toggle for including authentication in URL

NBTX:
This re-orders the sections as follows:
- Request
- *Code* (when made visible)
- Request Options (Authentication / Headers / Parameters)
- Response
- Collections
- History

Additionally, this adds a toggle to the Authentication section (the value of which is saved) that allows a user to toggle whether or not they want the authentication information in the URL.
_Authentication in the URL is also now turned **off** by default for security reasons._

Closes https://github.com/liyasthomas/postwoman/issues/178

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -33,6 +33,7 @@
         </nav>
       </div>
     </header>
+    <br>
     <nuxt id="main" />
     <footer>
       <!-- Top section of footer: GitHub/install links -->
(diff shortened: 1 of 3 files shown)


liyasthomas:
I think having request and response sections next to each other is better. Did you find difficulties to scroll down for request options..? I don't know.  Maybe maintaining the flow as in this PR is better. 👍

NBTX:
@liyasthomas I thought it made more logical sense to have request options to the request section - and then they're more intuitively obvious as request options.
Plus, I felt putting it before was okay given that the page auto-scrolls to the response section when you send a request.

liyasthomas:
Proceed 👍

PR #226, opened 2019-10-27
bodyParams and params equal when method change

breno-pereira:
I don't know if that makes sense, but I think when method change `bodyParams` and `params` should be the same

### Now (postwoman.io)
image

when I change to POST `bodyParams` is empty

image

### This PR
image

when I change to POST `bodyParams` are equal to `params`

image

### DIFF
--- a/pages/index.vue
+++ b/pages/index.vue
@@ -1531,8 +1531,13 @@ export default {
         this.setRouteQueryState();
     },
     methodChange() {
-      // this.$store.commit('setState', { 'value': ["POST", "PUT", "PATCH"].includes(this.method) ? 'application/json' : '', 'attribute': 'contentType' })
-      this.contentType = ["POST", "PUT", "PATCH"].includes(this.method) ? 'application/json' : '';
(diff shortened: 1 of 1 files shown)


liyasthomas:
Hi @breno-pereira thanks for the quick PR!

I've already gone through this use case once: `params` set the request body on a `POST`, and the query string on anything else.

But later, I came in front of API endpoints that want to use a query string and a request body at the same time. I don't know the extensibility of such use cases. Anyways this seems like a required feature now, so let's discuss over it.

NBTX:
I think the key problem here is just confusion over what the section does - and given its true purpose, it definitely does not make sense for `bodyParams` to equal `params` for POST requests - even if it does for requests without a body.

My proposal is that we land https://github.com/liyasthomas/postwoman/pull/189 and just show the Parameters section exclusively for HTTP verbs where a request body should be supplied and have URL/querystring parameters supplied in the URL bar OR add a new section for query parameters and keep what we currently have exclusively for body parameters.

liyasthomas:
Closing stale PRs

PR #231, opened 2019-10-30
Feature/pre request script

### LINKED ISSUE #218: Implement pre-request and post-request scripts (and request chaining)
Support for pre-request and post-request scripts should be added.

- [x] pre-request #231 
- [x] post-request

Essentially this means a user should be able to enter some JavaScript code they want executed before a request to generate values - e.g. a random UUID for a request header.

Additionally, a user should be able to enter some JavaScript code they want executed *after* a request *either* for the purposes of chaining requests (i.e. automatically setting parameters for a request to data returned from a previous request) *or* for the purposes of presenting data.

This should work like Postman's [Pre-Request Scripts](https://learning.getpostman.com/docs/postman/scripts/pre_request_scripts).

nickpalenchar:
fixes #218 

scrapped the old branch `feature/dynamic-headers` after merge woes (will delete soon).

Lots more to think about, but I feel it satisfies mvp (and would love to get this in master so I can take a break from keeping it up-to-date manually 😅)

Request section has a `code` button. When clicked, it enables pre-request scripts. A pre-request script section appears above where javascript can be entered.

pre-request script has access to a postwoman api object (`pw`) that can be interacted with. Right now the only thing to do is `pw.environment.set("key", "value")` to set a variable that can be used in areas of the request. Note this is native javascript, so any js method can be used. Final values are all treated as strings.

Test by setting something in the pre-request script `pw.environment.set("route", "/api/users");` and replacing parts of the request within double angle brackets (path = `<<route>>`). Response should function the same as if the final value was statically typed.

Additionally, requests made this way end up in history with the same `code` icon. Hovering over it informs the user that this entry in the history used pre-request scripts. Due to it's highly dynamic potential, you could say that note serves as a warning!

tangentially -- @NBTX can you make the request-chaining into its own issue? I realized it's quite separate from this one so it would be good to address that after this.

### DIFF
--- a/components/history.vue
+++ b/components/history.vue
@@ -12,6 +12,7 @@
       </li>
     </ul>
     <ul>
+      <li style="max-width: 44px"></li>
       <li @click="sort_by_label()">
         <label class="flex-wrap">
           Label
@@ -50,6 +51,12 @@
       :remain="Math.min(5, filteredHistory.length)"
     >
       <ul v-for="(entry, index) in filteredHistory" :key="index" class="entry">
+        <li style="width: 44px">
+          <button v-if="entry.usesScripts"
+            v-tooltip="'This entry used pre-request scripts.'"
+            class="icon"
+          ><i class="material-icons" style="z-index: 10050;">code</i></button>
+        </li>
         <li>
           <input
             aria-label="Label"
--- a/functions/preRequest.js
+++ b/functions/preRequest.js
@@ -1,0 +1,17 @@
+export default function getEnvironmentVariablesFromScript(script) {
+  let _variables = {};
+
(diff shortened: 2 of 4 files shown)


liyasthomas:
Works good!
I like the indicator in the history section 👍 

### `#todo` - after merging to master, I want this to be in production ASAP for more beta testing
- Toggling pre-script section **off** will remove already binded properties.
- `useHistory()` on history entries with pre-scripts will retain those pre-scripts. We prolly need to store pre-scripts too to history.

nickpalenchar:
Agreed on the post-merge todos! Additionally, we could also indicate this feature is not fully stable as-is. Something as simple as changing the header to `Pre-Request Script (experimental)` should be all that’s needed to sufficiently warn users.

PR #246, opened 2019-11-03
Make page changes more fluid

NBTX:
A very simple tweak; by making the pages a minimum height of 100vh, it makes the app feel smoother as the footer can be scrolled into view on any page and the app does not 'jump' in height when pages are changed.


liyasthomas:
*It's a simple spell but quite unbreakable.gif*

PR #258, opened 2019-11-04
Fix all contributors badge in README

NBTX:


NBTX:
Nevermind, I read up on how to change the template. Will re-open with a config change.

PR #276, opened 2019-11-07
Fix typo in README.md

genk1:
# Fixed Typo

## Before
``` _Histories can deleted one-by-one or all together_ ```

## AFTER
``` _Histories can be deleted one-by-one or all together_ ```


liyasthomas:
Will be fixed in next commit

PR #282, opened 2019-11-07
Enhancement: make saving of collections, requests, folder more convenient

elonmir:
It bugged me that I had to hit the save button for collections, requests etc manually with the mouse, so I quickly added the correct event handler for the enter key.


elonmir:
> LGTM. And this is v.handy 🙌🏼

Yep it is, and let me say, I just started using vue a few weeks ago :D So I am glad if it is a useful suggestion!

PR #292, opened 2019-11-09
Replace highlight.js with ace editor.

hosseinnedaee:
In this PR I removed highlight.js and implemented ace editor that support textarea highlighting.

resolving https://github.com/liyasthomas/postwoman/issues/266


pages/index.vue
```diff
@@ -568,14 +568,13 @@
...
-  ref="responseBody"
-  id="body"
-  rows="16"
-  placeholder="(waiting to send request)"
->{{response.body}}</code>
-            </pre>
+            <ResponseBody
+              :value="responseBodyText"
+              theme="dracula"
+              lang="json"
```
NBTX: What if the response body is not JSON?

liyasthomas:
@hosseinnedaee Yes, that's easily possible.
use `editor.session.setMode();`

refer:
- https://codepen.io/callmeaatrey/pen/dGPPjE
- https://ourcodeworld.com/articles/read/310/how-to-set-ace-editor-mode-according-to-the-file-extension-of-a-file

pages/index.vue
```diff
@@ -568,14 +568,13 @@
...
-              <code
-  ref="responseBody"
-  id="body"
-  rows="16"
-  placeholder="(waiting to send request)"
->{{response.body}}</code>
-            </pre>
+            <ResponseBody
+              :value="responseBodyText"
+              theme="dracula"
```
liyasthomas: making `theme` a prop is better since, while changing theme to light mode, we could fire up a light theme for the editor 👍

assets/css/styles.scss
```diff
@@ -517,6 +517,10 @@ pre {
   display: grid;
 }
 
+pre.ace_editor {
+  z-index: 0;
```
liyasthomas: `font-family: Roboto Mono,monospace;`

pages/index.vue
hosseinnedaee: the ace editor `setValue` method just gets _String_, so regardless of response type before passing the value to `ResponseBody` we convert it to _String_ .
the `lang`  determines the highlight syntax and how to show the passed string value.
the `lang` also check syntax and show error if it's wrong(but in response body I disabled it because it's just a readonly api response)

pages/index.vue
hosseinnedaee: absolutely👍
which theme is good for light mode?
[http://ajaxorg.github.io/ace-builds/kitchen-sink.html](http://ajaxorg.github.io/ace-builds/kitchen-sink.html)

liyasthomas: XCode 😊

hosseinnedaee: I think there are three options
1. we can use the selected PW theme in setting and based on that choose the proper theme for ace editor:
- kanda dark -> dracula
- clearly white -> xcode
- just black -> ???
2. use theme prop for pass any ace available themes
3. both above options. use selected PW theme to define ace theme but if a theme prop exist prefer that instead PW theme.

what do you think?

liyasthomas:
Providing both options is great! But i dont compel to implement both at once. Anyway, it's your feature branch 😊

liyasthomas:
I think we should land this to production and add more theming afterwards. @hosseinnedaee what do you think?

hosseinnedaee:
> I think we should land this to production ASAP and add more theming afterwards. @hosseinnedaee what do you think?
> 
> Will merge to master tomorrow.

I'm working on it, and hope some commits next hours.
anyway, we'll merge tomorrow.

pages/index.vue
hosseinnedaee: @NBTX I made `lang` prop dynamic, changing by response type.

hosseinnedaee:
Now ace editor theme change by PW changing.

PR #293, opened 2019-11-09
Icon and placement improvements

NBTX:
This improves the placement and choice of some of the icons. For example, the websocket icon 'cloud' has been replaced with 'power' which resembles a plug (more commonly associated with socket) and the twitter link has been moved to the top right with the GitHub link.

image
image


liyasthomas:
@NBTX got few broken UI on mobile. I'll rectify and merge.
Moving twitter button to top nav is not a better move. I put twitter at footer and GitHub at top so users will get the prime focus on GitHub

HTTP icon is so small. Not our fault, it's comparitively smaller in material icons. 
And Alien head icon now became our brand icon. I won't change that from http page to logo.

Everything else seems good.

NBTX:
May I propose centering the Twitter icon in the footer then? At least when the 'Install PWA' button is not present it looks a little odd in the current placement.

The HTTP icon could be enlarged with CSS. I do prefer the 'HTTP' icon as it makes the app more intuitive and the brand logo is not lost because it can be added to the header text.

liyasthomas:
Center Tweet logo ✔
Logo to HTTP icon 👎  (trust me. people like the logo in that place than a usual icon)

NBTX:
Okay.

PR #341, opened 2019-11-26
GraphQL Request Headers

AndrewBastin:
Added support for headers in GraphQL requests.

### DIFF
--- a/pages/graphql.vue
+++ b/pages/graphql.vue
@@ -315,15 +315,22 @@ export default {
           query: gql.getIntrospectionQuery()
         });
 
+        let headers = {};
+        this.headers.forEach((header) => {
+          headers[header.key] = header.value;
+        });
+
         const reqOptions = {
           method: "post",
           url: this.url,
           headers: {
-            ...this.headers,
+            ...headers,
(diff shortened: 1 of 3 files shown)


pages/graphql.vue
```diff
@@ -224,6 +319,7 @@ export default {
           method: "post",
           url: this.url,
           headers: {
+            ...this.headers,
```
liyasthomas: I couldn't spot headers on network tab, instead its index are appended 😄 

to reproduce:
0. Added few headers and its values
1. Get Scheme
2. Check request headers and see only its index are appended. Ex: 0, 1, 2...

Screen Shot 2019-11-26 at 9 39 47 AM

AndrewBastin:
Oops...

Rookie JS mistake...

Gimme 5 mins 😅

pages/graphql.vue
AndrewBastin: @liyasthomas I fixed it in a later commit...

Check it out...

liyasthomas:
@AndrewBastin 💯 you're free to merge.

PR #343, opened 2019-11-26
refactor: minor improvements

jamesgeorge007:
- Replace all occurrences of `var` with `const`
- Self closed components. (Vue.js best practices)
- Use `Array.includes()` instead of repeated conditional checks
- Make use of directive shorthand notations for consistency (Vue.js best practices).

### DIFF
--- a/pages/index.vue
+++ b/pages/index.vue
@@ -1205,7 +1205,7 @@ export default {
     },
     requestCode() {
       if (this.requestType === "JavaScript XHR") {
-        var requestString = [];
+        const requestString = [];
         requestString.push("const xhr = new XMLHttpRequest()");
         const user = this.auth === "Basic" ? "'" + this.httpUser + "'" : null;
         const pswd =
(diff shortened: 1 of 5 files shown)


pages/index.vue
```diff
@@ -1259,14 +1259,14 @@ export default {
         }
         return requestString.join("\n");
       } else if (this.requestType == "Fetch") {
-        var requestString = [];
-        var headers = [];
+        const requestString = [];
+        const headers = [];
```
liyasthomas: @jamesgeorge007 
```
Error: "headers" is read-only
```

Ran into this error on Generate Code modal

To reproduce:

0. Visit preview deployment: https://deploy-preview-343--postwoman.netlify.com
1. Make a change / send a request
2. Click "Generate Code"
3. Switch Request Type to `Fetch`
4. Other two (JavaScript and cURL) Request Types works fine. But Fetch code is not generated, and ran into error. Works on https://postwoman.io

jamesgeorge007: :+1:

PR #347, opened 2019-11-27
fix: SSE onclose handle

FatDoge:
Actually the sse doesn't have the onclose handle, call it before sse closed.


liyasthomas:
This was solved in #351

PR #350, opened 2019-11-27
Undo header/param/body param deletion

AndrewBastin:
Added the ability to undo header/param/body param deletion.

NOTE : Because toasts are for some reason not appearing in `index.vue`, I haven't tested the code used in that page, but it works in the GraphQL page. I am working really late, so I might have made some mistakes, reviewers, please do check :stuck_out_tongue:

### DIFF
--- a/pages/graphql.vue
+++ b/pages/graphql.vue
@@ -604,10 +604,22 @@ export default {
       return false;
     },
     removeRequestHeader(index) {
+      // .slice() is used so we get a separate array, rather than just a reference
+      const oldHeaders = this.headers.slice();
+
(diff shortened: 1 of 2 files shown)


liyasthomas:
There seems to a bug on toast in `index.vue` . Actually toast fired and as soon ait open, it closed. That's what happened. I couldn't find why it happened. Anyway, will merge this since there's no other problems.

liyasthomas:
I think the reason why toast closed quickly may be the `header`, `payload params` and `request params` are hidden at initial render and opened on some action which causing the mounted() to act async. Anyway I need to investigate more into it.

PR #366, opened 2019-11-28
I18n

### LINKED ISSUE #336: Internal server environment usage requirements
When I deploy on an internal server，I noticed that the postwoman request ```css``` and ```icon``` from ```fonts.googleapis.com```, but the internal server can not request the resources,so the layout become confusing.

Wish the all resources can be used offline.

Just copy the ```css``` and ```icon``` from network ,put them into the ```dist``` and rewrite the code where import the other internet resources.

**Additional context**

liyasthomas:
fixes #336


lang/fa-IR.js
```diff
@@ -1,3 +1,83 @@
 export default {
```
NBTX: Should we even keep this file if it's not translated?

lang/pt-BR.js
```diff
@@ -1,3 +1,83 @@
 export default {
```
NBTX: Again about including this file - or is there a list somewhere of what needs to be translated?

liyasthomas: I've already pinged the corresponding contributors, we may have to keep few limits / guidelines for i18n in near future.

PR #397, opened 2019-12-04
Fix need manual reload after switch lang

codeudan:
Fix need manual reload after switch lang. Now it will auto reload after switch lang, so user will see that immediately.


liyasthomas:
`location.replace(this.switchLocalePath(langCode));` is not the preferred way to switch locale. And this will reload the tab.

>Fix need manual reload after switch lang.

Currently, switching language doesn't need to manually reload o take effect changes.

PR #400, opened 2019-12-04
Custom methods support

### LINKED ISSUE #398: Custom request method
Some APIs use custom request methods. For example the API of vault uses `LIST` do deliver secrets:
https://www.vaultproject.io/api/secret/kv/kv-v2.html#list-secrets

It would be great to have a custom option that opens a text field where one can put in strings that describe the method.

Another solution is to just replace the dropdown with a text field.

I think for such APIs the only alternative is to use a different tool, but I really like postwoman :)

liyasthomas:
fixes #398

### DIFF
--- a/assets/css/styles.scss
+++ b/assets/css/styles.scss
@@ -706,9 +706,9 @@ ol li {
 }
 
 #send {
-  #hidden-message {
-    display: none;
-  }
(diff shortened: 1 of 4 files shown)


liyasthomas:
Retry DeepCode

PR #410, opened 2019-12-08
Fixing bug on request saving

### LINKED ISSUE #390: Save to collections after deleting all the collections causes an error page
**Describe the bug**
When trying to save to collection, after deleting all the existing collection throws 

> this.$store.state.postwoman.collections[this.$data.requestData.collectionIndex] is undefined

**To Reproduce**
1. Save a request to an existing collection.
2. Delete all the existing collections
3. Click on Save to Collections
4. Error is seen, in addition to that neither "Go Home" or "Reload" does not work.

**Expected behavior**
Either say that there is no collection OR Save without a collection OR prompt for new collection

**Screenshots**

image

 - OS: Windows 10
 - Browser: Firefox 
 - 70.0.1

adevr:
This bug appeared when saving a request with no collections available.
Fixes #390

### DIFF
--- a/components/collections/saveRequestAs.vue
+++ b/components/collections/saveRequestAs.vue
@@ -128,7 +128,11 @@ export default {
       const userSelectedAnyCollection =
         this.$data.requestData.collectionIndex !== undefined;
       if (!userSelectedAnyCollection) return [];
-
+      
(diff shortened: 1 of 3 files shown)


adevr:
#390

PR #416, opened 2019-12-10
Update id-ID.js

williamsp:
Fix untranslated "proxy"


williamsp:
Hi, thank you for approving my commit.
But, it seems some string is still not available on the lang file keys.
For example strings from History and Collection on the right panel.

liyasthomas:
Oh. Would you like to add those?

williamsp:
Once the keys are already included in the lang file, I could help you translating the values.

PR #440, opened 2019-12-18
I18n attribute added for Spanish language

JOsktgui:
- An attribute called `status_message` was added in the languages
- `I18n` branch updated with `master`

image


liyasthomas:
Will update with remaining locales later.

PR #476, opened 2020-01-07
OAuth 2.0/OIDC Access Token Retrieval Support

### LINKED ISSUE #337: Feature Request: Support OAuth2/OIDC
Currently, Postwoman supports Bearer Token authentication, but [the process of obtaining a token for OAuth2 is non-trivial for any real-life APIs](https://www.oauth.com/oauth2-servers/access-tokens/authorization-code-request/). 

In an ideal world, Postwoman would support OIDC for endpoint discovery, and Code Grant Types with PCKE (sorry, I'm not 100% sure on the correct terminology as different providers end up relabelling these this - see KeyCloak and IdentityServer4 for examples). It would also be great if there was support for Refresh Tokens, and JWT token introspection to support this.
The gold standard would probably also support OIDC Dynamic Client Registration, but I'm not sure how widly this is supported.
In addition, support for [OIDC Front Channel Logout](https://openid.net/specs/openid-connect-frontchannel-1_0.html) would be amazing.

At a minimum, OAuth2 authentication with Implicit Grants would support my use case (internal APIs), but will probably not support public APIs from Google or Facebook.

Postman has [support for OAuth2 with Code, Password and Client Credential grant types](https://learning.getpostman.com/docs/postman/sending-api-requests/authorization/#oauth-20), but it's annoying to use as:

  * Only the last settings are remembered, and so overwritten if you have to change them. If you're authenticating against different endpoints or with different credentials, this quick becomes unmanageable.
  * It doesn't support Refresh Tokens out of the box, so, for real-life APIs, you're left manually refreshing tokens all the time. 

**Additional context**

There's probably security implications of allowing Postwoman.io as a valid redirect_url for OAuth2 secured apps, but I don't really feel qualified to discuss this.

reefqi037:
This PR adds OAuth2.0/OIDC Access Token retrieval function to Postwoman 🎉 

### Main Feature
With this feature, OAuth 2.0 access token can be retrieved directly from inside Postwoman.
Since Postwoman is a web-based SPA, it have no way to hide client secret used in the normal OAuth 2.0 flows. Therefore, only **Authorization Code Flow + PKCE** is implemented.

### Additional Features
- [x] Automatic configuration using OpenID Connect Discovery endpoint ( Configure endpoints automatically! )
- [x] Access Tokens Management ( Save, reuse, and delete multiple access tokens )
- [x] Token Request Management ( Save, reuse and delete multiple authorization endpoints configuration )

### Screenshots
1. Main UI
oauth

2. Access Token Management UI
token-management

3. Token Request Management UI
token-req-management

### Limitations
- Users can only use OAuth Service Providers that provide PKCE support for client authentication.
- Users have to register [Postwoman](https://postwoman.io) as callback URL and trusted/allowed origin (CORS) in the OAuth application and Service Providers' settings. 

There should be room for improvements but for the time being, I think the core functions here should works ( tested with Okta and Auth0 as Service Providers ). Feel free to suggest and improve 👨‍💻 👩‍💻 🚀 

fixes #337


liyasthomas:
@reefqi037 I did few refactoring to `oauth.js` in https://github.com/liyasthomas/postwoman/commit/57f7621567d6182790142fd472e31e4fb9392297 can you please verify it. Usually i make dump mistakes 🤦‍♂

reefqi037:
@liyasthomas Yeah I have seen it and made some tests. No 🐛 found at the moment. All good! 👍

liyasthomas:
Thanks for the quick response

reefqi037:
No problem!

Bettelstab:
@reefqi037 

While I really appreciate your work (I'm just about to give it a try) I disagree with what you said about Authorization Code Flow + PKCE being the only possible flow. In my understanding, the reason you can't use a client secret in usual SPAs is that you would have to embed it in the code, so it would be accessible by the public, as you said.

In case of Postwoman, however, there is not that one client secret, but each user could add her own and save it client-side. Nobody else will ever see it. Should work perfectly fine.

PR #484, opened 2020-01-11
Porting (most of) code to typescript

### LINKED ISSUE #467: What about additional loaders: + Pug, TypeScript, SASS, material-vue ?
It may be good add some productivity tools to tool chain 

What about additional loaders: + Pug, TypeScript, SASS, material-vue ? All of these are proven good with both, Vue and Nuxt. I see no point to make branch on this. It is question of one npm -D.

Anything here is not a question at all, excepting vue-material support is newer than vuetify's, and later  is more nuxt-ish and have longer nuxt track/ this can be point of discussion but both are stable and  no very difference which will be accepted. All of this save on typing with no addition obligations.

Not if no. Let if yes

AndrewBastin:
**This PR intends to complete the Typescript work which began in the `typescript` branch.**

It ports every code file (except tests and pages/index.vue) to TypeScript with as much types provided.

There are certain things to note about the TypeScript addition to the project though.

### There is a change in the way components are imported.
Components before could be imported with this snippet
```js
import Section from "../components/section";
```
But from this PR on, it will be:
```js
import Section from "../components/section.vue";
```
For Vue components, you have to include the file extension.

### Default Vue component code is modified a bit
Vue component scripts are used to be written as
```js
<script>
export default {
...[[ Component Code ]]...
};
</script>
```
From this PR onwards it will be
```js
<script lang="ts">
import Vue from "vue";

export default Vue.extend({
...[[ Component Code ]]...
});
</script>
```
**NOTE : It is also best to include an empty component script even if the component has no code logic, sometimes the compiler doesn't pick up the Vue component otherwise**

### Directives and Middlewares have a new code format
The way directives and middlewares are written have been revamped to work with TypeScript, please check the code in the corresponding folders, to see how it is written in the new format 

## Some notes on the TypeScript and Vue interaction
- I noticed while porting that TypeScript compiler will freak out if it can't resolve/infer types from the component code, so if you find TypeScript giving you weird errors like (`foo is not a property of this`), try explicitly specifying types, that will usually resolve that sorta issues.

- While adding plugins, make sure to import them into vue-plugins.d.ts file and add it to the import declaration, sometimes, it will be automatically be picked up (its a bit finicky), if you still can't access the plugin, just add the plugin reference to the Vue interface present in the same file.

Now, this PR doesn't bring full TypeScript port, pages/index.vue is due a huge refactor and it will be rewritten in TypeScript during that process. Also, tests will still be written in JavaScript.

Also, another thing to note is that, I haven't fully brought types as much as possible, some places could use more typings (and less `any`), particularly the types regarding the store are implemented with `any` as a placeholder. This needs to be refactored into more meaningful type definitions in the future.

Soooo, that's about it,

Happy hacking :smile: 

This PR may fix #467


AndrewBastin:
@jamesgeorge007 I have brought back the dynamic imports in the recent commit e046f68 

Turns out somewhere in the configuring of the typescript and porting, I fixed the dynamic import issue :stuck_out_tongue:

AndrewBastin:
Umm, can I merge this to the typescript branch, before it gets more outdated ?

liyasthomas:
@AndrewBastin yes

PR #486, opened 2020-01-11
chore(oauth): Added method signatures as per JSDoc conventions

jamesgeorge007:
Follow up of #485 #476


assets/js/oauth.js
```diff
@@ -24,17 +30,30 @@ const sendPostRequest = async (url, params) => {
...
+ */
+
+const parseQueryString = searchQuery => {
+  if (searchQuery === "") {
     return {};
   }
-  const segments = string.split("&").map(s => s.split("="));
+  const segments = searchQuery.split("&").map(s => s.split("="));
   const queryString = {};
   segments.forEach(s => (queryString[s[0]] = s[1]));
```
jamesgeorge007:
```js
  const queryString = segments.reduce(
    (obj, el) => ({ ...obj, [el[0]]: el[1] }),
    {}
  );
```

jamesgeorge007: @liyasthomas what do you think about the above approach?

jamesgeorge007: Does it make the code hard to read:question:

liyasthomas: I thought `forEach` made more sense but going through `reduce` implementation, I think devs prefer `reduce` over `forEach` to make new object. So I :+1:  that.

liyasthomas:
Do not merge, there is an error.

PR #487, opened 2020-01-14
Network Strategies

AndrewBastin:
This PR intends to introduce Network Strategies into Postwoman.

Network Strategies are basically functions which take in the Postwoman Store and the request config and returns a Promise for the response. It generalizes the network request making process and makes further introduction of features easier.

Currently, this PR introduces 2 Network Strategies

- **AxiosStrategy** : Uses axios to send the request
- **ProxyStrategy** : Uses the postwoman proxy feature to send the request through a proxy.

Currently, Network Strategies allow its consumers to not worry about whether proxy is enabled, the `sendNetworkRequest` (found in functions/network.js) function will detect the appropriate strategy from the available and executes that. But in the future, when we plan to add new strategies (for e.g for native), we do not need to worry about rewriting the consumers to adopt those.


liyasthomas:
Axios progress bar seems to be broken.

To reproduce:
1. Visit deploy preview: https://deploy-preview-487--postwoman.netlify.com/
2. Send a request
3. Progress won't finish even after axios fetch response.

Except this, I couldn't find any bugs.

AndrewBastin:
Well, the AxiosStrategy doesn't use the `$axios` plugin thingy. Does the progress bar depend on that or something ? If so, we will need to fix that.

liyasthomas:
Well axios automatically binds progress to Nuxt's progress bar. We can turn that off, and invoke progress manually. Lemme find a fix for that.

functions/network.js
```diff
@@ -0,0 +1,13 @@
+import AxiosStrategy from "./strategies/AxiosStrategy";
+import ProxyStrategy from "./strategies/ProxyStrategy";
+
+
+const sendNetworkRequest = (req, store) => {
+  if (store.state.postwoman.settings.PROXY_ENABLED) {
+    return ProxyStrategy(req, store);
+  } else {
+    return AxiosStrategy(req, store);
+  }
```
jamesgeorge007: Avoid the `else` block.

```js
  if (store.state.postwoman.settings.PROXY_ENABLED) {
    return ProxyStrategy(req, store);
  }
  return AxiosStrategy(req, store);
```

functions/strategies/ProxyStrategy.js
```diff
@@ -0,0 +1,13 @@
+import axios from "axios";
+
+const proxyStrategy = async (req, store) => {
+  const proxyRes = await axios.post(
```
jamesgeorge007:
```suggestion
  const { data } = await axios.post(
```

functions/strategies/ProxyStrategy.js
```diff
@@ -0,0 +1,13 @@
...
+    store.state.postwoman.settings.PROXY_URL ||
+    "https://postwoman.apollotv.xyz/",
+    req
+  );
+
+  return proxyRes.data;
```
jamesgeorge007:
```suggestion
  return data;
```

functions/strategies/ProxyStrategy.js
jamesgeorge007: I'm a big fan of the object destructuring approach. It's quite subjective :sweat_smile:

functions/strategies/ProxyStrategy.js
AndrewBastin: Updated in the latest commit

functions/strategies/ProxyStrategy.js
AndrewBastin: Updated in the latest commit

functions/network.js
AndrewBastin: Updated in the latest commit

AndrewBastin:
@liyasthomas what about the progress bar issue ?

liyasthomas:
Fix coming in few mins

PR #510, opened 2020-01-21
Validate duplicate collections

### LINKED ISSUE #509: Multiple collections with the same name shouldn't exist
Assume there is a default collection named 'My Collection'. If he/she makes an attempt to create yet another collection under the same name it just happens which isn't intended.

jamesgeorge007:
closes #509

### DIFF
--- a/store/postwoman.js
+++ b/store/postwoman.js
@@ -111,6 +111,10 @@ export const mutations = {
   },
 
   addNewCollection({ collections }, collection) {
+    const duplicateCollection = collections.some(item => item.name === collection.name)
+    if (duplicateCollection) {
+      alert('Duplicate collection');
+    }
     collections.push({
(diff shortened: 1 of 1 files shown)


store/postwoman.js
```diff
@@ -111,6 +111,12 @@ export const mutations = {
   },
 
   addNewCollection({ collections }, collection) {
+    const { name } = collection
+    const duplicateCollection = collections.some(item => item.name === name)
+    if (duplicateCollection) {
+      alert('Duplicate collection');
```
jamesgeorge007: Have a better warning message in mind?

liyasthomas: Better is a toast.

jamesgeorge007: Done

PR #516, opened 2020-01-23
Validations for edit and create collections activity

### LINKED ISSUE #515: Warn the user if name field was left blank while creating a new collection
As of now, the user can create collection(s) leaving the name field blank which shouldn't be the case.

jamesgeorge007:
closes #515 
Follow up of #510

### DIFF
--- a/components/collections/addCollection.vue
+++ b/components/collections/addCollection.vue
@@ -57,7 +57,7 @@ export default {
   methods: {
     addNewCollection() {
       if (!this.$data.name) {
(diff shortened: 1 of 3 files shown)


liyasthomas:
Check for duplicates?

jamesgeorge007:
> Check for duplicates?

#510 has already landed which brought about validations in that regard.

liyasthomas:
@jamesgeorge007  there's  one edge case where user edits a collection name. Please add non-empty and duplication check on there too.

PR #517, opened 2020-01-23
Update asset/*. js to  asset/*. ts.

Sn005:
This PR is ./assets/js/*.js converted to * .ts.

Specifically, it is the following file.

- curlparser.ts
- oauth.ts
- pwa.ts

supplement.

How to convert to TypeScript.

If you make convert to TypeScript at once, you will have the following problems.
- Checks take a long time due to a large number of update files
- The Difference from The latest version occurs

So, I want to put out PR in detail for each directory.


AndrewBastin:
I will review this and get back to you.

PR #525, opened 2020-01-25
Add nl-BE translation

Jefwillems:
I added a file and entry in config for the nl-BE language. Almost everything is translated, except some words really don't make sense when you translate them to dutch, so i left those unchanged from english.


liyasthomas:
Hiya! Thanks for the tranlations, but `i18n` PRs are only accepted to [`i18n` branch](https://github.com/liyasthomas/postwoman/tree/i18n).

Feel free to close this PR and make the same to i18n branch 👋

Jefwillems:
Whoops, should've done that! I will!

PR #518, opened 2020-01-24
Feature/post request tests

nickpalenchar:
Closes #465 

Want to get this in while it's still in sync with master. Minimal but functional and ready to be used 😄 

The tests button is next to the pre-request script button, which opens a window to code tests. A commented out line given an example of a (failing) test. Comment it in to test the test. Other examples are documented [here](https://github.com/liyasthomas/postwoman/wiki/Post-Requests-Tests), and can all be tried.

Biggest todo still is saving to history. Will do later this weekend.

The UI is minimal and... a little funky.. could def use some formatting help by someone better at styling than I am!

In terms of functionality+docs I'll add a bit more incrementally once merged. This was a really fun one!
If someone else merges I recommend squashing and merging.. I made a lot of incremental commits.

PR #531, opened 2020-01-27
Enhancements

jamesgeorge007:
Follow up of #518 

- Typo fixes.
- Couple of other tweaks.


liyasthomas:
I suspect @nickpalenchar planned on a nested tests kinda for this. At least those naming looks like that. Anyway only merge in Nick's review approval.

PR #533, opened 2020-01-29
Modification of French translations

thomasbnt:


liyasthomas:
`i18n` translations PR should be based to i18n branch. Closing this and opening the same for i28n branch.

PR #543, opened 2020-01-31
Refactor

liyasthomas:
- Use template literals.
- Prefer Object destructuring assignment approach.
- `let` to `const`.
- Other minor tweaks.

### DIFF
--- a/components/ace-editor.vue
+++ b/components/ace-editor.vue
@@ -74,7 +74,9 @@ export default {
 
   methods: {
     defineTheme() {
-      if (this.theme) return this.theme;
+      if (this.theme) {
+        return this.theme;
+      }
       return (
         this.$store.state.postwoman.settings.THEME_ACE_EDITOR || DEFAULT_THEME
       );
--- a/components/ace-editor.vue
+++ b/components/ace-editor.vue
@@ -55,8 +55,8 @@ export default {
 
   mounted() {
     const editor = ace.edit(this.$refs.editor, {
-      theme: "ace/theme/" + this.defineTheme(),
-      mode: "ace/mode/" + this.lang,
+      theme: `ace/theme/${this.defineTheme()}`,
+      mode: `ace/mode/${this.lang}`,
       ...this.options
     });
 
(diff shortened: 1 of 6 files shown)


components/ace-editor.vue
```diff
@@ -74,13 +74,10 @@ export default {
...
-        return this.theme;
-      } else {
-        return (
-          this.$store.state.postwoman.settings.THEME_ACE_EDITOR || DEFAULT_THEME
-        );
-      }
+      if (this.theme) return this.theme;
+      return (
+        this.$store.state.postwoman.settings.THEME_ACE_EDITOR || DEFAULT_THEME
+      );
```
jamesgeorge007:
```js
if () {
  //
}
return ...
```

Having a consistent style across the codebase is really important. What do you think?

pages/index.vue
```diff
@@ -1885,54 +1883,33 @@ export default {
       if (this.requestType === "JavaScript XHR") {
         const requestString = [];
         requestString.push("const xhr = new XMLHttpRequest()");
-        const user =
-          this.auth === "Basic Auth" ? "'" + this.httpUser + "'" : null;
+        const user = this.auth === "Basic Auth" ? `'${this.httpUser}'` : null;
         const pswd =
-          this.auth === "Basic Auth" ? "'" + this.httpPassword + "'" : null;
+          this.auth === "Basic Auth" ? `'${this.httpPassword}'` : null;
```
jamesgeorge007: Ig we can go with the `if/else` approach here. Also regarding the names, `pswd` doesn't sound that good.

pages/index.vue
```diff
@@ -2295,9 +2264,8 @@ export default {
       }));
     },
     pathInputHandler() {
-      let queryString = this.getQueryStringFromPath(),
-        params = this.queryStringToArray(queryString);
-
+      let queryString = this.getQueryStringFromPath();
```
jamesgeorge007:
```suggestion
      const queryString = this.getQueryStringFromPath();
```

pages/index.vue
```diff
@@ -2295,9 +2264,8 @@ export default {
...
+      let params = this.queryStringToArray(queryString);
```
jamesgeorge007:
```suggestion
      const params = this.queryStringToArray(queryString);
```

pages/index.vue
liyasthomas: `queryString`'s scope is only required inside that block, so let's go wil `let`

PR #546, opened 2020-02-02
update e2e tests

yubathom:
Some updates to e2e tests:
 * fix error of dependency start-server-and-test update (`^1.10.8`) #544
 * add test to proxy feature
 * update test to url queries feature (test to add/remove passwords at url with the toggler)
 * use nuxt `build/start` to run tests instead of `nuxt` 
> Some tests were failing because the vm of the ci was loading the assets (and the page was not ready to run the first test). This is a screenshot from a test that runned at travis:

cypress-screenshot

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -98,3 +98,9 @@ sw.*
 
 # File explorer
 .directory
+
+# Tests screenshots
+tests/*/screenshots
+
+# Tests videos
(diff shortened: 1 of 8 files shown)


liyasthomas:
@yubathom  thank you for this PR.

PR #549, opened 2020-02-03
Show Ctrl instead of Command for shortcuts non-Apple platforms

AndrewBastin:
Well, this is pretty self explanatory and pretty small PR.

The keyboard shortcuts dialog will show the shortcuts with `Ctrl` instead of `⌘` when viewing the shortcuts on the non-Apple platforms.


liyasthomas:
Why I never think of this! 🤐🙏

PR #567, opened 2020-02-12
Add some missing Indonesian language

fikimaul:


liyasthomas:
`i18n`  PRs are only accepted to `[i18n](https://github.com/liyasthomas/postwoman/tree/i18n?files=1)

PR #571, opened 2020-02-15
:sparkles: Native share + updated meta description

liyasthomas:


layouts/default.vue
```diff
@@ -388,6 +388,15 @@
                         </svg>
                         <span>{{ $t("tweet") }}</span>
                       </button>
+                      <button
+                        v-if="navigatorShare"
```
jamesgeorge007: As of now, a toast shows up if the user's browser doesn't support Web Share API. We can either remove this conditional rendering stuff or get rid of the toast. WDYT?

liyasthomas: Remove the toast.

PR #572, opened 2020-02-15
fix: drop the toast which doesn't show up

jamesgeorge007:
Follow up of #571 

The `share` button seems to show up only if `Web Share API` is supported in the user's browser.

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -715,7 +715,7 @@ export default {
           .then(() => {})
           .catch(console.error);
       } else {
-      	this.$toast.info("Web Share API is not supported in your browser")
+      	// fallback
       }
     }
   },


layouts/default.vue
```diff
@@ -715,7 +715,7 @@ export default {
           .then(() => {})
```
jamesgeorge007: Some kind of acknowledgement would suit here. Say, logging to the console or maybe a toast.

jamesgeorge007: Also, it seems there're couple of best practices while making use of the `Web Share API`. Please refer this [blog post](https://alligator.io/js/web-share-api/)

liyasthomas: We don't have `canonical` URL for home page.
And regarding the "logging  or showing toast": well there's no need to notify the user on successful share event. I never find it useful.

PR #591, opened 2020-02-23
Environment Mangement

JacobAnavisca:
- [x] Add Environment
- [x] Edit Environment
- [x] Delete Environment
- [x] Sync Environments

- [x] Review


components/environments/editEnvironment.vue
```diff
@@ -0,0 +1,214 @@
...
+      </ul>
+    </div>
+    <div slot="footer">
+      <div class="flex-wrap">
+        <span></span>
+        <span>
+          <button class="icon" @click="hideModal">
+            {{ $t("cancel") }}
+          </button>
+          <button class="icon primary" @click="saveEnvironment">
```
NBTX: I notice you've re-used `saveEnvironment` for the edit environment modal - however this means that it forces you to enter the name again if you don't want to change the name (and it shows a 'Duplicate environment' toast when you do this.)

image

image
^ (*Upon entering the same name into the field, it does save however this message shows.*)

Would you please update the PR to fix this?

store/postwoman.js
```diff
@@ -74,6 +74,13 @@ export const state = () => ({
       requests: []
     }
   ],
+  environments: [
+    {
+      name: "My Env Variables",
```
NBTX: Can you rename this to 'My Environment Variables'?

store/postwoman.js
```diff
@@ -102,6 +109,63 @@ export const mutations = {
...
+
+  saveEnvironment({ environments }, payload) {
+    const { environment, environmentIndex } = payload;
+    const { name } = environment;
+    const duplicateEnvironment = environments.some(
+      item =>
+        item.environmentIndex !== environmentIndex &&
+        item.name.toLowerCase() === name.toLowerCase()
+    );
+    if (duplicateEnvironment) {
```
NBTX: This check is performed when you save an environment with a duplicate name but not when you create an environment with a duplicate name? Surely it should be both or neither?

store/postwoman.js
```diff
@@ -102,6 +109,63 @@ export const mutations = {
...
+
+  setVariableValue({ editingEnvironment }, { index, value }) {
+    editingEnvironment.variables[index].value = value;
+  },
+
+  removeVariable({ editingEnvironment }, variables) {
+    editingEnvironment.variables = variables;
+  },
+
+  addVariable({ editingEnvironment }, value) {
```
NBTX: See comment for `setVariableValue`

store/postwoman.js
```diff
@@ -102,6 +109,63 @@ export const mutations = {
...
+
+  setEditingEnvironment(state, value) {
+    state.editingEnvironment = { ...value };
+  },
+
+  setVariableKey({ editingEnvironment }, { index, value }) {
+    editingEnvironment.variables[index].key = value;
+  },
+
+  setVariableValue({ editingEnvironment }, { index, value }) {
```
NBTX: If at all possible, could we not wrap the variable in quotes unless it is in fact just a string?
Perhaps a way of testing this would be with JSON.parse and a try/catch.

```typescript
// Example 1:
// Returns "test"
testValue("test");

// Example 2:
// Returns 300 (as opposed to "300")
testValue("300");

// Example 3:
// Returns true (as opposed to "true")
testValue("true");

function testValue(myValue: string){
    try {
        return JSON.parse(myValue);
    } catch(ex) {
        // Now we know it's a string just leave it as a string value.
        return myValue;
    }
}
```

lang/en-US.js
```diff
@@ -45,6 +45,14 @@ export default {
   preview_html: "Preview HTML",
   history: "History",
   collections: "Collections",
+  environment: "Environment",
+  new_environment: "New Environment",
+  my_new_environment: "My New Environment",
+  edit_environment: "Edit Environment",
+  env_variable_list: "Variable List",
+  invalid_environment_name: "Please provide a valid name for the environment",
+  use_environment: "Use environment",
```
NBTX: To make this consistent, can you change this to "Use Environment" (title case)

components/environments/editEnvironment.vue
JacobAnavisca: I updated the pr to use the environments name as the value so it'll auto populate. I also updated the message handling so you shouldn't be seeing the duplicate message unnecessarily.

store/postwoman.js
JacobAnavisca: Added the duplicate name check for when you create an environment.

store/postwoman.js
JacobAnavisca: I left addVariable alone since it just automatically creates a basic empty string that can be updated to other types through the setVariableValue

store/postwoman.js
JacobAnavisca: pretty much just copy pasta'd this function to the postwoman store and set it up for the setVariableValue mutation

PR #594, opened 2020-02-24
POST request body editor reacts to the content type

AndrewBastin:
The Raw Request Body editor for POST requests is set to JSON mode and only provided the syntax highlighting mode for JSON. This PR intends to add the ability to for the user to update the syntax highlighting by updating the Content-Type.

So, for example, if the POST Content-Type is set to `application/xml`, the raw request editor updates the language mode to XML mode and if the request mode is set to `application/json`, it updates the mode to JSON Syntax Highlighting mode.

### DIFF
--- a/functions/editorutils.js
+++ b/functions/editorutils.js
@@ -1,0 +1,11 @@
+const mimeToMode = {
+  "text/plain": "plain_text",
+  "text/html": "html",
+  "application/xml": "xml",
+  "application/hal+json": "json",
+  "application/json": "json"
+}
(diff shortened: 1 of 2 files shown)


AndrewBastin:
@liyasthomas you should look into why the CI giving hick-ups all the time.... :roll_eyes:

liyasthomas:
it's an issue with CI itself. 😞

always pass on 2nd run. but don't know why this is happening. I'll issue a compliant.

PR #596, opened 2020-02-24
Feature: Add prettier/pretty-quick formatting w/ Husky pre-commit

dmitryyankowski:
Husky (https://github.com/typicode/husky) allows us to run commands before initial commits to the repo are made/PR's are pushed. In this scenario, husky runs `yarn prettier` to format all of the files within the repo before being staged. This is done in the `.huskyrc` file like so (this can be moved to the `package.json` file if we want):
```
{
  "hooks": {
    "pre-commit": "npm run prettier"
  }
}
```

I've used an opinionated Prettier config `.prettierrc` which look's like the following (this can be moved to the `package.json` file if we want):
```
{
  "trailingComma": "es5",
  "semi": false,
  "singleQuote": true,
  "printWidth": 100
}
```

The command that is added to the `package.json` is the following:
`"prettier": "prettier --write **/*.{html,js,json,vue}"`

This tells Prettier what files can be formatted.

**The reason for this PR is to clean up the code quality and formatting within the repo, to one that's consistent (updated by Prettier).**

**Note**: This does not affect how new code must be written or formatted. You can code/format in any way you would like, and Prettier will format it right before changes are committed to the repo/PR is made.


dmitryyankowski:
This is a substantial change. So discussion and updates are welcome :)

dmitryyankowski:
Not sure what's failing on the Netlify builds.. thought it might be the build.js file, but doesn't seem to be it.

dmitryyankowski:
Fixed 😃

liyasthomas:
@dmitryyankowski I guess we went with double quotes (`"`) for strings as a standard and better stick to it. so turn off `"singleQuote": true`

I would like to hear from @jamesgeorge007 on this whole PR 🤝

dmitryyankowski:
> @dmitryyankowski I guess we went with double quotes (`"`) for strings as a standard and better stick to it. so turn off `"singleQuote": true`
> 
> I would like to hear from @jamesgeorge007 on this whole PR 🤝

Damn the double quotes LOL ... haha no worries. Will make it double quotes!

jamesgeorge007:
Bit busy rn. Will take a look soon.

dmitryyankowski:
@jamesgeorge007 the reason i've added Prettier formatting is because of the different code styles for contributions pushed to the repo. Everyone has their own way of formatting; semi-colons, spaces, etc. Prettier will allow us to keep the repo consistent in code formatting/style. 💯

liyasthomas:
@dmitryyankowski sorry for the frequent PRs from upstream, we will merge this ASAP after review.

.huskyrc
```diff
@@ -0,0 +1,5 @@
+{
+  "hooks": {
+    "pre-commit": "npm run prettier"
```
jamesgeorge007: Please make sure that you add the respective script to `package.json` as well.

dmitryyankowski: Thanks for the suggestions! I saw `pretty-quick` before, but totally forgot it was a thing :D I've added it and updated it to run `pretty-quick` instead of `prettier` :) @jamesgeorge007

liyasthomas:
@dmitryyankowski 
> I've used an opinionated Prettier config .prettierrc which look's like the following (this can be moved to the package.json file if we want):

can you do me a favor? please consider moving the `.prettierrc` to `package.json` since there's already lot of config files in the root, it's getting outta control. You can push this along with #602

dmitryyankowski:
> @dmitryyankowski
> 
> > I've used an opinionated Prettier config .prettierrc which look's like the following (this can be moved to the package.json file if we want):
> 
> can you do me a favor? please consider moving the `.prettierrc` to `package.json` since there's already lot of config files in the root, it's getting outta control. You can push this along with #602

Done in https://github.com/liyasthomas/postwoman/pull/602 @ https://github.com/liyasthomas/postwoman/pull/602/commits/50339a2480205cd46c866aa314cb57d8d689b9c0 Moved both the `.prettierrc` and `.huskyrc` contents into the `package.json` file!

liyasthomas:
@all-contributors please add @dmitryyankowski for code

PR #598, opened 2020-02-24
Save GraphQL schema and response in state

dmitryyankowski:
Allows the user to go to any page, and back to the GraphQL page, without having to re-request the schema or response. This PR saves the schema and response into the `gql` state when they are requested.

**To test**: Go to the GraphQL page and request the schema or run a query. Go to any page, or refresh, and the schema and response should be there.


dmitryyankowski:
I'm only focusing on the GraphQL page at the moment. So unsure whether this could be applied to the index page, etc.

liyasthomas:
Hi @dmitryyankowski  there seems to be an issue with the CI that make tests fail without any reason. While I'm looking into it, you can reopen the PR.

dmitryyankowski:
Don't worry about it just yet! I realized I hadn't completed the new update. I forgot about syncing the types/queries/mutations on the right side. I will re-open once i'm fully finished :) Thanks @liyasthomas

PR #603, opened 2020-02-25
Added regex to handle url parts

### LINKED ISSUE #62: Fast URL entry

### LINKED ISSUE #581: Bigger URL and/or Path input field
I'm working on a project and I'm dealing with similar URLs of an API.
Postwoman only provides a small input field to add in the URL

image

I cannot properly see what URL I'm sending the request to.

Even the history section shows a small excerpt of the entire URL.

### Possible Solution

This is what postman does when thrown a big URL

image

JacobAnavisca:
Fix #62 
Fix #581

### DIFF
--- a/pages/index.vue
+++ b/pages/index.vue
@@ -1517,9 +1517,9 @@ export default {
   computed: {
     uri: {
       get() {
-        return this.$store.state.request.uri ?
-          this.$store.state.request.uri
-          : this.url + this.path;
+        return this.url + this.path ?
+          this.url + this.path
+          : this.$store.state.request.uri;
       },
       set(value) {
         this.$store.commit("setState", { value, attribute: "uri" })
--- a/pages/index.vue
+++ b/pages/index.vue
@@ -1493,6 +1493,7 @@ export default {
     selectedRequest(newValue, oldValue) {
       // @TODO: Convert all variables to single request variable
       if (!newValue) return
+      this.uri = newValue.url + newValue.path
       this.url = newValue.url
       this.path = newValue.path
       this.method = newValue.method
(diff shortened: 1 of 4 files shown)


liyasthomas:
@JacobAnavisca there seems to be a problem with parsing URL while passing `URL` via parameter. That's why tests are failing. I suggest you to run `npm run dev:e2e` for testing few use cases.

liyasthomas:
@JacobAnavisca pre-commit hooks have been introduced hence please rebase to current master and pull latest change, then send PR.

NBTX:
I don't think we should be using regex for this. JavaScript has a very mature URL API:
https://developer.mozilla.org/en-US/docs/Web/API/URL

JacobAnavisca:
main reason I'm using regex is to handle environment variables, like <<baseUrl>>/get

NBTX:
Could we substitute them before we try parsing?

JacobAnavisca:
How about regex as a fallback? like so:
`    uri: {
      get() {
        return this.$store.state.request.uri ?
          this.$store.state.request.uri
          : this.url + this.path;
      },
      set(value) {
        this.$store.commit("setState", { value, attribute: "uri" })
        let url;
        if (this.preRequestScript && this.showPreRequestScript) {
          const environmentVariables = getEnvironmentVariablesFromScript(this.preRequestScript);
          url = parseTemplateString(value, environmentVariables);
        }
        try {
          url = new URL(url);
          this.url = url.origin;
          this.path = url.pathname;
        } catch(error) {
          console.log(error)
          let uriRegex = value.match(
            /^((http[s]?:\/\/)?(<<[^\/]+>>)?[^\/]*|)(\/?.*)$/
          );
          this.url = uriRegex[1];
          this.path = uriRegex[4];
        }
      }
    }`

PR #605, opened 2020-02-25
JSON linting in the code editor

AndrewBastin:
This PR intends to add the ability for the Ace Code Editor to provide linting for JSON files in Postwoman.

You can try out the linter through any editor instance that specifies JSON as the language (so basically the GraphQL variable editor and the POST request content editor (when content type is set to `application/json`)).

### DIFF
--- a/components/ace-editor.vue
+++ b/components/ace-editor.vue
@@ -21,8 +21,10 @@
 <script>
 const DEFAULT_THEME = "twilight"
 
-import ace from "ace-builds"
-import "ace-builds/webpack-resolver"
+import ace from "ace-builds";
+import "ace-builds/webpack-resolver";
+import jsonParse from '../functions/jsonParse';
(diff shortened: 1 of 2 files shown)


dmitryyankowski:
@AndrewBastin i've added a new prop to the Ace Editor component (`<Editor />`). You can specify a `lint` prop like `<Editor :lint="false" />` and it will bypass the linting process. The reason for this was because the response box on the GraphQL page is supplied with `:lang="'json'"`. However when  the `runQuery` function is run .. it updates the response box to regular text `Loading...` (which isn't json), thus showing a linting error.

While the response box is in fact `json` for responses, we shouldn't run any linting on it anyways as the data is coming from an external data source.

This `lint` prop may come in handy for other editor boxes in the future, so I think it's a decent addition. 💯

PR #618, opened 2020-02-28
Link multiple auth providers

liyasthomas:
This is the hacky solution- Open alternate signin popup on error catch.
 - [x] Google
 - [x] GitHub
 - [ ] Combine login flow in `login.vue` and `settings.vue` page.

### DIFF
--- a/components/firebase/login.vue
+++ b/components/firebase/login.vue
@@ -66,8 +66,9 @@ export default {
               action: {
                 text: this.$t("yes"),
                 onClick: (e, toastObject) => {
-                  fb.writeSettings("syncHistory", false)
+                  fb.writeSettings("syncHistory", true)
                   fb.writeSettings("syncCollections", true)
+                  fb.writeSettings("syncEnvironments", true)
                   this.$router.push({ path: "/settings" })
                   toastObject.remove()
                 },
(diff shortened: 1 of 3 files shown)


dmitryyankowski:
How do I test this commit using the Netlify preview build link? @liyasthomas I press login with Google/Github and a box pops up, and then closes. Should I do it on localhost? Is Firebase Auth setup to allow logins from Localhost?

liyasthomas:
only production and localhost work with Firebase Auth

dmitryyankowski:
> only production and localhost work with Firebase Auth

Gotcha! Will take a look :)

dmitryyankowski:
@liyasthomas not sure how it was working on your local computer 😄 But I fixed a few scoping issues, and issues with toast notifications. If you notice.. on the live site, if you login; there's no toast that say's "Logged out" (When there should be). However; I fixed that. 

This PR should fix any issues I saw.

**Note**: I didn't look too far into how i18n works in this project, and if a key is missing in one of the lang JS files.. will it fallback to a key in the EN file? That's the only thing I'm wondering, but didn't look into. If there are isssues with users from other countries not seeing toast boxes.. that could be a root cause.

@liyasthomas

PR #620, opened 2020-02-28
Added shortcut to quickly run the GraphQL query

AndrewBastin:
This PR introduces a new shortcut to the GraphQL Query Editor which allows the user press Ctrl+Enter (Cmd+Enter on Mac) to run the query while editing in the query editor.


AndrewBastin:
@dmitryyankowski let @liyasthomas also review and merge the code.

PR #629, opened 2020-02-29
Ignore empty GQL Variable Strings

AndrewBastin:
This is a simple patch to ignore empty GraphQL variable strings.

That's about it. :sweat_smile:


liyasthomas:
Thank you all for the PRs and support. I'm having vacation and couldn't keep up with development. I'll make sure I review current PRs within 24hours.

PR #636, opened 2020-03-03
Refactor/ui

liyasthomas:


liyasthomas:
@NBTX @AndrewBastin Major UI refactor coming up.

UI elements including Header, Sidebar, etc will've seperate components.

PR #650, opened 2020-03-05
Added the ability to prettify GraphQL queries

AndrewBastin:
This PR intends to add the ability to do GraphQL query prettification. It adds a editor shortcut (`cmd-p` or `ctrl-p`) and also a button above the query editor to do prettification.

I am not a pretty big fan of the icon for the prettify button, I had that auto-fix magic wand icon in my mind but I think it is removed from the material icons collection now. @liyasthomas it would be awesome if you could replace that with something more aesthetically pleasing.

**NOTE to Translators: Added new entries for i10n**

### DIFF
--- a/components/graphql/queryeditor.vue
+++ b/components/graphql/queryeditor.vue
@@ -151,8 +151,8 @@ export default {
       name: "prettifyGQLQuery",
       exec: () => this.prettifyQuery(),
       bindKey: {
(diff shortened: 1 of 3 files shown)


dmitryyankowski:
@AndrewBastin you beat me to this. I thought about adding this the other day hehe :D Nice!

PR #655, opened 2020-03-07
Refactor/ui

liyasthomas:
- Introduces new Tabs component
- Replaced every old tab components with new one
- @AndrewBastin please check Tab implementations on GraphQL page


AndrewBastin:
@liyasthomas the new tabs have broken the type links in the docs panel, clicking on a type in the query tab for example doesn't switch to the types tab and scroll to the Film section.

liyasthomas:
@AndrewBastin I've fixed tab switch.

PR #693, opened 2020-03-20
Fix raw input (JSON)

leomp12:
Related issue: https://github.com/liyasthomas/postwoman/issues/597

As I've commented on commit https://github.com/liyasthomas/postwoman/commit/b14adc29f521271306970e896c45ffe46ee04848, the problem with JSON body and raw input is happening again after this commit, maybe I've named things badly (`canListParameters`) and you misunderstood me, so I'm suggesting another fix :stuck_out_tongue: 

I've realized that it's getting `rawInput` original state from Vuex, and then it could happen to have JSON content with raw input disabled and without the toggle component, this way it would not be possible to enable raw input at all...

Now I'm making sure `rawInput` is true always `canListParameters` is false, and kept (revert) `canListParameters` to be true only with `application/x-www-form-urlencoded`.

> With immediate `canListParameters` wacther, https://github.com/liyasthomas/postwoman/compare/master...leomp12:fix/raw-input?expand=1#diff-b19fa22add3de9cd0b98b05366479d92R1767 may be unnecessary, just to be really sure

### DIFF
--- a/pages/index.vue
+++ b/pages/index.vue
@@ -1434,8 +1434,8 @@ export default {
     },
     canListParameters: {
       immediate: true,
-      handler (canToggleRaw) {
-        this.rawInput = !canToggleRaw
+      handler (canListParameters) {
+        this.rawInput = !canListParameters
       }
     },
     contentType(contentType, oldContentType) {
@@ -1764,7 +1764,7 @@ export default {
     },
     rawInput: {
       get() {
-        return this.canListParameter ? this.$store.state.request.rawInput : true
+        return this.canListParameters ? this.$store.state.request.rawInput : true
       },
       set(value) {
         this.$store.commit("setState", { value, attribute: "rawInput" })
--- a/pages/index.vue
+++ b/pages/index.vue
@@ -1432,8 +1432,11 @@ export default {
         ])
       },
     },
-    canListParameters(canToggleRaw) {
-      this.rawInput = !canToggleRaw
+    canListParameters: {
+      immediate: true,
+      handler (canToggleRaw) {
+        this.rawInput = !canToggleRaw
+      }
     },
     contentType(contentType, oldContentType) {
       const getDefaultParams = contentType => {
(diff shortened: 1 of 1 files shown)


leomp12:
@liyasthomas lot of bad commits here, so if you agree with changes I would suggest to squash the PR :sweat_smile:

liyasthomas:
@leomp12 PR looks great! But I've been getting feedbacks from users suggesting I should enable RAW input for content-types other application/x-www-form-urlencoded. For example application/json should also support RAW payload, if not user need to enter payload params one-by-one.

Hope you can understand why I hotfixed that yesterday. Please add support for Raw input for content-types expect text/html and application/xml

leomp12:
Hey @liyasthomas , not sure if I undestand you :thinking:

This PR is exactly **ensuring JSON content will be with RAW payload**, I'm not suggesting remove raw input for JSON, is the opposite.

leomp12:
> But I've been getting feedbacks from users suggesting I should enable RAW input for content-types otherthan application/x-www-form-urlencoded

It's exactly what I'm doing here, forcing raw input for all content types other than `application/x-www-form-urlencoded`, then only form encoded will have raw input toggle to be able to enable/disable, in other cases raw input is always true (enabled) and with no option to disable.

With this PR, users would **not be able to disable RAW input for JSON**, do you disagree with that?

liyasthomas:
@leomp12 
> This PR is exactly ensuring JSON content will be with RAW payload, I'm not suggesting remove raw input for JSON, is the opposite.

I understand this PR exactly does that, but at the same time it also remove the toogler to disable key-value pair right? What I recommend i that, we should allow user to have the freedom to toggle between Raw input and key-value input (not jut raw input) for all content-type except text/plain and application/xml.

sorry if there was any confusion.

leomp12:
Ahhh ok, now I understand you 😃
Sorry for that, I think JSON was not working with parameters list at all (don't work on mey tests I think).
So maybe we can just set raw input enabled by default for JSON, but with the toggle, wdyt?

liyasthomas:
> So maybe we can just set raw input enabled by default for JSON, but with the toggle, wdyt? 

Perfect!

leomp12:
Great, I'll commit a fix tomorrow

leomp12:
Hello again @liyasthomas ,
Now it's showing the raw input toggle for both JSON and form encoded and it'll be raw by default only when `rawParams` is edited.
So even for JSON I'm keeping raw disabled by default (as it was), but when user enable raw input it'll be preserved through URL query, sounds good?

PR #695, opened 2020-03-21
Add application/vnd.api+json

### LINKED ISSUE #694: Support preview of JSON:API's "application/vnd.api+json" Content-Type
**Describe the bug**
When I issue a GET request to an endpoint responding with the [content type "application/vnd.api+json"](https://jsonapi.org/format/#content-negotiation-clients), I see a 200 success response, but the reponse field is blank (see screenshot).

When I click "copy response", and paste the value in a text editor, I see:

`[object Object]`

This leads me to believe that it isn't recognizing this particular content-type and doesn't know to parse it as JSON? Just a guess.

I've alluded to this in this issue: https://github.com/liyasthomas/postwoman/issues/634.

I believe other similar tools are able to recognize this.

**To Reproduce**
1. Make a request to a JSON API having a `application/vnd.api+json` Content-Type
2. Scroll down
3. Response body is blank but request is 200 okay.

**Screenshots**
image

 - OS: Catalina
 - Browser Chrome
 - Version v1.9.0

**Additional context**

allthesignals:
Add application/vnd.api+json to the places where content type matters.

Closes #694 

I noticed there are a lot of places where this is repeated. Should we place these in an constant somewhere? I'm not sure I understand where the different types matter...

### DIFF
--- a/pages/index.vue
+++ b/pages/index.vue
@@ -1439,6 +1439,7 @@ export default {
       const getDefaultParams = contentType => {
         switch (contentType) {
           case "application/json":
+          case "application/vnd.api+json":
           case "application/hal+json":
             return "{}"
           case "application/xml":
@@ -1463,7 +1464,8 @@ export default {
       } else {
         if (
           this.responseType === "application/json" ||
-          this.responseType === "application/hal+json"
+          this.responseType === "application/hal+json" ||
+          this.responseType === "application/vnd.api+json"
         ) {
           this.responseBodyText = JSON.stringify(this.response.body, null, 2)
           this.responseBodyType = "json"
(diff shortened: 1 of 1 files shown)


AndrewBastin:
Lol, I thought I was cleared to work on this. I kinda updated my development setup (software wise) and wanted to test it out. Nvm.

This implementation suffices, but I was thinking of capturing all the 'application/***+json' Content-Types using a REGEX or something.

But I think this does it for now.

allthesignals:
Oh sorry @AndrewBastin ! I thought I mentioned that in the issue... a regex is great approach, happy to help! Is there any important difference between "json-ish" content types?

liyasthomas:
https://stackoverflow.com/a/48704300

AndrewBastin:
@allthesignals 

Naah, I meant, for example, `application/vnd` is a family of Content-Types and the things after `application/vnd` can vary according to the vendors of the API, so I was thinking of solving this with a general solution that captures all `application/vnd` Content-Types together in like a single clause with a regex most probs.

But I think there needs to be a refactor in general around that content-type negotiation, as you said, it seems really repeated and really spaghetti like.

But I don't feel that is within the scope of this PR, but is something we have to do later.

Hence, I will be merging this PR.

AndrewBastin:
Actually @liyasthomas we can close this PR, we all do that general solution thingy in a refactor.

PR #698, opened 2020-03-23
Err french typo

thomasbnt:


liyasthomas:
Thanks for the quick PR. We acknowledge the issue and promise to patchit in the next commit.

PR #820, opened 2020-04-28
Realtime SocketIO support for json user input

feydan:
The SocketIo tab in realtime currently only supports string input even though the SocketIo library can support sending json data (implicitly encoding and decoding).  

This change attempts to parse the user input to json in order to send it as json instead of a string.  If the user input is not valid json, it will be sent as a string.

### DIFF
--- a/components/realtime/socketio.vue
+++ b/components/realtime/socketio.vue
@@ -144,13 +144,13 @@ export default {
             ts: new Date().toLocaleTimeString(),
           })
         })
-        this.io.on("connect_error", error => {
+        this.io.on("connect_error", (error) => {
           this.handleError(error)
         })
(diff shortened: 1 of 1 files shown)


liyasthomas:
@feydan Thank you for the quick PR, it's a nice enhancement.

feydan:
Thank you for the lightning fast response!

whyDontI:
Hey, @feydan It doesn't seem to work for me.
Data is still being sent as a string.
Can you please check once?
Let me know if I am missing something?

feydan:
Hi @whyDontI .  It is hard to know if you are missing something without additional context.

Are you using the SocketIO tab?  
How do you know it is sending as a string?
Do you have steps to recreate the issue?

Issue #834, opened 2020-05-01
URL Path Parameters


sboulema:
**Is your feature request related to a problem? Please describe.**
I am trying to convert from Postman to Postwoman. I love that I have a proper API testing tool that I can use on my laptop and on my phone and keep everything in sync!
I do however miss the easy way of setting url path parameters as is possible in Postman.

**Describe the solution you'd like**
More info on how path parameters work in Postman: 

> Path parameters form part of the request URL, and are referenced using placeholders preceded by : as in the following example: /customer/:id

https://learning.postman.com/docs/postman/sending-api-requests/requests/#sending-parameters

**Describe alternatives you've considered**
Either a similar solution with a special character or maybe a dropdown next to the current query parameters to indicate if it is a query or path parameter.

liyasthomas:
Hi @sboulema, Glad you're making use of PW.

1. To add path parameters, simply choose "Parameters tab" under request section and add parameters.

![image](https://user-images.githubusercontent.com/10395817/80800536-5b9d2700-8bc7-11ea-988f-3460de20210b.png)

2. There's also another way to append parameters that is to append parameters along the URL itself. Ex: `https://httpbin.org/get?param1=true&param2=false`

3. You can even use `Pre-request scripts` to add/remove parameters

![image](https://user-images.githubusercontent.com/10395817/80800111-4d9ad680-8bc6-11ea-9280-812ac8ec750d.png)

Read more on `Pre-request scripts`: https://github.com/liyasthomas/postwoman/wiki/Pre-Request-Scripts.

Hope issue is solved.

sboulema:
Thanks for the quick response!

Those parameters are added as query parameters though. I am looking for adding a parameter as part of the url path.

/petstore/pet/:petId

With a parameter petId defined as 42 for example.

Would lead to: /petstore/pet/42 as url.

The current parameters would give me:

/petstore/pet?petId=42 as url.

liyasthomas:
Yep, that's why we've `Pre-request scripts` which can be use to set environment variable. Check Environments Tab in right sidebar to add/export/import envs.

Read more on `Pre-request scripts`: https://github.com/liyasthomas/postwoman/wiki/Pre-Request-Scripts

![image](https://user-images.githubusercontent.com/10395817/80801120-0b26c900-8bc9-11ea-8114-1df1e69bc36f.png)

sboulema:
Ah, missed that! Thank you very much 😃

[...]

PR #840, opened 2020-05-03
URL Path Parameters #834

sboulema:
[PathParameter](https://i.imgur.com/teJ3MiD.png)

Adds a type dropdown to the current parameter list. Based on the type, the parameter either gets included as a query parameter or gets added as an environment variable so it can be used in the current url path template replacement. 
Parameters without a type are treated as query parameters so the implementation is backwards-compatible.

This way the parameter gets saved with the request and does not have to be saved with the environment.


liyasthomas:
@sboulema I can't reproduce this PR's intended behaviour on the deploy preview: https://deploy-preview-840--postwoman.netlify.app/

Environment variables added from parameters list with `Path` type were not recognised.

sboulema:
Wow sorry it only worked in a couple of scenarios. Updated the PR works much more reliably now!

anstosa:
Looks like this was merged but I don't see it in the UI. Has it moved or changed? Or is setting runtime environment variables via pre-request scripts still the preferred way of implementing URL path parameters?

PR #851, opened 2020-05-07
Environment variables in collections

### LINKED ISSUE #642: Environment variables in collections
I have a request with path ```<<root>>/public/blogger```, which I want to run against different environments. I save this request in a collection. I would expect to be able to change the environment (and, hence, `root`) again later. But when I load the saved request, `root` is already replaced with the value it had when I saved the request.

Video: https://fashiontale.org/index.php/s/Aaz77P7breqAeK8

sboulema:
Fixes saving a request when using environment variables.

url and path variables get overwritten with replaced values, so before saving we reconstruct them using the original uri variable.

This PR intends to fix #642.

### DIFF
--- a/functions/utils/uri.js
+++ b/functions/utils/uri.js
@@ -1,0 +1,13 @@
+export function parseUrlAndPath(value) {
+    let result = {}
+    try {
+        let url = new URL(value)
+        result.url = url.origin
+        result.path = url.pathname
+    } catch (error) {
+        let uriRegex = value.match(/^((http[s]?:\/\/)?(<<[^\/]+>>)?[^\/]*|)(\/?.*)$/)
+        result.url = uriRegex[1]
+        result.path = uriRegex[4]
+    }
+    return result;
+}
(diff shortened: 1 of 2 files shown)


liyasthomas:
Hi @sboulema , thanks for the quick PR, I like the simplicity of the implementation. But there's an issue. While restoring a request from collections, the `env variables` are still translated to it's `evn value`. This shouldn't be happening. When we reuse a saved collection request - it should restore URL with saved env quotation and not its evn value.

To reproduce:
## **1: create and save a request with env variables to collection.**
image

## **2: click saved request from collection to restore it.**
image

As you can notice, when we restore the request - the `env variables` have been replace with `env values`. The expected behaviour is: `env variable` should be maintained.

I think we can store the env variable as it is (without converting to env values) while saving to collections. What do you say about that? I wanna hear from you.

sboulema:
Thanks for the very clear reproducing steps!
Helped in finding a couple of more cases were the URL is geting saved wrongly.

Should be ok now :)

And yes I completely agree, it should always save non-replaced version of the uri and only use replaced version when doing the actual request and showing in the history list.

PR #915, opened 2020-05-30
Cancellable Requests

### LINKED ISSUE #909: Button to cancel requests
Sometimes wrong requests take forever to get a response and I just want to cancel it but there is no button to cancel it I have to wait for it to error out or refresh the tab

AndrewBastin:
This PR intends to add the ability to cancel running requests.

@liyasthomas what do you think about the cancel button ? should it be red or something ?

The browser extensions don't currently support cancellable requests, the cancel attempts will just be ignored in case of extensions for now, the support implementation will be rolled with extension version v0.10 which is a bit delayed as I am waiting for approval for the extension to the Microsoft Edge add-on store.

This PR fixes #909


liyasthomas:
@AndrewBastin no need of red bg color. "Cancel" isn't a errored state - moreover, there's a red theme in our themes. feel free to merge the PR.

PR #923, opened 2020-06-05
GraphQL response options only visible when a response is shown

AndrewBastin:
GraphQL response options (copy response and download response JSON) is only visible when a response is available.

### DIFF
--- a/pages/graphql.vue
+++ b/pages/graphql.vue
@@ -225,6 +225,7 @@
                 class="icon"
                 @click="downloadResponse"
                 ref="downloadResponse"
+                v-if="response"
                 v-tooltip="$t('download_file')"
               >
                 <i class="material-icons">get_app</i>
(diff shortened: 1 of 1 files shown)


liyasthomas:
Hey @AndrewBastin just want to mention that `graphql` dependency has a `v15.0.0` update which I think is not stable. And also, the whole new monorepo Graphiql would be our best bet. Graphiql does have some issues with validation language services - I bumped the version but it seems graphql v15. 0.0 support isn't stable.

Just fyi.

AndrewBastin:
@liyasthomas thanks for the heads up. Can you elaborate on the relevant issues ?

We are using `graphql` dependency solely for language services, which should be powering any GraphQL based tool.

If it is a problem, we can push down the version.

liyasthomas:
The issue is with graplql language validation import. I'm checking what went wrong. I'll update once it's resolved.

PR #955, opened 2020-06-21
Binary Responses & Response Lenses

### LINKED ISSUE #929: Support downloading binary responses
**Describe the bug**
When a binary file (e.g. an image) is requested, the binary content of the response is properly displayed in the response body window, however if the **Download file** button is clicked, a corrupted blob of data is downloaded instead of the actual binary file that was requested.

**To Reproduce**
1. Go to postwoman.io
2. Set `http://tim.org/images/TIMtitle.GIF` as the request URL
3. Click **Send**
4. When the response has arrived, click on the **Download file** button above the response body textfield
5. Examine contents of the downloaded binary file (e.g. with a hex editor), or just try to open the image

**Expected behavior**
The same binary file gets downloaded that was requested from the remote server.

**Screenshots**
Hexdumps of the original and postwoman-downloaded image file.
Left: original image file, downloaded via *right-click, Save image as...*
Right: image file downloaded via postwoman.io

image

**Desktop:**
 - OS: Windows 10 1909
 - Browser: Chrome
 - Version: 83.0.4103.97

**Additional context**
When looking at the corrupted file, notice the double-quote in the very beginning and the very end, the unicode-escaped sequences and the occasional new `EF BF BD` byte triplets. Its size is also 2-3 times bigger than that of the original file.

AndrewBastin:
This PR brings in two major updates to Postwoman both dealing with how we parse responses.

The first major update is the switch to binary responses for data. The current response system expects only string responses and acts really funky for binary responses. The new response system expects data to be stored in pure [ArrayBuffer](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer) objects. The implementation for this is ongoing and the dependent like the [Browser Extensions](https://github.com/AndrewBastin/postwoman-extension) and [Proxywoman](https://github.com/postwoman-io/proxywoman) are expected to be updated to comply with the system.

The second major update deals with a change in how we present responses. Certain responses, especially those with binary content expect a special way to render contents the current implementation for specializing for different types of contents are not really extensible to allow support for these varied and completely different response types. Hence, a concept of a Response Lens is being introduced, it is simply an extensible mechanism to give certain Content Types to specify a special way to render and allow multiple ways of representation for the same data. The Lens system expects the response data to be represented in the new response system.

This PR intends to fix #929.

Deploy preview: https://deploy-preview-955--postwoman.netlify.app

**MVP**
- [x] Basic response parser and renderer
  - [x] RAW
  - [x] JSON
  - [x] HTML
  - [x] XML
  - [x] Image
- [x] Actions
  - [x] Copy to clipboard / Beautify / Download on RAW / JSON
  - [x] Preview on HTML
  - [x] Preview / Download on Binaries (Image, GIF etc..)
- [x] Browser extension support
  - [x] Firefox
  - [x] Chrome (implemented on v0.12^)
- [x] Proxy support @NBTX 
- [x] Beautify / Refactor / Lint code

| Site | Extension OFF & Poxy OFF | Extension ON & Proxy OFF | Extension OFF & Proxy ON | Extension ON & Poxy ON |
| ----------- | ----------- | -----------  | -----------  | -----------  |
| [Live](https://postwoman.io/) | [x] working | [x] working | [x] working | [ ] failed |
| [#955](https://deploy-preview-955--postwoman.netlify.app/) | [x] working | [x] working | [x] working | [x] working |


AndrewBastin:
@liyasthomas you got the preview content source wrong, the body is an array buffer now, I have amended the fix to your commit, will force push it right now.

AndrewBastin:
@liyasthomas can you revert the a885e77 commit, I envisioned RAW lens as a place to show raw unprocessed data just as how it was fetched, all the post processing prettification stuff should only happen within each lenses for each.

What you think ?

liyasthomas:
Well the default data is shown as received.. Prettify button is only shown for JSON type. I think it's better to offer a Prettify button in case of large arrays as response

AndrewBastin:
Well, I suppose then the use can just use the JSON lens instead right ? We are already prettifying there. Plus syntax highlighting of course.

I envision Raw just to be, well, Raw.

liyasthomas:
Okay. makes sense. Then guess there's no need for Prettify button in json either. Cause editor already prettify the content right?
@AndrewBastin

AndrewBastin:
Yup, yeah, that is the thing with lenses right, we can render specifically to present for each content type, and Raw will show the actual response in case you need it.

liyasthomas:
@AndrewBastin I've add a basic Header component in `/lenses`. It's okay to movethe component elsewhereifyou want :+1:

AndrewBastin:
Umm, the header lens implementation looks nice actually, I am fine with it

AndrewBastin:
> It might be a good idea to have common logic within a `mixin`.
> `components/lenses/renderers/*.vue`

Well, yeah, I was thinking about it, most of the lenses are working with string content, so we could implement a mixin to generalize that. I will work on that.

liyasthomas:
@AndrewBastin 
on https://deploy-preview-955--postwoman.netlify.app
Method: `post`
URL: `https://httpbin.org/get`
produce `405` which is expected, but its HTML lens is not rendering. Is it just me or can you replicate it? I afraid this was introduced in mixins.

AndrewBastin:
@liyasthomas it works fine for me, lenses shouldn't just not work, it looks at the content type.

NBTX:
Does the preview at https://deploy-preview-955--postwoman.netlify.app/ expect the Proxy to return a binary response as expected by this PR (i.e. can it be used for testing)?

AndrewBastin:
@NBTX, nope, not yet, if you want I will add the relevant flag thingy right now as we discussed in Discord.

AndrewBastin:
@NBTX I have pushed the code with the proposed binary response support for proxy.

Requests expecting binary response add an extra field `wantsBinary` with the a value of `true`. The proxy should (for now atleast) check the presence of this field and if so, return the response's `data` field as a base64 string and add a new field to the response with the field `isBinary` with a value of true. If not, it should still act the same as before. You can check the commit diffs to understand it from the code perspective and feel free to shoot questions here.

AndrewBastin:
@liyasthomas can you actually implement a proper styling for the warning :sweat_smile:

liyasthomas:
@NBTX Chrome extension issue has  been fixed. Looking forward on proxy support to merge this PR.

AndrewBastin:
Umm, @liyasthomas there is a small bug left to fix which is rolled to Firefox but on Chrome yet which I will upload for review soon and then everything is green from the extensions side.

tests/e2e/integration/proxy.spec.js
```diff
@@ -11,15 +11,15 @@
...
-      .get("#send")
-      .click()
-      .get("#response-details-wrapper", { timeout: 24000 })
-      .should("be.visible")
-      .should(($wrapper) => {
-        expect($wrapper).to.contain("Hello World")
-      })
-  })
-})
+// describe("Proxy enabled - external request", () => {
```
NBTX, line 14: Does this need to be re-enabled?

helpers/utils/b64.js
```diff
@@ -0,0 +1,31 @@
+export const decodeB64StringToArrayBuffer = (input) => {
+  const bytes = Math.floor((input / 4) * 3)
```
NBTX, line 2: This should be `input.length`, not `input`, otherwise you just get NaN and `bytes` is always interpreted as 0.
```suggestion
  const bytes = Math.floor((input.length / 4) * 3)
```

components/ui/ace-editor.vue
```diff
@@ -76,6 +76,7 @@ export default {
       })
     },
     lang(value) {
+      console.log("ace/mode/" + value)
```
NBTX, line 79: Can we remove this - or add a prefix such as `Current editor:`?
```suggestion
      console.log("[INFO] Current Editor: ace/mode/" + value)
```

```suggestion
```

liyasthomas:
Okay. I will complete testing and solve issues by tomorrow.

tests/e2e/integration/proxy.spec.js
liyasthomas: Need total rewrite if tests. Will look into it after merge.

components/ui/ace-editor.vue
AndrewBastin: Umm, that was a log I forgot to remove while updating Ace Editor component code, I am committing the removal of that log.

tests/e2e/integration/proxy.spec.js
AndrewBastin: Well, actually I think it is better to do this before merging ?

helpers/utils/b64.js
AndrewBastin: Oooh, good catch!

AndrewBastin:
The 0.14 Chrome Extension is out!

Which means this PR is ready for review! (finally)

tests/e2e/integration/proxy.spec.js
liyasthomas: Let's land this fist. Need to tweak response lens further to support tests.

liyasthomas:
@NBTX 

1. url: https://deploy-preview-955--postwoman.netlify.app/
3. turn on proxy (https://deploy-preview-955--postwoman.netlify.app/settings).
2. make a default GET request.
3. notice 2 extra spaces have been appended at eof producing invalid json warning.

image

SamJakob:
@liyasthomas I've resolved the issue with the null bytes - the proxy was base64 encoding with padding, so I switched to 'raw encoding' (i.e. base64 without padding.)

NBTX:
@liyasthomas I've resolved the issue with the null bytes after the response data. The base64-encoding was including padding, so I switched to 'raw encoding'. (i.e. encoding without padding.)

PR #974, opened 2020-07-04
feat(tabs): add keyboard accessibility

jyash97:
Accessing tabs through `tabIndex` and pressing `space` or `enter`.
https://www.loom.com/share/6732ad5285f34518a37e0ad677f0b66f

### DIFF
--- a/components/ui/tabs.vue
+++ b/components/ui/tabs.vue
@@ -2,7 +2,13 @@
   <div class="tabs-wrapper">
     <div class="tabs">
       <ul>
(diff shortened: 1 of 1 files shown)


liyasthomas:
Thank for the quick pull request - it looks good to me.

But there's already a native `@keyup.enter` event listener for keypress event on Vue. Also the mentioned focus class is not working - better to use `:focus` selector for that event. 

I've fixed the errors and added the feature in https://github.com/liyasthomas/postwoman/commit/84eb935cec8953e509ca147bfd37c16e454be220 - thus closing this pull request for now.

PR #993, opened 2020-07-15
fix: replaceWithJSON used wrong commit name

perseveringman:
replaceWithJSON function used wrong commit name. 

replaceWithJSON


liyasthomas:
@perseveringman  thanks for the quick pull request. All LGTM.

PR #1007, opened 2020-07-20
Improvements on the generated doc ui

perseveringman:
1. Improve the layout of the doc.
2. Add an text-anchor to the doc title.
3. Adjust the padding of the elements.


liyasthomas:
I'm looking for a better UI implementation. Current implementation is not upto the mark. Let this pull request stay while I work on the ne UI.

PR #1033, opened 2020-07-30
Add trailing backslash to generated cURL code for easier paste-and-execute

ushuz:
And stop generating Content-Length header so it's easier to modify the generated code. Corresponding HTTP clients (like cURL) should handle Content-Type pretty well anyway.

### DIFF
--- a/pages/index.vue
+++ b/pages/index.vue
@@ -1897,7 +1897,7 @@ export default {
           requestString.push(`  -H 'Content-Type: ${this.contentType}; charset=utf-8'`)
           requestString.push(`  -d '${requestBody}'`)
         }
-        return requestString.join(" \n")
+        return requestString.join(" \\\n")
       }
     },
     tokenReqDetails() {
--- a/pages/index.vue
+++ b/pages/index.vue
@@ -1825,7 +1825,6 @@ export default {
           } else if (this.contentType.includes("x-www-form-urlencoded")) {
             requestBody = `"${requestBody}"`
           }
-          requestString.push(`xhr.setRequestHeader('Content-Length', ${requestBody.length})`)
           requestString.push(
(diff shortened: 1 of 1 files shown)


pages/index.vue
```diff
@@ -1879,28 +1877,27 @@ export default {
...
         if (["POST", "PUT", "PATCH", "DELETE"].includes(this.method)) {
           const requestBody = this.rawInput ? this.rawParams : this.rawRequestBody
-          requestString.push(`  -H 'Content-Length: ${requestBody.length}' \n`)
-          requestString.push(`  -H 'Content-Type: ${this.contentType}; charset=utf-8' \n`)
-          requestString.push(`  -d '${requestBody}' \n`)
+          requestString.push(`  -H 'Content-Type: ${this.contentType}; charset=utf-8'`)
+          requestString.push(`  -d '${requestBody}'`)
         }
-        return requestString.join("").slice(0, -2)
+        return requestString.join(" \\\n")
```
liyasthomas, line 1900: A `\\` is additional. Please remove it. Only `\n` is enough.

ushuz: @liyasthomas No, adding a trailing backslash `\\` to each line is the whole point, otherwise you can't just paste-and-execute a multiline shell command.

liyasthomas: oops. I'll revert my commit.

pages/index.vue
```diff
@@ -1897,7 +1897,7 @@ export default {
           requestString.push(`  -H 'Content-Type: ${this.contentType}; charset=utf-8'`)
           requestString.push(`  -d '${requestBody}'`)
         }
-        return requestString.join(" \\\n")
+        return requestString.join(" \n")
```
liyasthomas, line 1900: reverting last commit.
```suggestion
        return requestString.join(" \\\n")
```

PR #1115, opened 2020-08-25
support user to customize the path to socket-io server

HerbertChu:
Add one more column to input customize path
image


liyasthomas (approved):
LGTM

PR #1126, opened 2020-08-29
tests: improve test suite

jamesgeorge007:
Use `toBe()` matcher for primitives.

### DIFF
--- a/components/ui/__tests__/autocomplete.spec.js
+++ b/components/ui/__tests__/autocomplete.spec.js
@@ -8,6 +8,9 @@ const props = {
   source: ["app", "apple", "appliance", "brian", "bob", "alice"],
 }
 
+// ["pp", "pple", "ppliance", "lice"]
+const suggestionStr = props.source.filter((str) => str.startsWith("a")).map((str) => str.slice(1))
+
 const factory = (props) =>
   mount(autocomplete, {
     propsData: props,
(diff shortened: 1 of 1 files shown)


liyasthomas (approved):

AndrewBastin:
> Use `toBe()` matcher for primitives.

About the `toBe()` usage. Yes, for primitives it is fine, but I want tests to be consistent and its behaviour easily predictable. `toBe()` checks for referential equality in non-primitive cases which is a really obscure behaviour.

I want to move the development process to a more TDD style and expect tests to be written for all the new code and I believe `toBe()` adds a wee bit of confusion to people on its behaviour.

Hence, I want to reserve `toBe()` just for referential equality checks.

@liyasthomas

liyasthomas:
I guess `toBe()` suits best for primitives like strings, numbers and booleans. For objects, and everything else I prefer `toEqual()`. Wdy say @jamesgeorge007 ?

AndrewBastin:
@liyasthomas that is what James prefers too, but I am suggesting to keep `toEqual()` for primitives as well.

liyasthomas:
Okay, I second Andrew if somehow `toBe()` cause any confusion for a new user.

components/ui/__tests__/autocomplete.spec.js
```diff
@@ -8,6 +8,9 @@ const props = {
   source: ["app", "apple", "appliance", "brian", "bob", "alice"],
 }
 
+// ["pp", "pple", "ppliance", "lice"]
+const suggestionStr = props.source.filter((str) => str.startsWith("a")).map((str) => str.slice(1))
```
jamesgeorge007, line 12: Should we keep this change?

AndrewBastin: Yup, this is fine.

PR #1141, opened 2020-09-04
Lossless optimized images

tiagofilipesilva:
Hi,
The image size was optimized without quality loss.


liyasthomas (approved):
@tiagofilipesilva this is a nice addition. but I wonder how you do the optimization? Is it using any manual tools or is there any automated process (GitHub action, npm package etc)?

tiagofilipesilva:
Hi @liyasthomas!
These were done "manually" via Papa's Best Optimizer. Kudos for that tool and your project!

https://papas-best.com/optimizer_en

MyIsaak:
Hi @liyasthomas and @tiagofilipesilva,

Would you say its reasonable to convert all images into AVIF format for those browsers that support it? Then for older browsers the default image formats can be used.

tiagofilipesilva:
Hi @MyIsaak, I'm not the project owner but in my opinion anything that is _lossless_, _open_ and _has a decent support base_ should be used.

I don't think that AVIF is supported yet by Mozilla so I believe that we should use only things that have support base, because otherwise we would have to think on fallbacks, and the [fallback of the fallback](http://camendesign.com/code/video_for_everybody/test.html), etc.

But again, that is just my opinion.

MyIsaak:
@tiagofilipesilva I'm not saying we should fully replace images to AVIF, but use multiple image formats (AVIF, WEBP included) and let the browser choose which one it supports with the tags mentioned earlier.

Of course this might be tiresome to convert all images into their formats which increases the repo size, however this could also be done as a build step. Using `sharp` to convert and compress all images is an idea.

tiagofilipesilva:
Judging from @liyasthomas 's track record I think he can make it happen! :+1:

Issue #850, opened 2020-05-07
Import of Postman collection with Nested folders is failing


nabsha:
**Describe the bug**
If there are more than 1 level of sub folders in postman collection, then import fails without errors.
[parent-folder-3.postman_collection.txt](https://github.com/liyasthomas/postwoman/files/4593909/parent-folder-3.postman_collection.txt)

**To Reproduce**
1. Import the attached file which is in postman format 2.1 version 
2. See error

**Expected behavior**
File to be imported successfully

**Screenshots**

**Desktop (please complete the following information):**
 - OS: mac OS
 - Browser chrome
 - Version : master

liyasthomas:
@JacobAnavisca

mahoriR:
@liyasthomas  Looks like @JacobAnavisca  is already working on this. I would be happy to contribute. I can work on weekends.

AndrewBastin:
@JacobAnavisca are you still working on this ?

@mahoriR if you are interested, you could take this up!

mahoriR:
> @mahoriR if you are interested, you could take this up!

@AndrewBastin  Sure. Thanks.

liyasthomas:
Hi @mahoriR is there any updates in this issue? Let me know if you need any help from our side.

mahoriR:
> Hi @mahoriR is there any updates in this issue? Let me know if you need any help from our side.

I'll be working on this 13th Sept onwards.  Unfortunately blocked with office work till them.

exussum12:
Hoppscotch doesnt seem to support sub folders like this. Ive got the import working, but the front end stops working.

So its not really a postman import issue, its more a limit of Hoppscotch.

liyasthomas:
Yes, as of now we don't support nested subfolders. Any idea how we can toggle this issue or will think about adding nested folders.

exussum12:
I can't see another ticket for nested sub folders, This ticket would need to be put on hold until that happens. The import completely stops the front end working currently and you need to clear the local storage to continue.

Obviously thats not a great solution, so I think implement nested folders, and then look at this issue again. Whats your thoughts on that ?

liyasthomas:
Yes, you're right. Nested subfolders should be a feature.

exussum12:
For now, ive started a tool to convert thefolder structure.

The example here looks like this when imported

![image](https://user-images.githubusercontent.com/1102850/92299186-0fcd0680-ef48-11ea-94c1-38b3b869de02.png)


[...]

PR #1150, opened 2020-09-08
Handle postman subfolders when importing json

florinmirosnicencu:
Hello everybody, 

I have created this pr in order to fix this issue #850 .

It's based on the idea mentioned by @exussum12, but instead of a space between subfolder names I have chosen -- for better clarity.

Looking forward to your feedback.


components/collections/import-export-collections.vue
```diff
@@ -204,6 +205,10 @@ export default {
             name: "",
             requests: [],
           }
+      if (folders) {
+        //pick up collection name even when all children are folders
+        postwomanCollection[0].name = collection.info ? collection.info.name : ""
```
florinmirosnicencu, line 210: FYI: This is a bugfix for the case where a collection that contains only folders is imported, and in that case the name for taht collection  will not be populated.

exussum12 (approved):

liyasthomas (approved):

PR #1180, opened 2020-09-21
Add check for active suggestion

MattJarman:
I noticed when trying to tab to the next input in the headers section, it would paste `undefined` into the current input. I've added a check that restores the tab functionality if there are no active suggestions.

Please let me know if there are any issues with this, thanks!


liyasthomas (approved):
@MattJarman thanks for noticing this issue and making this quick PR.

MattJarman:
Thanks!

PR #1203, opened 2020-10-01
Add missing id translation on history component

fikimaul:
Add missing id translation on history component


liyasthomas (approved):
LGTM. Thanks for this quick PR.

Issue #1188, opened 2020-09-25
Add search (live filter?) box for GraphQL schema panel


Qix-:
**Is your feature request related to a problem? Please describe.**

The schema fetching in Hoppscotch is very helpful and (mostly) well laid-out. However, it appears to be using a virtual list of some sort and thus isn't ctrl+f searchable, making large schemas incredibly hard to traverse.

This, paired with the fact it can't be resized (it's a tiny sliver for me) and that clicking types and whatnot doesn't seem to scroll to the appropriate positions, makes it for kind of a clunky experience - especially when the API in question doesn't provide formal docs and instead simply points to the in-schema documentation.

**Describe the solution you'd like**

Basically this.

![image](https://user-images.githubusercontent.com/885648/94298672-d005a780-ff66-11ea-8d5b-d32d1311eafb.png)

**Describe alternatives you've considered**

Nothing really, feels like this is a core feature that'd be abundantly useful.

Alternatively, a tree-like viewer instead of a long virtual list, where items can be expanded to include documentation notes and whanot (collapsed by default).

**Additional context**

Only appears to be missing for GraphQL. The component in the screenshot is just hacked in, copied from another tab.

liyasthomas:
Thanks for reporting your pain points regarding GraphQL page. We're working on an overhaul on GraphQL page and hoping to bring the best experience. Will use this issue as a base to track it's progress.

AndrewBastin:
I think this will be good for Hacktoberfest as I am preoccupied with working on the Teams backend.

The implementation is pretty trivial actually.

LeoMartinDev:
I'll give it a try 👌

LeoMartinDev:
I'm wondering, should it filters only on `fieldName` ? Maybe we could also search in the description, not sure about this

Qix-:
Being able to fuzzy search for any of the content would be groundbreaking. Some of the schemas that get imported are _massive_ and it's not always clear what the field name will be.

LeoMartinDev:
> Being able to fuzzy search for any of the content would be groundbreaking. Some of the schemas that get imported are _massive_ and it's not always clear what the field name will be.
> 
> Just my $0.02.

You mean, including mutations args, fields types, etc.. ? Got something workin, filtering on description & field/mutation/type name

Qix-:
Yeah, basically any part of the spec that is returned would be searchable in theory - within reason, of course.

LeoMartinDev:
[...]

PR #1220, opened 2020-10-03
Add search input & capability to graphql types and fields

LeoMartinDev:
Related issue #1188 

The PR adds basic filtering for `queryFields`, `mutationFields`, `subscriptionFields` and `gqlTypes`.

The value provided to the search input is used to match on `fieldName` and `description` fields.

`gqlTypes` are handled differently as we also filter on `gqlTypes._fields` array.

Some improvements for an other PR could be to highlight matched text (?)
Also, `gqlTypes._fields` filtering could be improved by keeping only matched fields as it can be confusing to see the whole Schema and having to scroll until you find concerned field

**Unfiltered results**

image

**Filtered results**

image

### DIFF
--- a/components/graphql/field.vue
+++ b/components/graphql/field.vue
@@ -27,7 +27,8 @@
 
 <style scoped lang="scss">
 .field-highlighted {
-  color: #ffff00;
+  @apply border-b-2;
+  @apply border-acColor;
 }
 </style>
 
--- a/components/graphql/field.vue
+++ b/components/graphql/field.vue
@@ -1,6 +1,6 @@
 <template>
   <div class="p-2 m-2 border-b border-dashed border-brdColor">
-    <div class="field-title">
+    <div class="field-title" :class="{ 'field-highlighted': isHighlighted }">
       {{ fieldName }}
       <span v-if="fieldArgs.length > 0">
         (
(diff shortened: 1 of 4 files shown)


AndrewBastin:
@LeoMartinDev congrats on your first contribution to Hoppscotch!!!!

This is really good stuff!

Umm, the types search is a bit confusing for now. But we are sorta thinking of a overhaul to the GraphQL page anyways. To make it a lot more simplified.

To solve the issue with types search for now, can we highlight matched fields with an outline or some color effect ? If you are able to pull that off too, this would be perfect, or I will eventually come around and implement that.

LeoMartinDev:
> @LeoMartinDev congrats on your first contribution to Hoppscotch!!!!
> 
> This is really good stuff!
> 
> Umm, the types search is a bit confusing for now. But we are sorta thinking of a overhaul to the GraphQL page anyways. To make it a lot more simplified.
> 
> To solve the issue with types search for now, can we highlight matched fields with an outline or some color effect ? If you are able to pull that off too, this would be perfect, or I will eventually come around and implement that.

Thanks, glad to help 😄 
I thought about the highlighting feature and I have two solutions :
1. creating a new prop `hightlight-name` to the `field` component and set it to `true` from the graphql page
2. add a new key to the `grapqlTypes._fields` like `{ fieldNameToDisplay: '<span class="highlighted">Test</span>' }` and render it using `v-html`

AndrewBastin:
The first one seems better.

LeoMartinDev:
@liyasthomas I just added the highlighting functionality, any thought about the styling ? For now the text is just yellow 😅

image

liyasthomas:
I'll look into it.

liyasthomas (approved):
LGTM

cc: @AndrewBastin

Issue #1222, opened 2020-10-03
HTTP PAGE : Headers suggestions are overlapped by the response section title


LeoMartinDev:
**Describe the bug**
The headers suggestion in the HTTP page are overlapped by the response section title

**To Reproduce**
1. Go to 'Home'
2. Click on 'Headers'
3. Click on 'Add new'
4. Click on the header name field
5. See error

**Screenshots**
![image](https://user-images.githubusercontent.com/7238941/94997421-3a43cb00-05ab-11eb-917d-43cbf4bd513c.png)

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Chrome
 - Version 85

tonybatty:
I have fixed this issue on PR #1223

PR #1223, opened 2020-10-03
Fix: http page headers suggestions overlapped issue

tonybatty:
Fixes issue #1222

The header dropdown was being cut off due to the parent container .tabs having overflow set to hidden so have removed this property from the styles and this has fixed the issue:

Screenshot 2020-10-03 at 19 38 27


liyasthomas (approved):
Thanks. For this quick pr.

PR #1248, opened 2020-10-07
Fix form post data parsing

yq314:
This PR is to fix 2 issues in the `nodejs-request` codegen.

1. form data with more than 1 key/val pairs are not parsed properly
before:
Screenshot 2020-10-07 at 11 07 00 AM

after:
Screenshot 2020-10-07 at 11 05 03 AM

2. XML is not quoted properly so it generates broken code
before:
Screenshot 2020-10-07 at 11 06 47 AM

after:
Screenshot 2020-10-07 at 11 05 19 AM


liyasthomas (approved):
LGTM

PR #1249, opened 2020-10-07
add codegen for nodejs native and instrument snapshot tests

yq314:
This PR adds codegen for NodeJs Native and also performs snapshot test on the generated code.

I put the test data in `__fixtures__` so it can be reused in tests for other generators.

Screenshot 2020-10-08 at 12 09 24 AM


yq314:
Retry DeepCode

liyasthomas (approved):
Thanks for this PR, also kudos for tests.

PR #1255, opened 2020-10-09
Run tests against all codegen instead of individual ones

yq314:
Realized we don't need to write tests for generators separately as they are going to be exactly the same.

Instead, we just need to run the same set of tests against all generators at one go.

By doing this, discovered a bug in the curl generator that `auth` is not being passed.


liyasthomas (approved):
Thanks for this. You contributions are very helpful.

Issue #1190, opened 2020-09-27
Kick off Hacktoberfest 👩‍💻👨‍💻👌 with the next best thing DEVs will ❤️


liyasthomas:
### Are you looking to contribute to an open source project in this Hacktoberfest? or want to get started with latest technologies like VueJS & NuxtJS?

We're developing Hoppscotch, an open source API request builder for the web, and would love to count you as a contributor to improve our components & applications.

Web app: [hoppscotch.io](http://hoppscotch.io)
GitHub: @hoppscotch

_24k+ stars on GitHub • # 2 product of the day on Product Hunt • featured in Indie Hackers, Hacker news, Hacker noon, YouTube & open source podcast etc._

## **Get started 🚀**

As the project’s size grew quite a bit the past couple of months, you might ask yourself, where to get started 🤔?

**Meet the technology stack:**

0. HTML + SCSS + JavaScript
1. [NuxtJS: The Intuitive Vue Framework](https://nuxtjs.org)
2. [TailwindCSS](https://tailwindcss.com)

Check out [How to set-up Hoppscotch development environment in 5 mins →](https://github.com/hoppscotch/hoppscotch#browser-based-development-environment)

## 1. Codegen

At Hoppscotch, we want to help developers to go from discovering an API to experimenting with it and then to make actually using the API in their code as frictionless as possible. To support that, Hoppscotch had the ability to generate code for JavaScript (XHR and Fetch API) and cURL. We recognize that a lot of devs of different backgrounds use Hoppscotch. So recently, we modularized and separated the code generation system (only for REST currently, WS and GraphQL coming soon!) into different code generator modules which can be easily added to and extended upon.

As a result now one can easily add code generation for multiple programming languages (and/or libraries, or even programs), even if you don't have much knowledge about the rest of the Hoppscotch codebase.

#### Step 1: Create a code generator object
Code generator objects are placed on [`helpers/codegen/generators`](https://github.com/hoppscotch/hoppscotch/tree/master/helpers/codegen/generators) folder, they have a specified `id`, `name` (display name) and a generator function which has a parameter, `context` containing all the data about the request (refer, the existing generators to see which all data is available). The generator function should at the end return a string which is the final generated request code. The generator functions do not currently offer async support, but if you want async support, ping @AndrewBastin.

#### Step 2: Register your new codegen object
[...]

PR #1258, opened 2020-10-09
fix:add 'helpers' folder to docker compose volumes

Ldoppea:
Hi,

I was playing with the repository so I can learn how to add a code generator for the issue https://github.com/hoppscotch/hoppscotch/issues/1190 and found that the `helpers` folder was not mapped in `docker-compose.yml`.

I added it so hot-reload can work while editing code generators.

Seems like `functions`, `lang` and `netlify` folders are missing too. I didn't add them as I don't need them for now and I don't know docker enough to know what is the impact but I can edit the PR if you want them to be added.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,6 +15,7 @@ services:
       - "./static:/app/static"
       - "./store:/app/store"
       - "./components:/app/components"
+      - "./helpers:/app/helpers"
     ports:
       - "3000:3000"
     environment:


liyasthomas:
Thanks for noticing that. Can you also confirm Docker compose is working? Follow these steps: https://github.com/hoppscotch/hoppscotch#docker-compose

Ldoppea:
Yes it is working.

To be honest I did not succeeded to build the app using `npm run dev` due to missing `@nuxtjs/pwa` package. I don't know why but `npm install` does not retrieve the pwa package on my computer (node v14.5.0 and npm  v6.14.5 on windows 10). But I did not investigate enough to create an issue.

So I tried the `docker-compose up` option (I'm a newbie to Docker). It worked from the first try but then it took me like 1hour to understand why my new code generator was not appearing in the result. Restarting the `docker-compose up` command did not work, I finally found that I had to remove the container image and then recreate it from scratch, which took a lot of time. Later I found that adding the new volume worked very well.

liyasthomas (approved):
Thanks for this PR.

PR #1287, opened 2020-10-18
Replace headers in view

### LINKED ISSUE #1257: Headers not replaced with Variables in the code export view
**Describe the bug**
Headers not replaced with Variables

**To Reproduce**
add `<<variable>>` as the value of a header, with `<<variable>>` set to something, press the export code button

**Expected behavior**
value should be something instead of `<<variable>>`

exussum12:
fixed #1257

Cant add this in the same place as the URL as you end in an infinite
loop


liyasthomas (approved):
LGTM. I like the implementation.

PR #1288, opened 2020-10-18
Added schema polling from a GraphQL

AndrewBastin:
This PR intends to change the "Get Schema" button on the GraphQL page to instead do a recurring poll of the schema, this really helps in development settings when the schema might change multiple times over the same session.


liyasthomas:
deepcode retry

liyasthomas (approved):
LGTM.

PR #1346, opened 2020-11-26
duplicate identifier for requests fixed

### LINKED ISSUE #1345: Duplicate request identifier
**Describe the bug**
There are two keys to identify a request depending on where they are shown. It's called `name` in the collections listing whereas `label` in the request section. 

**To Reproduce**
1. Simply create a collection with a request.
2. Select the request.
3. Check if the label in the request gets changed or not.

**Expected behavior**
Remove the duplicate `label` and follow the `name` since collections and folders do as well.

hariaakash:
Fixes #1345

### DIFF
--- a/components/collections/save-request-as.vue
+++ b/components/collections/save-request-as.vue
@@ -17,12 +17,11 @@
     <div slot="body">
       <ul>
         <li>
-          <label for="selectLabel">{{ $t("label") }}</label>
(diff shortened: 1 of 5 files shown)


liyasthomas:
Need to run some tests and check backward compatibility. Will review in couple of days.

liyasthomas (approved):
LGTM. Thanks for noticing the redundant key (`label`) in request object and coming with this quick fix PR.

PR #1354, opened 2020-12-01
Web Worker regex test

### LINKED ISSUE #1329: Freeze of application when trying to enter query arguments in WebSocket URL
**Describe the bug**

Typing `?` or `=` in WebSocket URLs cause freezes or let the whole application hang.

**To Reproduce**

Type a long URL as WebSocket URL. For example `ws://localhost:8000/ws/2f2dffb6-55e0-49b4-a879-8d2470dbed12`.
Now add a `?` to the end, you'll experience a freeze. Type a query argument name and try to enter `=`.
You'll experience a freeze again. I didn't wait this time since it didn't resolve after over a minute.

**Expected behavior**

To don't freeze.

**GIFs**

On the last freeze at the end of the GIF I was trying to insert `=`.

 - OS: Windows 10 Version 20H2
 - Browser: Chrome
 - Version: 86

**Additional context**

This doesn't happen on shorter URLs. For example no way to reproduce with the example URL `wss://echo.websocket.org` by just trying to add a query parameter here.
But can also be reproduced by just adding some more characters as path. The longer the whole URL gets, the worse the freeze on entering `?` or `=`.
Also happens in `Socket.IO` tab, doesn't happen for SSE and MQTT.
Also not for other top-level tabs outside of `Realtime`.

hariaakash:
Fixes #1329 
Also moved package installation in docker file before adding packages.


liyasthomas (approved):
LGTM. Thanks @hariaakash for this PR.

Squash and merging all commits.

PR #1358, opened 2020-12-01
export markdown from docs

hariaakash:
Documentation generator also generates markdown code for easy documentation export (instead of cli)


liyasthomas:
@hariaakash This is soo cool 👏👏

But we need to verify the best implementation of this feature. An "exportable Markdown" is okay - but I guess there should be no friction for users to make use of this awesome feature.

@AndrewBastin @athul wdyt

AndrewBastin:
Yayaya. Pretty awesome 🚀

hariaakash:
The idea was to easily get a static site up using the index.html which is already available with the cli. If we can document this down, then it would be neat. Another thought was to directly use the collectection.json file to directly generate docs and use it instead of the markdown file.

templates/template.md
```diff
@@ -0,0 +1,183 @@
...
+```
+{{/testScript}}
+
+{{/requests}}
+
+{{/collections}}
+
+---
+
+Generated by [HoppScotch CLI](https://github.com/hoppscotch/hopp-cli) ❤️
```
athul, line 183: @hariaakash This PR is quite cool. I tried the Deploy preview and  found it quite interesting. Saw this in the bottom of the Generated Docs 🙈

athul (commented):
Love what you did with the Markdown Template from the CLI:smile:. But there seems to be a lot of whitespace being generated(I was planning to fix that in the CLI).

Possibly the templating could do that. I'm not  familiar with mustache though.

liyasthomas:
How about this. Instead of showing the raw MD code, what if we allow users to download the `collections.md` - ready to be hostable in gist like MD render supported pages.

liyasthomas (approved):
LGTM 🚀

PR #1378, opened 2020-12-11
capitalise API if you don't mind

khalby786:


liyasthomas:
Oh, that wont be necessary.

PR #1372, opened 2020-12-09
Refactor: Modularize home page

liyasthomas:
- [x] Move `headers` section to component (@liyasthomas)

- [x] Move `parameters` section to component (@liyasthomas)

- [ ] Move `authentication` section to component

  - [x] Move `auth token list` modal to component (@liyasthomas)

  - [ ] Move `auth token generator` modal to component

  - [ ] Move `auth token request list` modal to component

- [ ] Move `pre request script` / `environments` section to component

- [ ] Move `post request script` / `tests` section to component

- [x] Move `notes` section to component (@liyasthomas)

- [x] Move `import cURL` modal to component (@liyasthomas)

- [x] Move `codegens` modal to component (@liyasthomas)

**Others:** (optional)

- [ ] Better variable names

PR #1388, opened 2020-12-15
feat: input toggle for parameters and headers

liyasthomas:
- [x] move `bodyParams` to separate component - missed in #1372 
- [x] introducing new key `active: true` (default: true) in:

  - [x] `params`

  image

  - [x] `headers`

  image

  - [x] `bodyParams`

  image

- [x] checks

  - [x] restore from history

  - [x] restore from collections

  - [x] queryString

  - [x] codegen

- [x] backward compatibility (when requests imported from History / Collections) might break, toggling 	active	 doesn't work on them since there's no such key in old format. yelp : (

> @AndrewBastin should we change `active` key to some non-destructive key like `disabled` which is `false` by default..?


AndrewBastin:
Umm, I don't understand why we have to worry about imports from old versions breaking, we just need to transform the data during the import process to the new structure, that would be enough right ? Or is there any other concern I am not seeing ?

liyasthomas:
Yep, I think that would be enough. On it.

PR #1392, opened 2020-12-18
Unit tests for test functions toBe, toHaveProperty

carlos-ds:
I was looking into writing some unit tests for the test functions that are available in the application. And this is what I've come up with so far. Feel free to add your questions and remarks. If this solution seems ok, then I'll also add some more tests for the other functions (toBeLevelxxx, toBeType, toHaveLength) in another PR.


AndrewBastin (approved):
Umm, I have fixed some of the stuff and made sure everything is working fine.

Just a few things to remember with what we are aiming as a sorta code guide for our test specs:

- use `toEqual` instead of `toBe` in test cases (I mean in Jests, not HS tests). This helps us not get caught off guard with the referential behaviours of `toBe`.
- If you make changes to the exports make sure you check the imports of the changed export is not broken and update it accordingly if so (here, you changed the `runTestScriptsWithVariables` function to a named export from default export, which broke the import in `index.vue`)

Other than that, the tests were pretty stellar!
Thank you so much for contributing unit tests to Hoppscotch. We are kinda busy with new features that we can't find time to improve our poor test coverage and these contributions really help ^_^

Hope to see more in the future!

carlos-ds:
Thanks for the feedback @AndrewBastin, that's really valuable. Pretty obvious that I should've used toEqual instead of toBe, I lost it there 🤦

Based on your feedback, I'll also add tests for the other test functions in another PR.

PR #1393, opened 2020-12-18
Fix when folder is having no request property

farhan-tariq:


liyasthomas:
@farhan-tariq can you describe what issue you were having to come up with this PR 🙂

farhan-tariq:
One of the recent upstream change caused this component to fail, if the folder did not have the request property. 

It's just some extra validation to check that the properties exist before checking their lengths

liyasthomas (approved):
LGTM.

PR #1407, opened 2021-01-02
Fix a missing close parenthesis before '.then'

sakurayang:
Origin generators:
```javascript
axios.post('address', {
    // request body
},{
  headers : {
    // request head
}.then(response => {
    console.log(response);
}).catch(error => {
    console.log(error);
})
```
after fix:
```javascript
axios.post('address', {
    // request body
},{
  headers : {
    // request head
}).then(response => {
    console.log(response);
}).catch(error => {
    console.log(error);
})
```


liyasthomas (approved):
Thanks for noticing the typo. LGTM.

PR #1438, opened 2021-01-24
feat: better media types detection for JSON, XML and HTML lenses

dunglas:
Hi!

Currently, the XML and JSON lenses only support a list of hardcoded media types, but they are able to render any format having the registered `+xml` and `+json` media suffixes.

For instance, [JSON-LD](https://json-ld.org/) (`application/ld+json`) and [ActivityPub](https://www.w3.org/TR/activitypub/) (`application/activity+json`) responses aren't rendered by the lens while there are perfectly suitable, and having the JSON preview helps a lot.

This PR uses a regex to detect JSON and XML based media types.

Example with [API Platform](https://api-platform.com), which uses JSON-LD by default:

Before:

image

After:

image

Edit: I also added support for XHTML to the HTML lens.


AndrewBastin (dismissed):
Looks super good! Thank you for the implementation!

Also, @liyasthomas Husky pre commit hook is being a bit weird, isn't the `.husky` folder supposed to be gitignore-ed ?

liyasthomas:
Yes, `.husky/` supposed yo be git ignored. Let me take a look how it behaves on a fresh install. As of now I'll just delete it from this PR.

liyasthomas (approved):
LGTM

PR #1439, opened 2021-01-24
feat: add JSON-LD to the list of known content types

dunglas:
Add [JSON-LD](https://json-ld.org/) to the list of known content types. It's a standard and a format commonly used by web APIs and in the Linked Data ecosystem (Schema.org, SOLID, ActivityPub...).

It's also used by default by [API Platform](https://api-platform.com) that I maintain. I will probably create more following PR to ensure that Hoppscotch and API Platform play well together!


liyasthomas (approved):
Thanks for the PR.

> on a side note: API platform looks slick 🚀

PR #1440, opened 2021-01-24
fix: status code detection

dunglas:
The current regular expressions detecting status code classes are incorrect. For instance `422` (which is returned by Rails, Symfony and API Platform when a request is invalid because of invalid user data) is detected as belonging to the `success` class instead of `client error` because it contains the number `2`.

The regex should be something like `/^4\d{2}$/` but using a RegExp here isn't necessary, so I switched to simple numerical comparisons (which are faster).

image

### DIFF
--- a/helpers/findStatusGroup.js
+++ b/helpers/findStatusGroup.js
@@ -1,36 +1,37 @@
-const statusCategories = [
-  {
-    name: "informational",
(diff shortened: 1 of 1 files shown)


liyasthomas (approved):
Thanks for noticing that edge case. Thanks for the PR.

PR #1441, opened 2021-01-25
Accurate response time and size

liyasthomas:
- Response time
  - [x] Axios strategy

  - [x] Extension strategy

- Response size
  - [x] Axios strategy

  - [x] Extension strategy

- Refactor response section


AndrewBastin (dismissed):
Looks good to me! 💯

AndrewBastin:
@liyasthomas @NBTX 

Can we figure out a way for the proxy to do timing from the proxy server ? Aren't we basically doing the timing from the browser still ?

liyasthomas:
There's [Server timing API](https://web.dev/custom-metrics/#server-timing-api) which does this. But since proxy request is generated from client, the actual time required to get response will be: client <=> proxy + proxy <=> endpoint. Which is better when scouts does this thing. Any thoughts @NBTX ?

AndrewBastin:
I mean, since, the proxy is the one that actually makes a request in the proxy mode. Isn't the time taken by the proxy to actually complete the request somewhat relevant ?

liyasthomas:
I guess we should indicate both 👍

Issue #1452, opened 2021-01-29
On updating the value of a path parameter the API request is not updated. 


oshhh:
**Describe the bug**
Un changing the value of the path parameter, the request is not updated and on sending the request, the request goes according to the previous value of the path parameter.

**To Reproduce**
1. Create a request containing a path parameter and set the parameter value to val1.
2. Send request.
3. Change the value of the path parameter.
4. Send the request again
4. See error: request sent according to previous value of parameter.

Also visible from the show code feature. The value of the parameter not updated in the show code feature.

**Expected behavior**
The URL should be updated on changing the path parameter.

**Screenshots**
![Screenshot 2021-01-30 at 2 28 25 AM](https://user-images.githubusercontent.com/45964755/106326676-df915600-62a2-11eb-89f1-52bf6b2e9be2.png)

Issue #1451, opened 2021-01-29
"<<" and ">>" surrounding a path parameter encoded on saving to collection


oshhh:
**Describe the bug**
On saving a request to a collection, it gets encoded. "<<" and ">>" around path_parameters also get encoded and hence a saved request is not usable directly if it contains path parameters.

**To Reproduce**
1. Create a request with URL "https://url.com/<<parameter>>" and create a path parameter. 
2. Save request to a collection.
3. Now open the saved request
4. See error: request URL is now "https://url.com/%3C%3Cparameter%3E%3E".

**Expected behavior**
URL should be "https://url.com/<<parameter>>".

**Screenshots**
![Screenshot 2021-01-30 at 2 04 24 AM](https://user-images.githubusercontent.com/45964755/106325057-321d4300-62a0-11eb-86c2-7a25bd44e6ff.png)

liyasthomas:
Thanks for noticing that issue.

PR #1453, opened 2021-01-29
fixed path parameter encoding and URL updation

oshhh:
Changed the parseTemplateString function. new URL(string).toString() would take care of bad format URLs instead of calling encodeURI explicitly as was done in this [commit](https://github.com/hoppscotch/hoppscotch/commit/9ac0cf140657167974d7b8b7015c851347139dcd#diff-ed425e40c260457bd7cafb88cf0ddc51b1895c787b93dc81c6fc31bf9496f911) to ensure double encoding doesn't occur on the URLs.

Also changed the way updation happens for URLs. On getting a URL, I am regenerating it from the URI and parameters instead of creating it in advance. Creating in advance is an issue when the parameter values are changed as mentioned in issue #1452

Solves #1451 #1452

### DIFF
--- a/helpers/templating.js
+++ b/helpers/templating.js
@@ -3,5 +3,6 @@ export default function parseTemplateString(string, variables) {
     return string
   }
   const searchTerm = /<<([^>]*)>>/g // "<<myVariable>>"
-  return decodeURI(encodeURI(string)).replace(searchTerm, (match, p1) => variables[p1] || "")
+  string = new URL(string).toString();
+  return decodeURI(string).replace(searchTerm, (match, p1) => variables[p1] || "")
 }
--- a/helpers/utils/uri.js
+++ b/helpers/utils/uri.js
@@ -3,7 +3,7 @@ export function parseUrlAndPath(value) {
   try {
     let url = new URL(value)
     result.url = url.origin
-    result.path = url.pathname
+    result.path = decodeURIComponent(url.pathname)
   } catch (error) {
     let uriRegex = value.match(/^((http[s]?:\/\/)?(<<[^\/]+>>)?[^\/]*|)(\/?.*)$/)
     result.url = uriRegex[1]
--- a/helpers/utils/uri.js
+++ b/helpers/utils/uri.js
@@ -3,7 +3,7 @@ export function parseUrlAndPath(value) {
   try {
     let url = new URL(value)
     result.url = url.origin
-    result.path = decodeURIComponent(url.pathname)
+    result.path = url.pathname
   } catch (error) {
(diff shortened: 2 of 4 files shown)


liyasthomas (dismissed):
LGTM

oshhh:
@liyasthomas added a few more changes to ensure code generation feature doesn't break. Instead of storing a decoded path, I've changed the parseTemplateString function. new URL(string).toString() would take care of bad format URLs instead of calling encodeURI explicitly as was done in this [commit](https://github.com/hoppscotch/hoppscotch/commit/9ac0cf140657167974d7b8b7015c851347139dcd#diff-ed425e40c260457bd7cafb88cf0ddc51b1895c787b93dc81c6fc31bf9496f911).

Also changed the way updation happens for URLs. On getting a URL, I am regenerating it from the URI and parameters instead of creating it in advance. Creating in advance is an issue when the parameter values are changed as mentioned in issue #1452

helpers/templating.js
```diff
@@ -3,5 +3,6 @@ export default function parseTemplateString(string, variables) {
     return string
   }
   const searchTerm = /<<([^>]*)>>/g // "<<myVariable>>"
-  return decodeURI(encodeURI(string)).replace(searchTerm, (match, p1) => variables[p1] || "")
+  string = new URL(string).toString();
```
liyasthomas, line 6: This seems to be a breaking change.
Can you assign a new flag for storing `string` instead of mutating it with the original parameter.

Deploy preview: https://deploy-preview-1453--hoppscotch.netlify.app (check console logs)
Refer: https://stackoverflow.com/questions/518000/is-javascript-a-pass-by-reference-or-pass-by-value-language

oshhh: @liyasthomas explored the error some more and it turned out the error was because of the interdependency between url and uri (and hence a recursive call of getters). Generally an update in uri leads to an update in url but on startup we give url and calculate uri from url. Now I am making sure url is depends on the value of uri and not vice versa. Please have a look and let me know if it seems fine now.

liyasthomas: Ahh, well, here's the [deploy preview](https://deploy-preview-1453--hoppscotch.netlify.app). It's failing to construct URL object from string.

oshhh: @liyasthomas I think you'd have to open the preview in incognito, the link you shared has an empty request appended to it. The [preview](https://deploy-preview-1453--hoppscotch.netlify.app/) is working fine from my side.

liyasthomas: Can you share a screenshot of the response section with the default request set up?

`GET https://httpbin.org/get`

oshhh: @liyasthomas I see what you meant, sorry. I assumed parseTemplateString is used only to parse URLs. It's breaking where the path is being passed.

PR #1488, opened 2021-02-15
[Feat: formData] Add chips to indicate chosen files in formData

nelsontky:
image

Per the image above, chips have been added to denote the files that have been chosen. This feature is still currently incomplete.

Todo:

- [x] Pressing close button on chip will lead to removal of file
- [x] Set max width of request body value field when file names are too long
- [x] Check preservation of responsive design
- [x] ~~Make sure Vuex persist works with `File` objects~~
File objects cannot be recreated, and this thus means that we will not be able to restore a user's files in a request after a reload. FormData file fields also will not be able to be synced/exported. Thus, we should exclude file fields from vuex and store them instead in a component's state.

### DIFF
--- a/components/http/http-body-parameters.vue
+++ b/components/http/http-body-parameters.vue
@@ -54,12 +54,14 @@
           @keyup.prevent="setRouteQueryState"
         />
         <div v-else class="file-chips-container">
-          <deletable-chip
-            v-for="(file, i) in Array.from(bodyParams[index].value)"
-            :key="`body-param=${index}-file-${i}`"
-          >
-            {{ file.name }}
-          </deletable-chip>
+          <div class="file-chips-wrapper">
+            <deletable-chip
+              v-for="(file, i) in Array.from(bodyParams[index].value)"
+              :key="`body-param=${index}-file-${i}`"
+            >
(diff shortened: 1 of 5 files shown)


liyasthomas:
I've updated the file chip component with matching styles 🚀 

---
image

liyasthomas:
Fixed mobile responsiveness and overflow on large file names :)

---

image

nelsontky:
Was kinda busy at work today so didnt have much time to work on this! I should be able.to.contribute more tomorrow!

nelsontky:
I've implemented the file deletion feature. Additionally, since `File` objects cannot be serialized nor constructed without having the bytes that make up the file itself, file inputs are excluded from all forms of persistence, namely via `vuex-persist`, via route queries, or via saving to collections.

I may have found a bug in the route queries too, in that for POST requests, the browser URL query params do not update immediately when the http body parameters are updated. I will observe the issue more and will open an issue if this issue hasn't been brought up before!

liyasthomas:
@nelsontky this feature is awesome.

> URL query params do not update immediately when the http body parameters are updated.

This haven't reported before.

PR #1491, opened 2021-02-17
[Feat/form data] Remove file fields from collections

nelsontky:
Since `File` is not serializable, form data fields with files cannot be saved into a collection. Thus, files are filtered out before a request is added to a collection.


liyasthomas (approved):
Thanks.

PR #1547, opened 2021-03-18
add UI for overwriting request

rohanrajpal:
user can now select the request to overwrite for my-collections

**checks:**

- [ ] remove use-request trigger from `<Request />` component while it's inside save request modal.
- [ ] move trigger from `.row-wrapper` to `<button>` component.

### DIFF
--- a/components/collections/collection.vue
+++ b/components/collections/collection.vue
@@ -132,7 +132,7 @@
             :collection-index="collectionIndex"
             :folder-index="-1"
             :folder-name="collection.name"
-            :request-index="index"
+            :request-index="collectionsType.type === 'my-collections' ? index : request.id"
             :doc="doc"
             :saveRequest="saveRequest"
             @edit-request="editRequest($event)"
--- a/components/collections/collection.vue
+++ b/components/collections/collection.vue
@@ -8,7 +8,6 @@
       @drop="dragging = false"
       @dragleave="dragging = false"
       @dragend="dragging = false"
-      @click="$emit('select-folder', { name: '', id: collection.id, reqIdx: '' })"
     >
       <button class="icon" @click="toggleShowChildren">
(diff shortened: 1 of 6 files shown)


components/collections/collection.vue
```diff
@@ -8,7 +8,7 @@
       @drop="dragging = false"
       @dragleave="dragging = false"
       @dragend="dragging = false"
-      @click="$emit('select-folder', { name: '', id: collection.id })"
+      @click="$emit('select-folder', { name: '', id: collection.id, reqIdx: '' })"
```
liyasthomas, line 11: `.row-wrapper` is used for drag and drop requests into collections and folders. I think it's better to move `@click` trigger to `<button>` component inside `.row-wrapper`.

components/collections/collection.vue
```diff
@@ -8,7 +8,7 @@
...
     >
       <button class="icon" @click="toggleShowChildren">
```
liyasthomas, line 13:
```
      <button class="icon" @click="toggleShowChildren">
```
Button component already has `@click="toggleShowChildren"` trigger. `select-folder` trigger should be appended to this `@click`.

liyasthomas: Also on a second thought, it's efficient to trigger the new `select-folder` trigger only if the component is inside save request modal. I mean there's a flag that tells `<Collection />` component that it's being displayed inside save request modal right? Use that to conditionally trigger `select-folder` trigger.

liyasthomas (changes requested):

components/collections/collection.vue
rohanrajpal: on it

components/collections/collection.vue
rohanrajpal: got it, yeah conditionally emitting sounds better

liyasthomas (approved):

PR #1554, opened 2021-03-21
Show sub-folders first instead of requests

farhan-tariq:
Sub-folders show at the bottom of the list. You have to scroll all the way down to reach a folder you want to access.
Depends on user preference I believe but makes more sense the other way round

Before
image

After
image


liyasthomas (approved):
This makes more sense, since in collections, we're showing folders first. Thanks for the quick PR.

PR #1562, opened 2021-03-24
Sync bodyParams and rawBodyParams

### LINKED ISSUE #1435: Provide a visual editor for JSON/KV request bodies
Users cannot send JSON to an endpoint if the endpoint requires numbers to be JSON numbers. Currently Hoppscotch serializes all parameters as JSON strings. 

User input:
image
Expected behaviour: user can either select the parameter type and/or the value gets automatically converted into the proper type. 

Can always use the raw request bodies, but that is less convenient.

Reflex-Gravity:
This PR fixes #1435

I have added the option to sync across Request Body params and Request Body raw Input.

I believe that even though the code works, I may have not followed the code standard of this code base. Let me know if I have to do any changes.


liyasthomas (approved):

liyasthomas:
There's a active <=> inactive toggler for key-value parameters. I guess we should respect that before converting to raw body. Just check if the item in key-value pair has `active` property. If there exists an `active property, check if it's true. if true - convert to raw. also if there doesn't exist `active` property - also convert to raw.`

components/http/BodyParameters.vue
```diff
@@ -184,6 +176,21 @@ export default {
         fileIndex,
       })
     },
+    updateBodyParams(event, index, type){
+      this.$store.commit(type, {
+        index,
+        value: event.target.value,
+      })
+      let rawParmas = {}
+        this.$store.state.request.bodyParams.forEach(_param=>{
```
liyasthomas, line 185:
```suggestion
        this.$store.state.request.bodyParams
        .filter((item) => (item.hasOwnProperty("active") ? item.active == true : true))
        .forEach(_param=>{
```

liyasthomas: Checks active property before making raw body from key-value pair.

liyasthomas (changes requested):

components/http/BodyParameters.vue
snapeur82: Yessss,,,
Love un marci

snapeur82:
Yessss,,,

Le jeu. 25 mars 2021 14 h 04, Liyas Thomas ***@***.***> a
écrit :

Reflex-Gravity:
> There's an active <=> inactive toggler for key-value parameters. I guess we should respect that before converting to raw body. Just check if the item in key-value pair has `active` property. If there exists an `active property, check if it's true. if true - convert to raw. also if there doesn't exist `active` property - also convert to raw.`
> 
> Also `delete` and `active <=> inactive` toggle doesn't seems to be updating raw body since it's tied to `@change` event listener which only activates when there's a user input in input box. Need to add events on both active toggle and delete button.

Sure. I'll update these too.

AndrewBastin (approved):
I don't see anything generally wrong. I am approving this.

liyasthomas:
@Reflex-Gravity let me know if you need any help with those pending tweaks.

Reflex-Gravity:
> @Reflex-Gravity let me know if you need any help with those pending tweaks.

I have done the requested changes.

liyasthomas (approved):

liyasthomas:
@Reflex-Gravity thank you.

PR #1564, opened 2021-03-26
Refactor: Combine rest and graphql collections

IshaGupta18:
Passed a `type` parameter to differentiate between rest and graphql collections.

All files expect `index.vue` can be deleted in the `components/collections/graphql` folder.

**checks:**
- [x] check if `flag <=> collectionType` is breaking any previous collections operations
- [x] rest collection
- [x] graphql collections


components/collections/Add.vue
```diff
@@ -58,19 +59,22 @@ export default {
     syncCollections() {
       if (fb.currentUser !== null && this.SYNC_COLLECTIONS) {
         fb.writeCollections(
-          JSON.parse(JSON.stringify(this.$store.state.postwoman.collections)),
-          "collections"
+          this.$props.type == "rest"
+            ? JSON.parse(JSON.stringify(this.$store.state.postwoman.collections))
+            : JSON.parse(JSON.stringify(this.$store.state.postwoman.collectionsGraphql)),
+          this.$props.type == "rest" ? "collections" : "collectionsGraphql"
```
liyasthomas, line 65: this is an overkill. can't we use something like this..?

```suggestion
          JSON.parse(JSON.stringify(this.$props.type == "rest" ? this.$store.state.postwoman.collections : this.$store.state.postwoman.collectionsGraphql), this.$props.type == "rest" ? "collections" : "collectionsGraphql")
```

liyasthomas (approved):
LGTM

PR #1585, opened 2021-04-10
refactor GraphQL queries in collection.vue

rohanrajpal:
Shift the queries to inside the component for both requests and folders


liyasthomas (approved):
LGTM

PR #1643, opened 2021-05-16
Removed Redundant code in nuxt.config.js

bhavukkalra:
nuxt.config.js seems to already have access to all the environment variables present in root directory's .env file via `process.env.VARIABLE_NAME`. So, this require statement can be removed

### DIFF
--- a/nuxt.config.js
+++ b/nuxt.config.js
@@ -1,5 +1,3 @@
-require("dotenv").config()
-
 // Common options
 export const options = {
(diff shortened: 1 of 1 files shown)


liyasthomas:
I guess you need to prepend `require("dotenv").config()` to use `.env` file in `nuxt.config.js`.

Refer nuxt/dotenv: https://github.com/nuxt-community/dotenv-module#using-env-file-in-nuxtconfigjs

PR #1696, opened 2021-06-01
Changing Hoppscotch Title from Hs to Hoppscotch for all sizes

JoelJacobStephen:
Changed the Title from Hs to Hoppscotch for all sizes. I feel that using Hs is a waste of space especially on devices that have huge width like the iPad. I also tried testing this on most devices including small devices like the iPhone SE and the Galaxy Fold. It worked fine for all devices that I tested.


liyasthomas:
@JoelJacobStephen thank you for this quick PR 🙌

liyasthomas (approved):
LGTM

Issue #1691, opened 2021-05-31
Error when saving requests with multipart/form-data content type


kaname-png:
**Describe the bug**
When creating a `multipart/form-data` content type request it works correctly, but saving the request and selecting again this saved request changes the content type to `application/json`, it is very annoying to be changing the content type all the time.

**To Reproduce**
1. Go to 'Create a request'
2. Click on 'Change content type to `multipart/form-data`'
3. Click on 'Save request'
4. Go to the 'Collections and click on the saved request'

**Expected behavior**
Correctly display the request with the content type you assign before saving the request.

**Screenshots**
![wmtraJ0k6p](https://user-images.githubusercontent.com/56084970/120213655-adf8af00-c1f0-11eb-92d2-d638d25b2e3a.gif)

**Desktop:**
 - OS: Windows 10
 - Browser Microsoft Edge
 - Version 91.0.864.37

liyasthomas:
Sorry you faced such behavior. We'll look into this and get back to you as soon as possible.
Thank you for your patience.

nelsontky:
I was the guy who did the `multipart/form-data` feature. Maybe I could look into it!

PR #1703, opened 2021-06-05
fix: prevent overwrite of content type

nelsontky:
This PR fixes issue #1691. This fix prevents the changing of HTTP method to overwrite the content type of a request if the content type is already set.  Moreover, the `AutoComplete` component watches for updates to its `value` prop so as to update the input field contents when the `value` prop changes.


liyasthomas (approved):
LGTM

PR #1709, opened 2021-06-08
Made changes to Black Mode to make UI elements more visible.

JoelJacobStephen:
The sections in the current Black Mode is not quite visible as it's too dark. So made changes to the some of the colors in the black mode to make the sections and tab headers pop put more from the black background.


liyasthomas (approved):
Thank you for the PR - LGTM.

PR #1727, opened 2021-06-19
Cleaner Docker run command

d4v3y0rk:
added two flags to the docker run command to clean up the container after it has been run and also assign it a name to help manage it while it is running.


liyasthomas (approved):
Thank you for this quick PR - it's a nice addition. LGTM.

PR #1737, opened 2021-07-01
fix: duplicated input ids

s-r-x:
There are some duplicated inputs' ids on the page. 
To reproduce:
- click "save to collections"(gql or home page - that's doesn't matter) and check out the id of the "Request name" input.
- click "new" and compare the "Label" input id with the one from the first step. they are the same.
Sometimes duplicated ids  can lead to some really weird behavior. in this case if you click on the label from the second step - input from the first one gets focused.

### DIFF
--- a/components/collections/SaveRequest.vue
+++ b/components/collections/SaveRequest.vue
@@ -9,15 +9,15 @@
       </div>
     </template>
     <template #body>
-      <label for="selectLabel">{{ $t("token_req_name") }}</label>
+      <label for="selectLabelSaveReq">{{ $t("token_req_name") }}</label>
       <input
-        id="selectLabel"
+        id="selectLabelSaveReq"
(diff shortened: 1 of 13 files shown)


components/collections/SaveRequest.vue
```diff
@@ -9,15 +9,15 @@
...
+      <label for="selectLabelSaveReq">{{ $t("token_req_name") }}</label>
       <input
-        id="selectLabel"
+        id="selectLabelSaveReq"
         v-model="requestData.name"
         class="input"
         type="text"
         @keyup.enter="saveRequestAs"
       />
-      <label for="selectLabel">Select location</label>
```
s-r-x, line 20: i'm not quite sure what is the purpose of this one

liyasthomas (approved):
LGTM

PR #1742, opened 2021-07-14
fix: undefined duration

s-r-x:
Screenshot_20210714_103933

duration is always undefined when there was any error during request


liyasthomas (approved):
LGTM

PR #1762, opened 2021-08-06
fix: pkce params match rfc naming convention

hughobrien:
RFC: https://datatracker.ietf.org/doc/html/rfc7636


liyasthomas (approved):
Thanks for this quick PR. LGTM.

PR #1764, opened 2021-08-07
fix: explicit package.json deps: allows pnpm import/install

hughobrien:
these deps are transitive, but pnpm is strict about that so they are missing when using pnpm import/install. This PR makes them explicit.


liyasthomas:
We're considering officially supporting `pnpm` for the next major release. We'll make sure all deps are pnpm compatible and merge along with this PR. Thanks for the quick PR - we'll get back to this after 1 week.

liyasthomas:
With the release of v2 - this PR is now absolute.

We've added "self-hosting" in our roadmap for the next milestone releases. Watch this space for out-of-the-box support to `pnpm` soon.

PR #1765, opened 2021-08-07
fix: oauth token exchange param naming convention

hughobrien:
https://developer.okta.com/docs/guides/implement-auth-code-pkce/exchange-code-token/


liyasthomas (approved):
LGTM.

hughobrien:
thanks for the quick merges! (and the great tool)

PR #1766, opened 2021-08-07
fix: bad var names from oauth well.known config

hughobrien:
e.g. https://swagger.io/docs/specification/authentication/openid-connect-discovery/

I smell a rogue linter..,


liyasthomas (approved):
Thanks for noticing these bugs which are not easily discoverable. Typo or something might've caused this. LGTM.

PR #1769, opened 2021-08-09
fix: broken settings persistence on the home page

s-r-x:
The problem:
https://user-images.githubusercontent.com/41614937/128659587-6ebd297b-7f61-4db1-bf04-41c47e4d76f9.mp4

[This line](https://github.com/hoppscotch/hoppscotch/blob/main/newstore/localpersistence.ts#L32) always returns true since vuexData.postwoman.settings is an object(somehow)
[This one](https://github.com/hoppscotch/hoppscotch/blob/main/newstore/localpersistence.ts#L30) also always returns true because we are using shallow comparison for different objects, so _.isEmpty is a better choice here.
I don't fully understand the background of these migrations, but i took responsibility to refactor this file a little bit(the postwoman part of it), looks like it shouldn't break anything

### DIFF
--- a/newstore/localpersistence.ts
+++ b/newstore/localpersistence.ts
@@ -2,7 +2,7 @@
 
 import clone from "lodash/clone"
 import assign from "lodash/assign"
-import eq from "lodash/eq"
+import isEmpty from "lodash/isEmpty"
 import {
   settingsStore,
   bulkApplySettings,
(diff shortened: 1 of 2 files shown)


AndrewBastin:
The core reason why this happened is because of a thing with Vuex regenerating the settings data everytime on save rather than the migration logic failing. But even then, your implementation for migrations seems a bit more stable-ish.

Thank you for your PR @s-r-x !

AndrewBastin (approved):

PR #1789, opened 2021-08-28
fix: fix typos in simplified Chinese translation

kdxcxs:


liyasthomas:
@kdxcxs thank you for the quick translation and PR.

PR #1794, opened 2021-08-31
Update korean translation

dksgusals99:


liyasthomas (approved):
Thanks for the quick PR. LGTM.

PR #1797, opened 2021-08-31
Fix some mistranslations an typos.

jenskeiner:
Hi, while skimming through the German version of the app, I noticed some mistranslated words and sentences as well as some typos. Tried to fix up all I could find, though I would swear there's more polishing that could be done. However, I haven't really seen every corner of the app yet, so these changes are hopefully at least a good first step to improve the wording.

### DIFF
--- a/locales/de.json
+++ b/locales/de.json
@@ -1,44 +1,44 @@
 {
   "action": {
-    "cancel": "Stornieren",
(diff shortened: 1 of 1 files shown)


liyasthomas:
Thanks for the quick PR. Translations are updated once every week. Looking forward to having more contributions from you.

PR #1805, opened 2021-09-04
chore: more accurate chinese translation

Cluas:


Cluas:
I see that there are other branches that have been optimised, I'll close this pr @liyasthomas

PR #1815, opened 2021-09-06
chore(i18n): fixed french translations

bokub:
I tried hopscotch in french and found out the translations were really bad 😄 

I fixed the most obvious ones, but I didn't have time to go deeper. I'm sure there are still a lot of errors that didn't catch my attention, but these files are loooong!

I might open another PR in a few days, but maybe not, depends if I find some free time.

P.S: I'm french.


liyasthomas:
Thanks for the effort and this quick PR. Really appreciate it.

Btw, sorry — our translations are mostly machine translated and I admit it won't be the best. But thanks to contributors like you.

bokub:
You're welcome!

I know that Home Assistant (which I used to contribute to), use [Lokalize](https://lokalise.com/) to manage translations.

You should give it a look IMO!

liyasthomas (approved):

liyasthomas:
Will definitely take a look into it. Currently, we use: [GitLocalize](https://gitlocalize.com/repo/6398)

bokub:
Oooh, I didn't realize you were already using a tool.

Maybe you should put a link in the app, something like "Help us translate !", and maybe not use translations made by a machine and use English as a fallback 🤷 

I don't know for other languages, but for french, some buttons or links were unable to understand, and I had to switch to English to understand what they meant 😬

PR #1844, opened 2021-10-01
Remove random </br>s in locale footers

khalby786:
In certain languages, the "Help & feedback" option in the footer will have a `</br>` tag in it as text (which is not even valid HTML). In one case, `</br>` was literally translated into some other language. I'm guessing this is from before Hoppscotch's recent UI update(?).

image


liyasthomas (approved):
Hi @khalby786, thanks for this quick PR. LGTM.

Issue #1847, opened 2021-10-02
Enable copy/duplicate menu option for environments


StephaneBischoff:
**Is your feature request related to a problem? Please describe.**
Its a bit of a pain to manage environments usually. You sometime start off writing a few variables for an environment and then need to do the same for the PROD, QA and other environments. 

Why require users to rewrite the same variable name all over again?

**Describe the solution you'd like**
I would like to see an option that copies/duplicate your environment.

**Describe alternatives you've considered**
None for the moment

**Additional context**
Here is what I am thinking:

- Add a duplicate option in the context menu
- When click, creates a new copy of the environment with a different name '{Old name} Copy 1'
    - If copied again then name '{Old name} Copy 2'

![copy-paste](https://user-images.githubusercontent.com/9750058/135728960-ba29b9d9-6ba4-41bd-b42e-b65dc9ffcb97.png)

StephaneBischoff:
Hi. I would like to add this feature if its ok?

liyasthomas:
Hi @StephaneBischoff, you're free to pick up this feature. Do let us know if you got stuck.

PR #1848, opened 2021-10-02
feat: add ability to copy environments

StephaneBischoff:
Submitted PR for feature request: #1847 

Copy Env

### DIFF
--- a/packages/hoppscotch-app/components/environments/Environment.vue
+++ b/packages/hoppscotch-app/components/environments/Environment.vue
@@ -43,6 +43,17 @@
         />
         <SmartItem
           v-if="!(environmentIndex === 'Global')"
+          svg="copy"
+          :label="`${$t('action.copy')}`"
+          @click.native="
+            () => {
+              duplicateEnvironment()
+              $refs.options.tippy().hide()
+            }
+          "
+        />
+        <SmartItem
+          v-if="!(environmentIndex === 'Global')"
           svg="trash-2"
           color="red"
           :label="`${$t('action.delete')}`"
@@ -66,7 +77,10 @@
 
 <script lang="ts">
 import { defineComponent, PropType } from "@nuxtjs/composition-api"
-import { deleteEnvironment } from "~/newstore/environments"
+import {
+  deleteEnvironment,
+  duplicateEnvironment,
+} from "~/newstore/environments"
 
 export default defineComponent({
   props: {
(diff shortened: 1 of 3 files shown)


packages/hoppscotch-app/components/environments/Environment.vue
```diff
@@ -41,6 +41,17 @@
             }
           "
         />
+        <SmartItem
+          v-if="!(environmentIndex === 'Global')"
+          svg="copy"
+          :label="`${$t('action.copy')}`"
```
liyasthomas, line 47: I prefer the button should read "Duplicate" instead of "Copy".
Feel free to add a new string entry to our English translation file, here under `action`: https://github.com/hoppscotch/hoppscotch/blob/main/packages/hoppscotch-app/locales/en.json

packages/hoppscotch-app/newstore/environments.ts
```diff
@@ -76,6 +77,26 @@ const dispatchers = defineDispatchers({
...
+      return {
+        environments,
+      }
+    }
+    return {
+      environments: [
+        ...environments,
+        {
+          ...cloneDeep(newEnvironment),
+          name: `${newEnvironment.name} - ${environments.length + 1}`,
```
liyasthomas, line 95: Appending `environments.length + 1` to duplicated entry's name doesn't feel right to me, instead, how about appending the incrementing count if the environment has entries with the same name.

Ex: Before duplicating `Foo`
Environments:
- Foo
- Bar
- Choco
- Baz

After duplicating `Foo`
Environments:
- Foo - 2  // instead of `Foo - 5`<-total environments.length+1
- Bar
- Choco
- Bax

Ideally the above should make more sense with duplicate's naming, or we can prefer something very simple like `Duplicate of <old environment name>` for the new entry's name.

liyasthomas (changes requested):
Hi @StephaneBischoff, thanks for this quick PR - really appreciate it. I've shared few things that would improve the feature UX. Let me know what you think.

packages/hoppscotch-app/components/environments/Environment.vue
StephaneBischoff: Yea I agree. Will do the change. 
What about the other translations?

packages/hoppscotch-app/newstore/environments.ts
StephaneBischoff: Thanks @liyasthomas Yes, good suggestions. I would prefer going with something very simple. I like you 2 option :)

packages/hoppscotch-app/components/environments/Environment.vue
liyasthomas: You can omit other translations except English. Other will be updated once every week.

liyasthomas (approved):
LGTM

AndrewBastin (approved):

PR #1850, opened 2021-10-03
Feature/1842 make body stored on query vars

### LINKED ISSUE #1842: Body is not stored at query vars when using "Copy link" feature
**Describe the bug**
I would like to use the Hoppscotch query string API to build Hoppscotch links in my application so we can debug REST calls from our HTTP test reports. 

The idea behind is to make a "Play" button in our reports that link to a pre-filled Hoppscotch API tester. We already can do this but we can only fill "parameters" (query string), "headers", "method", and "endpoint". There is no such "body" parameter to pre-fill the body of the request. 

**To Reproduce**
Open a new Hoppscotch session and fill endpoint, method, headers, parameters and body. You will see that the link at the address bar will include all of those information but the body part.

**Expected behavior**
Include "body" parameter with the url-encoded body value. 

**Additional context**
I don't know if this is a bug report or a feature request since I don't know if this is made on purpose. But it seems weird for us that such as an important part as the body is not present at the Hoppscotch query vars API.

Thank you, greetings.

mandaputtra:
Related issue #1842. Tell me if there is any test that I should change, or provide. I'm new on testing (never done UI testing before)

fixes #1842


AndrewBastin (approved):
LGTM 💯 

Thank you for the quick PR @mandaputtra!!!! 🤗 

Will merge as soon as the checks get done :)

Issue #1828, opened 2021-09-13
Better error messages for Graphql


kaname-png:
**Is your feature request related to a problem? Please describe.**
Currently the application only says to press F12 to see the error, but the error is not readable.

# hoppscotch
![image](https://user-images.githubusercontent.com/56084970/133028055-f17bf29b-b641-4e60-a758-7e4fc49f849d.png)

# Graphql Playground
![image](https://user-images.githubusercontent.com/56084970/133028177-e6093d7e-18ef-45eb-b43f-597c36b37def.png)

With the Graphql Playground error message you can easily know what is wrong, while with the hoppscotch error message you don't know what is wrong because the error is not readable.

**Describe the solution you'd like**
Display a more readable error message to understand what is going wrong.

**Describe alternatives you've considered**
None.

**Additional context**
No.

H-thompson:
I would like to take this feature on if it's not already being worked on. As part of the process of pulling through the error data I will ensure that it fits with the current localisation structure.

liyasthomas:
Hi @H-thompson thank for showing the interest to work on this.

StephaneBischoff:
Hi @H-thompson @liyasthomas proposed a very simple fix. Let me know if you want some more improvements.

kaname-png:
No, that's fine, thanks!

PR #1852, opened 2021-10-03
Show graphql error message

StephaneBischoff:
Fix proposal (very simple) for request #1828 

I was not clear from the original ticket what the requirements were. 
- Like the toast message format might not be ideal. Maybe could indicate that it was a HTTP 400 error for example instead of the entire error message. 
- Also, should we continue to say to go and check the logs for futher details?
- I did not test with proxies. Seems like the request made is different in that case. (not using `responseType: "arraybuffer"`

Let me know what you think. thanks

error-message

### DIFF
--- a/packages/hoppscotch-app/components/graphql/RequestOptions.vue
+++ b/packages/hoppscotch-app/components/graphql/RequestOptions.vue
@@ -447,7 +447,7 @@ const runQuery = async () => {
       icon: "done",
     })
   } catch (e: any) {
-    response.value = `${e}. ${t("error.check_console_details")}`
+    response.value = `${e}.\n\n${t("error.check_console_details")}`
     nuxt.value.$loading.finish()
 
     $toast.error(`${e} ${t("error.f12_details")}`, {
--- a/packages/hoppscotch-app/components/graphql/RequestOptions.vue
+++ b/packages/hoppscotch-app/components/graphql/RequestOptions.vue
@@ -447,10 +447,12 @@ const runQuery = async () => {
       icon: "done",
     })
   } catch (e: any) {
-    response.value = `${e}.\n\n${t("error.check_console_details")}`
+    response.value = `${e}`
     nuxt.value.$loading.finish()
 
-    $toast.error(`${e} ${t("error.f12_details")}`, {
+    $toast.error(
+      `${t("error.something_went_wrong")}. ${t("error.check_console_details")}`,
+      {
       icon: "error_outline",
     })
     console.error(e)
--- a/packages/hoppscotch-app/components/graphql/RequestOptions.vue
+++ b/packages/hoppscotch-app/components/graphql/RequestOptions.vue
@@ -453,8 +453,9 @@ const runQuery = async () => {
     $toast.error(
       `${t("error.something_went_wrong")}. ${t("error.check_console_details")}`,
       {
-      icon: "error_outline",
-    })
+        icon: "error_outline",
+      }
+    )
     console.error(e)
   }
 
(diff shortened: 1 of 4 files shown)


liyasthomas:
@StephaneBischoff, to follow the best practices — you can try making seperate branches for each new features and thereby keep the commit history clean.

On this PR, you can rebase to the current main branch and fix the commit history.

kaname-png:
In my opinion it would be nice if the error message was formatted.

packages/hoppscotch-app/helpers/strategies/AxiosStrategy.js
```diff
@@ -52,12 +52,13 @@ const axiosWithoutProxy = async (req, _store) => {
...
       responseType: "arraybuffer",
     })
-
     return res
   } catch (e) {
     if (axios.isCancel(e)) {
       // eslint-disable-next-line no-throw-literal
       throw "cancellation"
+    } else if (e.response.data) {
+      throw new Error(Buffer.from(e.response.data, "base64"))
```
liyasthomas, line 61:
```suggestion
      throw new Error(Buffer.from(JSON.stringify(e.response.data, null, 2), "base64"))
```

This should beautify the error message.

StephaneBischoff: Just saw that there are unit tests. Will add those.

AndrewBastin:
I actually think the response actually should reflect the error JSON itself purely to not mess with the syntax highlighting. Instead we can have an indicator somewhere and a toast or something like that (@liyasthomas thoughts ?)

StephaneBischoff:
> In my opinion it would be nice if the error message was formatted.

Here is new formatting. Let me know what you think? thanks.

image

liyasthomas:
i guess @AndrewBastin is right. We shouldn't append anything to error message and it's better to show the error object as it is. Though, a JSON formatting for readability won't do any harm.

@StephaneBischof, we can remove the prefix: "Error:" and suffix: "Check console for details" strings from error object.

packages/hoppscotch-app/components/graphql/RequestOptions.vue
```diff
@@ -447,7 +447,7 @@ const runQuery = async () => {
       icon: "done",
     })
   } catch (e: any) {
-    response.value = `${e}. ${t("error.check_console_details")}`
+    response.value = `${e}.\n\n${t("error.check_console_details")}`
     nuxt.value.$loading.finish()
 
     $toast.error(`${e} ${t("error.f12_details")}`, {
```
liyasthomas, line 453:
```suggestion
    response.value = `${e}`
    nuxt.value.$loading.finish()

    $toast.error(
      `${t("error.something_went_wrong")}. ${t("error.check_console_details")}`,
      {
```

packages/hoppscotch-app/components/graphql/RequestOptions.vue
```diff
@@ -447,10 +447,12 @@ const runQuery = async () => {
...
-    response.value = `${e}. ${t("error.check_console_details")}`
+    response.value = `${e}`
     nuxt.value.$loading.finish()
 
-    $toast.error(`${e} ${t("error.f12_details")}`, {
+    $toast.error(
+      `${t("error.something_went_wrong")}. ${t("error.check_console_details")}`,
+      {
       icon: "error_outline",
     })
```
liyasthomas, line 457:
```suggestion
    $toast.error(
      `${t("error.something_went_wrong")}. ${t("error.check_console_details")}`,
      {
        icon: "error_outline",
      }
    )
```

liyasthomas (approved):
LGTM.

Issue #1840, opened 2021-09-27
Control arrow right and left still bind to change history


mandaputtra:
**Describe the bug**
Mostly when I navigate the text, I use `CTRL + <-` or `CTRL + ->` to navigate per word, but when I use it on the body editor it navigate through my browser history not per word.

**To Reproduce**
1. Open Hoppsctoot
2. Go to body 
3. Change content type to application/json, 
4. Write stuff,
5. Navigate using CTRL + Right  or CTRL + Left

**Expected behavior**
Navigating per word

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.10
 - Browser: Brave latest, Chromium (94) latest.
 - Version  V1.29.81 (Sep 14, 2021)

**Additional context**
I thik I can contribute to solve this issue. Tell me If this is necessary I'll work on it.

liyasthomas:
hi @mandaputtra, thanks for reporting this behavior. We'd be happy if you can figure out a fix to this.

I'd like to mention our universal action handler and resolver logic which is probably the root cause of this behavior.

`actions.ts` defines the key bindings and it's corresponding action keywords.
Here it's `navigation.jump.back` and `navigation.jump.forward`
https://github.com/hoppscotch/hoppscotch/blob/728515c2255692f8a6a2cbe8866bd8d5e418d2cf/packages/hoppscotch-app/helpers/actions.ts#L30-L31

`default.vue` defines the corresponding action handlers.

https://github.com/hoppscotch/hoppscotch/blob/dcbb17b1642d07da0a22ef3171e4572dd82dd93d/packages/hoppscotch-app/layouts/default.vue#L127-L132

A universal fix to actions interfering the native behaviors like navigate per word is to disable them when the focus is in `textareas`, `input`, `contenteditable` etc. This has been already implemented in `isTypableElement()` in `dom.ts`.

https://github.com/hoppscotch/hoppscotch/blob/dcbb17b1642d07da0a22ef3171e4572dd82dd93d/packages/hoppscotch-app/helpers/utils/dom.ts#L5-L19

Guess this will be helpful.

deepto98:
@liyasthomas Hi, I'd like to work on this issue. If it isn't assigned already, can you please assign this to me?

mandaputtra:
Go ahead if you want. I actually want to work on this weekend. Got so much to do on weekday :((

deepto98:
[...]

PR #1862, opened 2021-10-08
route navigation no longer happening in typable areas

rishabhkalra96:
The PR fixes the issue #1840 which earlier was navigating the user to previous / next route on clicking ctrl + left / right arrows on linux and optn + left / right arrows on mac for all browsers.


mandaputtra:
This is fixed the issue but I can't do another shortcut when I'm in the text area or input. Could you make it the other short cut work? 

https://user-images.githubusercontent.com/23342943/136638045-af7dadde-ec01-4ca0-a84d-4c51df6a18c2.mp4

liyasthomas:
@mandaputtra, I guess it's safe to disable all app shortcuts while the user is focused on native elements like inputs, textareas etc. This way, we can make sure the natives shortcuts are not being overridden with app shortcuts.

Issue #1875, opened 2021-10-14
[feature]: Ability to save API response


bhav15h:
**Is your feature request related to a problem? Please describe.**
I would like to be able to save an API response similar to the feature in postman. 
1. This will then allow us to preview the API instead of having to do a request to get the response. 
2. Also this is helpful whereby if you use a token to authenticate and the token has expired, instead of reauthenticate to view a response you get to quickly see the structure and response of the API.

**Describe the solution you'd like**
The solution for this is when a request to an API has returned a result we have a new button to save the response. This would be nested under the saved API request. 
We should be able to rename the saved response as there could be different responses from 1 request. e.g if an exception was raised in the api we could receive a different response.

**Describe alternatives you've considered**
At the moment the only alternative is to save the response in a file locally, but this is very hectic and not easily sharable to other people in the team as we may have a long list of 100 APIs. 

**Additional context**
Postman seem to have done it very well.

sguzmanm:
Hi, this seems interesting, mind if I look into it?

liyasthomas:
@sguzmanm yes, feel free to pick this up.

Selvestr1:
> **Is your feature request related to a problem? Please describe.** I would like to be able to save an API response similar to the feature in postman.
> 
> 1. This will then allow us to preview the API instead of having to do a request to get the response.
> 2. Also this is helpful whereby if you use a token to authenticate and the token has expired, instead of reauthenticate to view a response you get to quickly see the structure and response of the API.
> 
> **Describe the solution you'd like** The solution for this is when a request to an API has returned a result we have a new button to save the response. This would be nested under the saved API request. We should be able to rename the saved response as there could be different responses from 1 request. e.g if an exception was raised in the api we could receive a different response.
> 
> **Describe alternatives you've considered** At the moment the only alternative is to save the response in a file locally, but this is very hectic and not easily sharable to other people in the team as we may have a long list of 100 APIs.
> 
> **Additional context** Postman seem to have done it very well.

m-roberts:
+1 for this feature. Postman has the ability to save an API response as an "example" under requests in a collection.

AndrewBastin:
This is something we have debated a lot on whether or not to implement, even had the PR in. But what we think is "Examples" as a thing within requests are more of a documentation thing and not exactly a "Testing" thing.

[...]

PR #1894, opened 2021-10-24
Save example for request

sguzmanm:
Hi, this is a draft for saving examples for a request. Honestly I have too many doubts regarding the change and how big it will be so I'm leaving here a demo of the current functionality to check if the change is going the right way and how I can improve it for merging: 

https://user-images.githubusercontent.com/25377562/138613704-314cb555-8ce9-4278-98b3-b8774194e132.mov

So things I want to add to this change for a V1 version: 
- Change the example name

Doubts I have:
1. Define what features should be visible when looking at a request example (for example, I'm not sure if it makes sense to add params when looking at an example) 
2. In which scenarios is saveRequest prop for components set to true? This is used mostly on the code for the Collections list in the right
image
3. As the code is today, to support saving examples we would have to change all lens renderers, user and global collections, and REST and GraphQL request views. Is that OK? Also I think we can do this in multiple PRs or just add the example feature with some sort of BETA flag to see if users like it and change other views progressively
4. Is there any test suite for adding new changes to the repo? I see that there are multiple triggers or listeners involved, like Firebase for instance

Contributes to #1875


Sismars19:
Hi, how are you? Do you want to have a good investment ? please click this one, thank you very much

https://vuuzletvph.com/what-is-the-secret-of-nft-tokens-and-why-they-are-bought-for-millions-of-dollars/

What is the secret of NFT tokens and why they are bought for millions of dollars - Vuuzle.TV Philippines
vuuzletvph.com

What is the secret of NFT tokens and why they are bought for millions of dollars - Vuuzle.TV Philippines
vuuzletvph.com

liyasthomas:
This PR has gone stale, hence closing. We're working on another viable solution. Stay tuned for future updates.

codeasashu:
@liyasthomas Any update on this?

Issue #1928, opened 2021-11-02
When l click "Bulk edit", the params l input in "key-value edit" will not be showed.


renmu123:
**Describe the bug**
When l click "Bulk edit", the params l input in "key-value edit" will be not showed.

**Expected behavior**
Params l inputed in "key-value edit" can be edited in "Bulk edit" just like postman

**Screenshots**
![image](https://user-images.githubusercontent.com/26896790/139840409-7021a782-16b7-43c1-af5d-a63252a17b2b.png)
![image](https://user-images.githubusercontent.com/26896790/139840504-7a225fbe-d765-43ba-b59c-777174b932e3.png)

**Desktop (please complete the following information):**
 - OS: Win01
 - Browser Chrome PWA
 - Version 2.0

liyasthomas:
Hi @renmu123, thanks for requesting this feature.

We're aware of this behaviour and will try to improve Bulk Editor with two-way binded data support in near future.

PR #1930, opened 2021-11-02
sync request params with bulk editor.

0xc0Der:
in this PR:

- synced Parameters with bulk editor.
- enabled clearAll button in bulk editor.

fixes issue #1928.


packages/hoppscotch-app/components/http/Parameters.vue
```diff
@@ -204,32 +203,40 @@ useCodemirror(bulkEditor, bulkParams, {
...
+      `${!item.active ? "// " : ""}${item.key}: ${item.value}`
     )
-      addParam()
-  },
-  { deep: true }
-)
+  else params.splice(index, 1)
+
+  bulkParams.value = params.join("\n")
+}
```
liyasthomas, line 218: Hi @0xc0Der, this doesn't ignore the empty fields in the list and sets the last entry which has a key-value pair.

I guess we should keep the flexibility to keep empty entries in key-value pairs as well as empty lines in bulk editors since timing those is causing Cumulative Layout Shift (CLS) throughout the UI.

What do you think?
Everything else looks good though.

0xc0Der: so, the empty fields you added at the end which went away when you started editing is an expected behavior considering the structure of the code. then the field you added got appended at the end.
did i get that right?
keeping empty fields at the end is not important in my opinion _we have an "add" button_ but the empty fields in the middle are preserved as you see.
am i off point here? please correct me if i am wrong.

0xc0Der: if you want to preserve these empty fields, i am working on it.
another problem to think about:
bulk editor contents does not get saved in local storage, then it does not get set at startup.

liyasthomas (approved):
LGTM

PR #1936, opened 2021-11-04
fix: bind key-value with bulk editor at startup.

0xc0Der:
further testing of my last PR. I found a **huge bug** that causes the key-value pairs get deleted if i try to edit them after a fresh reload.
the PR fixes it completely.

in this PR:
- fix `editBulk[...]Line` algorithm.
- add `onMounted` function to initialize bulk editor at startup.


liyasthomas (approved):
Thanks for this quick PR @0xc0Der.

0xc0Der:
happy to help. sorry for the bug :smile:.

Issue #1891, opened 2021-10-21
Html preview view area is not full height


iagormoraes:
**Describe the bug**
By getting a HTML response from any endpoint call, the preview is not getting the full height of available space as shown in the screenshot.

**To Reproduce**
1. Make request to `https://github.com/hoppscotch/hoppscotch/issues`
2. Click on `Show Preview`
3. You should see the height of the preview not fitting the available space

**Expected behavior**
The users should be able to see the preview in full height of available space.

**Screenshots**
<img width="1679" alt="Screen Shot 2021-10-21 at 08 18 35" src="https://user-images.githubusercontent.com/13892132/138268136-4dd0953e-34cb-4d35-8837-deed57b40e57.png">

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser Safari
 - Version 2.0

**Smartphone (please complete the following information):**

**Additional context**

liyasthomas:
Hi @rishabhkalra96, let me know if you've any trouble with this issue. Would love to help you out.

iagormoraes:
I can help also in this issue, just let me know 🙂

AndrewBastin:
@iagormoraes sure you can work on this!

iagormoraes:
I can't run dev due to some error of package, I saw we use workspaces for it but it seems to be not recognizing it:

```
packages/hoppscotch-app do-dev$ pnpm run dev
[99 lines collapsed]
│ [dev:nuxt       ]  WARN  in /workspace/hoppscotch/node_modules/.pnpm/yargs-parser@20.2.9/node_modules/yargs-parser/build/lib/index.js
│ [dev:nuxt       ] 
│ [dev:nuxt       ] 
│ [dev:nuxt       ]  ERROR  Failed to compile with 1 errors
│ [dev:nuxt       ] 
│ [dev:nuxt       ] This dependency was not found:
│ [dev:nuxt       ] 
│ [dev:nuxt       ] * @hoppscotch/js-sandbox in ./helpers/RequestRunner.ts
│ [dev:nuxt       ] 
│ [dev:nuxt       ] To install it, you can run: npm install --save @hoppscotch/js-sandbox
└─ Running...
```

It happens both local and in cloud with Gitpod, someone knows that?

liyasthomas:
[...]

PR #1938, opened 2021-11-04
fix: 100% HTML preview area size

iagormoraes:
This PR solves the problem of making the HTML and avoid the same to happen on the other panes.

closes https://github.com/hoppscotch/hoppscotch/issues/1891


liyasthomas (approved):
@iagormoraes thanks for this quick PR - this looks great.
All looks good to me.

PR #1940, opened 2021-11-05
i18n: Update Chinese translations

jks15satoshi:


liyasthomas:
Thanks for this quick translation.

Issue #1834, opened 2021-09-23
can't use an environment variable in request body


zecrazytux:
**Describe the bug**
environment <<variable>> should be useable within a raw request body, but it isn't

**To Reproduce**
create an env variable, eg "user"
use it (<<user>>) into url: it works
use it (<<user>>) into the raw request body textarea: it is not replaced with var content

**Expected behavior**
<<user>> should be replaced with env var content

**Screenshots**
![image](https://user-images.githubusercontent.com/929169/134485341-41ac29aa-e420-40e9-8237-f8e83339143a.png)

**Desktop (please complete the following information):**
chrome

liyasthomas:
@zecrazytux thank you reporting this behaviour. We'll be taking a look into this as soon as possible.

dhivyasreedhar:
Hi, can I work on this issue?

liyasthomas:
Hi @dhivyasreedhar, thanks for showing interest in working on this issue. You're free to take up this. I'm assigning it to you, [contact us](https://github.com/hoppscotch/hoppscotch#support) if you need any help.

rishabhkalra96:
Hi @liyasthomas, Is this issue still been looked into? 
I'd like to take this up if @dhivyasreedhar is no longer working on it.

liyasthomas:
Hi @rishabhkalra96, I guess you can pick up this, thanks for showing interest on contributing to Hoppscotch.

0xc0Der:
shouldn't this be closed?

widdlepaws:
hello :) when can we expect it to be delivered ? really looking forward to it ! 
Last release got packaged too early :') https://github.com/hoppscotch/hoppscotch/releases/tag/v2.1.0

liyasthomas:
This feature is already live on the production deployment: Hoppscotch.io. You can fetch the `latest` tag if you're on Docker or fetch the latest `main` branch from Git to enjoy the latest changes.

humbertoarizah:
Hi, I know this issue is already closed, but I tried to access to env variables within the body and it has been impossible. I have this notation in the body, for example:
{"account" : {
    "username": "<<new-username>>",
    "email": "<<new-email>>"
}}
And my response is not taking the env variables but the string (<<new-username>>).
I don´t know if I´m doing anything wrong.
I would appreciated any help.

n-alonso:
[...]

PR #1942, opened 2021-11-06
allow environment variables in request body.

0xc0Der:
solves issue #1834.


AndrewBastin (changes requested):
There are multiple things wrong with this PR.

First of all the functionality does not work.

Steps for reproduction:
recording

(open image in new tab for more clarity)

Secondly, this is breaking functionality with FormData editor when using environment variables.

Steps of reproduction:
recording

(open image in a new tab for more clarity)

0xc0Der:
hmmm. investigating.

0xc0Der:
fixed.
- the first problem was because u did not track env vars changes.
- the second is not mine. it in the `EnvInput` code.

0xc0Der:
i guess i missed up the branch. :smile:

packages/hoppscotch-app/components/http/RawBody.vue
```diff
@@ -111,6 +119,25 @@ useCodemirror(
...
+}
+
+watch(envVariables, () => {
+  updateBody(rawBodyData.value)
+})
+
+watch(rawBodyData, (newValue) => {
+  updateBody(newValue)
+})
+
```
AndrewBastin, line 140: This is totally not the right approach to implementing this in Hoppscotch.

Hoppscotch processes requests in a step-by-step process which are broadly classified into.

- Composing (The user is still building the request, did not press the Send button yet)
- Initial Validation (Deciding whether the request should even run, when the Send button is pressed)
- Pre-Request Script Execution (A Pre-Request script is executed if defined)
- Environment Injection (After pre-request script we have all the data necessary to fill the environment variable holes, we fill them)
- Execution (Actual network request is ran)

When the send request action is fired, the pre-request scripts are initially ran. [Pre-Request scripts](https://github.com/hoppscotch/hoppscotch/wiki/Pre-Request-Scripts) allow for environment variables to be created and updated dynamically, those environments are all aggregated into an ephemeral list of environmental variables as specified here https://github.com/hoppscotch/hoppscotch/blob/6d54f21c1eed4df152b95e7c79a92d9b0fe1ae09/packages/hoppscotch-app/helpers/RequestRunner.ts#L49-L53

Your code tries to resolve environment variables during the composing phase itself, which doesn't take into account the dynamic variables which can be defined by the pre-request scripts.

AndrewBastin (changes requested):

packages/hoppscotch-app/components/http/RawBody.vue
0xc0Der: noted. re-investigating.

AndrewBastin (approved):
LGTM 💯 

Once again thank you for the PR @0xc0Der 🤗 🤗

PR #1957, opened 2021-11-17
chore(i18n): update ru translations

Craiqser:


liyasthomas (approved):
@Craiqser thanks for this quick PR. Translations will be updated once every week.

PR #1958, opened 2021-11-17
chore(i18n): update it translations

BrainCrumbz:
Hi there,
sorry for the unsolicited PR, I hope that's ok.
We found that some italian translations looked funny, so we though to give it a go to the whole it.json file.
HTH


liyasthomas (approved):
@BrainCrumbz thanks for this quick PR. Translations will be updated once every week.

Issue #1980, opened 2021-11-28
[bug]: Request hangs if the pre-request script is not valid


andriuspetrauskis:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When I enter pre-request script some invalid javascript code, I expect app to show an error. Instead the request just keeps loading.

### Steps to reproduce

1. Go to https://hoppscotch.io/
2. Click Pre-request Script
3. Enter any random text
4. Press send

### Environment

Production

### Version

Cloud

PR #1981, opened 2021-11-28
fix: show an error when pre-request script fails

andriuspetrauskis:
Closes [1980](https://github.com/hoppscotch/hoppscotch/issues/1980)


liyasthomas:
Only add the files that have been changed in the specified task. It seems like 90+ files has been changed in this PR (most probably due to a project-wide lint command). Either force-push the required file changes or feel free to open another PR with relevant files changes.

Opening a draft PR would be better for us to review them.

PR #2021, opened 2021-12-17
feat: (authentication) Api key based authentication [#2000]

### LINKED ISSUE #2000: [feature]: Api Key based authentication

### Summary

Hello,

We see many apis and api gateways using api key as a standard authentication mechanism, it would be awesome to support this authentication method in hoppscotch. This would require the ability to specificy the place where to set the api key (header or path) and under which name (different vendor and implementation use different header name).

Currently APIs using API Keys passed through headers have their authentication specified as any other header when they actually concern Authentication. To have a consistent handling of Authentication it should be made possible to specify API Key as an autentication method in hoppscotch

EdwinFajardoBarrera:
Closes #2000  

### Description

This is the PR linked to the addition of support api-key based authentication.

### Checks

### Additional Information

Added:
image
image

image

### DIFF
--- a/packages/hoppscotch-app/helpers/utils/EffectiveURL.ts
+++ b/packages/hoppscotch-app/helpers/utils/EffectiveURL.ts
@@ -100,6 +100,47 @@ export function getEffectiveRESTRequest(
           envVariables
         )}`,
       })
+    } else if (request.auth.authType === "api-key") {
+      const { key, value, addTo } = request.auth
+
+      if (addTo === "Header") {
+        effectiveFinalHeaders.push({
+          active: true,
+          key,
+          value: parseTemplateString(value, envVariables),
+        })
+
+        const paramIdx = request.params.findIndex((el) => el.key === key)
+
+        request.params[paramIdx] = {
+          ...request.params[paramIdx],
+          active: false,
+        }
+      } else if (addTo === "Query params") {
+        const paramIdx = request.params.findIndex((el) => el.key === key)
+
+        if (paramIdx === -1) {
+          request.params.pop()
+          request.params.push({
+            active: true,
+            key,
+            value: parseTemplateString(value, envVariables),
+          })
+        } else {
+          request.params[paramIdx] = {
+            ...request.params[paramIdx],
+            active: true,
+          }
+        }
+      }
+    }
+  } else {
+    const key = request.auth?.key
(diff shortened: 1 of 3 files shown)


liyasthomas:
@EdwinFajardoBarrera I've shared my suggestions on this PR, do let me know if those sound good to you. Except transferring the logic of appending the parameter to the actual parameters list at the time of sending the request all looks good to me. Because if we append the parameter to the actual parameters section and choose to change the `Pass by` value again, the added parameter still stays in the actual parameters section. The ideal solution is to not add parameters to the actual parameters section instead add them to effective parameters. More info in the review description. Thank you.

EdwinFajardoBarrera:
I think now it works well, using `effectiveFinalParams` avoided all the problems. I would like to read your comments in order to create the PR.

liyasthomas (changes requested):

liyasthomas (approved):
LGTM

packages/hoppscotch-app/helpers/utils/EffectiveURL.ts
```diff
@@ -100,6 +112,22 @@ export function getEffectiveRESTRequest(
...
         )}`,
       })
+    } else if (request.auth.authType === "api-key") {
+      const { key, value, addTo } = request.auth
+
+      if (addTo === "Header") {
+        effectiveFinalHeaders.push({
+          active: true,
+          key: parseTemplateString(key, envVariables),,
+          value: parseTemplateString(value, envVariables),
```
liyasthomas, line 122:
```suggestion
          key: parseTemplateString(key, envVariables),
          value: parseTemplateString(value, envVariables),
```

packages/hoppscotch-app/helpers/utils/EffectiveURL.ts
```diff
@@ -100,6 +112,22 @@ export function getEffectiveRESTRequest(
...
+        })
+      } else if (addTo === "Query params") {
+        effectiveFinalParams.push({
+          active: true,
+          key,
+          value,
```
liyasthomas, line 128:
```suggestion
          key: parseTemplateString(key, envVariables),
          value: parseTemplateString(value, envVariables),
```

liyasthomas (changes requested):

liyasthomas (approved):

PR #2048, opened 2021-12-28
Bug/body missing onimport

anwarulislam:
Found two bugs while importing cURL and Postman collection into Hoppscotch. when importing cURL the request doesn't appear with its payload. And while importing postman collection it's failed because of some condition. I've already fixed those two bugs.


liyasthomas (approved):
LGTM. Thank you for noticing the error and this quick PR.

PR #2074, opened 2022-01-20
Feature: hopp-cli in TypeScript

devblin:
### Description
New hopp-cli written in TypeScript

> Note: Work in progress


packages/hoppscotch-cli/src/utils/checks.ts
```diff
@@ -0,0 +1,202 @@
...
+    }
+  }
+  if (!param.x.body || typeof param.x.body !== "object") {
+    return false;
+  } else {
+    if (typeof param.x.body.contentType === "undefined") return false;
+    if (typeof param.x.body.body === "undefined") return false;
+  }
+  return true;
+}
```
AndrewBastin, line 108: Although, implementation is a lot more sound (because it actually checks each value), we have an implementation for `isHoppRESTRequest` in `@hoppscotch/data` which should be used instead.

packages/hoppscotch-cli/src/utils/test-parser.ts
```diff
@@ -0,0 +1,82 @@
...
+export const testParser = async (
+  testScriptPair: TestScriptPair
+): Promise<number> => {
+  let totalFailing: number = 0,
+    testRes: number = 0;
+  const testScriptExecPromise = execTestScript(
+    testScriptPair.testScript,
+    testScriptPair.response
+  );
+  const testScriptExecRes = await testScriptExecPromise();
```
AndrewBastin, line 72:
```suggestion
  const testScriptExecRes = await execTestScript(
    testScriptPair.testScript,
    testScriptPair.response
  )();
```

packages/hoppscotch-cli/src/utils/test-parser.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  if (testScriptExecRes._tag === "Right") {
```
AndrewBastin, line 74: Don't check tags directly. Use the fp-ts Either functions.

```ts
import * as E from "fp-ts/Either"

if (E.isRight(testScriptExecRes)) {
}
```

AndrewBastin (changes requested):
Some minor corrections for now. We can simplify the code in a lot of places by relying more on FP-TS stuff.

devblin:
@AndrewBastin , I have pushed changes with requested changes, and I will do refactoring to simplify using fp-ts.
I am using [this](https://gcanti.github.io/fp-ts/) fp-ts doc for reference.

packages/hoppscotch-cli/src/handlers/command.ts
```diff
@@ -0,0 +1,107 @@
+import fs from "fs/promises";
+import inquirer from "inquirer";
+import chalk from "chalk";
+import { array as A, string as S } from "fp-ts";
```
AndrewBastin, line 4:
```suggestion
import * as A from "fp-ts/Array";
import * as S from "fp-ts/string";
```

packages/hoppscotch-cli/src/handlers/error.ts
```diff
@@ -0,0 +1,35 @@
+import chalk from "chalk";
+import { either as E } from "fp-ts";
```
AndrewBastin, line 2:
```suggestion
import * as E from "fp-ts/Either";
```

packages/hoppscotch-cli/src/utils/checks.ts
```diff
@@ -0,0 +1,112 @@
...
+import { errors } from ".";
+import {
+  HoppRESTRequest,
+  translateToNewRESTCollection,
+  HoppCollection,
+  isHoppRESTRequest,
+} from "@hoppscotch/data";
+import { Either, left, right } from "fp-ts/lib/Either";
+import * as S from "fp-ts/string";
+import { pipe } from "fp-ts/lib/function";
```
AndrewBastin, line 13:
```suggestion
import * as E from "fp-ts/Either";
import * as S from "fp-ts/string";
import { pipe } from "fp-ts/function";
```

packages/hoppscotch-cli/src/utils/getters.ts
```diff
@@ -0,0 +1,299 @@
+import chalk from "chalk";
+import { pipe } from "fp-ts/lib/function";
```
AndrewBastin, line 2:
```suggestion
import { pipe } from "fp-ts/function";
```

packages/hoppscotch-cli/src/utils/mutators.ts
```diff
@@ -0,0 +1,94 @@
+import { Environment } from "../interfaces";
+import { flow } from "fp-ts/lib/function";
```
AndrewBastin, line 2:
```suggestion
import { flow } from "fp-ts/function";
```

packages/hoppscotch-cli/src/utils/test-parser.ts
```diff
@@ -0,0 +1,78 @@
+import { pipe } from "fp-ts/lib/function";
+import { array as A, either as E } from "fp-ts";
```
AndrewBastin, line 2:
```suggestion
import { pipe } from "fp-ts/function";
import * as A from "fp-ts/Array";
import * as E from "fp-ts/Either";
```

AndrewBastin (changes requested):
Just some minor corrections.

devblin:
> Just some minor corrections.

Updated 👍🏽

packages/hoppscotch-cli/src/utils/mutators.ts
```diff
@@ -0,0 +1,93 @@
...
+ */
+export const flattenRequests = (
+  collections: HoppCollection<HoppRESTRequest>[]
+): TE.TaskEither<HoppCLIError, RequestStack[]> =>
+  pipe(
+    collections,
+    A.map(requestsParser),
+    TE.sequenceArray,
+    TE.map(RA.toArray),
+    TE.map(A.flatten)
```
AndrewBastin, line 92: Instead of mapping twice, we can just map once and do this operation together.

```suggestion
    TE.map(flow(RA.toArray, A.flatten))
```

packages/hoppscotch-cli/src/utils/checks.ts
```diff
@@ -0,0 +1,103 @@
...
+  pipe(
+    path,
+    TE.fromPredicate(S.isString, () => error({ code: "NO_FILE_PATH" })),
+    TE.chainW(checkFileURL)
+  );
+
+/**
+ * Check if given data is HoppCLIError.
+ * @param error Error data to check.
+ * @returns
```
AndrewBastin, line 99: Either remove the line or fill up the info here.

packages/hoppscotch-cli/src/utils/checks.ts
```diff
@@ -0,0 +1,103 @@
...
+
+export const isExpectResultPass = (
+  expectResult: string
+): E.Either<boolean, boolean> =>
+  expectResult === "pass" ? E.right(true) : E.left(false);
+
+/**
+ * Checks if given error is CommanderError with exitCode 0.
+ * @param error Error data to check.
+ * @returns
```
AndrewBastin, line 76: Either remove the line or fill up the info here.

packages/hoppscotch-cli/src/utils/checks.ts
```diff
@@ -0,0 +1,103 @@
...
+ */
+export const isSafeCommanderError = (error: any) => {
+  if (error instanceof CommanderError && error.exitCode === 0) {
+    process.exit(0);
+  }
+};
```
AndrewBastin, line 82: The name of the function makes it seem that the function returns a boolean while it doesn't.
I am not really sure the this function should exit the program by itself.

packages/hoppscotch-cli/src/utils/checks.ts
```diff
@@ -0,0 +1,103 @@
...
+        (fullPath) => extname(fullPath) === ".json",
+        (fullPath) => error({ code: "FILE_NOT_JSON", path: fullPath })
+      )
+    )
+  );
+
+export const isExpectResultPass = (
+  expectResult: string
+): E.Either<boolean, boolean> =>
+  expectResult === "pass" ? E.right(true) : E.left(false);
```
AndrewBastin, line 71: why ? 🥲 

Remember, Eithers are meant to abstract error handling, this function has no reasonable error case, its a yes or no question. So using Eithers for this is dumb.

Also see comments on [test.ts](tests.ts)

packages/hoppscotch-cli/src/utils/checks.ts
```diff
@@ -0,0 +1,103 @@
...
+import * as E from "fp-ts/Either";
+import * as TE from "fp-ts/TaskEither";
+import { error, HoppCLIError } from "../types";
+
+/**
+ * Typeguard to check valid Hoppscotch REST Collection
+ * @param param The object to be checked
+ * @returns Boolean value corresponding to the validity check
+ */
+export function isRESTCollection(param: any): E.Either<boolean, boolean> {
```
AndrewBastin, line 19: why does isRESTCollection return an either, it is a yes or no question. Not a failable operation. So avoid the use of Either for situations like these.

AndrewBastin, line 19: We can convert this function to be a proper type guard which can confirm the type casting as well (`param is HoppCollection<HoppRESTRequest>` as the function return type ?)

AndrewBastin, line 19: `any` is a type which is really frowned upon.

Take the scenario where `null` or `undefined` or some other random value is passed as the `param` value., because of the any type, TypeScript will ignore type checking. Instead go for the `unknown` type. 

See: [https://mariusschulz.com/blog/the-unknown-type-in-typescript](https://mariusschulz.com/blog/the-unknown-type-in-typescript)

packages/hoppscotch-cli/src/utils/mutators.ts
```diff
@@ -0,0 +1,93 @@
...
+ * @param path Collection json file path.
+ * @returns For successful parsing we get array of HoppCollection<HoppRESTRequest>,
+ */
+export const parseCollectionData = (
+  path: string
+): TE.TaskEither<HoppCLIError, HoppCollection<HoppRESTRequest>[]> =>
+  pipe(
+    // Trying to read give collection json path.
+    TE.tryCatch(
+      () => pipe(path, fs.readFile),
```
AndrewBastin, line 53: You can refrain from using `pipe` function for short pipelines

```suggestion
      () => fs.readFile(path)
```

packages/hoppscotch-cli/src/utils/mutators.ts
```diff
@@ -0,0 +1,93 @@
...
+      (reason) => error({ code: "UNKNOWN_ERROR", data: E.toError(reason) })
+    ),
+
+    // Checking if parsed file data is array.
+    TE.map((a) => pipe(a.toString(), JSON.parse)),
```
AndrewBastin, line 58: What if `JSON.parse` fails ? This can happen when the JSON is malformed.

packages/hoppscotch-cli/src/utils/mutators.ts
```diff
@@ -0,0 +1,93 @@
...
+    TE.chainW(
+      TE.fromPredicate(
+        (data) => Array.isArray(data),
+        (_) => error({ code: "MALFORMED_COLLECTION", path })
+      )
+    ),
+
+    // Validating collections to be HoppRESTCollection.
+    TE.chainW((collections) =>
+      pipe(
```
AndrewBastin, line 68: Going with my correction on `isRESTCollection` where it is a type guard, we can express it as a type guard that makes sure this is a REST Collection.
This pipe will then look like this:

```ts
        pipe(
          collections,
          TE.fromPredicate(
            flow(A.every(isRESTCollection)),
            () => error({ code: "MALFORMED_COLLECTION", path })
          )
        )
```

(I haven't checked the validity of the above code (because markdown), but it should be something similar)

packages/hoppscotch-cli/src/utils/mutators.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  return formData;
+};
+
+/**
+ * Parses provided error message to maintain hopp-error messages.
+ * @param e Custom error data.
+ * @returns Parsed error message without extra spaces.
+ */
+export const parseErrorMessage = (e: any) => {
```
AndrewBastin, line 30: Try to use `unknown` instead of `any`

packages/hoppscotch-cli/src/utils/mutators.ts
```diff
@@ -0,0 +1,93 @@
...
+  let msg: string;
+  if (e instanceof Error) {
+    const x = e as NodeJS.ErrnoException;
```
AndrewBastin, line 33: I don't understand this cast, what if the error is not an `ErrnoException`, why can't we instead directly do instanceof `e` to `ErrnoException`.

packages/hoppscotch-cli/src/utils/request.ts
```diff
@@ -0,0 +1,336 @@
...
+  pipe(
+    TE.tryCatch(
+      async () => {
+        if (A.isNonEmpty(requests)) {
+          const tableStream = getTableStream();
+          responseTableOutput.header(tableStream);
+
+          const testScriptData = await pipe(
+            requests,
+            A.map((request) =>
```
AndrewBastin, line 290: You can use `flow` here.

packages/hoppscotch-cli/src/utils/request.ts
```diff
@@ -0,0 +1,336 @@
...
+const preRequestScriptRunner = (
+  request: EffectiveHoppRESTRequest
+): TE.TaskEither<HoppCLIError, EffectiveHoppRESTRequest> =>
+  pipe(
+    TE.of(request),
+    TE.map((req) => req.preRequestScript),
+    TE.chain((preScript) =>
+      runPreRequestScript(preScript, { global: [], selected: [] })
+    ),
+    TE.map((envs) => Object({ name: "Env", variables: envs.selected })),
```
AndrewBastin, line 263: Using the object constructor makes this object lose type information (this TE now returns an `any`)

```suggestion
      TE.map((envs) => ({ name: "Env", variables: envs.selected }))
```

packages/hoppscotch-cli/src/utils/request.ts
```diff
@@ -0,0 +1,336 @@
...
+    TE.chainW((env) =>
```
AndrewBastin, line 264: Use `TE.chainEitherKW` instead

packages/hoppscotch-cli/src/utils/checks.ts
devblin: Yup, actually Commander throws error weirdly, even if its not supposed to be error. And within handlError, we use this to exclude those type of CommanderError (with exitCode = 0), and just exit else print the error.

packages/hoppscotch-cli/src/utils/request.ts
```diff
@@ -0,0 +1,336 @@
...
+    ),
+    A.map(preRequestScriptRunner),
+    TE.sequenceArray,
+    TE.map(RA.toArray),
+    TE.map(
+      A.map((effHoppRequest) =>
+        createRequest(`${rootPath}/${collection.name}`, effHoppRequest)
+      )
+    ),
+    TE.bindTo("PARSED_REQUESTS"),
```
AndrewBastin, line 233: What is the point of binding if you are not aggregating TEs. You are just chaining them in a sequence. A normal chain should suffice right ?

packages/hoppscotch-cli/src/utils/checks.ts
devblin: Yeah, now I think its dumb 😢.

packages/hoppscotch-cli/src/utils/checks.ts
devblin: OK.

packages/hoppscotch-cli/src/utils/request.ts
```diff
@@ -0,0 +1,336 @@
...
+    : req.params;
+  const reqHeaders = A.isNonEmpty(req.effectiveFinalHeaders)
+    ? req.effectiveFinalHeaders
+    : req.headers;
+  config.url = S.isEmpty(req.effectiveFinalURL)
+    ? req.endpoint
+    : req.effectiveFinalURL;
+  config.method = req.method as Method;
+
+  for (const x of reqParams) {
```
AndrewBastin, line 58: Use a pipeline for this ?

packages/hoppscotch-cli/src/utils/request.ts
```diff
@@ -0,0 +1,336 @@
...
+    if (x.active) {
+      if (!config.params) {
+        config.params = {};
+      }
+      if (x.key) config.params[x.key] = x.value;
+    }
+  }
+  for (const x of reqHeaders) {
```
AndrewBastin, line 66: Use a pipeline for this

packages/hoppscotch-cli/src/utils/mutators.ts
devblin: Checked, this works.

packages/hoppscotch-cli/src/utils/test.ts
```diff
@@ -0,0 +1,207 @@
...
+    ),
+
+    /**
+     * Reducing tests-report to calculate total failing and
+     * tests-report-size.
+     */
+    TE.chainW(
+      flow(
+        A.reduce({ failing: 0, testsReportSize: 0 }, (prev, testReport) =>
+          Object({
```
AndrewBastin, line 145: You can remove the `Object` constructor there (but keep the brackets)

packages/hoppscotch-cli/src/utils/mutators.ts
devblin: Oh, ok this should be also within tryCatch.

packages/hoppscotch-cli/src/utils/test.ts
```diff
@@ -0,0 +1,207 @@
...
+ * HoppCLIError - Due to some errors.
+ */
+export const runTests = (
+  tests: TestScriptData[]
+): TE.TaskEither<HoppCLIError, null> =>
+  pipe(
+    tests,
+    A.map(testRunner),
+    TE.sequenceArray,
+    TE.map(RA.toArray),
```
AndrewBastin, line 123: Use a single `map` and then chaining the operation by `flow`

packages/hoppscotch-cli/src/utils/test.ts
```diff
@@ -0,0 +1,207 @@
...
+    /**
+     * Recursively parsing test-results
+     * to obtain test-report array.
+     */
+    TE.map((testResult) => testResult.tests),
+    TE.chainW(
+      flow(
+        A.map(testDescriptorParser),
+        T.sequenceArray,
+        T.map(RA.flatten),
```
AndrewBastin, line 96: Single `map` and `flow`.

packages/hoppscotch-cli/src/utils/test.ts
```diff
@@ -0,0 +1,207 @@
...
+          failing: testMetrics.failing,
+          passing: testMetrics.passing,
+          descriptor: testDescriptor.descriptor,
+          expectResults: testDescriptor.expectResults,
+        }) as TestReport
+    ),
+    T.map((testReport) =>
+      A.isNonEmpty(testReport.expectResults) ? [testReport] : []
+    ),
+    T.bindTo("TEST_REPORT"),
```
AndrewBastin, line 46: Why bind if you are not aggregating and instead sequentially processing things.

packages/hoppscotch-cli/src/utils/test.ts
```diff
@@ -0,0 +1,207 @@
...
+      pipe(
+        testDescriptor.children,
+        A.map(testDescriptorParser),
+        T.sequenceArray,
+        T.map(RA.flatten),
+        T.map(RA.toArray),
+        T.map(A.concat(TEST_REPORT))
+      )
+    ),
+    T.bindTo("TESTS_REPORT"),
```
AndrewBastin, line 63: Why do you bind here ?

packages/hoppscotch-cli/src/utils/test.ts
```diff
@@ -0,0 +1,207 @@
...
+     * Recursive call to testDescriptorParser on testDescriptor's
+     * children; The result is concated with TEST_REPORT to generate
+     * final output, binded to TESTS_REPORT.
+     */
+    T.chain(({ TEST_REPORT }) =>
+      pipe(
+        testDescriptor.children,
+        A.map(testDescriptorParser),
+        T.sequenceArray,
+        T.map(RA.flatten),
```
AndrewBastin, line 58: The three maps can be converted into a single map.

packages/hoppscotch-cli/src/utils/index.ts
```diff
@@ -0,0 +1,7 @@
+export * from "./constants";
+export * from "./checks";
+export * from "./request";
+export * from "./test";
+export * from "./getters";
+export * from "./mutators";
+export * from "./functions/array";
```
AndrewBastin, line 7: Why are we exporting like this ?

packages/hoppscotch-cli/src/types/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from "./errors";
+export * from "./request";
+export * from "./response";
```
AndrewBastin, line 3: Why are we exporting like this

packages/hoppscotch-cli/src/interfaces/request.ts
```diff
@@ -0,0 +1,34 @@
+import { AxiosPromise, AxiosRequestConfig } from "axios";
+import { HoppRESTRequest } from "@hoppscotch/data";
+
+/**
+ * Request stack interface.
```
AndrewBastin, line 5: If you want describe things in documentation, make sure you are descriptive about it, what you wrote is just literally what the interface is, not what it does or any other helpful information, it is not adding anything other than noise to the code.

packages/hoppscotch-cli/src/interfaces/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from "./request";
+export * from "./response";
```
AndrewBastin, line 2: Why are we exporting ?

packages/hoppscotch-cli/src/utils/mutators.ts
devblin: Actually, `ErrnoException` is just an interface within `NodeJS` namespace, but `ErrnoException` extends `Error` interface.
And the `Error` here is the decalaration of `ErrorConstructor` (which initiates Error related props),
And thats why this way to use `ErrnoException` has it has more properties which are usefull to create error message.

AndrewBastin (changes requested):

packages/hoppscotch-cli/src/utils/checks.ts
AndrewBastin: In that case, either rephrase the function correctly to explain that, this just looks like a type guard if we do not look at the implementation now.

packages/hoppscotch-cli/src/utils/request.ts
devblin: Ok.

packages/hoppscotch-cli/src/utils/request.ts
devblin: Sure

packages/hoppscotch-cli/src/utils/request.ts
devblin: Actually, I used bindTo, to help me keep track of output so far, visible to me while working on next part. If It seems ambiguous, I will update 👍🏽

packages/hoppscotch-cli/src/utils/checks.ts
devblin: Ok, this would be better.

packages/hoppscotch-cli/src/utils/checks.ts
devblin: OK

packages/hoppscotch-cli/src/interfaces/request.ts
devblin: Sorry for this, I will make sure to go through undocumented code.

packages/hoppscotch-cli/src/utils/mutators.ts
AndrewBastin: aaah gotcha. that makes sense, I would suggest writing a utility type guard for it though, this entire function looks kinda funky.

packages/hoppscotch-cli/src/utils/request.ts
AndrewBastin: Umm, do not use binds for letting you know things, that is what comments are for 🤣
You can place a comment near the chain/map to explain what and why.

packages/hoppscotch-cli/src/interfaces/request.ts
AndrewBastin: Not the undocumented code bit. But how useful is a comment like this:

```js
// x is a number
const x = 42;
```

Instead you should be explaining what x is doing and what is its purpose etc.

devblin: Yes, got it, I will make the documentation more descriptive.

packages/hoppscotch-cli/src/utils/index.ts
devblin: Actually, I currently import functions based on their dir path (like `import { .... } from /interface`). So, thats why I have exported the functions within /SOME_DIR to /SOME_DIR/index.ts

AndrewBastin: Yeah gotcha, this is technically not a problem. But let's not do this.

The reason why I say we shouldn't is because of two reasons.
1. This encourages sensible file names and file structures.
2. The imports on the user side will have more contextual information about the thing they are importing. For example, in utils, we have various family of functions in different files each encomposing different behaviours. Now when we see the filename in the import, we get a more clear picture of what that function might do, for example, we know functions in `mutators` can mutate stuff (etc.)

We follow the opposite behaviour in library code though (you will see similar exports in `@hoppscotch/data` for example), the reasoning there is, when we are writing libraries, we shouldn't expose unwanted the internals of the library, we should only expose concerns. Since the CLI is not a library (or an intermediary product) and is a end product where there will be a lot more complex and cross-cutting concerns in the future, I like the code to be more explicit.

packages/hoppscotch-cli/src/utils/mutators.ts
```diff
@@ -0,0 +1,74 @@
...
+      (reason) => error({ code: "UNKNOWN_ERROR", data: E.toError(reason) })
+    ),
+
+    // Checking if parsed file data is array.
+    TE.chainEitherKW((data) =>
+      pipe(
+        data.toString(),
+        J.parse,
+        E.map((jsonData) => (Array.isArray(jsonData) ? jsonData : [jsonData])),
+        E.mapLeft((e) => error({ code: "SYNTAX_ERROR", data: E.toError(e) }))
```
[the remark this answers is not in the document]
devblin: Done 👍🏽

AndrewBastin (approved):
LGTM :100:

AndrewBastin (approved):
LGTM 💯 

(finally)

PR #2083, opened 2022-01-24
feat: Ensure Method In Endpoint

### LINKED ISSUE #2058: [feature]: explicit UI for protocol (e.g. http / https)

### Summary

Thank you for the wonderful project!

As a new user it took me a little while to figure out why my calls to `localhost` or `127.0.0.1` weren't working, even with the browser extension installed.  I eventually tried typing `http://` before the domain and it worked.

I think of the URL input the same way as I think of the browser URL input (but with the ability to specify method, body, header, etc).  Given that, I was expecting the request to work even without an explicit specification of `http://`.

I appreciate there may be a use case for explicitly not having a protocol specified in a request, so maybe this could be accomplished with a new UX element that provides a select dropdown, defaulting to http, providing `http` and `https` as pre-provided values, but also capable of freeform input and editing as a text field?

It seems safe to bet that 99.99% of the time your users are going to want either http or https prefixed to their URLs.  A reasonable portion of new users (and even veteran users) will probably not realize / remember that `http` has to be manually entered as part of the URL the first time they try.

The space will have to be taken up regardless (since http:// will have to be manually typed), so having a UX element would:

1) Save keystrokes.
2) Communicate the requirement of a protocol selection.
3) Achieve that requirement without user input.

This will also help new users verify their setup faster -- right now a new user would have reasonable cause to assume that the failed call is due to misconfiguration of the browser extension (or server) as opposed to realizing that the request itself has simply been improperly formed.

RishabhAgarwal-2001:
Closes #2058

### Description

While sending a HTTP request without a specified method, the method is determined using domain name and appended to the request. Following is the way of determining the request method - 

* If domain name is either `localhost` or period delimited numbers (eg. `127.0.0.1` or any IP address), `HTTP` protocol is used
* In all other cases `HTTPS` protocol is used

The updated endpoint is reflected in the URL input box.

### Checks


AndrewBastin (approved):
lgtm 💯

PR #2088, opened 2022-01-25
fix: parsing of protocol correctly

### LINKED ISSUE #2066: [feature]: curl can't parse the url without the protocol header correctly.

### Summary

anwarulislam:bug/body-missing-onimport https://github.com/hoppscotch/hoppscotch/pull/2048
I have cleared the cache and reloaded, but I still can't parse the request body. In addition, I can't parse the url without the protocol header correctly.

curl -X PUT "localhost:9200/megacorp/employee/1?pretty" -H 'Content-Type: application/json' -d'
{
    "first_name" : "John",
    "last_name" :  "Smith",
    "age" :        25,
    "about" :      "I love to go rock climbing",
    "interests": [ "sports", "music" ]
}
'

kyteinsky:
Closes #2066 

### Description

Domain with no protocol supplied was not parsed correctly and resulted in null. This issue has been fixed in this commit. Moreover, params with no values now do not carry double quotes in the key section and are placed at the end of the URL itself instead of the parameters section. Basic auth (username and password) has been added.

### Checks


AndrewBastin (changes requested):
@kyteinsky 

Looks good!

I think this would be a good opportunity to refactor a bit of odd quirks with this since we are already working on this.

I am basing the quirks based on the following command as an example.

```
curl -X PUT "localhost:9200/megacorp/employee/1?pretty" \
-H 'Content-Type: application/json' \
-d '{
"first_name" : "John",
"last_name" : "Smith",
"age" : 25,
"about" : "I love to go rock climbing",
"interests": [ "sports", "music" ]
}'
```

Quirk 1: Since `pretty` does not have a value, it doesn't fit neatly with the key value system of the parameters we have, hence, I think it would be cool if the field remains in the URL itself.

Quirk 2: Repetition of the Content-Type header. In the `HoppRESTReqBody` data structure, we do have a slot for content type which is injected to the request, hence defining a Content-Type header again in the headers field is not needed, hence it would be cool if we can remove that behaviour.

That is about it as far as I can see.

kyteinsky:
> Quirk 2: Repetition of the Content-Type header. In the HoppRESTReqBody data structure, we do have a slot for content type which is injected to the request, hence defining a Content-Type header again in the headers field is not needed, hence it would be cool if we can remove that behaviour.

Content-Type is not attached to headers in the final object sent to `makeRestRequest()`. It is stored from the output of parsedCurl back to `body.contentType`. In the example given, `-H 'Content-Type: application/json'` passes down the content type to `parsedCurl.headers`, so just a check is made to look if that is present, otherwise the default type: application/json is assumed and dealt with.

```typescript
const contentType: Exclude<ValidContentTypes, "multipart/form-data"> =
  parsedCurl.headers?.contentType || "application/json"
```

kyteinsky:
Just found out one more thing: `-u username:password` does not work as well.

packages/hoppscotch-app/components/http/ImportCurl.vue
```diff
@@ -116,7 +129,64 @@ const handleImport = () => {
...
+        ...(username.length > 0 && {
+          username,
+          password,
+        }),
+        ...(user.length > 0 && {
+          username: user.split(":")[0],
+          password: user.split(":")[1],
+        }),
+      } as HoppRESTAuth
+    }
```
AndrewBastin, line 163: Split this into a separate function ?

Keeping common logic isolated allows us to iterate more quicker

packages/hoppscotch-app/helpers/contentParser.ts
```diff
@@ -0,0 +1,118 @@
+import { HoppRESTReqBody } from "~/../hoppscotch-data/dist"
```
AndrewBastin, line 1:
```suggestion
import { HoppRESTReqBody } from "@hoppscotch/data"
```

AndrewBastin (changes requested):

packages/hoppscotch-app/helpers/curlparser.ts
```diff
@@ -2,23 +2,16 @@ import * as URL from "url"
 import * as querystring from "querystring"
 import * as cookie from "cookie"
 import parser from "yargs-parser"
+import { detectContentType, parseBody } from "./contentParser"
+import { isType } from "./typeutils"
+import { HoppRESTReqBody } from "~/../hoppscotch-data/dist"
```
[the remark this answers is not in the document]
kyteinsky: Ahh, I will be mindful of those auto-imports from now on.

packages/hoppscotch-app/helpers/typeutils.ts
```diff
@@ -18,3 +18,6 @@ export const valueIncludes = <T, V extends readonly T[]>(
   obj: T,
   values: V
 ): obj is V[number] => values.includes(obj)
+
+export const isType = <Type>(variable: any): variable is Type =>
+  <Type>variable !== undefined
```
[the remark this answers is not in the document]
kyteinsky: I couldn't find any generic way to type check. Should I replace all instances with if-else checks? This indeed isn't working against other types but makes typescript happy somehow :laughing:

packages/hoppscotch-app/helpers/contentParser.ts
```diff
@@ -0,0 +1,118 @@
...
+    return {
+      ...(!!tempBody && { parsedJSON: JSON.stringify(tempBody, null, 2) }),
+      ...(!tempBody && { error: true }),
+    }
+  } catch (e) {
+    return {
+      error: true,
+    }
+  }
+}
```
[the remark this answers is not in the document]
kyteinsky: Woah, an uncanny resemblance to Rust's Option. I will make sure to learn fp-ts after this.

packages/hoppscotch-app/components/http/ImportCurl.vue
kyteinsky: Do you mind any suggestions on where I should place the function?

AndrewBastin: Well, I mean like, this logic is pretty self contained right, so I was thinking more like, take this entire bit out and make it as a function in the same file itself. So basically it abstracts the whole Auth parsing bits into a single function.

packages/hoppscotch-app/helpers/curlparser.ts
```diff
@@ -2,105 +2,77 @@ import * as URL from "url"
...
-  curlCommand = curlCommand.replace(/-d /, "--data ")
+  const replaceables: { [key: string]: string } = {
+    "--request": "-X",
+    "--header": "-H",
+    "--url": "",
+    "--form": "-F",
+    "--data-raw": "--data",
+    "--data": "-d",
+    "--user": "-u",
+  }
```
AndrewBastin, line 40: I don't think the explicit type definition is necessary ?

Also, I think we can move this variable outside the function as this is a constant and we don't need to create the same thing all the time.

packages/hoppscotch-app/helpers/curlparser.ts
```diff
@@ -2,105 +2,77 @@ import * as URL from "url"
...
-  curlCommand = curlCommand.replace(/ -XPUT/, " -X PUT")
-  curlCommand = curlCommand.replace(/ -XPATCH/, " -X PATCH")
-  curlCommand = curlCommand.replace(/ -XDELETE/, " -X DELETE")
+  curlCommand = curlCommand.replace(/-XPOST/, " -X POST")
+  curlCommand = curlCommand.replace(/-XGET/, " -X GET")
+  curlCommand = curlCommand.replace(/-XPUT/, " -X PUT")
+  curlCommand = curlCommand.replace(/-XPATCH/, " -X PATCH")
+  curlCommand = curlCommand.replace(/-XDELETE/, " -X DELETE")
   curlCommand = curlCommand.trim()
   const parsedArguments = parser(curlCommand)
```
AndrewBastin, line 53: This entire bit of code seems to be related to pre-processing the cURL command. Can we just break this entire bit off into a separate function outside ?

packages/hoppscotch-app/helpers/contentParser.ts
```diff
@@ -0,0 +1,267 @@
...
+        O.filter((pairs) => pairs !== null && pairs.length > 0),
+        O.map((pairs) => pairs.map((p) => p.replace("=", ": ")).join("\n"))
+      )
+    }
+
+    case "multipart/form-data": {
+      return pipe(
+        O.Do,
+
+        O.bind("boundary", () =>
```
AndrewBastin, line 173: Some comments here of what is happening would be a bit useful

packages/hoppscotch-app/helpers/types/CurlParserResult.ts
```diff
@@ -0,0 +1,22 @@
+import { HoppRESTReqBody } from "@hoppscotch/data"
+
+export type curlParserRequest = {
+  url: string
+  urlWithoutQuery: string
+  compressed: boolean
+  query: any // change it
+  headers: any // change it
```
AndrewBastin, line 8: ?

packages/hoppscotch-app/components/http/ImportCurl.vue
```diff
@@ -124,7 +138,35 @@ const handleImport = () => {
...
+          isFile: false,
+          key,
+          value: parsedCurl.multipartUploads[key],
+        })
+      }
+      finalBody = {
+        contentType: "multipart/form-data",
+        body: ydob,
+      } as HoppRESTReqBodyFormData
+    }
```
AndrewBastin, line 169: It would be cool if we don't keep all the processing steps in the UI, let the UI worry about presentation and working with the state stores.

If we can actually have a separate function which can take the input string and spit out the HoppRESTRequest object, that would be the most ideal situation

AndrewBastin (changes requested):
Overall, a great attempt, but lets just get some stuff done 😅

packages/hoppscotch-app/helpers/functional/debug.ts
```diff
@@ -7,3 +10,20 @@ export const trace = <T>(x: T) => {
...
+ * @returns The parameter `x` passed to this
+ */
+export const namedTrace =
+  (name: string) =>
+  <T>(x: T) => {
+    const printFn = (val?: T | string) => console.log(`${name}: ${val ?? x}`)
+    pipe(
+      O.tryCatch(() => JSON.stringify(x)),
+      O.match(printFn, printFn)
+    )
```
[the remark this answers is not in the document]
kyteinsky: Sometimes it just printed `[object Object]`, so I did that.

It also had a filter to check if it is `not an array` and of type `object`, but tryCatch handled it regardless and there was no performance requirement since we were debugging, so I removed that O.filter.

packages/hoppscotch-app/helpers/functional/json.ts
```diff
@@ -0,0 +1,9 @@
+import * as O from "fp-ts/Option"
+
+/**
+ * Checks and Parses JSON string
+ * @param str Raw JSON data to be parsed
+ * @returns Option type with some(JSON data) or none
+ */
+export const safeParseJSON = (str: string): O.Option<string> =>
```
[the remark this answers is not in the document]
kyteinsky: Oh yeah, I guess I overlooked it.

`O.Option<object>` works?

packages/hoppscotch-app/helpers/curlparser.ts
```diff
@@ -2,105 +2,77 @@ import * as URL from "url"
...
-        ? joinDataArguments(parsedArguments["data-ascii"])
-        : parsedArguments["data-ascii"],
-      dataArray: Array.isArray(parsedArguments["data-ascii"])
-        ? parsedArguments["data-ascii"]
-        : null,
-      isDataBinary: false,
-    }
-  }
-}
+const joinDataArguments = (dataArguments: string[]) => dataArguments.join("&")
```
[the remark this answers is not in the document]
kyteinsky: Since it was there since ancient times, I had left it there :rofl: 

Will replace it.

packages/hoppscotch-app/helpers/curlparser.ts
kyteinsky: > I don't think the explicit type definition is necessary ?

It creates these type errors:

```
Element implicitly has an 'any' type because expression of type 'string' can't be used to index type
'{ "--request": string; "--header": string; "--url": string; "--form": string; "--data-raw": string; "--data": string; "--user": string; }'.

No index signature with a parameter of type 'string' was found on type
'{ "--request": string; "--header": string; "--url": string; "--form": string; "--data-raw": string; "--data": string; "--user": string; }'.ts(7053)
```

Even though `r` is taken from replaceables
```ts
  for (const r in replaceables) {
    for (const sym of ["'", '"', " "])
      curlCommand = curlCommand.replace(RegExp(r + sym), replaceables[r] + sym)
  }
```

packages/hoppscotch-app/helpers/curlparser.ts
kyteinsky: You got it.

packages/hoppscotch-app/helpers/functional/debug.ts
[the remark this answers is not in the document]
kyteinsky: @AndrewBastin

kyteinsky:
Finally! 

_sigh of relief_

packages/hoppscotch-app/helpers/contentParser.ts
```diff
@@ -0,0 +1,274 @@
...
+ * Detects the content type of the input string
+ * @param rawData String for which content type is to be detected
+ * @returns Content type of the data
+ */
+export function detectContentType(
+  rawData: string
+): HoppRESTReqBody["contentType"] {
+  let contentType: HoppRESTReqBody["contentType"]
+
+  if (safeParseJSON(rawData)._tag === "Some") {
```
AndrewBastin, line 19: Do not use tags!

Use `O.isSome`

packages/hoppscotch-app/helpers/contentParser.ts
```diff
@@ -0,0 +1,274 @@
...
+    contentType = "application/json"
+  } else if (/<\/?[a-zA-Z][\s\S]*>/i.test(rawData)) {
+    if (prettifyXml(rawData)._tag === "Some") {
```
AndrewBastin, line 22: Use `O.isSome`

packages/hoppscotch-app/helpers/contentParser.ts
```diff
@@ -0,0 +1,274 @@
...
+    case "application/hal+json":
+    case "application/ld+json":
+    case "application/vnd.api+json":
+    case "application/json": {
+      return pipe(
+        rawData,
+        safeParseJSON,
+        O.map((parsedJSON) => JSON.stringify(parsedJSON, null, 2)),
+        // O.getOrElse(() => O.some("{}")),
+        O.match(() => O.some("{}"), O.some)
```
AndrewBastin, line 155: `O.getOrElse(() => "{}")`

packages/hoppscotch-app/helpers/functional/debug.ts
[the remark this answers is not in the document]
AndrewBastin: > Sometimes it just printed [object Object], so I did that.

My suggestion won't. Try it 😉 
NOTE that it is ```console.log(`${name}: `, x)``` not ```console.log(`${name}: ${x}`)```
(x is a second parameter)

packages/hoppscotch-app/helpers/types/CurlParserResult.ts
```diff
@@ -0,0 +1,27 @@
...
+  compressed: boolean
+  queries: HoppRESTParam[]
+  hoppHeaders: HoppRESTHeader[]
+  method: RESTMethod
+  contentType: HoppRESTReqBody["contentType"]
+  body: HoppRESTReqBody["body"]
+  cookies:
+    | {
+        [key: string]: string
+      }
```
AndrewBastin, line 21: Use `Record` type ?

packages/hoppscotch-app/helpers/types/RESTMethod.ts
```diff
@@ -0,0 +1,11 @@
...
+  | "get"
+  | "post"
+  | "put"
+  | "patch"
+  | "delete"
+  | "head"
+  | "connect"
+  | "options"
+  | "trace"
+  | "custom"
```
AndrewBastin, line 11: method param is a string for a reason ;)
"CUSTOM" is not a valid REST Method, instead when the method is set to CUSTOM, the user can edit the Method field to use a custom text. Hence, any valid string can happen.

AndrewBastin (changes requested):

AndrewBastin (approved):

PR #2104, opened 2022-01-31
feat: handle test script execution failure

### LINKED ISSUE #2079: [bug]: Tests not updating with faulty code

### Current behavior

When I add a faulty test condition, it won't get updated.

```js
pw.test("Status code is 200", ()=> {
    pw.expect(pw.response.status).toBe(200);
});

pw.test("Response is an array", () => {
  const tasks = pw.response.body.json();
  pw.expect(tasks).toBeType("array");
});
```

Results in an outdated (previous, previous, previous, ...) title and test condition.
For example above code shows me twice `Expected '200' to be '200'` (I had `pw.expect(pw.response.status).toBe(200);` in the second test before).

### Steps to reproduce

- Add 2 tests
- Run the request
- Check the test results
- Edit the second test
- Run the request
- Check the test result

### Environment

Production

### Version

Cloud

RishabhAgarwal-2001:
Closes #2079 

### Description
Existing code didn't handle the case when execution of test script was not successful. In case of a faulty test/post-request script, the users were greeted with the results of the latest valid test script. 
With this PR, the case of script execution failure is now covered. Users will see an error message instead of old test results.

### Checks

### Additional Information

N/A


liyasthomas (dismissed):
LGTM.

packages/hoppscotch-app/helpers/types/HoppTestResult.ts
```diff
@@ -13,4 +13,5 @@ export type HoppTestResult = {
   tests: HoppTestData[]
   expectResults: HoppTestExpectResult[]
   description: string
+  scriptError?: boolean
```
AndrewBastin, line 16: Why is this optional ? Couldn't we just set this to false on error ?

AndrewBastin (changes requested):

packages/hoppscotch-app/helpers/types/HoppTestResult.ts
RishabhAgarwal-2001: I thought of only including it in case of an error. While we can certainly keep this `false` for correct test results. I made it optional to make less changes to existing code.

RishabhAgarwal-2001: Offlined. Making `scriptError` field non-optional seems a better solution in maintaining long term code consistency. 
Required changes have been made.

liyasthomas (approved):

AndrewBastin (approved):
LGTM 💯

PR #2110, opened 2022-02-06
feat: nested variable name resolution

SentientMatt:
### Description
This PR adds handling for nested(dynamic) variable names.

This code should handle infinite depth in actual variable value resolution, but only support 2 levels deep for highlighting and tooltips due to the restrictions on using Regex directly without regression for matching <<>>.

### Checks
- [ ] My code requires changes to the documentation (Don't believe so, though potentially?)
- [x ] All the tests have passed

### Additional Information
Previously <<<<**env**>>_var>> would not resolve
Now if <<**env**>> is defined as "dev" at environment level and <<dev_var>> is defined at global, this will resolve

### DIFF
--- a/packages/hoppscotch-app/components/smart/EnvInput.vue
+++ b/packages/hoppscotch-app/components/smart/EnvInput.vue
@@ -28,6 +28,7 @@ import IntervalTree from "node-interval-tree"
 import debounce from "lodash/debounce"
 import isUndefined from "lodash/isUndefined"
 import { tippy } from "vue-tippy"
+import { parseTemplateString } from "../../helpers/templating"
 import { aggregateEnvs$ } from "~/newstore/environments"
 import { useReadonlyStream } from "~/helpers/utils/composables"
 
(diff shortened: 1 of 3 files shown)


AndrewBastin:
Hi @SentientMatt.

In a funny twist of things, we actually have an internal branch to implement this feature 🤣 which we were planning to merge on Monday (a couple more kinks to work out).

I will still keep the PR open for a while to see if any of your improvements can be bought over, but still at the end we will be closing this PR without merging.

SentientMatt:
Hi @AndrewBastin,

No worries, that works out well for me too 😀.

liyasthomas:
Implemented in https://github.com/hoppscotch/hoppscotch/commit/7fc757d79469d7bb72d8737aa87387fc693e4ac5, hence closing.

SentientMatt:
Hi,

I know this is a million month later, but finally got back to properly testing again. I see this works for variable resolution in the body, but not in the url. Could we also please introduce path and query parameters?

AndrewBastin:
> I see this works for variable resolution in the body, but not in the url.
@anwarulislam can you look into this ?

> Could we also please introduce path and query parameters?
I didn't exactly get that part ?

PR #2126, opened 2022-02-20
Enable image lens for WebP

kidonng:
Closes # 

### Description

With all major browsers have added support for more than one year (Safari being the latest), we can safely assume WebP is available most of the time.

👇 Meta warning: below is a WebP image showing WebP support

[](https://caniuse.com/?search=webp)

If the browser really doesn't support WebP, they just get a alt icon and nothing really breaks. That's still better than showing the raw binary:

image

### Checks


AndrewBastin:
@kidonng thank you for the quick PR.

WebP is indeed awesome and it was an oversight not to support it in the Image Lens 😛

Merging.

AndrewBastin (approved):

PR #2131, opened 2022-02-25
refactor: combine add and edit environment modals

kyteinsky:
### Description

Combines add and edit modals for ease of use.

@liyasthomas @AndrewBastin 

Please check if the PR does what it needs to.


packages/hoppscotch-app/components/environments/Details.vue
```diff
@@ -219,6 +226,16 @@ export default defineComponent({
         return
       }
 
+      if (!this.envList.find((env) => env.name === this.name)) {
```
liyasthomas, line 229: We can't use `env.name` since multiple environments can have the same name. Use `index` instead

kyteinsky: Oh good catch!

liyasthomas (approved):
LGTM

AndrewBastin (changes requested):
There is now an issue with saving an environment

Steps to reproduce
1. Create an environment and add some variables.
2. Save the environment
3. Now click on the newly created environment to edit it.
4. Observe that the environment variables not appearing.

kyteinsky:
@AndrewBastin I am really out of it these days, huh

liyasthomas (approved):
LGTM

AndrewBastin (changes requested):
One minor correction and then one missed implementation.

Steps to reproduce:
1. Disable any active environment selection (set to no environment)
2. Create a REST Request, add a pre-request script with the given content
```js
pw.env.set("x", "y");
```
3. Execute the request. 

The execution will lead to the assignment of the variable to environment, but since none is selected, there should be an alert in the "Test Results" tab asking what to do with it. If we selected, Create Environment there, it opens up the Edit Environment Modal but without proper configuration (It shows "Edit" instead of "New", it doesn't populate the environment configuration that will be added etc.)

AndrewBastin (changes requested):
It would be cool if on clicking "Create New Envrionment" from Test Result Alert, it fills up the environment variables the Test Results have found out onto the modal.

kyteinsky:
One thing I find inconsistent is the `Global` environment. Consider this scenario:

* Don't select any environment
* Run a pre-request script to set an env variable `variable: value`
* Click on `Add to Global` button in Test Results tab
* *`variable` gets added to `Global`*
* Now change `value` to `value2` in pre-request scripts and click `Send`
* *`value` is changed to `value2` inside `Global` env*
* Now change the `variable` (key) field in pre-request scripts to `somekey` and click `Send`
* *Test Results show that no env is selected*

AndrewBastin:
@kyteinsky Yup, that is an interesting observation.

@liyasthomas, did you miss the code to make sure changes to the global only not trigger that alert box ?

liyasthomas:
> One thing I find inconsistent is the `Global` environment. Consider this scenario:
> 
> * Don't select any environment
> * Run a pre-request script to set an env variable `variable: value`
> * Click on `Add to Global` button in Test Results tab
> * _`variable` gets added to `Global`_
> * Now change `value` to `value2` in pre-request scripts and click `Send`
> * _`value` is changed to `value2` inside `Global` env_
> * Now change the `variable` (key) field in pre-request scripts to `somekey` and click `Send`
> * _Test Results show that no env is selected_

I guess this is the expected behavior. If you did not select any environment, we'll prompt you to add the new variable to either "Global" or "Create new environment".
As you changed the variable key from `variable` to `somekey`, Hoppscotch treat it as a new variable addition thus, must ask what to do with the new environment variables.

AndrewBastin:
@liyasthomas ahh gotcha. that situation is handled, I thought @kyteinsky was talking about Global updations (which we do ignore).

@kyteinsky, if you change the key, that is considered a new key.

Global variables are kinda transparent and affects all the requests, in order to prevent the user absolutely not knowing about it, unlike an environment variable being set, all ways of setting Global Variables are intentionally explicit (requiring user's consent). Selecting no environment doesn't mean Global is the executing environment, it doesn't have the same behaviors as a regular environment. There is always an affinity to the executing environment for making changes.

kyteinsky:
Oh okay. It just seemed a bit queer to me that this happened. If that is the expected behaviour, then there's no issue. Although, imo replacing `Select environment` with `Global` would've cleared things up a bit for the end user.

@AndrewBastin @liyasthomas

liyasthomas:
Ahh, `Global` environments are not something that users can "switch context into" something but rather a set of environment variables that's available in all environment contexts.

kyteinsky:
I see. Now everything seems to come together. The way `Global` behaves.

AndrewBastin (dismissed):
LGTM 💯 

@liyasthomas a final check pass ?

liyasthomas (approved):

AndrewBastin (approved):

PR #2145, opened 2022-03-02
chore: introduce curl parser tests and minor changes

### LINKED ISSUE #2132: [bug]: Save option overrides the request name

### What I've been going through:

When I have a REST request saved in a collection with a certain name `xyz`, and I change some parameters of the request and then save it again the name of the request inside the collection is changed to `Untitled request`.

### What I expected to happen:

The request should've been saved with the same name (`xyz`).

### Steps to reproduce

1. Save a REST request with a name other than `Untitled request`
2. Change the request
3. Save it again

### Environment

Production

### Version

Local

### LINKED ISSUE #2155: [bug]: Lost URimport cURL

### Current behavior

When I use 'import cURL' then the URL path is lost

### Steps to reproduce

1. Click to 'import cURL'
2. input: curl 'https://muxueqz.top/skybook.html'
3. URL was truncate to 'https://muxueqz.top', but it should is 'https://muxueqz.top/skybook.html'

### Environment

Production

### Version

Cloud

kyteinsky:
### Description
Curl parser tests and other related changes.

@AndrewBastin 

### Checks

fixes #2132, fixes #2155


kyteinsky:
Covers 90% of curl parser lines.

AndrewBastin (changes requested):
So I have just commited some failing test cases which shouldn't fail as per the official cURL manpage ([https://curl.se/docs/manpage.html](https://curl.se/docs/manpage.html)).

Just to help you out, some quick ways to get some cURL commands to test.
1. Use Hoppscotch's "Show Code" feature. (Its there in the arrow menu in the Send button).
2. Use DevTools, Network Tab. In that you can right click on a network request and copy the cURL version of it.

Another thing (I have mentioned this before), the content-type should be set by the body field. It should not be repeated again in the headers once defined there.

Also another thing I noticed is that, when we generate imported request, for things the cURL import command doesn't specify, it should fetch the request info fields from `defaultRESTRequest` value we have defined here ([https://github.com/hoppscotch/hoppscotch/blob/d2c4d210e9bdc885b8b06b16557a671bd7f629ad/packages/hoppscotch-app/newstore/RESTSession.ts#L27](https://github.com/hoppscotch/hoppscotch/blob/d2c4d210e9bdc885b8b06b16557a671bd7f629ad/packages/hoppscotch-app/newstore/RESTSession.ts#L27)). Instead of manually specifying it (curlparser.ts > requestToHoppRequest), this way we can change the default values later on and not worry about rewriting the curl part again.

kyteinsky:
@AndrewBastin 

I tried something like this for inheriting the defaults:
```ts
  return makeRESTRequest({
    ...getDefaultRESTRequest(),
    ...(endpoint && { endpoint }),
    ...(method && { method }),
    ...(params && { params }),
    ...(headers && { headers }),
    ...(auth && { auth }),
    ...(finalBody?.body && { body: finalBody }),
  })
```

But this does not realize the dream since I cannot not set a default value due to type constraints, like `finalBody` has `HoppRESTReqBody` type and has to have a default value in order to be used inside the `makeRESTRequest` function.

AndrewBastin:
@kyteinsky One approach you can take is by using nullish coalescing ? If we are not planning to override, we can set those fields to null ? and then:

```js
  const default = getDefaultRESTRequest()
  return makeRESTRequest({
    ...default,
    endpoint: endpoint ?? default.endpoint,
    method: method ?? default.method,
    // And so on....
  })
```

kyteinsky:
yeah that's the point, the code I put out does the same thing. The problem is I can't assign method as null or undefined since it's type is string in the parseCurl method.

AndrewBastin:
Yeah, so shift how those variables work, allow for them to express you not having a value.
I don't understand what's weird about that.
Or another thing you can do is to assign those variables initially the value from default and then mutate it.

AndrewBastin (dismissed):

AndrewBastin (approved):

AndrewBastin (approved):

PR #2187, opened 2022-03-22
refactor: replace deprecated String.prototype.substr()

CommanderRoot:
### Description

[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.

### Checks


AndrewBastin (approved):
LGTM 💯

AndrewBastin:
Thank you for your quick PR @CommanderRoot ! ❤️

PR #2191, opened 2022-03-22
feat: override content-type

### LINKED ISSUE #2121: [bug]: cannot override Content-Type

### Current behavior

I need to send a WSDL request as `text/xml`, but that's not available from the drop-down types, so I chose `application/xml` and tried overriding it in the Headers section. The request is still sent as `application/xml` and not with the header I requested.

### Steps to reproduce

1. create a POST request
2. under "body" select `application/xml` and paste an XML content
3. under "headers" add `Content-Type: text/xml`
4. send request
5. check Network to see that `application/xml` was still used as a type

### Environment

Production

### Version

Cloud

anwarulislam:
Closes #2121 

### Description

Overriding `Content-Type` in Headers.

### Checks

### Additional Information

Previously, we were automatically assigning `Content-Type` from the body. No one could able to override the `Content-Type` from the header if they needed. They could be able to write the `Content-Type` but that wouldn't be prioritized. In this PR we basically fixed the issue. Checking if there's any `Content-Type` in headers then we're not adding Request Body's content-type with the request. Rather, the `Content-Type` from the header is being added in the actual request.


AndrewBastin (changes requested):
Please look into the comments for some fixes.

Also, I am not really satisfied with the behavior right now for multiple reasons.

1. If I press override and a valid override is applied, it would be cool if the button indicated that, the content-type is being overriden to this value, instead of just being an "Override" button.
2. If I create an override (which adds the header), then I remove the "Content-Type" header, clicking Override again doesn't do anything. Something would be nice there.

@liyasthomas can you chip in on the expected Override button behaviour as well ?

packages/hoppscotch-app/components/http/Body.vue
```diff
@@ -80,19 +91,45 @@
...
+  restHeaders$,
+  setRESTContentType,
+  setRESTHeaders,
+} from "~/newstore/RESTSession"
 
-export default defineComponent({
-  setup() {
-    return {
-      validContentTypes: Object.keys(knownContentTypes),
+const emit = defineEmits(["changeTab"])
```
AndrewBastin, line 107: Vue for TypeScript has an alternative way to define emits and props which tell the proper type signature of the emit functions.
Please check out: https://vuejs.org/api/sfc-script-setup.html#typescript-only-features

packages/hoppscotch-app/components/http/Body.vue
```diff
@@ -80,19 +91,45 @@
...
 
-      contentType: useStream(restContentType$, null, setRESTContentType),
-    }
+const validContentTypes = Object.keys(knownContentTypes)
+const contentType = useStream(restContentType$, null, setRESTContentType)
+
+// The functional headers list (the headers actually in the system)
+const headers = useStream(restHeaders$, [], setRESTHeaders) as Ref<
```
AndrewBastin, line 113: The type casting here is redundant.

packages/hoppscotch-app/components/http/Body.vue
```diff
@@ -80,19 +91,45 @@
...
+  HoppRESTHeader[]
+>
+
+const overridenContentType = ref<string | null>(null)
+
+watch(
```
AndrewBastin, line 119: Can't `overridenContentType` just be a computed property ?
Also, during the rewrite, try using `fp-ts` to build the result.

packages/hoppscotch-app/components/http/Body.vue
```diff
@@ -49,6 +49,17 @@
             />
           </div>
         </tippy>
+        <ButtonSecondary
+          v-tippy="{ theme: 'tooltip' }"
+          :title="
+            overridenContentType
+              ? 'The Content-Type is being overriden to ' + overridenContentType
```
AndrewBastin, line 56: Is this i18n friendly ?

cc: @liyasthomas :eyes:

AndrewBastin (changes requested):

packages/hoppscotch-app/components/http/Body.vue
```diff
@@ -80,19 +99,63 @@
...
+const headers = useStream(restHeaders$, [], setRESTHeaders)
+
+const overridenContentType = ref<string>("")
+
+watch(
+  headers,
+  (allHeaders: HoppRESTHeader[]) => {
+    overridenContentType.value = pipe(
+      allHeaders,
+      A.findFirst((h) => h.key.toLowerCase() === "content-type"),
```
liyasthomas, line 136: `overridenContentType` should check the `active` boolean value in headers entry.

liyasthomas: Please use `findLast()`
```suggestion
      A.findLast((h) => h.key.toLowerCase() === "content-type"),
```
Last occurance has highest preference if there're multiple entries with same key.

packages/hoppscotch-app/components/http/Body.vue
```diff
@@ -80,19 +99,56 @@
...
+const validContentTypes = Object.keys(knownContentTypes)
+const contentType = useStream(restContentType$, null, setRESTContentType)
+
+// The functional headers list (the headers actually in the system)
+const headers = useStream(restHeaders$, [], setRESTHeaders)
+
+const overridenContentType = computed(() =>
+  pipe(
+    headers.value,
+    A.findLast((h) => h.key.toLowerCase() === "content-type"),
```
liyasthomas, line 131:
```suggestion
    A.findLast((h) => h.key.toLowerCase() === "content-type" && h.active),
```
`overridenContentType` should check the active boolean value in the headers entry. cc: @anwarulislam

liyasthomas (changes requested):

liyasthomas (approved):
LGTM.

PR #2234, opened 2022-04-05
feat: add toInclude() function

### LINKED ISSUE #2221: [feature]: Add expect().toInclude([string]) method

### Summary

Add a `toInclude()` method to `pw.expect()`

(`toContain()` could work too, but [JS itself](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes) uses `includes()`, not `contains()`.)

I want to assert that an unstructured page contains a certain string of text.

(For example, I want to assert that `/attractions?q=egypt` should contain `Pyramids of Giza`)

### Current workaround

Currently, you can do the following:
```
pw.expect(pw.response.body.includes("example")).toBe(true)
```

This is suboptimal for two reasons:
1) It doesn't produce a detailed assertion message.
2) It's a bit verbose. 🙂 

### Contribution notes

I'm happy to take a first-pass at an implementation myself, if a maintainer is OK with it! 😄

ace-n:
Closes #2221

### Description
Adds `toInclude` operator to testing suite

### Checks

- [ ] My code requires changes to the documentation **(N/A)**
- [ ] I have updated the documentation as required **(N/A)**

### Additional Information
`"123".toInclude(123)` will return `true` even though their types are different (`string` vs. `int`).

I included a comment about this, with a link to the [relevant JS spec](https://tc39.es/ecma262/multipage/text-processing.html#sec-string.prototype.includes).

If this is an issue, we should switch to `indexOf` instead of `includes`.


AndrewBastin:
Can we also extend support for the `toInclude` to work with Arrays as `Array.prototype.includes` is also a thing and it would totally remove any confusion ?

AndrewBastin (dismissed):
Thank you a well written and thought out PR! 🤗 

LGTM 💯

PR #2241, opened 2022-04-07
feat: add `New Request` button for folder and collection

### LINKED ISSUE #2156: [feature]: Add 'new request' at collection level

### Summary

- Add 'new request' at collection.
- Right now only a folder can be created inside of another collection.
- Would be nice have a new feature to add an empty request at collection level.

- It is confusing at times while saving the request.
- It would be great if a new empty request would be instantiated when working on a new request.

kyteinsky:
### Description

  New Request button is now available in the kabab menu
  for both folders and collections in REST as well as
  GraphQL.
  It creates a new request, saves it to the store and
  points to it.

Closes #2156

### DIFF
--- a/packages/hoppscotch-app/components/collections/AddRequest.vue
+++ b/packages/hoppscotch-app/components/collections/AddRequest.vue
@@ -38,6 +38,7 @@ import { defineComponent } from "@nuxtjs/composition-api"
 export default defineComponent({
   props: {
     show: Boolean,
+    folder: { type: Object, default: () => {} },
     folderPath: { type: String, default: null },
     collectionIndex: { type: Number, default: null },
   },
@@ -54,6 +55,7 @@ export default defineComponent({
       }
       this.$emit("add-request", {
         name: this.name,
+        folder: this.folder,
         path: this.folderPath || `${this.collectionIndex}`,
       })
       this.hideModal()
--- a/packages/hoppscotch-app/components/collections/AddRequest.vue
+++ b/packages/hoppscotch-app/components/collections/AddRequest.vue
@@ -32,38 +32,40 @@
   </SmartModal>
 </template>
 
-<script>
-import { defineComponent } from "@nuxtjs/composition-api"
+<script setup lang="ts">
+import { useRESTRequestName } from "~/newstore/RESTSession"
+import { useI18n, useToast } from "~/helpers/utils/composables"
 
-export default defineComponent({
-  props: {
-    show: Boolean,
-    folder: { type: Object, default: () => {} },
-    folderPath: { type: String, default: null },
-    collectionIndex: { type: Number, default: null },
-  },
-  data() {
-    return {
-      name: null,
-    }
-  },
-  methods: {
-    addRequest() {
-      if (!this.name) {
-        this.$toast.error(`${this.$t("error.empty_req_name")}`)
-        return
-      }
-      this.$emit("add-request", {
-        name: this.name,
-        folder: this.folder,
-        path: this.folderPath || `${this.collectionIndex}`,
-      })
-      this.hideModal()
-    },
-    hideModal() {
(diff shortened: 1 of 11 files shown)


AndrewBastin (changes requested):
Some minor corrections.

packages/hoppscotch-app/components/collections/index.vue
```diff
@@ -651,13 +668,68 @@ export default defineComponent({
...
+      // TODO: check if the request being worked on
+      // is being overwritten (selected or not)
+      const { folder, path } = payload
+      this.$data.editingFolder = folder
+      this.$data.editingFolderPath = path
+      this.displayModalAddRequest(true)
+    },
+    onAddRequest({ name, folder, path }) {
+      const newRequest = {
+        ...getRESTRequest(),
```
AndrewBastin, line 688:
```suggestion
        ...cloneDeep(getRESTRequest()),
```

AndrewBastin (changes requested):
Really minor changes required.

packages/hoppscotch-app/components/collections/graphql/AddRequest.vue
```diff
@@ -0,0 +1,78 @@
...
+}>()
+
+const emit = defineEmits<{
+  (e: "hide-modal"): void
+  (e: "add-request", v: object): void
+}>()
+
+const name = ref("")
+
+watch(props, (v) => {
```
AndrewBastin, line 57:
```suggestion
watch(() => props.show, (show) => {) // 'show' will have the value of the show prop.
```

packages/hoppscotch-app/components/collections/graphql/AddRequest.vue
```diff
@@ -0,0 +1,78 @@
...
+  if (v.show) {
+    name.value = useGQLRequestName().value
```
AndrewBastin, line 59: Do not use composables like this, because many composables are only intended to be called during setup and not afterwards in the lifecycle.
Also, for this function, it creates a sort of a leak, as each time, a stream connection is created again when the watcher is called.

You don't have to subscribe to a stream of values here (which is what the composable does), we only want the value at that instant.
Hence, instead of doing this, we can just get the present GQL Session from `getGQLSession()` and then get the name from its request.
That gives us the value at that instant, which is what we need.

packages/hoppscotch-app/components/collections/AddRequest.vue
```diff
@@ -0,0 +1,82 @@
...
+}>()
+
+const emit = defineEmits<{
+  (e: "hide-modal"): void
+  (e: "add-request", v: object): void
+}>()
+
+const name = ref("")
+
+watch(props, (v) => {
```
AndrewBastin, line 61: Same issues as given on the counterpart of this modal in GQL. Please look into them.

Here you can use `getRESTRequest()` to get the request at that instant.

AndrewBastin (changes requested):
Corrections needed.

Using composables (`useXXX` functions) outside the context of a setup function is not a good practice.

packages/hoppscotch-app/components/collections/graphql/AddRequest.vue
kyteinsky: > Also, for this function, it creates a sort of a leak, as each time, a stream connection is created again when the watcher is called.

Ahh yes, I didn't look up the function.

packages/hoppscotch-app/components/collections/graphql/AddRequest.vue
kyteinsky: What is the performance difference if we use `onBeforeUpdate` instead of `watch`?

AndrewBastin (approved):
lgtm :100:

PR #2267, opened 2022-04-14
refactor : migrate components to script setup on ts

nivedin:
### Description
Migrate button and environment components to script setup with typescript.

### Checks

### DIFF
--- a/packages/hoppscotch-app/components/environments/Details.vue
+++ b/packages/hoppscotch-app/components/environments/Details.vue
@@ -142,8 +142,8 @@ const toast = useToast()
 const props = withDefaults(
   defineProps<{
     show: boolean
-    action: string
-    editingEnvironmentIndex: number | string | null
+    action: "edit" | "new"
+    editingEnvironmentIndex: number | "Global" | null
     envVars: () => Environment["variables"]
   }>(),
   {
--- a/packages/hoppscotch-app/components/environments/Details.vue
+++ b/packages/hoppscotch-app/components/environments/Details.vue
@@ -2,7 +2,7 @@
   <SmartModal
     v-if="show"
     dialog
-    :title="$t(`environment.${action}`)"
+    :title="t(`environment.${action}`)"
     @close="hideModal"
   >
     <template #body>
@@ -20,24 +20,24 @@
             @keyup.enter="saveEnvironment"
           />
           <label for="selectLabelEnvEdit">
-            {{ $t("action.label") }}
+            {{ t("action.label") }}
           </label>
         </div>
         <div class="flex items-center justify-between flex-1">
           <label for="variableList" class="p-4">
-            {{ $t("environment.variable_list") }}
+            {{ t("environment.variable_list") }}
           </label>
           <div class="flex">
             <ButtonSecondary
               v-tippy="{ theme: 'tooltip' }"
-              :title="$t('action.clear_all')"
+              :title="t('action.clear_all')"
               :svg="clearIcon"
               @click.native="clearContent()"
             />
             <ButtonSecondary
               v-tippy="{ theme: 'tooltip' }"
               svg="plus"
-              :title="$t('add.new')"
+              :title="t('add.new')"
               @click.native="addEnvironmentVariable"
             />
           </div>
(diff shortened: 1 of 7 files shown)


packages/hoppscotch-app/components/environments/Details.vue
```diff
@@ -130,136 +130,146 @@ import {
...
+} from "~/helpers/utils/composables"
+
+const t = useI18n()
+const toast = useToast()
+
+const props = withDefaults(
+  defineProps<{
+    show: boolean
+    action: string
+    editingEnvironmentIndex: number | string | null
```
AndrewBastin, line 146: Correction. The only string possible (as per the old code) is "Global". Use TypeScript's String Literal Types for it.
Read: https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types

`editingEnvironmentIndex: number | "Global" | null`

packages/hoppscotch-app/components/environments/Environment.vue
```diff
@@ -104,47 +105,49 @@ import {
...
-      } else duplicateEnvironment(this.environmentIndex)
-    },
-  },
-})
+const t = useI18n()
+const toast = useToast()
+
+const props = defineProps<{
+  environment: Environment
+  environmentIndex: number | string | null
```
AndrewBastin, line 115: Use String Literal Type `"Global"` instead of `string`.
See the similar comment on `components/environments/Details.vue`.

packages/hoppscotch-app/components/environments/Environment.vue
```diff
@@ -104,47 +105,49 @@ import {
...
+
+const duplicateEnvironments = () => {
+  if (props.environmentIndex === "Global") {
+    createEnvironment(`Global - ${t("action.duplicate")}`)
+    setEnvironmentVariables(
+      environmentsStore.value.environments.length - 1,
+      getGlobalVariables().reduce<ReduceReturnType>((gVars, gVar) => {
+        gVars.push({ key: gVar.key, value: gVar.value })
+        return gVars
+      }, [])
```
AndrewBastin, line 149:
```suggestion
      cloneDeep(getGlobalVariables()) // import cloneDeep from "lodash/cloneDeep"
```

packages/hoppscotch-app/components/environments/index.vue
```diff
@@ -120,65 +120,63 @@
...
+)
 
-    const globalEnvironment = computed(() => ({
-      name: "Global",
-      variables: globalEnv.value,
-    }))
+const showModalImportExport = ref(false)
+const showModalDetails = ref(false)
+const action = ref("new")
+const editingEnvironmentIndex = ref<string | number | undefined>("Global")
```
AndrewBastin, line 159: Only a single string value is possible for the type. "Global" . This can be represented by using TypeScript's String Literal Types
(see comment in `components/environments/Environment.vue` for further explanation)

ALSO, are `environmentIndex` values of type `"Global" | number | undefined` or `"Global" | number | null`. In other places it is marked as `null` see what is the actual value flowing and correct to that type.
undefined is not the same as null in JavaScript and TypeScript upholds that.

packages/hoppscotch-app/components/environments/index.vue
```diff
@@ -120,65 +120,63 @@
...
+  setCurrentEnvironment
+)
 
-    const globalEnvironment = computed(() => ({
-      name: "Global",
-      variables: globalEnv.value,
-    }))
+const showModalImportExport = ref(false)
+const showModalDetails = ref(false)
+const action = ref("new")
```
AndrewBastin, line 158: We can lock the value of action with a Union Type of String Literals as well.

Here, the type will be `"new" | "edit"`.
here you will have to use a `ref<"new" | "edit">` style as TypeScript will assume `string`

AndrewBastin (changes requested):
Some minor issues here and there to correct. Mostly regarding TypeScript.

packages/hoppscotch-app/components/environments/index.vue
nivedin: In the old file, the value was "Global" | number | undefined but in `components/environments/Details.vue` it is null I think null will be more accurate in this case.

liyasthomas (approved):
LGTM

PR #2268, opened 2022-04-16
feat: loading states for modal buttons

kyteinsky:
### Description

Adds loading state to buttons in different modals.

### DIFF
--- a/packages/hoppscotch-app/components/collections/Add.vue
+++ b/packages/hoppscotch-app/components/collections/Add.vue
@@ -26,6 +26,7 @@
       <span>
         <ButtonPrimary
           :label="$t('action.save')"
+          :loading="loadingState"
           @click.native="addNewCollection"
         />
         <ButtonSecondary
@@ -43,6 +44,7 @@ import { defineComponent } from "@nuxtjs/composition-api"
 export default defineComponent({
   props: {
     show: Boolean,
+    loadingState: Boolean,
   },
   data() {
     return {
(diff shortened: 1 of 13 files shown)


kyteinsky:
`SmartConfirmModal` would need modifications if delete functionality also needs loading.

AndrewBastin (approved):
lgtm :100:

ask and wait for @liyasthomas to know whether we should do anything about the Delete option.

Although, I would think it will not hurt to do that.

liyasthomas:
@kyteinsky it would be great if we can introduce loading state in delete confirmation modals too. Mainly the delete request, delete collection, delete folder prompts on teams.

kyteinsky:
@liyasthomas There are two ways to add loading in delete confirmation modals, one by introducing an optional prop in `SmartConfirmModal` and the other by creating a new component altogether. Which path should I take?

liyasthomas:
Introduce a prop in `SmartConfirmModal` will be the best way to do it.

kyteinsky:
Should I change the Folder delete logic in the respective files `CollectionMyFolder` and `CollectionTeamsFolder` or move them to `index.vue`? Or maybe move the delete functions for Request and Collection to their respective files?

AndrewBastin:
@kyteinsky you can maintain whatever way it is right now and not worry about where the logic is (although, all of them should be in index.vue). We are planning a major rewrite of the collection tree structure and we will clear most of it during that time.

AndrewBastin (approved):
lgtm 💯

PR #2270, opened 2022-04-17
fix: add aria labels to forms

### LINKED ISSUE #2137: [bug]: A lot of controls are not labeled for accessibility

### Current behavior

When I view hoppscotch.io with a screen reader, a lot of buttons and anchor tags on the page have no discernible text, making it impossible to figure out what these controls are for. I expect these to be properly labeled as per WCAG standards so people who don't rely on their eyesight can make use of the tool.  This is easy enough to fix by just setting a textual label on these components, can even be an aria-label which only labels elements for screen readers and won't take up visual space.

### Steps to reproduce

1. Load Narrator/VoiceOver/Orca/NVDA depending on what your OS is.
2. Visit hoppscotch.io.
3. Tab through the controls on the page. This is not how screenreader users generally navigate a page, but it will give a decent idea of just how many elements are unlabeled.

### Environment

Production

### Version

Cloud

sweetkane:
Closes #2137 

### Description
Add code to make all forms on page have an aria label equivelent to their placeholder text. This fix works as long as forms have descriptive placeholder text. There may still be more screen reader accessibility improvements to make in the future, but it looks like every component is properly labeled for screen reader use at the moment. 

### Checks

#### Google Lighthouse accessibility rating before change
Lighthouse pre
#### Google Lighthouse accessibility rating after change
Lighthouse post

### DIFF
--- a/packages/hoppscotch-app/components/smart/EnvInput.vue
+++ b/packages/hoppscotch-app/components/smart/EnvInput.vue
@@ -123,6 +123,7 @@ const envTooltipPlugin = new HoppReactiveEnvPlugin(envVars, view)
 
 const initView = (el: any) => {
   const extensions: Extension = [
+    EditorView.contentAttributes.of({ "aria-label": props.placeholder }),
     inputTheme,
     tooltips({
       position: "absolute",


liyasthomas (approved):
LGTM 💯

PR #2280, opened 2022-04-22
fix bug when update env first row

17307:
The original judgment logic resulted in the first row of environment variables never being updated.

Closes # 

### Checks


AndrewBastin:
@17307 

Thank you for the fix, this was a really stupid miscalculation which shouldn't have happened. But this is not a good way to patch it and it seems you have not allowed maintainers to edit your PR content. So I am closing this PR and pushing the patch.

PR #2321, opened 2022-05-04
feat: Shortcodes UI implementation

nivedin:
### Description
This PR introduces a feature in which authenticated user can view their created shortcodes and can delete or copy them.

### Checks


packages/hoppscotch-app/helpers/shortcodes/ShortCodeListAdapter.ts
```diff
@@ -0,0 +1,82 @@
...
+        this.timeoutHandle = setTimeout(() => func(), POLL_DURATION)
+      }
+    }
+
+    func()
+  }
+
+  public dispose() {
+    this.isDispose = true
+    clearTimeout(this.timeoutHandle as any)
```
AndrewBastin, line 49: TypeScript is raising a valid point here, what if there is no timeout handle (for e.g, already disposed or when not initialized). Should we call `clearTimeout` then ?

packages/hoppscotch-app/pages/profile.vue
```diff
@@ -273,6 +355,39 @@ const sendEmailVerification = () => {
...
+const loading = computed(
+  () => adapterLoading.value && myShortCodes.value.length === 0
+)
+
+const deleteShortCode = (codeID: string) => {
+  pipe(
+    backendDeleteShortcode(codeID),
+    TE.match(
+      (err) => {
+        toast.error(`${t("error.something_went_wrong")}`)
```
AndrewBastin, line 376: We can detect exactly what happened here. For example, if its a network error (see err.type)

packages/hoppscotch-app/pages/profile.vue
```diff
@@ -273,6 +355,39 @@ const sendEmailVerification = () => {
...
+        console.error(err)
+      },
+      () => {
+        toast.success(`${t("shortcodes.deleted")}`)
+        refetchShortcodes()
```
AndrewBastin, line 381: Do we have to refetch the entire list again just for deletion ? The adapter should handle deletion without refetching everything.

packages/hoppscotch-app/pages/profile.vue
```diff
@@ -273,6 +355,39 @@ const sendEmailVerification = () => {
...
+      }
+    )
+  )()
+}
+
+const refetchShortcodes = () => {
```
AndrewBastin, line 387: This function doesn't need to exist.

AndrewBastin (changes requested):

Also, please maintain a single naming convention across the codebase, for example, the backend operations define it as a 'Shortcode', not a 'ShortCode' or 'Short Code'. Different naming styles are being used throughout the code here, normalize it into a single style.

(it should be Shortcode/shortcode, not shortCode/short-code/Shortcode etc.)

liyasthomas:
@nivedin this branch seems broken. Please rebase to the latest `main` and update the PR.

PR #2324, opened 2022-05-05
fix : save request popup bug

### LINKED ISSUE #2318: [bug]: The `Save request` popups too many times

### Current behavior

When I save the current request and I click to another request the pop up will ask me to save the request. But I already saved from the save button. Most of the times if I click save it will create a new request

This happens almost every time I change from one request to another

### Steps to reproduce

1. Make request 1
2. Click `Save` to save the request
3. Click on another request 
4. The pop up is shown
image

### Environment

Release

### Version

Self-hosted

nivedin:
Closes #2318 

### Description
This PR fixes the unsaved change popup bug in which the popup tends to trigger even though the request was saved.PR also fix a small bug that occurred when changing context while saving the request.  

### Checks


AndrewBastin (approved):
lgtm :100:

PR #2366, opened 2022-05-26
fix: curl parser url sanitisation

kyteinsky:
- All Unsafe characters are removed from the url string following RFC 1738
- All the non-tagged arguments are checked for possible presence of url
- Some more local url checks were added.

Reference: [RFC 1738](http://www.faqs.org/rfcs/rfc1738.html)


AndrewBastin (approved):
lgtm :100:

PR #2370, opened 2022-05-27
fix: curl parser json detection issues

kyteinsky:
### Description:

Replace `\\"` with `"` for detection of JSON content.

### DIFF
--- a/packages/hoppscotch-app/helpers/curl/sub_helpers/contentParser.ts
+++ b/packages/hoppscotch-app/helpers/curl/sub_helpers/contentParser.ts
@@ -7,8 +7,7 @@ import { tupleToRecord } from "~/helpers/functional/record"
 import { safeParseJSON } from "~/helpers/functional/json"
 import { optionChoose } from "~/helpers/functional/option"
 
-const isJSON = (jsonString: string) =>
-  pipe(jsonString.replaceAll('\\"', '"'), safeParseJSON, O.isSome)
+const isJSON = flow(safeParseJSON, O.isSome)
 
 const isXML = (rawData: string) =>
   pipe(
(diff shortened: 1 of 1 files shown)


packages/hoppscotch-app/helpers/curl/sub_helpers/contentParser.ts
```diff
@@ -158,14 +158,12 @@ const getXMLBody = (rawData: string) =>
...
-    O.map((parsedJSON) => JSON.stringify(parsedJSON, null, 2)),
-    O.getOrElse(() => "{ }"),
-    O.of
-  )
+const getFormattedJSON = flow(
+  safeParseJSON,
+  O.map((parsedJSON) => JSON.stringify(parsedJSON, null, 2)),
+  O.getOrElse(() => "{ }"),
+  O.of
+)
```
AndrewBastin, line 166: Why wrap this in Option if it is never going to be having a None state ?

AndrewBastin (changes requested):

packages/hoppscotch-app/helpers/curl/sub_helpers/contentParser.ts
kyteinsky: I thought it would be nice that content type remained application/json since the parser did not display any errors. A subtle hint.

AndrewBastin: Umm okay, I will allow for now because this is just a minor thing.
But this is an anti-pattern and should not be repeated.

AndrewBastin (approved):
lgtm 💯

packages/hoppscotch-app/helpers/curl/sub_helpers/contentParser.ts
kyteinsky: hmm

PR #2372, opened 2022-05-27
feat: export hoppscotch collections as openapi document

### LINKED ISSUE #2092: [feature]: Export collections as Postman accepted file (Raml / Curl, OpenApi)

### Summary

Export collections as Postman accepted file (Raml / Curl, OpenApi/ etc)

This would be a good solution for somebody to quickly jump between Postman and Hopscotch.
For example: I spent a lot of time setting up a collection with a lot of requests and only after that I realized that in Hoppscotch I couldn't get all the info i wanted from the response. ( "time" / "duration" expanded info for a request - check picture). Following this discovery, I tried to export the collection and then to **migrate it to Postman unsuccesfully**.

postman_duration_example

amk-dev:
Closes #2092 

### Description
feature to export Hoppscotch collections as an Open API document.

### Checks

### Additional Information

Problems to tackle / limitations 

~~1. No Support For Enviornment Variables
    1. Openapi doesn't support enviornment variables anywhere other than the server url
    2. One way to tackle this is to replace env variables with their values when exporting, but sensitive env variables may get exported, not good from a security standpoint. but seems like the practical thing to do.
2. Openapi doesn't support giving values for authentication. eg: we can define the auth scheme is basic auth, but won't be able to give the username and password.
    1. Workaround might be to generate the headers and add them to the headers object~~
3. When we convert a hopp collection to open api document, there are possible chances for missing data,
    for eg: lets say we have two requests 
            1.POST https://jsonplaceholder.com/posts
            2.POST https://someotherinfo.com/posts with some different body
            So right now we don't have a way to represent these two requests in one openapi document. workaround might be to generate multiple openapi documents


packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,213 @@
...
+  HoppRESTReqBody,
+  knownContentTypes,
+} from "@hoppscotch/data"
+import { pipe } from "fp-ts/function"
+import * as TE from "fp-ts/TaskEither"
+import * as E from "fp-ts/Either"
+import * as A from "fp-ts/Array"
+import * as RA from "fp-ts/ReadonlyArray"
+import { OpenAPIV3 } from "openapi-types"
+import { isPlainObject, omit } from "lodash"
```
AndrewBastin, line 14:
```suggestion
import isPlainObject from "lodash/isPlainObject"
import omit from "lodash/omit"
```

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,213 @@
...
+      pipe(
+        paths,
+        A.reduce({}, (allPaths: OpenAPIV3.PathsObject, path) => ({
+          ...allPaths,
+          [path.pathname]: omit(path, "pathname"),
+        })),
+        generateOpenApiDocument,
+        E.right
+      )
+    )
```
AndrewBastin, line 207: We are wrapping this block in a Right Either at the end, so basically this is acting like a map operation. So, why not just use map

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,213 @@
...
+    A.flatten,
+    E.sequenceArray,
+    E.map(RA.toArray),
+    E.chainW((paths) =>
+      pipe(
+        paths,
+        A.reduce({}, (allPaths: OpenAPIV3.PathsObject, path) => ({
+          ...allPaths,
+          [path.pathname]: omit(path, "pathname"),
+        })),
```
AndrewBastin, line 203: Can be simplified into a flow

```suggestion
    E.chainW(
      flow( // import { flow } from "fp-ts/function"
        paths,
        A.reduce({}, (allPaths: OpenAPIV3.PathsObject, path) => ({
          ...allPaths,
          [path.pathname]: omit(path, "pathname"),
        })),
```

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,213 @@
...
+              },
+            ],
+          },
+          E.right
+        )
+    ),
+    E.chainW(({ basePathObject, method }) =>
+      pipe(
+        E.tryCatch(
+          () => pipe(request.body, generateOpenApiRequestBody),
```
AndrewBastin, line 172:
```suggestion
          () => generateOpenApiRequestBody(request.body),
```

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,213 @@
...
+          () => "INVALID_BODY"
```
AndrewBastin, line 173:
```suggestion
          () => "INVALID_BODY" as const
```

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,213 @@
...
+          [path.pathname]: omit(path, "pathname"),
+        })),
+        generateOpenApiDocument,
+        E.right
+      )
+    )
+  )
+
+const exporter: HoppExporter<HoppCollection<HoppRESTRequest>[]> = (content) =>
+  pipe(content, convertHoppToOpenApiCollection, JSON.stringify, TE.right)
```
AndrewBastin, line 211: Can be simplified with a flow

```suggestion
const exporter: HoppExporter<HoppCollection<HoppRESTRequest>[]> = 
  flow(content, convertHoppToOpenApiCollection, JSON.stringify, TE.right)
```

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,213 @@
...
+    "get",
+    "post",
+    "patch",
+    "delete",
+    "options",
+    "head",
+    "patch",
+    "trace",
+  ].includes(method)
+}
```
AndrewBastin, line 111:
```suggestion
const isValidMethod = (method: string): method is OpenAPIV3.HttpMethods =>
  [
    "get",
    "post",
    "patch",
    "delete",
    "options",
    "head",
    "patch",
    "trace",
  ].includes(method)
```

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,213 @@
...
+    E.bind("contentType", () =>
+      pipe(
+        hoppRequestBody.contentType,
+        E.fromPredicate(
+          (contentType) => !!(contentType && knownContentTypes[contentType]),
+          () => "INVALID_CONTENT_TYPE"
+        )
+      )
+    )
+  )
```
AndrewBastin, line 61: ?

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,213 @@
+/* eslint-disable */
```
AndrewBastin, line 1: Not exactly a good sign for whether you are doing things right 🤣

AndrewBastin (changes requested):

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
amk-dev: was getting a lot of undef and unused variable errors while experimenting. thats why 😂. 
actually this is wip. i had to kind of do some clean up on my pc, so i just commited it as a wip commit. i'll rebase it.

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,397 @@
...
+const isFormDataBody = (
+  body: HoppRESTReqBody
+): body is HoppRESTReqBodyFormData => body.contentType === "multipart/form-data"
+
+const isEmptyBody = (body: HoppRESTReqBody): body is HoppRestReqBodyEmpty =>
+  !(body.body && body.contentType)
+
+type RequestBodyGenerationErrors = "INVALID_CONTENT_TYPE" | "INVALID_BODY"
+
+export const generateOpenApiRequestBody = (
```
AndrewBastin, line 126: Comments explaining what we are doing here please :)

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,397 @@
...
+  hoppRequestBody: HoppRESTReqBody
+): E.Either<RequestBodyGenerationErrors, OpenAPIV3.RequestBodyObject> =>
+  pipe(
+    hoppRequestBody,
+    E.fromPredicate(isEmptyBody, () => "NON_EMPTY_BODY"),
```
AndrewBastin, line 131: `as const` assertion

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,397 @@
...
+      ApiKeyAuth: {
+        type: "apiKey",
+        name: "api-key-header-name",
+        in: "header",
+      },
+    },
+  },
+})
+
+type PathGenerationErrors =
```
AndrewBastin, line 284: Type names should be singular

```suggestion
type PathGenerationError =
```

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,397 @@
...
+  )
+
+const isFormDataBody = (
+  body: HoppRESTReqBody
+): body is HoppRESTReqBodyFormData => body.contentType === "multipart/form-data"
+
+const isEmptyBody = (body: HoppRESTReqBody): body is HoppRestReqBodyEmpty =>
+  !(body.body && body.contentType)
+
+type RequestBodyGenerationErrors = "INVALID_CONTENT_TYPE" | "INVALID_BODY"
```
AndrewBastin, line 124: Type names should be singular

```suggestion
type RequestBodyGenerationError = "INVALID_CONTENT_TYPE" | "INVALID_BODY"
```

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
amk-dev: sure. will do

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,397 @@
...
+    E.map(RA.toArray),
+    E.map(
+      flow(
+        A.reduce({}, (allPaths: OpenAPIV3.PathsObject, path) => ({
+          ...allPaths,
+          [path.pathname]: omit(path, "pathname"),
+        })),
+        generateOpenApiDocument
+      )
+    )
```
AndrewBastin, line 388: Multiple maps can be put together into a single map.

```suggestion
    E.map(
      flow(
      	RA.toArray,
        A.reduce({}, (allPaths: OpenAPIV3.PathsObject, path) => ({
          ...allPaths,
          [path.pathname]: omit(path, "pathname"),
        })),
        generateOpenApiDocument
      )
    )
```

AndrewBastin (changes requested):
Some minor things to correct

packages/hoppscotch-app/helpers/import-export/export/index.ts
```diff
@@ -1,18 +1,62 @@
...
 
-export type HoppExporter<T> = (content: T) => TE.TaskEither<string, string>
+export type ExportError = "INVALID_EXPORTER" | "IMPORT_ERROR"
+
+export type HoppExporter<T> = (
+  content: T
+) => TE.TaskEither<
+  ExportError | HoppToOpenAPIConversionError,
+  HoppCollection<HoppRESTRequest>[] | OpenAPIV3.Document
+>
```
AndrewBastin, line 16: This type has so many issues.

There needs to be more thought into how the type came about.

I will walk through the explanation.

Exporting is a process by which we take a Collection(s) (which will be a HoppCollection) and spits out an aribtrary thing.

Hence, an exporter can be defined as a function which takes in **a HoppCollection array** and spits out the result. We can definitely say that the result is a TaskEither to allow for error cases and also allow the exporter to be async if it needs to be. 

Now the problem is that TaskEither types need 2 bits of information, the error case type and the success case type. For Error case, we can't exactly say what the error of an exporter can be, it can even introduce its own errors, we don't want to express all the errors possible in the type because that is counter intuitive, instead we can take it as an argument to define an Exporter.

The second type is the next thing, we can allow that to be a type parameter as well, but the problem with that is that every exporter is going to return a different value making it difficult on the UI side code to judge how to implement the save operation.

If we go through the UI side code, you can see that it is at the end converted to a Blob and then downloaded. So why can't we define an exporter as a function which takes a collection and spits out an either which may return a Blob. This way all exporters will have the same standard return type and the UI side implementation will become super simplified.

```suggestion
export type HoppExporter<
  ReqType extends HoppRESTRequest | HoppGQLRequest,
  ErrorType extends string
> = (collections: Array<HoppCollection<ReqType>>) => TE.TaskEither<ErrorType, Blob>
```

packages/hoppscotch-app/components/collections/ImportExport.vue
```diff
@@ -316,10 +325,40 @@ const importToTeams = async (content: HoppCollection<HoppRESTRequest>) => {
...
 
-  const dataToWrite = collectionJson.value
+  const dataToWrite = await getDataToWrite(exporterId)(myCollections.value)()
+
+  if (E.isRight(dataToWrite)) {
+    writeExport(JSON.stringify(dataToWrite.right))
+  } else {
+    failedExport(dataToWrite.left)
+  }
+}
```
AndrewBastin, line 338: fp-ts implementation should look like

```ts
const exportJSON = (exporterId: string) => {
  getJSONCollection()

  return pipe(
    myCollections.value,
    getDataToWrite(exporterId),
    TE.match(failedExport, flow(JSON.stringify, writeExport))
  )()
}
```

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,406 @@
...
+  string,
+  unknown
+]): [string, OpenAPIV3.SchemaObject] => {
+  return [
+    key,
+    isPlainObject(value)
+      ? {
+          properties: {
+            ...generateEntries(value as object),
+          },
```
AndrewBastin, line 54: Can be simplified.

```suggestion
          properties: generateEntries(value as object),
```

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,406 @@
...
+    }))
+  )
+
+const isValidContentType = (
+  contentType: unknown
+): contentType is keyof typeof knownContentTypes =>
+  !!(
+    contentType &&
+    typeof contentType === "string" &&
+    objHasProperty(contentType, "string")(knownContentTypes)
```
AndrewBastin, line 105: Why are we doing this check ?

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,406 @@
...
+          () => "INVALID_CONTENT_TYPE" as const
+        ),
+        E.chainW((contentType) =>
+          pipe(
+            hoppRequestBody,
+            O.fromPredicate(isNonFormDataBody),
+            O.chain((hoppRequestBody) =>
+              pipe(
+                hoppRequestBody,
+                O.fromPredicate(
```
AndrewBastin, line 159: `pipe` can be replaced by a `flow`.

```suggestion
            O.chain(
              flow(
                O.fromPredicate(
```

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,406 @@
...
+                ),
+                O.chain(({ body }) => safeParseJSON(body)),
+                O.map(
+                  (body): OpenAPIV3.RequestBodyObject => ({
+                    content: {
+                      [contentType]: {
+                        schema: {
+                          properties: {
+                            ...generateEntries(body),
+                          },
```
AndrewBastin, line 171: Can be simplified.

```suggestion
                          properties: generateEntries(body),
```

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,406 @@
...
+                hoppRequestBody,
+                O.fromPredicate(isFormDataBody),
+                O.map(
+                  (body): OpenAPIV3.RequestBodyObject => ({
+                    content: {
+                      [contentType]: {
+                        schema: {
+                          properties: {
+                            ...generateFormDataEntries(body),
+                          },
```
AndrewBastin, line 206: Can be simplified.

```suggestion
                          properties: generateFormDataEntries(body),
```

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,406 @@
...
+  pipe(
+    hoppAuth.authType,
+    O.fromPredicate((authType) => authType === "none"),
+    O.map(() => ({})),
+    O.alt(() =>
+      pipe(
+        hoppAuth.authType,
+        O.fromPredicate(
+          (authType): authType is "basic" | "api-key" | "bearer" =>
+            authType !== "none" && authType !== "oauth-2"
```
AndrewBastin, line 233: Not future proof.

```suggestion
          (authType): authType is "basic" | "api-key" | "bearer" =>
            ["basic", "api-key", "bearer"].includes(authType)
```

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,406 @@
...
+        ),
+        O.map((authType) =>
+          pipe(
+            {
+              basic: "basicAuth",
+              "api-key": "ApiKeyAuth",
+              bearer: "bearerAuth",
+            },
```
AndrewBastin, line 241: Can be moved out and const asserted.

packages/hoppscotch-app/locales/en.json
```diff
@@ -214,7 +214,19 @@
...
     "gist_created": "Gist created",
     "require_github": "Login with GitHub to create secret gist",
-    "title": "Export"
+    "title": "Export",
+    "hopp_export_name": "Export as Hoppscotch Collections",
+    "hopp_export_title": "The collections will be exported as json",
+    "hopp_export_label": "Export as Hoppscotch Collections",
+    "openapi_export_name": "Export as an Openapi Document",
+    "openapi_export_title": "The collections will be exported as an openapi document",
+    "openapi_export_label": "Export as Openapi document",
```
AndrewBastin, line 223: ~~Openapi~~ -> OpenAPI

packages/hoppscotch-app/locales/en.json
```diff
@@ -214,7 +214,19 @@
     "create_secret_gist": "Create secret Gist",
     "gist_created": "Gist created",
     "require_github": "Login with GitHub to create secret gist",
-    "title": "Export"
+    "title": "Export",
+    "hopp_export_name": "Export as Hoppscotch Collections",
+    "hopp_export_title": "The collections will be exported as json",
```
AndrewBastin, line 219: ~~json~~ -> JSON

packages/hoppscotch-app/locales/en.json
```diff
@@ -214,7 +214,19 @@
...
+    "hopp_export_label": "Export as Hoppscotch Collections",
```
AndrewBastin, line 220: Repetition ?

packages/hoppscotch-app/locales/en.json
```diff
@@ -214,7 +214,19 @@
...
+    "openapi_export_name": "Export as an Openapi Document",
+    "openapi_export_title": "The collections will be exported as an openapi document",
+    "openapi_export_label": "Export as Openapi document",
+    "error_generic": "Something went wrong",
```
AndrewBastin, line 224: Repetition. Generic Error is already defined in `error.something_went_wrong`

packages/hoppscotch-app/locales/en.json
```diff
@@ -214,7 +214,19 @@
...
+    "error_invalid_auth": "Invalid Authentication in one of the requests",
+    "error_invalid_body": "Invalid Body in one of the requests",
+    "error_invalid_url": "Invalid Url in of the requests",
```
AndrewBastin, line 227: ~~Url~~ -> URL

packages/hoppscotch-app/helpers/import-export/export/index.ts
amk-dev: Yup. I agree. I Intented the exporter to be kept as a seperate from the download functionality. just something that converts hopp collection to openapi/whatever format. and have the download/blob thingy on the importExport file. but this complicates the UI part a little.

i've updated the implementation.

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
amk-dev: ohh, my bad

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
amk-dev: just making sure we're dealing with a valid content type

packages/hoppscotch-app/helpers/import-export/export/index.ts
```diff
@@ -1,18 +1,65 @@
...
+  ? Exporter
+  : never
+
+export type RESTCollectionExporterError =
+  | (_RESTCollectionExporter["exporter"] extends () => Promise<
+      HoppExporter<HoppRESTRequest, infer ErrorType>
+    >
+      ? ErrorType
+      : never)
+  | ExportError
```
AndrewBastin, line 65: We can just index the Array by the number type. Simplifiying above bits.

```suggestion
export type RESTCollectionExporterError =
  | (typeof RESTCollectionExporters[number]["exporter"] extends () => Promise<
      HoppExporter<HoppRESTRequest, infer ErrorType>
    >
      ? ErrorType
      : never)
  | ExportError
```

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
AndrewBastin: Use `isValidContentType` instead.

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,480 @@
...
+        ),
+        E.chainW((contentType) =>
+          pipe(
+            hoppRequestBody,
+            O.fromPredicate(isNonFormDataBody),
+            O.chain((hoppRequestBody) =>
+              pipe(
+                hoppRequestBody,
+                O.fromPredicate(() => isContentTypeJSON(contentType)),
+                O.chain(flow(({ body }) => safeParseJSON(body))),
```
AndrewBastin, line 236: You can drop the flow here.

```suggestion
                O.chain(({ body }) => safeParseJSON(body)),
```

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,480 @@
...
+                O.map(generateOpenAPIBodyForJSON(contentType)),
+                O.alt(() =>
+                  pipe(
+                    hoppRequestBody,
+                    O.fromPredicate(isContentTypeUrlEncoded),
+                    O.map(({ body }) => generateOpenAPIBodyForUrlEncoded(body)),
+                    O.alt(() =>
```
AndrewBastin, line 243: Shouldn't this `alt` be on the outer level ?

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,480 @@
...
+              )
+            ),
+            E.fromOption(() => "INVALID_BODY" as const)
+          )
+        )
+      )
+    )
+  )
+
+const OpenAPIAuthNames = {
```
AndrewBastin, line 268: Compile time constant should be UPPER_SNAKE_CASE

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,480 @@
...
+        ),
+        O.map((authType) =>
+          pipe(OpenAPIAuthNames, (auths) => ({
+            [auths[authType]]: [],
+          }))
+        )
+      )
+    ),
+    E.fromOption(() => "INVALID_AUTH" as const)
+  )
```
AndrewBastin, line 296: More cleaner implementation ?

```suggestion
  hoppAuth.authType === "none"
    ? E.right({})
    : pipe(
        hoppAuth.authType,
        O.fromPredicate(
          (authType): authType is "basic" | "api-key" | "bearer" =>
            ["basic", "api-key", "bearer"].includes(authType)
        ),
        O.map((authType) =>
          pipe(OpenAPIAuthNames, (auths) => ({
            [auths[authType]]: [],
          }))
        ),
        E.fromOption(() => "INVALID_AUTH" as const)
      )
```

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,480 @@
...
+      ApiKeyAuth: {
+        type: "apiKey",
+        name: "api-key-header-name",
+        in: "header",
+      },
+    },
+  },
+})
+
+type PathGenerationError =
```
AndrewBastin, line 338: PathGenerationError ?

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,480 @@
...
+      pipe(
+        E.tryCatch(
+          () => new URL(request.endpoint),
+          () => "INVALID_URL" as const
+        )
+      )
+    ),
+
+    E.bindW("openApiBody", () =>
+      pipe(request.body, generateOpenApiRequestBody)
```
AndrewBastin, line 369: You can drop pipes for small pipelines

```suggestion
    generateOpenApiRequestBody(request.body)
```

packages/hoppscotch-app/helpers/import-export/export/openapi.ts
```diff
@@ -0,0 +1,480 @@
...
+
+    E.bindW("openApiHeaders", () =>
+      pipe(request.headers, generateOpenApiHeaders, E.right)
+    ),
+
+    E.bindW("openapiQueryParams", () =>
+      pipe(request.params, generateOpenApiQueryParams, E.right)
+    ),
+
+    E.bindW("openapiAuth", () => pipe(request.auth, generateOpenApiAuth)),
```
AndrewBastin, line 380:
```suggestion
    E.bindW("openapiAuth", () => generateOpenApiAuth(request.auth)),
```

AndrewBastin (changes requested):
look into the given corrections

AndrewBastin (changes requested):
as discussed in the call

vaaski:
what happened to this?

ahmed2m:
Interested in this as well!

ChuTungLinh:
any news regarding to this  ?

PR #2410, opened 2022-06-11
feat: checks only if not default param

moritzmock:
Related to issue 2207

### Description

The issue 2207 added a warning for "unsaved changes might get lost", however, I found it annoying that the message poped up also at the default value...

### Checks

### DIFF
--- a/packages/hoppscotch-app/components/collections/my/Request.vue
+++ b/packages/hoppscotch-app/components/collections/my/Request.vue
@@ -322,6 +322,7 @@ const setRestReq = (request: any) => {
   )
 }
 
+/** Loads request from the save once, checks for unsaved changes, but ignores default values */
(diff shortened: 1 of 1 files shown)


packages/hoppscotch-app/components/collections/my/Request.vue
```diff
@@ -362,6 +362,9 @@ const selectRequest = () => {
       setRESTSaveContext(null)
     }
   }
+  if (getRESTRequest().endpoint === getDefaultRESTRequest().endpoint) {
```
AndrewBastin, line 365: This only checks for the same endpoints, why shouldn't we just check against the entire default request itself ?

AndrewBastin (changes requested):

PR #2425, opened 2022-06-13
feat: display response size with a bigger unit in tooltip

nicognaW:
### Description

The response size is always displayed using larger units like KB or MB in other API testing tools I've used. But hoppscotch shows them up with only a B. I added a tooltip to display the response size in reasonable units.

For responses with a size of 0b to 1000b, the tooltip won't show up. For those with a size of 1000b to 1000000b, kB will be shown. For those larger than 1000000b, mb will be shown. 

I am no expert on the front-end technologies, so I just imitated the current codes. If you don't like these codes, please keep the idea even if you close the pr.

### DIFF
--- a/packages/hoppscotch-app/components/http/ResponseMeta.vue
+++ b/packages/hoppscotch-app/components/http/ResponseMeta.vue
@@ -117,7 +117,11 @@
             <span class="text-secondary"> {{ t("response.time") }}: </span>
             {{ `${response.meta.responseDuration} ms` }}
           </span>
-          <span v-if="response.meta && response.meta.responseSize">
+          <span
+            v-if="response.meta && response.meta.responseSize"
+            v-tippy="responseSizeTippy"
+            :title="responseSizeHint"
+          >
             <span class="text-secondary"> {{ t("response.size") }}: </span>
             {{ `${response.meta.responseSize} B` }}
           </span>
@@ -140,6 +144,36 @@ const t = useI18n()
 const props = defineProps<{
   response: HoppRESTResponse
 }>()
+
+const responseSizeTippy = computed(() => {
+  if (
+    props.response.type === "loading" ||
+    props.response.type === "network_fail" ||
+    props.response.type === "script_fail" ||
+    props.response.type === "fail"
+  )
+    return
+  if (Number(props.response.meta.responseSize) <= 1000)
+    return {
+      onShow: () => {
+        return false
+      },
+    }
+  else return { theme: "tooltip", allowHTML: true }
+})
+
+const responseSizeHint = computed(() => {
+  if (
+    props.response.type === "loading" ||
+    props.response.type === "network_fail" ||
+    props.response.type === "script_fail" ||
+    props.response.type === "fail"
+  )
+    return
+  const size = Number(props.response.meta.responseSize)
+  if (size >= 100000) return (size / 1000000).toFixed(2) + " MB"
+  if (size >= 1000) return (size / 1000).toFixed(2) + " KB"
+})
 
 const statusCategory = computed(() => {
   if (
--- a/packages/hoppscotch-app/components/http/ResponseMeta.vue
+++ b/packages/hoppscotch-app/components/http/ResponseMeta.vue
@@ -120,7 +120,7 @@
           <span
             v-if="response.meta && response.meta.responseSize"
             v-tippy="
-              responseSizeReadable
+              readableResponseSize
                 ? { theme: 'tooltip' }
                 : { onShow: () => false }
             "
@@ -128,8 +128,8 @@
           >
             <span class="text-secondary"> {{ t("response.size") }}: </span>
             {{
-              responseSizeReadable
-                ? responseSizeReadable
+              readableResponseSize
+                ? readableResponseSize
                 : `${response.meta.responseSize} B`
             }}
           </span>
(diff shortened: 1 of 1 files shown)


packages/hoppscotch-app/components/http/ResponseMeta.vue
```diff
@@ -141,6 +145,36 @@ const props = defineProps<{
...
 
+const responseSizeTippy = computed(() => {
+  if (
+    props.response.type === "loading" ||
+    props.response.type === "network_fail" ||
+    props.response.type === "script_fail" ||
+    props.response.type === "fail"
+  )
+    return
+  if (Number(props.response.meta.responseSize) <= 1000)
```
AndrewBastin, line 156: Redundant Number constructor.
We are using TypeScript which guarantees `props.response.meta.responseSize` is a number.

```suggestion
  if (props.response.meta.responseSize <= 1000)
```

packages/hoppscotch-app/components/http/ResponseMeta.vue
```diff
@@ -141,6 +145,36 @@ const props = defineProps<{
...
+
+const responseSizeHint = computed(() => {
+  if (
+    props.response.type === "loading" ||
+    props.response.type === "network_fail" ||
+    props.response.type === "script_fail" ||
+    props.response.type === "fail"
+  )
+    return
+  const size = Number(props.response.meta.responseSize)
```
AndrewBastin, line 173: Redundant Number constructor. We are using TypeScript which guarantees `props.response.meta.responseSize` is a number.

```suggestion
  const size = props.response.meta.responseSize
```

packages/hoppscotch-app/components/http/ResponseMeta.vue
```diff
@@ -141,6 +145,36 @@ const props = defineProps<{
...
 }>()
 
+const responseSizeTippy = computed(() => {
+  if (
+    props.response.type === "loading" ||
+    props.response.type === "network_fail" ||
+    props.response.type === "script_fail" ||
+    props.response.type === "fail"
+  )
+    return
```
AndrewBastin, line 155: Implicit undefined return. Since we return values in other code paths, it is best if you mention the return value here too (implcitly it will be undefined).

```suggestion
    return undefined
```

packages/hoppscotch-app/components/http/ResponseMeta.vue
```diff
@@ -141,6 +145,36 @@ const props = defineProps<{
...
+  if (Number(props.response.meta.responseSize) <= 1000)
+    return {
+      onShow: () => {
+        return false
+      },
+    }
+  else return { theme: "tooltip", allowHTML: true }
+})
```
AndrewBastin, line 163: Existence of `responseSizeTippy` is against the Vue ethos. This should be in the template and can be simplified if based against `responseSizeHint`.

```vue
<span
            v-if="response.meta && response.meta.responseSize"
            v-tippy="
              responseSizeHint
                ? { theme: 'tooltip', allowHTML: false }
                : { onShow: () => false }
            "
            :title="responseSizeHint"
          >
```

packages/hoppscotch-app/components/http/ResponseMeta.vue
```diff
@@ -141,6 +145,36 @@ const props = defineProps<{
...
+    props.response.type === "fail"
+  )
+    return
+  if (Number(props.response.meta.responseSize) <= 1000)
+    return {
+      onShow: () => {
+        return false
+      },
+    }
+  else return { theme: "tooltip", allowHTML: true }
```
AndrewBastin, line 162: `allowHTML` should not be turned on unless we have to render HTML content as it can lead to XSS issues.

packages/hoppscotch-app/components/http/ResponseMeta.vue
```diff
@@ -141,6 +145,36 @@ const props = defineProps<{
...
+    props.response.type === "network_fail" ||
+    props.response.type === "script_fail" ||
+    props.response.type === "fail"
+  )
+    return
+  const size = Number(props.response.meta.responseSize)
+  if (size >= 100000) return (size / 1000000).toFixed(2) + " MB"
+  if (size >= 1000) return (size / 1000).toFixed(2) + " KB"
+})
+
```
AndrewBastin, line 177: Implicitly returns undefined if now of the above 2 ifs match, which I think so is the intention but that should be expressed explicitly.
```suggestion
  if (size >= 100000) return (size / 1000000).toFixed(2) + " MB"
  if (size >= 1000) return (size / 1000).toFixed(2) + " KB"
  
  return undefined
})

```

packages/hoppscotch-app/components/http/ResponseMeta.vue
```diff
@@ -141,6 +145,36 @@ const props = defineProps<{
...
+const responseSizeHint = computed(() => {
+  if (
+    props.response.type === "loading" ||
+    props.response.type === "network_fail" ||
+    props.response.type === "script_fail" ||
+    props.response.type === "fail"
+  )
+    return
```
AndrewBastin, line 172: Implicit undefined return. Since we return values in other code paths, it is best if you mention the return value here too (implcitly it will be undefined).

```suggestion
    return undefined
```

AndrewBastin (changes requested):
@nicognaW please look into the given remarks and correct accordingly please.

Also, won't it be better if we flip the logic then ? (Show the KB/MB size data first and show bytes on hover) (cc: @liyasthomas)

Thank you for your effort by the way! ❤️

nicognaW:
@AndrewBastin Huge thanks for the review and suggestions. I fixed the code mentioned in the remarks and marked them resolved.

And actually, the flip logic idea also came to me the moment after I posted the PR, so I am working on it. Please give me some time🙏.

AndrewBastin (approved):
lgtm 💯

AndrewBastin (approved):
there was a minor typo I just patched. now should be fine. lgtm 💯 

Thank you for the PR @nicognaW

PR #2428, opened 2022-06-15
fix: resolve removing body parameters in requests (#2390)

### LINKED ISSUE #2390: [bug]: Trying to delete parameter hangs the application or doesn't work

I know this is two issues, but those are definitely connected. If it's required then we could divide this into two.

### Current behavior

Issue 1
When I try to delete non last parameter from request body list the application hangs.
I expect parameter to be deleted from the list.

Issue 2
When I delete last one and then try to delete one that is now last I get the message, but parameter stays. After I try to change it then it dissapears.

Possibly important:
- I use PWA on chrome
- Request is saved in a team collection
- Happens only with some parameters (last in the list works ok), at least when it's the first parameter I try to delete and don't try to delete more after deleting that one
- This request is using body with multipart/form-data

### Steps to reproduce

Issue 1
1. Create new request
2. Add body of type multipart/form-data
3. Add two parameters
4. Try to remove first one

Issue 2
1. Create new request
2. Add body of type multipart/form-data
3. Add two parameters
4. Remove last one
5. Try to remove new last one

image

### Environment

Production

### Version

Cloud

vaugenwake:
Closes #2390 

### Description
This PR resolves an issue with removing a parameters for request body parameters causing the browser to crash and hang. The solution to this is to add an `id` parameter to parameters so that each item is unique in the view and gets re-rendered on change.

[x] Fixed removing a body parameter from REST Request

### Checks

### DIFF
--- a/packages/hoppscotch-app/components/http/BodyParameters.vue
+++ b/packages/hoppscotch-app/components/http/BodyParameters.vue
@@ -39,7 +39,7 @@
     >
       <div
         v-for="(param, index) in workingParams"
-        :key="`param-${index}`"
+        :key="`param=${param.id}-${index}`"
         class="flex border-b divide-x divide-dividerLight border-dividerLight draggable-content group"
       >
         <span>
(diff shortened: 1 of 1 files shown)


vaugenwake:
@AndrewBastin Yeah all looks good from my tests

AndrewBastin:
wokey. looks good at my end too. merging.

once again, thank you for the PR @vaugenwake! nice catch!

AndrewBastin (approved):
lgtm 💯

PR #2454, opened 2022-06-23
fix: environment variables save without pressing 'save' button

### LINKED ISSUE #2436: [bug]: environment variables are eager to save without pressing 'save' button

### Current behavior

When I edit environment variables and don't hit 'save' button it will save without any warning.

I expect when I edit environment variables, I MUST hit 'save' button to keep editing changes.

Also why these important and sensitive data is editable by default? They must be editable only when user clicks on some 'edit' button or something like that. Changing these kind of data shouldn't be this much unprotected in my opinion.

### Steps to reproduce

1. Create an environment variable
2. Close and then reopen environment panel
3. Change one of them
4. Don't hit 'save' and just close environment panel
5. Check that variable that was edited, It is changed.

### Environment

Production

### Version

Cloud

nivedin:
Closes #2436 

### Description
This PR fixes a bug that saved the environment variables without pressing the 'save' button.

### Checks

### DIFF
--- a/packages/hoppscotch-app/components/environments/Details.vue
+++ b/packages/hoppscotch-app/components/environments/Details.vue
@@ -139,6 +139,8 @@ import {
   useToast,
 } from "~/helpers/utils/composables"
 
+type EnvironmentVariable = { id: number; env: Environment["variables"] }
+
 const t = useI18n()
 const toast = useToast()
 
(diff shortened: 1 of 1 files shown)


AndrewBastin (approved):
lgtm :100:

PR #2527, opened 2022-07-23
feat: added delay flag in @hoppscotch/cli and related tests

### LINKED ISSUE #2507: [feature]: `--delay` option to `hopp test`

### Summary

Option to set ms delay between requests on CLI:
```bash
hopp test --delay 100 <hoppscotch collection file>
```

For large APIs (and therefore large collections) `hopp test` may fail with rate limit errors. With a delay option would be possible to respect rate limit and test entire collection as expected :smi

devblin:
Closes #2507 

### Description
This feature introduces `-d, --delay` flag to allow providing delay (in milliseconds) between execution of consecutive requests.

### Checks


devblin:
@AndrewBastin

packages/hoppscotch-cli/src/utils/request.ts
```diff
@@ -358,3 +369,21 @@ export const getRequestMetrics = (
...
+const axiosDelay = (config: RequestConfig): AxiosPromise => {
+  return new Promise(async (resolve, reject) => {
+    try {
+      setTimeout(() => {
+        resolve(axios(config));
+      }, config.delay);
+    } catch (error) {
+      reject(error);
+    }
+  });
```
AndrewBastin, line 388: The try-catch is redundant and will not catch anything. as the setTimeout will run and the block execution will be complete before the axios execution occurs. 

Along with that the unused async/await on the async function.

Also, I think its a bad idea to delay axios calls, that feels weird, let the delay apply to one hoppscotch request instead of an axios one.

AndrewBastin (changes requested):

AndrewBastin (approved):
lgtm 💯

Issue #2424, opened 2022-06-13
[bug]: cURl paser body error


renmu123:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

when I import cURL, the bosy parse error.

### Steps to reproduce

Click to 'import cURL'
input: curl
```
curl 'http://api.test' \
  -H 'Accept: application/json, text/plain, */*' \
  -H 'Accept-Language: zh,zh-CN;q=0.9,en;q=0.8,zh-TW;q=0.7,ja;q=0.6' \
  -H 'Content-Type: application/x-www-form-urlencoded' \
  -H 'DNT: 1' \
  -H 'Origin: http://localhost:8000' \
  -H 'Proxy-Connection: keep-alive' \
  -H 'Referer: http://localhost:8000/' \
  -H 'User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/102.0.0.0 Safari/537.36' \
  --data-raw 'name=11&logistics_id=4&port_id=3' \
  --compressed \
  --insecure
```

Body was translate to "name: 11&logistics_id" instead of 
"name: 11 
logistics_id: 4
port_id: 3'"

### Environment

Release

### Version

Local

AndrewBastin:
cc: @kyteinsky

John-Weak:
Can you please assign this to me?

Also I was looking into tests of curl parser and found out that only one test contains  **--data-raw** ,
which is of type --data-raw '{"id":4,"crm_company_id":4,"industry_primary_id":2,"contacts":[]}',

Should I open a pull request to include the problematic type format:
 --data-raw 'name=11&logistics_id=4&port_id=3'

kyteinsky:
@John-Weak I am afraid that's not the issue here. The curl parser replaces all `--data-raw` tags with `-d` tags. But yeah one test sure is missing.

Last I looked into it there were issues with [this function](https://github.com/hoppscotch/hoppscotch/blob/main/packages/hoppscotch-app/helpers/curl/curlparser.ts#L75) and [this block](https://github.com/hoppscotch/hoppscotch/blob/main/packages/hoppscotch-app/helpers/curl/curlparser.ts#L94-L101) getting executed. The problem is I don't remember why that else if block is there in the first place (@AndrewBastin would be smiling rn).

John-Weak:
[...]

PR #2528, opened 2022-07-24
fix: curl parser x-www-form-urlencoded body parsing

kyteinsky:
x-www-form-urlencoded body is parsed correctly now with/without content-type header provided.
Solves #2424


AndrewBastin (approved):
lgtm 💯

PR #2544, opened 2022-07-31
refactor: revamped collection panel UI for mobile

JoelJacobStephen:
### Description

Implemented a new collection panel layout for mobile, instead of using a 3rd pane to display collections.

### Checks

### DIFF
--- a/packages/hoppscotch-app/components/app/PaneLayout.vue
+++ b/packages/hoppscotch-app/components/app/PaneLayout.vue
@@ -67,6 +67,7 @@
           <slot name="primary" />
         </Pane>
         <Pane
+          v-if="mdAndLarger || (!mdAndLarger && !loading && hasResponse)"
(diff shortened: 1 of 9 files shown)


anwarulislam (commented):
The change seems okay to me. But, one thing looks odd.
image
We actually don't need to show this part when we're in mobile view since no utility of this. We need to show a response after getting the response from API server. If no response is there then we can hide this part on mobile. For this, we can't add parameters to body/headers on mobile currently. It's overlapping the UI

anwarulislam (changes requested):
Handling `HttpResponse` isn't the responsibility of `PaneLayout` component. We need to somehow handle it from the parent component where we're using `AppPaneLayout`.

liyasthomas:
@JoelJacobStephen, rebase and then mark `read for review`

PR #2556, opened 2022-08-03
feat: remember pane sizes

### LINKED ISSUE #2456: [feature]: Remember pane sizes

### Summary

Nowadays, when you launch Hoppscotch, you need to resize the different panes to your liking every time.
It would be a nice feature to save the different pane sizes (for instance, in local storage) in order to restore the visual configuration the next time the app is opened:

hoppscotch

Every time the app is opened from scratch, the panes need to be resized manually. This small **QoL** improvement solves this issue.

sagar290:
Closes  #2456 

### Description
Save the different pane sizes (for instance, in local storage) in order to restore the visual configuration the next time the app is opened

### Checks

### DIFF
--- a/packages/hoppscotch-app/components/app/PaneLayout.vue
+++ b/packages/hoppscotch-app/components/app/PaneLayout.vue
@@ -6,7 +6,7 @@
       '!flex-row-reverse': SIDEBAR_ON_LEFT && mdAndLarger,
     }"
     :horizontal="!mdAndLarger"
-    @resize="setPaneEvent($event)"
+    @resize="setPaneEvent($event, 'vertical')"
   >
     <Pane
       :size="PANE_MAIN_SIZE"
@@ -62,6 +62,8 @@ const SIDEBAR = useSetting("SIDEBAR")
 
 const slots = useSlots()
 
+const hasSidebar = computed(() => !!slots.sidebar)
+
 const props = defineProps({
   layoutId: {
     type: String,
@@ -85,38 +87,36 @@ if (!COLUMN_LAYOUT.value) {
   PANE_MAIN_BOTTOM_SIZE.value = 50
 }
 
-function setPaneEvent(
-  event: PaneEvent[],
-  _type: "vertical" | "horizontal" = "vertical"
-) {
+function setPaneEvent(event: PaneEvent[], _type: "vertical" | "horizontal") {
   if (!props.layoutId) return
-
-  setLocalConfig(
-    `${props.layoutId}-pane-config-${_type}`,
-    JSON.stringify(event)
-  )
+  const STORAGE_KEY = `${props.layoutId}-pane-config-${_type}`
+  setLocalConfig(STORAGE_KEY, JSON.stringify(event))
 }
 
 function populatePaneEvent() {
   if (!props.layoutId) return
 
-  const storageKey = `${props.layoutId}-pane-config`
-  const verticalPaneEvent = getLocalConfig(`${storageKey}-vertical`)
-  if (verticalPaneEvent) {
-    const paneData: PaneEvent[] = JSON.parse(verticalPaneEvent)
-    PANE_MAIN_SIZE.value = paneData[0].size
-    PANE_SIDEBAR_SIZE.value = paneData[1].size
+  const verticalPaneData = getPaneData("vertical")
+  if (verticalPaneData) {
+    const [mainPane, sidebarPane] = verticalPaneData
+    PANE_MAIN_SIZE.value = mainPane?.size
+    PANE_SIDEBAR_SIZE.value = sidebarPane?.size
   }
 
-  const horizontalPaneEvent = getLocalConfig(`${storageKey}-horizontal`)
-  if (horizontalPaneEvent) {
-    const paneData: PaneEvent[] = JSON.parse(horizontalPaneEvent)
-    PANE_MAIN_TOP_SIZE.value = paneData[0].size
-    PANE_MAIN_BOTTOM_SIZE.value = paneData[1].size
+  const horizontalPaneData = getPaneData("horizontal")
+  if (horizontalPaneData) {
+    const [mainTopPane, mainBottomPane] = horizontalPaneData
+    PANE_MAIN_TOP_SIZE.value = mainTopPane?.size
+    PANE_MAIN_BOTTOM_SIZE.value = mainBottomPane?.size
   }
 }
 
-populatePaneEvent()
+function getPaneData(_type: "vertical" | "horizontal"): PaneEvent[] | null {
+  const STORAGE_KEY = `${props.layoutId}-pane-config-${_type}`
+  const paneEvent = getLocalConfig(STORAGE_KEY)
+  if (!paneEvent) return null
+  return JSON.parse(paneEvent)
+}
 
-const hasSidebar = computed(() => !!slots.sidebar)
+populatePaneEvent()
 </script>
--- a/packages/hoppscotch-app/components/app/PaneLayout.vue
+++ b/packages/hoppscotch-app/components/app/PaneLayout.vue
@@ -6,27 +6,27 @@
       '!flex-row-reverse': SIDEBAR_ON_LEFT && mdAndLarger,
     }"
     :horizontal="!mdAndLarger"
-    @resize="setVerticalPanelInfo"
+    @resize="setVerticalPaneEvent"
   >
     <Pane
-      :size="PANE_MAIN_SIZE.value"
+      :size="PANE_MAIN_SIZE"
       min-size="65"
       class="hide-scrollbar !overflow-auto flex flex-col"
     >
       <Splitpanes
         class="smart-splitter"
         :horizontal="COLUMN_LAYOUT"
-        @resize="setHorizontalPanelInfo"
+        @resize="setHorizontalPaneEvent"
       >
         <Pane
-          :size="PANE_MAIN_TOP_SIZE.value"
+          :size="PANE_MAIN_TOP_SIZE"
           min-size="50"
           class="hide-scrollbar !overflow-auto flex flex-col"
         >
           <slot name="primary" />
         </Pane>
         <Pane
-          :size="PANE_MAIN_BOTTOM_SIZE.value"
+          :size="PANE_MAIN_BOTTOM_SIZE"
           min-size="20"
           class="flex flex-col hide-scrollbar !overflow-auto"
         >
@@ -36,7 +36,7 @@
     </Pane>
     <Pane
       v-if="SIDEBAR && hasSidebar"
-      :size="PANE_SIDEBAR_SIZE.value"
+      :size="PANE_SIDEBAR_SIZE"
       min-size="20"
       class="hide-scrollbar !overflow-auto flex flex-col"
     >
@@ -49,8 +49,7 @@
 import { Splitpanes, Pane } from "splitpanes"
 import "splitpanes/dist/splitpanes.css"
 import { breakpointsTailwind, useBreakpoints } from "@vueuse/core"
-import { computed, useSlots } from "@nuxtjs/composition-api"
-import { ref } from "@vue/runtime-dom"
+import { computed, useSlots, ref } from "@nuxtjs/composition-api"
 import { useSetting } from "~/newstore/settings"
 import { setLocalConfig, getLocalConfig } from "~/newstore/localpersistence"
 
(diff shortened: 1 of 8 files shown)


packages/hoppscotch-app/components/app/PaneLayout.vue
```diff
@@ -56,5 +61,73 @@ const SIDEBAR = useSetting("SIDEBAR")
...
+    size: 65,
+  },
+]
+
+let panelHorizontalInfo: any = [
+  {
+    max: 100,
+    min: 0,
+    size: 75,
+  },
```
anwarulislam, line 82: Instead of hard-coding these values in template why are you using an array?

packages/hoppscotch-app/components/app/PaneLayout.vue
```diff
@@ -56,5 +61,73 @@ const SIDEBAR = useSetting("SIDEBAR")
...
+  {
+    max: 100,
+    min: 0,
+    size: 65,
+  },
+]
+
+function setVerticalPanelInfo(event: any) {
```
anwarulislam, line 90: There should be an explicit type for this event. If there's no explicit type then we should put `unknown` as type.

packages/hoppscotch-app/components/app/PaneLayout.vue
```diff
@@ -56,5 +61,73 @@ const SIDEBAR = useSetting("SIDEBAR")
...
+  // eslint-disable-next-line no-restricted-globals
```
anwarulislam, line 91: We shouldn't disable any rules define by ESLint unless we have valid reason.

packages/hoppscotch-app/components/app/PaneLayout.vue
```diff
@@ -56,5 +61,73 @@ const SIDEBAR = useSetting("SIDEBAR")
...
+  localStorage.setItem("vertical_panel_info", JSON.stringify(event))
```
anwarulislam, line 92: Don't use `locaStorage`. Follow the ESLint error where we mentioned to use `localpersistance.ts` function.

packages/hoppscotch-app/components/app/PaneLayout.vue
```diff
@@ -56,5 +61,73 @@ const SIDEBAR = useSetting("SIDEBAR")
...
+}
+
+function getVerticalPanelInfo() {
+  // eslint-disable-next-line no-restricted-globals
```
anwarulislam, line 96: We shouldn't disable any rules define by ESLint unless we have valid reason.

packages/hoppscotch-app/components/app/PaneLayout.vue
```diff
@@ -56,5 +61,73 @@ const SIDEBAR = useSetting("SIDEBAR")
...
+  let panelDataFromStorage = localStorage.getItem("vertical_panel_info")
```
anwarulislam, line 97: Don't use `locaStorage` directly. Follow the ESLint error where we mentioned to use `localpersistance.ts` function.

packages/hoppscotch-app/components/app/PaneLayout.vue
```diff
@@ -56,5 +61,73 @@ const SIDEBAR = useSetting("SIDEBAR")
...
+    setVerticalPanelInfo(panelVerticalInfo)
+    getVerticalPanelInfo()
+  }
+  panelVerticalInfo = panelDataFromStorage
+}
+
+function setHorizontalPanelInfo(event: any) {
+  // eslint-disable-next-line no-restricted-globals
+  localStorage.setItem("horizontal_panel_info", JSON.stringify(event))
+}
```
anwarulislam, line 111: The same repeatative errors. Follow above comments.

packages/hoppscotch-app/components/app/PaneLayout.vue
```diff
@@ -56,5 +61,73 @@ const SIDEBAR = useSetting("SIDEBAR")
...
+
+function getHorizontallPanelInfo() {
+  // eslint-disable-next-line no-restricted-globals
+  let panelDataFromStorage = localStorage.getItem("horizontal_panel_info")
```
anwarulislam, line 115: The same thing here as well. Follow above commments.

anwarulislam (changes requested):
We're using this `AppPaneLayout` for many pages. And we should remember pane sizes for every page separately. Your implementation is just remembering globally. So, whenever we change pane size it affects all the pages wherever we use the `PaneLayout`. 

Instead of doing it globally we can get the `layout-id` as prop from the parent component. And use that prop as a prefix to define keys. Also, don't use `localStorage` directly to the component. We already have the implementation to store values in `localpersistence.ts` file. You can use that function to store.

packages/hoppscotch-app/components/app/PaneLayout.vue
```diff
@@ -6,21 +6,28 @@
...
     >
-      <Splitpanes class="smart-splitter" :horizontal="COLUMN_LAYOUT">
+      <Splitpanes
+        class="smart-splitter"
+        :horizontal="COLUMN_LAYOUT"
+        @resize="setHorizontalPanelInfo"
+      >
         <Pane
-          :size="COLUMN_LAYOUT ? 45 : 50"
+          :size="PANE_MAIN_TOP_SIZE.value"
```
anwarulislam, line 22: Remove `.value`

packages/hoppscotch-app/components/app/PaneLayout.vue
```diff
@@ -6,21 +6,28 @@
...
+          min-size="50"
           class="hide-scrollbar !overflow-auto flex flex-col"
         >
           <slot name="primary" />
         </Pane>
         <Pane
-          :size="COLUMN_LAYOUT ? 65 : 50"
+          :size="PANE_MAIN_BOTTOM_SIZE.value"
```
anwarulislam, line 29: Remove `.value`. Also, we need to consider the `COLUMN_LAYOUT` condition here. Either in the script or in template. I think it should be in script.

packages/hoppscotch-app/components/app/PaneLayout.vue
```diff
@@ -43,7 +50,9 @@ import { Splitpanes, Pane } from "splitpanes"
 import "splitpanes/dist/splitpanes.css"
 import { breakpointsTailwind, useBreakpoints } from "@vueuse/core"
 import { computed, useSlots } from "@nuxtjs/composition-api"
+import { ref } from "@vue/runtime-dom"
```
anwarulislam, line 53: We need to import `ref` from `@nuxtjs/composition-api` instead of `@vue/runtime-dom`

packages/hoppscotch-app/components/app/PaneLayout.vue
```diff
@@ -56,5 +65,58 @@ const SIDEBAR = useSetting("SIDEBAR")
...
+type PanelInfo = {
+  max: number
+  min: number
+  size: number
+}
+
+const PANE_SIDEBAR_SIZE = ref()
+const PANE_MAIN_SIZE = ref()
+const PANE_MAIN_TOP_SIZE = ref()
+const PANE_MAIN_BOTTOM_SIZE = ref()
```
anwarulislam, line 84: Provide default value for Pane Sizes. eg. 75 for main pane.

anwarulislam (changes requested):
We still have the same issue. the default value is getting applied to every page if there's no `layout-id`. Instead of doing that way just do not store/restore values if there's no `layout-id`. And go with the default values.

anwarulislam:
@AndrewBastin, it seems okay to me.

packages/hoppscotch-app/components/app/PaneLayout.vue
```diff
@@ -6,21 +6,28 @@
...
-      <Splitpanes class="smart-splitter" :horizontal="COLUMN_LAYOUT">
+      <Splitpanes
+        class="smart-splitter"
+        :horizontal="COLUMN_LAYOUT"
+        @resize="setHorizontalPaneEvent"
+      >
         <Pane
-          :size="COLUMN_LAYOUT ? 45 : 50"
+          :size="PANE_MAIN_TOP_SIZE"
+          min-size="50"
```
anwarulislam, line 23: Remove min-size from here.

packages/hoppscotch-app/components/app/PaneLayout.vue
```diff
@@ -6,21 +6,28 @@
...
           class="hide-scrollbar !overflow-auto flex flex-col"
         >
           <slot name="primary" />
         </Pane>
         <Pane
-          :size="COLUMN_LAYOUT ? 65 : 50"
+          :size="PANE_MAIN_BOTTOM_SIZE"
+          min-size="20"
```
anwarulislam, line 30: Remove min-size from here as well

packages/hoppscotch-app/components/app/PaneLayout.vue
```diff
@@ -56,5 +64,62 @@ const SIDEBAR = useSetting("SIDEBAR")
...
+if (!COLUMN_LAYOUT) {
+  PANE_MAIN_TOP_SIZE.value = 50
+  PANE_MAIN_BOTTOM_SIZE.value = 50
+}
+
+function setVerticalPaneEvent(event: PaneEvent[]) {
+  if (!props.layoutId) return
+
+  setLocalConfig(`${PANE_STORAGE_KEY}_vertical`, JSON.stringify(event))
+}
```
anwarulislam, line 95: `setVerticalPaneEvent` and `setHorizontalPaneEvent` are doing the same thing. just storing value to storage. Instead of creating two function we can go with one function. You can name that `storePaneEvent`. It will have two parameters `event` and `type`. For horizontal type will be `hr` and for vertical it will be `vr`. Then pass the argument from template itself.

packages/hoppscotch-app/components/app/PaneLayout.vue
```diff
@@ -56,5 +64,62 @@ const SIDEBAR = useSetting("SIDEBAR")
...
+  size: number
+}
+
+const PANE_SIDEBAR_SIZE = ref(25)
+const PANE_MAIN_SIZE = ref(75)
+const PANE_MAIN_TOP_SIZE = ref(45)
+const PANE_MAIN_BOTTOM_SIZE = ref(65)
+const PANE_STORAGE_KEY = `${props.layoutId}-app-pane-layout`
+
+if (!COLUMN_LAYOUT) {
```
anwarulislam, line 86: `COLUMN_LAYOUT` is a reactive value. You have to access the value from `.value` property here. Change this to `COLUMN_LAYOUT.value`

packages/hoppscotch-app/components/app/PaneLayout.vue
```diff
@@ -56,5 +64,62 @@ const SIDEBAR = useSetting("SIDEBAR")
...
+  max: number
+  min: number
+  size: number
+}
+
+const PANE_SIDEBAR_SIZE = ref(25)
+const PANE_MAIN_SIZE = ref(75)
+const PANE_MAIN_TOP_SIZE = ref(45)
+const PANE_MAIN_BOTTOM_SIZE = ref(65)
+const PANE_STORAGE_KEY = `${props.layoutId}-app-pane-layout`
```
anwarulislam, line 84: Remove this line. We can access prop directly from methods. `layoutId` could be undefined. In that case no meaning of concatination.

packages/hoppscotch-app/components/app/PaneLayout.vue
```diff
@@ -56,5 +64,62 @@ const SIDEBAR = useSetting("SIDEBAR")
...
+
+if (!COLUMN_LAYOUT) {
+  PANE_MAIN_TOP_SIZE.value = 50
+  PANE_MAIN_BOTTOM_SIZE.value = 50
+}
+
+function setVerticalPaneEvent(event: PaneEvent[]) {
+  if (!props.layoutId) return
+
+  setLocalConfig(`${PANE_STORAGE_KEY}_vertical`, JSON.stringify(event))
```
anwarulislam, line 94: Don't mix the naming with kebab-case and snake_case. Use either one. IMO the better naming would be like this `${props.layoutId}-pane-config-vr`

anwarulislam (changes requested):
Few more changes are required

packages/hoppscotch-app/components/app/PaneLayout.vue
anwarulislam: Here vr and hr could be dynamic if you go with the same method for both.

packages/hoppscotch-app/components/app/PaneLayout.vue
sagar290: In that case, can set a condition if `layoutId` undefined no storage key will set?

packages/hoppscotch-app/components/app/PaneLayout.vue
sagar290: sure

anwarulislam (dismissed):
Seems okay to me. @AndrewBastin @liyasthomas

AndrewBastin (changes requested):
Just minor variable naming scheme corrections.

Also, this is only implemented for documentation.vue as of right now, other pages should also implement this thing.

(note to reviewer: @anwarulislam make sure you check variable naming schemes on your reviews as well)

anwarulislam:
> (note to reviewer: @anwarulislam make sure you check variable naming schemes on your reviews as well)

Noted!

AndrewBastin (approved):
lgtm 💯 

@liyasthomas a final round of checks and you can merge it in.

liyasthomas (approved):
LGTM 💯

PR #2575, opened 2022-08-10
feat: added support for path variables

### LINKED ISSUE #2086: [feature]: Support for Path Variables

### Current behavior

When I import the petstore collection and I try to use a method with a path variable inside, the app does not provide an easy way to provide them

Capture d’écran 2022-01-25 133509

I know I can simply override the url in the request, but this is not the "goal" 

What I'd like is that if I setup a parameter with the same name as in the path, it is used instead of being used as an additional query param

### Steps to reproduce

Import petstore openapi
Load the uploadFile request
Add a baseUrl env var (this will turn the baseUrl variable blue in the path)

=> when I provide a parameter with the name petId, it should be used in the path (and marked as blue)

### Environment

Production

### Version

Local

MasonJason23:
Closes #2086 

### Description

Created a secondary UI component under Query Parameters as "My Variables" which has the same functionality as Environment Variables. Allows users to make variables to use within the request input bar using curly brackets `{{}}`.

### TODO
- [ ] Add a warning message for recursive error loops

### Checks

### Additional Information

- Added a new custom type (`HoppRESTVar`) and modified the data structure of the existing data structure (`HoppRESTRequest`).
### "My Variable" Feature in Action
My Variables in Action


packages/hoppscotch-app/helpers/utils/EffectiveURL.ts
```diff
@@ -29,6 +29,7 @@ export interface EffectiveHoppRESTRequest extends HoppRESTRequest {
   effectiveFinalHeaders: { key: string; value: string }[]
   effectiveFinalParams: { key: string; value: string }[]
   effectiveFinalBody: FormData | string | null
+  effectiveFinalVars: { key: string; value: string }[]
```
AndrewBastin, line 32: If this is equivalent to `request.vars`, why does this exist ?

packages/hoppscotch-data/src/environment.ts
```diff
@@ -59,36 +65,71 @@ export const parseBodyEnvVariables = (
...
     )
-    depth++
+    depthEnv++
   }
 
-  return depth > ENV_MAX_EXPAND_LIMIT
-    ? E.left(ENV_EXPAND_LOOP)
-    : E.right(result)
+  if (myVariables) {
+    while (result.match(REGEX_MY_VAR) != null && depthVar <= ENV_MAX_EXPAND_LIMIT) {
```
AndrewBastin, line 89: We have to consider cases where path variables expand to environment variables, we have to resolve then those as well.

Dockerfile
```diff
@@ -22,7 +22,7 @@ RUN pnpm i --unsafe-perm=true
 ENV HOST 0.0.0.0
 EXPOSE 3000
 
-RUN mv packages/hoppscotch-app/.env.example packages/hoppscotch-app/.env
+RUN mv packages/hoppscotch-app/.env packages/hoppscotch-app/.env
```
AndrewBastin, line 25: Why is the rename ?

AndrewBastin (changes requested):
Please look into the suggestions.

Dockerfile
MasonJason23: Not sure. This was probably something during the beginning when the team started working on this and we were still figuring stuff out. I just undid the changes now.

.github/workflows/deploy-netlify.yml
```diff
@@ -19,7 +19,7 @@ jobs:
           run_install: true
 
       - name: Setup Environment
-        run: mv packages/hoppscotch-app/.env.example packages/hoppscotch-app/.env
+        run: mv packages/hoppscotch-app/.env packages/hoppscotch-app/.env
```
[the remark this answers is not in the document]
MasonJason23: Not sure. This was probably something during the beginning when the team started working on this and we were still figuring stuff out. I just undid the changes now.

packages/hoppscotch-app/components/http/Parameters.vue
```diff
@@ -1,363 +1,13 @@
...
-          filled
-          class="mb-4"
-          @click.native="addParam"
-        />
-      </div>
-    </div>
+  <div>
+    <HttpQueryParams />
+    <br />
+    <HttpVariables />
```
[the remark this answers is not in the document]
MasonJason23: 👍

AndrewBastin:
In between the review limbo of Vue 3 + Vite and the major amount of changes because of it, I think this PR was neglected and got a bit stale. I think it will be a bit of work to rework this into a fully functioning PR against the revamped codebase.

Hence, I am merging this in, but to an internal branch so that someone from the Hoppscotch Team can eventually rework this PR to be compliant with the latest codebase.

Thank you for your effort guys! <3

We hope we can land this in production eventually so your work isn't in vain, even if not, most probably your code will be used as a reference to implement the final implementation.

PR #2606, opened 2022-08-22
fix: form data with same key send only last one

### LINKED ISSUE #2571: [bug]: Multiple form data with the same name only send the last one

### Current behavior

When using `application/x-www-from-urlencoded` to send form data, if you fill in duplicate key names, only the last key will be sent.

|key|value|
|---|---|
|id|1|
|id|2|

Should be encoded as `id=1&id=2`,
but is actually encoded as `id=2`.

### Steps to reproduce

1. Enter test url
2. Choose Body
3. Change Content-Type to `application/x-www-from-urlencoded`
4. Fill in two values ​​with the same key
5. Send,and you can see the wrong Body on the server

### Environment

Release

### Version

Cloud

nivedin:
Closes #2571 

### Description
This PR fixes the issue when using `application/x-www-from-urlencoded` as content-type to send form data, when you fill in duplicate key names, only the last key was sent.

### Checks

### DIFF
--- a/packages/hoppscotch-app/helpers/functional/record.ts
+++ b/packages/hoppscotch-app/helpers/functional/record.ts
@@ -13,7 +13,7 @@ export const tupleToRecord = <
   tuples: [KeyType, ValueType][]
 ): Record<KeyType, ValueType> =>
   tuples.length > 0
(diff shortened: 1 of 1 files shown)


AndrewBastin (approved):
lgtm 💯

PR #2664, opened 2022-09-14
fix: allow posting empty key files in multipart/form-data

### LINKED ISSUE #2293: [bug]: Cannot POST a file to a REST endpoint using multipart/form-data

### Current behavior

I operate a REST endpoint that takes a file as input, and would like to test it using a afile.  If I curl this command, I will get my file back from httpbin.org in the response body:

           curl -vks -X POST -F data=@petstore_swagger.json  https://httpbin.org/post

  This issue:

	https://github.com/hoppscotch/hoppscotch/issues/1679

  suggests that this is supported in Hoppscotch, however I get a different result. 

  The HTTP response shows an empty file:
      
 "files": { }

  And says my content-type was: `"Content-Type": "multipart/form-data; boundary=----WebKitFormBoundaryKQngLTS8PbjO5yW7",`
  
My expectation was that I would see the same "files" attribute that I see in the response body from my curl command, i.e. that my file body would be echoed back to me.  (see steps to reproduce below)

### Steps to reproduce

  I create the equivalent of this request in the form at https://hoppscotch.io/ on a new REST request:

               1) Select POST
               2) Set the URL to https://httpbin.org/post
               3) Set content-type to multipart/form-data
               4) Click "Choose Files" and specify my file
               5) Click "send"

### Environment

Production

### Version

Cloud

fumblehool:
### Description
`form-data` param have following properties - 
```
{
    active: boolean,
    key: string,
    value: string,
    isFile: boolean,
    value: array
}
```

There was a check to remove blank parameter. This check is based on `key`.
However, in case of file upload - `key` is empty but `isFile` is set to True.

This PR updates those checks.

### DIFF
--- a/packages/hoppscotch-app/src/components/http/BodyParameters.vue
+++ b/packages/hoppscotch-app/src/components/http/BodyParameters.vue
@@ -250,7 +250,7 @@ watch(
       workingParams.value,
       A.filterMap(
         flow(
-          O.fromPredicate((e) => e.entry.key !== ""),
+          O.fromPredicate((e) => e.entry.key !== "" || e.entry.isFile),
           O.map((e) => e.entry)
         )
       )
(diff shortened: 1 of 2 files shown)


fumblehool:
Working API call after the update.
Screenshot 2022-09-14 at 11 17 53 PM

AndrewBastin (approved):
lgtm 💯

PR #2679, opened 2022-09-19
fix: parse env with dash from request body

### LINKED ISSUE #2665: [bug]: Env variable with dash in its name doesn't work in JSON body

### Current behavior

When I use an env variable with a **dash in its name** in a JSON body, the variable does not become substituted with the actual value. Even though the variable is correctly highlighted green. 

However when I use a env variable without dashes in its name in a JSON body, the variable becomes substituted with the actual value as expected.

------

https://github.com/hoppscotch/hoppscotch/issues/1834
https://github.com/hoppscotch/hoppscotch/pull/1942

### Steps to reproduce

1. Create an env variable with dashes in its name
2. Create a request with a json body which uses this env variable
3. Send request – variable doesn't become substituted with the actual value

### Environment

Production

### Version

Cloud

anwarulislam:
Closes #2665  

### Description

When I use an env variable with a dash in its name in a JSON body, the variable does not become substituted with the actual value. Even though the variable is correctly highlighted green. With this PR this issue get resolved.

### Checks

### DIFF
--- a/packages/hoppscotch-app/src/helpers/utils/EffectiveURL.ts
+++ b/packages/hoppscotch-app/src/helpers/utils/EffectiveURL.ts
@@ -12,7 +12,6 @@ import {
   HoppRESTReqBody,
   HoppRESTRequest,
   parseTemplateString,
-  parseBodyEnvVariables,
   Environment,
   HoppRESTHeader,
   HoppRESTParam,
@@ -280,7 +279,9 @@ function getFinalBodyFromRequest(
       ),
       toFormData
     )
-  } else return parseBodyEnvVariables(request.body.body, envVariables)
+  }
+
+  return parseTemplateString(request.body.body, envVariables)
 }
 
 /**


packages/hoppscotch-app/src/helpers/utils/EffectiveURL.ts
```diff
@@ -280,7 +279,9 @@ function getFinalBodyFromRequest(
       ),
       toFormData
     )
-  } else return parseBodyEnvVariables(request.body.body, envVariables)
+  }
+
+  return parseTemplateString(request.body.body, envVariables)
```
AndrewBastin, line 284: We have `parseBodyEnvVariables` which handle a special case we have with body  where if we don't get a match, we will leave the environment variables as is and don't parse them down like we do in other places.

Hence, what is to be fixed is not here, but on the `parseBodyEnvVariables` function.

AndrewBastin (changes requested):

AndrewBastin:
Rejecting this PR as this is a trivial fix and the fix proposed in this PR is not optimal.

PR #2706, opened 2022-09-29
chore: exclude robots.txt and sitemap.xml from using fallback navigation

amk-dev:
Closes # 

### Description

robot.txt and sitemap.xml won't follow service worker fallbacks.


liyasthomas (approved):
LGTM 💯

PR #2790, opened 2022-10-16
chore: update env copy instructions

TheHollidayInn:
Closes # 

### Description

When following the current .env file instructions, this will lead you to have a removed .env.example which can accidentally be committed. So this is a minor change in the instructions.

### Checks

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,7 +279,9 @@ _Add-ons are developed and maintained under **[Hoppscotch Organization](https://
 
 ## **Developing**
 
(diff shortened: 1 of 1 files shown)


liyasthomas:
`.env` file should be present on `packages/hoppscotch-app` directory. This instruction is clearly stated in README.
 
I believe the addition in this PR is unnecessary, hence closing.

TheHollidayInn:
The Readme instructions have you alter a file this updated has you copy so no git changes are added

AndrewBastin:
@liyasthomas, I think @TheHollidayInn is in the right here, the current phrasing of the step makes it feel like we have to update `.env.example` eventhough it is just supposed to be a reference file.

PR #2791, opened 2022-10-16
feat: use environment variable to specify proxyscotch access token

### LINKED ISSUE #2203: [bug]: How to specify the token for self-hosted proxyscotch?

### Current behavior

There's no way to specify a token for proxyscotch.
So I get response:
```
(Proxy Error) Unauthorized request; you may need to set your access token in Settings.
```

### Steps to reproduce

1. Setup a proxyscotch instance by `proxyscotch-server-* --host 0.0.0.0:9159 --token 123456`
2. Connect hoppscotch to the instance by specifying the PROXY URL as 'http://localhost:9159'
3. Call any API from hoppscotch.

### Environment

Production

### Version

Self-hosted

biondizzle:
Closes #2203

### Description
Currently no way of specifying the access token to a self-hosted version of Proxyscotch. This adds an optional environment variable to do so.

This will pass `accessToken` to the Proxy where it will get filled into this struct:
https://github.com/hoppscotch/proxyscotch/blob/2e94def1c28106f556718feb33232a8d1a8471b5/libproxy/proxy.go#L30

Which will get set when the request body is decoded in these two spots:
https://github.com/hoppscotch/proxyscotch/blob/2e94def1c28106f556718feb33232a8d1a8471b5/libproxy/proxy.go#L197
https://github.com/hoppscotch/proxyscotch/blob/2e94def1c28106f556718feb33232a8d1a8471b5/libproxy/proxy.go#L206

And then be validated here:
https://github.com/hoppscotch/proxyscotch/blob/2e94def1c28106f556718feb33232a8d1a8471b5/libproxy/proxy.go#L216

### Checks

### Additional Information
I don't know typescript so not sure if I needed to add `accessToken` to https://github.com/hoppscotch/hoppscotch/blob/6e7d28db7b8f9ce3fe5735bcb59e2c32195111c0/packages/hoppscotch-app/src/helpers/strategies/AxiosStrategy.ts#L16

Also, not sure if the environment variable needs to be added to any of the workflows YAMLs but decided not to add it since clearly the current prod version of this has been working fine without the need for this environment variable. Assuming that's because https://proxy.hoppscotch.io/ is secured via incoming source restriction. 

Just let me know what you want changed and I'll do it real quick, or feel free to modify it to your heart's content.


AndrewBastin:
@NBTX Can you review this as well as this is proxyscotch related ?

packages/hoppscotch-app/src/helpers/strategies/AxiosStrategy.ts
```diff
@@ -29,6 +29,7 @@ const getProxyPayload = (
   let payload: ProxyPayloadType = {
     ...req,
     wantsBinary: true,
+    accessToken: (!import.meta.env.VITE_PROXYSCOTCH_ACCESS_TOKEN) ? "" : import.meta.env.VITE_PROXYSCOTCH_ACCESS_TOKEN,
```
AndrewBastin, line 32: You can use Nullish Coalescing.

```suggestion
    accessToken: import.meta.env.VITE_PROXYSCOTCH_ACCESS_TOKEN ?? ""
```

AndrewBastin (changes requested):

packages/hoppscotch-app/src/helpers/strategies/AxiosStrategy.ts
biondizzle: I thought JS couldn't do this. Must be new. That's awesome! Updated!

AndrewBastin:
@biondizzle would love your opinion on whether the Access Token also should be a setting in the settings page as well. Like, to allow for hoppscotch.io users to make use of it as well for self hosted proxies ?

biondizzle:
@AndrewBastin when you go to [hoppscotch.io/settings](https://hoppscotch.io/settings), whether you're logged in or not, you're not able to change the Proxy URL. The cursor is set to cancel, however; you're able to change it on the selfhosted version.

This makes sense. Keeping that purely server-side is a great security measure and if users were able to change the URL but not the access token, then it's easy enough to steal the access token because they can point the URL to their own server and the server-side code will send along the access token too. 

Allowing users to change the proxy URL on [hoppscotch.io/settings](https://hoppscotch.io/settings) and allowing them to enter in an access token will require some rework of the UI to more easily convey the difference between "Default https://proxy.hoppscotch.io/" and "Custom with Access Token". There would need be a dropdown or some kind of action for the user to use a custom proxy with a custom access token, so that way on the server-side, it's easy to control the parameters of what's being sent out where:

Default = Use https://proxy.hoppscotch.io/ with XXX token REGARDLESS of what the user inputs. All of that is ignored when DEFAULT is selected

Custom = Use ONLY the user input values.

This would be a cool addition to the product because companies using the product could have a custom proxy with all the networking handled so it's within their company network to more easily give access of examples to their APIs they don't provide publicly (i.e. Dev/QA/UAT versions of an API could correspond to a Dev/QA/UAT proxy allowed to access those environments).

AndrewBastin:
Oh okay, true. I guess that makes sense, so to keep this PR lean and focus on just providing for the immediate need, we will merge this in and I will keep tabs on implementing a Settings based override later on to expand upon the need.

AndrewBastin (approved):

biondizzle:
@AndrewBastin  BTW. Great Job on this product. It kicks the pants off Postman and has made life tremendously easier to get people examples of how some internal APIs work (especially those with poor documentation) and setup some QA tasks without having to send people cURL commands in bashscripts or go walking them through setting up Postman.

I have plans to add in an optional login system that doesn't require Firebase so it's more easily self hosted in any environment (if that's a feature you'd be willing to merge, if not; no sweat, it will just help me personally). Just need to find the time 😂

If you ever want to discuss anything, I'm available on Discord: DonnyDonato#8403 or Telegram: https://t.me/passsnap

Thanks for taking the time to review my MR and get this feature in!

AndrewBastin:
@biondizzle aww, thanks dude! its an effort by the whole team and the community which lead to where we are.

You don't need to work on the login system, we are working on it and that is one of our major projects we are working on right now. Since you said you are using Hoppscotch within your company, we are doing a survey of sorts to make the experience better for you guys as we are now aiming to prep Hoppscotch to support enterprise adoption and use cases and we would love you and your company's feedback, please do reach out to me using the Hoppscotch Discord Server! <3

@liyasthomas follow up!

PR #2810, opened 2022-10-21
fix: useCodemirror getting non strings as value

amk-dev:
**Description**

fix HOPPSCOTCH-APP-43


AndrewBastin (approved):
lgtm 💯 

@amk-dev I am reiterating the importance of documenting the PR opening comment, please specify what you are changing, why and the before and after behaviour changes as this is a public record

PR #2813, opened 2022-10-23
Feat: Add inline environment variable edit

### LINKED ISSUE #2607: [feature]: ability to inline edit environment variable

### Summary

When we hover over the variable we see the value. It would be great if we can show an edit button to edit the value inline or at least prompt the user to edit that specific variable in the current modal implementation.

It will be a cool feature that will increase developers' productivity. They can easily change the value of the variable

franciscoemanuel:
Closes #2607 

### Description

- Adds edit action on environment variable tooltip:
image

- When the edit icon is clicked, the environment edit modal is opened with the variable value auto selected
image

### Checks

### Additional Information
- Implements two new actions handler to invoke the environment edit modals;
- Changes the approach on `environments/index.vue` to conditionally render env components to use v-show instead of v-if, so the modals can be called even when the correspondent tab is not active;
- Adds option to pass parameters to `invokeAction` method
- Adds `selectText` prop to EnvInput so it can auto select the text on the input when it's rendered

### DIFF
--- a/packages/hoppscotch-app/src/components/environments/index.vue
+++ b/packages/hoppscotch-app/src/components/environments/index.vue
@@ -155,9 +155,9 @@
         @update-selected-team="updateSelectedTeam"
       />
     </div>
-    <EnvironmentsMy v-if="environmentType.type === 'my-environments'" />
+    <EnvironmentsMy v-show="environmentType.type === 'my-environments'" />
     <EnvironmentsTeams
-      v-else
+      v-show="environmentType.type === 'team-environments'"
       :team="environmentType.selectedTeam"
       :team-environments="teamEnvironmentList"
       :loading="loading"
--- a/packages/hoppscotch-app/src/components/environments/index.vue
+++ b/packages/hoppscotch-app/src/components/environments/index.vue
@@ -155,9 +155,9 @@
         @update-selected-team="updateSelectedTeam"
       />
     </div>
-    <EnvironmentsMy v-show="environmentType.type === 'my-environments'" />
+    <EnvironmentsMy v-if="environmentType.type === 'my-environments'" />
     <EnvironmentsTeams
-      v-show="environmentType.type === 'team-environments'"
+      v-if="environmentType.type === 'team-environments'"
       :team="environmentType.selectedTeam"
       :team-environments="teamEnvironmentList"
       :loading="loading"
--- a/packages/hoppscotch-app/src/components/environments/index.vue
+++ b/packages/hoppscotch-app/src/components/environments/index.vue
@@ -155,9 +155,9 @@
         @update-selected-team="updateSelectedTeam"
       />
     </div>
-    <EnvironmentsMy v-if="environmentType.type === 'my-environments'" />
+    <EnvironmentsMy v-show="environmentType.type === 'my-environments'" />
     <EnvironmentsTeams
-      v-if="environmentType.type === 'team-environments'"
+      v-show="environmentType.type === 'team-environments'"
       :team="environmentType.selectedTeam"
       :team-environments="teamEnvironmentList"
       :loading="loading"
--- a/packages/hoppscotch-app/src/components/environments/my/index.vue
+++ b/packages/hoppscotch-app/src/components/environments/my/index.vue
@@ -135,6 +135,6 @@ defineActionHandler("modals.my.environment.edit", (args?: any[]) => {
       return environment.name === envName
     }
   )
-  editEnvironment(envIndex >= 0 ? envIndex : envName)
+  editEnvironment(envIndex >= 0 ? envIndex : "Global")
 })
 </script>
--- a/packages/hoppscotch-app/src/components/environments/my/index.vue
+++ b/packages/hoppscotch-app/src/components/environments/my/index.vue
@@ -128,8 +128,7 @@ const resetSelectedData = () => {
 }
 
 defineActionHandler("modals.my.environment.edit", (args?: any[]) => {
-  const envName: any = args?.[0]
-  const variableName: string = args?.[1]
+  const [envName, variableName]: any = args
   editingVariableName.value = variableName
   const envIndex: number = environments.value.findIndex(
     (environment: Environment) => {
--- a/packages/hoppscotch-app/src/components/environments/my/index.vue
+++ b/packages/hoppscotch-app/src/components/environments/my/index.vue
@@ -64,6 +64,7 @@
       :show="showModalDetails"
       :action="action"
       :editing-environment-index="editingEnvironmentIndex"
+      :editing-variable-name="editingVariableName"
       @hide-modal="displayModalEdit(false)"
     />
     <EnvironmentsImportExport
@@ -83,6 +84,8 @@ import { useI18n } from "~/composables/i18n"
 import IconArchive from "~icons/lucide/archive"
 import IconPlus from "~icons/lucide/plus"
 import IconHelpCircle from "~icons/lucide/help-circle"
+import { Environment } from "@hoppscotch/data"
+import { defineActionHandler } from "~/helpers/actions"
 
 const t = useI18n()
 const colorMode = useColorMode()
@@ -100,6 +103,7 @@ const showModalImportExport = ref(false)
 const showModalDetails = ref(false)
 const action = ref<"new" | "edit">("edit")
 const editingEnvironmentIndex = ref<number | "Global" | null>(null)
+const editingVariableName = ref<string>("")
 
 const displayModalAdd = (shouldDisplay: boolean) => {
   action.value = "new"
@@ -122,4 +126,16 @@ const editEnvironment = (environmentIndex: number | "Global") => {
 const resetSelectedData = () => {
   editingEnvironmentIndex.value = null
 }
+
+defineActionHandler("modals.my.environment.edit", (args?: any[]) => {
+  const envName: any = args?.[0]
+  const variableName: string = args?.[1]
+  editingVariableName.value = variableName
+  const envIndex: number = environments.value.findIndex(
+    (environment: Environment) => {
+      return environment.name === envName
+    }
+  )
+  editEnvironment(envIndex >= 0 ? envIndex : envName)
+})
 </script>
--- a/packages/hoppscotch-app/src/components/environments/my/index.vue
+++ b/packages/hoppscotch-app/src/components/environments/my/index.vue
@@ -103,7 +103,7 @@ const showModalImportExport = ref(false)
 const showModalDetails = ref(false)
 const action = ref<"new" | "edit">("edit")
 const editingEnvironmentIndex = ref<number | "Global" | null>(null)
-const editingVariableName = ref<string>("")
+const editingVariableName = ref("")
 
 const displayModalAdd = (shouldDisplay: boolean) => {
   action.value = "new"
--- a/packages/hoppscotch-app/src/components/environments/my/index.vue
+++ b/packages/hoppscotch-app/src/components/environments/my/index.vue
@@ -127,14 +127,16 @@ const resetSelectedData = () => {
   editingEnvironmentIndex.value = null
 }
 
-defineActionHandler("modals.my.environment.edit", (args?: any[]) => {
-  const [envName, variableName]: any = args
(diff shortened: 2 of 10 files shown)


franciscoemanuel:
@anwarulislam

franciscoemanuel:
@anwarulislam Do you have any suggestions about the code?

anwarulislam:
> @anwarulislam Do you have any suggestions about the code?

I have tested the feature. And it's really good so far. The only thing I noticed is the props name. Since you're passing a `boolean` value in porps `select-text` the name should be `is-text-selected` or something meaningful. Other than that, the functionality is okay. And it's working fine.

anwarulislam (commented):
I have tested the feature. And it's really good so far. The only thing I noticed is the props name. Since you're passing a `boolean` value in porps `select-text` the name should be `is-text-selected` or something meaningful. Other than that, the functionality is okay. And it's working fine.

franciscoemanuel:
@anwarulislam Agreed, I've renamed the prop. Do you have any other comment?

anwarulislam:
@liyasthomas I have tested this feature. And it's good to go.

packages/hoppscotch-app/src/helpers/actions.ts
```diff
@@ -55,8 +57,8 @@ export function bindAction(action: HoppAction, handler: () => void) {
   activeActions$.next(Object.keys(boundActions) as HoppAction[])
 }
 
-export function invokeAction(action: HoppAction) {
-  boundActions[action]?.forEach((handler) => handler())
+export function invokeAction(action: HoppAction, args?: any[]) {
```
AndrewBastin, line 60: The action system is not supposed to be having an `args` parameter, even if allowed which can be arranged, I am not a fan of it being there without type safety.

packages/hoppscotch-app/src/helpers/editor/extensions/HoppEnvironment.ts
```diff
@@ -58,27 +59,43 @@ const cursorTooltipField = (aggregateEnvs: AggregateEnvironment[]) =>
...
+        editIcon.addEventListener("click", () => {
+          const isPersonalEnv =
+            envName === "Global" || selectedEnvType !== "TEAM_ENV"
+          const action = isPersonalEnv ? "my" : "team"
+          invokeAction(`modals.${action}.environment.edit`, [
+            envName,
+            parsedEnvKey,
+          ])
+        })
+        editIcon.innerHTML = `<i class="inline-flex items-center px-1 ml-2 -my-1 text-base border-x material-icons border-secondary">edit</i>`
```
AndrewBastin, line 95: @liyasthomas we still support material icons usage like this ? We downloading the font still ?

packages/hoppscotch-app/src/helpers/editor/extensions/HoppEnvironment.ts
```diff
@@ -58,27 +59,43 @@ const cursorTooltipField = (aggregateEnvs: AggregateEnvironment[]) =>
...
+      const parsedEnvKey = text.slice(start - from, end - from)
 
-      const envValue =
-        aggregateEnvs.find(
-          (env) => env.key === text.slice(start - from, end - from)
-          // env.key === word.slice(wordSelection.from + 2, wordSelection.to - 2)
-        )?.value ?? "Not found"
+      const tooltipEnv = aggregateEnvs.find(
+        (env) => env.key === parsedEnvKey
+        // env.key === word.slice(wordSelection.from + 2, wordSelection.to - 2)
```
AndrewBastin, line 66: Redundant comment

packages/hoppscotch-app/src/components/smart/EnvInput.vue
```diff
@@ -42,6 +42,7 @@ const props = withDefaults(
     styles?: string
     envs?: { key: string; value: string; source: string }[] | null
     focus?: boolean
+    isTextSelected?: boolean
```
AndrewBastin, line 45: I am not a fan of using props with this naming scheme to define state-y operations.

Props are philosophically meant to be representative of the rendering state by name association but this doesn't align well with that concept.

Either, rename the props to express these actions are done on mount (`focusOnMount`, `selectTextOnMount` etc.) or move the move and focus options to component methods and expose it so the parent can call it via a template ref.

I know `focus` is from existing code but it was not being used even apparently so it slipped my radar I guess (@anwarulislam)

packages/hoppscotch-app/src/components/environments/teams/index.vue
```diff
@@ -143,6 +145,7 @@ const showModalImportExport = ref(false)
 const showModalDetails = ref(false)
 const action = ref<"new" | "edit">("edit")
 const editingEnvironment = ref<TeamEnvironment | null>(null)
+const editingVariableName = ref<string>("")
```
AndrewBastin, line 148: You can drop implicit generic declaration if the type is known

```suggestion
const editingVariableName = ref("")
```

packages/hoppscotch-app/src/components/environments/teams/index.vue
```diff
@@ -178,4 +181,14 @@ const getErrorMessage = (err: GQLError<string>) => {
     }
   }
 }
+
+defineActionHandler("modals.team.environment.edit", (args?: any[]) => {
+  const [envName, variableName]: any = args
+  editingVariableName.value = variableName
+  const teamEnvToEdit: TeamEnvironment | undefined =
```
AndrewBastin, line 188: Since, the assigned expression already gives info about what is the type of `teamEnvToEdit`. You can drop the explicit type declaration

```suggestion
  const teamEnvToEdit =
```

packages/hoppscotch-app/src/components/environments/teams/index.vue
```diff
@@ -178,4 +181,14 @@ const getErrorMessage = (err: GQLError<string>) => {
...
+    props.teamEnvironments.find((environment: TeamEnvironment) => {
```
AndrewBastin, line 189: Same as above, the type of `environment` is automatically inferrable, hence you can drop the type definition.

```suggestion
    props.teamEnvironments.find((environment) => {
```

packages/hoppscotch-app/src/components/environments/my/index.vue
```diff
@@ -100,6 +103,7 @@ const showModalImportExport = ref(false)
 const showModalDetails = ref(false)
 const action = ref<"new" | "edit">("edit")
 const editingEnvironmentIndex = ref<number | "Global" | null>(null)
+const editingVariableName = ref<string>("")
```
AndrewBastin, line 106:
```suggestion
const editingVariableName = ref("")
```

packages/hoppscotch-app/src/components/environments/my/index.vue
```diff
@@ -122,4 +126,15 @@ const editEnvironment = (environmentIndex: number | "Global") => {
...
+
+defineActionHandler("modals.my.environment.edit", (args?: any[]) => {
+  const [envName, variableName]: any = args
+  editingVariableName.value = variableName
+  const envIndex: number = environments.value.findIndex(
+    (environment: Environment) => {
+      return environment.name === envName
+    }
+  )
+  editEnvironment(envIndex >= 0 ? envIndex : envName)
```
AndrewBastin, line 138: Since you have defined `envName` as `any`. You have to provide guarantee that `envName` is either a number or `"Global"` to satisfy the type contract.
Please make sure to add the assertion that `envName` is `"Global"` before calling `editEnvironment` to make sure you keep its contract and not bring possibility of UB.

packages/hoppscotch-app/src/components/environments/index.vue
```diff
@@ -155,9 +155,9 @@
         @update-selected-team="updateSelectedTeam"
       />
     </div>
-    <EnvironmentsMy v-if="environmentType.type === 'my-environments'" />
+    <EnvironmentsMy v-show="environmentType.type === 'my-environments'" />
     <EnvironmentsTeams
-      v-else
+      v-show="environmentType.type === 'team-environments'"
```
AndrewBastin, line 160: Why are you changing this from `v-if` to `v-show` ?

AndrewBastin (changes requested):
I have some gripes with this implementation. Most notably is the issue with the actions getting option to get args, which I am more or less okay with, but I am against it being just typed with `any`.

I can come up with a revamped type for the action argument which is typesafe when I get time though.

Other than that, I am not a big fan of the popup UI, but I will let @liyasthomas to be a judge of that as that is his area.

@anwarulislam please note this errors to improve your reviewing insights for future reviews. Improving type safety is a key thing for us to keep this spaghetti of a codebase rolling 🙏

packages/hoppscotch-app/src/components/environments/index.vue
franciscoemanuel: @AndrewBastin so the modals can be called even when the correspondent tab is not active;

packages/hoppscotch-app/src/helpers/actions.ts
franciscoemanuel: @AndrewBastin could you provide further details about the action arguments solution you mentioned in your first comment? I think that usingarguments in action handlers allow them to be more flexible and prevent scenarios of having to store data in a more global scope.

packages/hoppscotch-app/src/components/environments/index.vue
AndrewBastin: Is there any case where not using a `v-show` causes an issue ? Because I still don't exactly understand why it is needed, not rendering the modal components if not required is better than v-show for this case, unless we are using some mounted state stuff.

It would be awesome if you could provide a steps to repro for a condition which breaks when you use v-show.

packages/hoppscotch-app/src/helpers/actions.ts
AndrewBastin: Yeah, well, the thing is we don't want to use `any` in our codebase as much as possible as it removes all the contract guarantees TypeScript can enforce and this leads to our code being more prone to errors. Hence, we try to type check all our code as much as possible to make sure our code is free of errors. For example, in the review only I pointed out how accidentally calling with the `envName` not being "Global" or something is totally possible.

No worries, I have some time today to come up with a type that can express the action handler types safely.

AndrewBastin:
@franciscoemanuel can you enable "Allow edits from maintainers" so I can push the type safe action argument system.

You can follow this guide: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests

franciscoemanuel:
@AndrewBastin It's enabled
Captura de tela de 2022-11-01 19-21-36

packages/hoppscotch-app/src/components/environments/index.vue
franciscoemanuel: @AndrewBastin To reproduce you can select the tab for "team environment" and if you try to edit some global variable the modal is not shown, because `EnvironmentsMy` where the modal is originally from is not in the dom with `v-if`

AndrewBastin: Yeah, okay, makes sense. I am not a fan of Global being hidden in Teams though as they are being shared, but that is not in scope for this PR, hence, I will allow it.

packages/hoppscotch-app/src/helpers/editor/extensions/HoppEnvironment.ts
AndrewBastin: We should remove material icon font usage. But that is a separate PR, just note this.

packages/hoppscotch-app/src/helpers/actions.ts
AndrewBastin: I hae pushed the update which implements this.

AndrewBastin (approved):
Everything looks good to me so far, but I am not a big fan of the popup UI, can we have some touchup over there @liyasthomas ?

Else, this can be merged.

liyasthomas (approved):
LGTM 💯

PR #2827, opened 2022-10-26
fix: codemirror field overflow

### LINKED ISSUE #2823: [bug]: Overflowed fields doesn't wrap and pushes the UI elements away

### Current behavior

When I do type more than what the fields can cover, the fields overflow, however, the overflowed fields keeps displaying previous words thus pushing away some of the elements of the browser away.

What I expect is to not make these fields not push away the UI and should the fields respect their currently set height and width even when overflowed.

### Steps to reproduce

1. Go to the REST section.
2. On the query, select any of the following: **parameters**, **body**, **headers**, or **authorization**.
image
3. Attempt to overflow the fields by typing than what it can over (_visually_).
4. Do the same for the other candidates.

## Parameters
msedge_QcuafLt5yV

## Authorization
explorer_ewwfUtBiXK

## Header
Notepad_7e0ShuXFna

### Notes
- Pre-request Script and Tests don't have this layout bug.
- As seen in the header, the header field can retain the width of the field, but in terms of its value, however, does not.

### Environment

Production

### Version

Cloud

nivedin:
Closes #2823 

### Description
This PR fixes the issue where some codemirror fields doesn't wrap and pushes the UI elements away when overflowing.

### Checks

### DIFF
--- a/packages/hoppscotch-app/src/components/smart/EnvInput.vue
+++ b/packages/hoppscotch-app/src/components/smart/EnvInput.vue
@@ -1,8 +1,8 @@
 <template>
   <div
-    class="relative py-4 flex items-center flex-1 flex-shrink-0 overflow-auto whitespace-nowrap"
+    class="relative flex items-center flex-1 flex-shrink-0 py-4 overflow-auto whitespace-nowrap"
   >
-    <div class="absolute flex flex-1">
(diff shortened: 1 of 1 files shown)


liyasthomas (approved):
LGTM 💯

PR #2836, opened 2022-10-28
chore: add polyfill for replaceAll

amk-dev:
**Before**
we were using String.prototype.replaceAll at some places, but some people where using old browsers without it.

**After**
this PR adds the polyfill for replaceAll


AndrewBastin (approved):
lgtm 💯 

@amk-dev reminder to mark the `replaceAll` related issues on Sentry to be marked as resolved with the merge release

PR #2847, opened 2022-10-31
fix: errors on opening json outline

amk-dev:
**Before**
Calling `tippyActions.focus()` was failing because tippyActions was a ref inside a v-for, and hence an array of tippyActions refs.

**After**
Use indexes to access refs inside a v-for


AndrewBastin (approved):
lgtm 💯

PR #2857, opened 2022-11-18
feat: parse query parameters

levz0r:
### Description

This PR adds a small, yet convenient UX improvement that is similar to Postman's behavior: when pasting and/or modifying the request URL, the query parameters will be parsed and modified accordingly in the query parameters section.

### Checks

### DIFF
--- a/packages/hoppscotch-app/src/components/http/Request.vue
+++ b/packages/hoppscotch-app/src/components/http/Request.vue
@@ -244,7 +244,7 @@ import IconFolderPlus from "~icons/lucide/folder-plus"
 import { computed, ref, watch } from "vue"
 import { isLeft, isRight } from "fp-ts/lib/Either"
 import * as E from "fp-ts/Either"
-import { cloneDeep } from "lodash-es"
+import { cloneDeep, debounce } from "lodash-es"
 import { refAutoReset } from "@vueuse/core"
 import {
   updateRESTResponse,
@@ -258,6 +258,7 @@ import {
   getRESTRequest,
   restRequest$,
   setRESTSaveContext,
+  setRESTParams,
 } from "~/newstore/RESTSession"
 import { editRESTRequest } from "~/newstore/collections"
 import { runRESTRequest$ } from "~/helpers/RequestRunner"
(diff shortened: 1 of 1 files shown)


AndrewBastin:
We had thought about this sort of an implementation before but we decided against as we thought of the URL entry as something more static and fixed instead of being a dynamic sort of a thing like how we envision the parameters section as.

For example, if I enter in `?name=andrew`, the fields do get populated but it removes existing params added.

Hoppscotch doesn't consider the Parameter section as a reflection of the URL bar unlike what Postman does, we consider them as separate things, one meant to be more static, other meant to be more dynamic.

If we are to consider moving ahead with this implementation, I would prefer just having a Computed Params sorta subsection  (similar to Computed Headers (try using authentication or POST body and look at the headers section)) where the URL entries are also shown.

Not in terms of this PR, but in general: we are not looking forward to building an open-source Postman here, most of our users switch from Postman hating its direction, it doesn't make sense we just add their stuff without scrutiny.

levz0r:
@AndrewBastin thank you for the detailed response. 
I share Your opinion about Postman, but I think this particular feature of Postman is useful.
What if this feature would be added under experiments and its usage would be measured before migrating into core functionality?

AndrewBastin:
@levz0r what about a middle ground to this as I mentioned in my comment.

We see the URL part of the request a bit differently in the sense that we expect it to not change a lot and be minimal. You can introduce parameters there, but usually the norm is to put things which are static there. You can think it more like an 'endpoint' field.

So, what I propose is a more middle approach like what we have for headers:
image

We internally call it Computed Headers, where these headers are applied due to some indirect side effect (in this example, it is because I set Body type to Form Data and added Basic Auth). We can introduce something similar for Parameters and maybe even allow for additional features around like moving parameters from URL to Params and back.

Because, the implementation you are proposing is basically against how Hoppscotch used to work from the beginning, so introducing this change will either cause inconsistencies (based on this experiement being enabled or not) and many other edge cases your implementation in this PR isn't currently handling.

AndrewBastin:
Closing PR due to inactivity.

Issue #2859, opened 2022-11-21
[bug]: Deleted form parameters are still sent.


mfusigh:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

Submitting a request after deleting form data variables still submits the deleted variables.

### Steps to reproduce

1. Create and send POST request with several multipart/form-data variables.
2. Delete a variable using the trash icon next to it.
3. Send the request again and the deleted variable is still sent.

I did a sample request to Google sending three variables in the body, then deleted the "c" variable by using the trash icon, then sent the request. After the first and second request I click on "Show code" and got this back for both:

```
first request:
curl --request POST \
  --url https://google.com/ \
  --header 'Authorization: Bearer ' \
  --header 'Content-Type: multipart/form-data' \
  --form a=1 \
  --form b=2 \
  --form c=3


second request after deleting "c":
curl --request POST \
  --url https://google.com/ \
  --header 'Authorization: Bearer ' \
  --header 'Content-Type: multipart/form-data' \
  --form a=1 \
  --form b=2 \
  --form c=3
```

The second request should not have had the "c=3" line since it was not represented anywhere in the UI anymore.

### Environment

Production

### Version

Cloud

AndrewBastin:
Fixed in #2861

PR #2861, opened 2022-11-24
fix: delete not working properly on request body

amk-dev:
fixs #2859

**Before**

See the mentioned issue for more info.

we were using ```Array.splice``` to remove an element from an array. but this was not triggering a watcher on the array.

**After**

we now use ```Array.filter``` to remove an element from an array and the reactivity works


AndrewBastin (approved):
lgtm 💯

PR #2863, opened 2022-11-29
feat: ability to delete user account and data

anwarulislam:
### Description

This feature addition now allows users to delete their user data from firebase and their association with Teams. If the user is an OWNER of the team we don't allow operation to proceed.

### Checks


liyasthomas (dismissed):
LGTM 💯

packages/hoppscotch-app/src/pages/profile.vue
```diff
@@ -285,23 +217,101 @@
...
 import {
   currentUser$,
   probableUser$,
   setDisplayName,
   setEmailAddress,
+  signOutUser,
   verifyEmailAddress,
 } from "~/helpers/fb/auth"
 
+import { pipe } from "fp-ts/lib/function"
```
AndrewBastin, line 312:
```suggestion
import { pipe } from "fp-ts/function"
```

packages/hoppscotch-app/src/pages/profile.vue
```diff
@@ -344,6 +356,56 @@ const loadingCurrentUser = computed(() => {
...
+      () => {
+        deletingUser.value = false
+        showDeleteAccountModal.value = false
+        toast.success(t("settings.account_deleted"))
+        signOutUser()
+        router.push(`/`)
+      }
+    )
+  )()
+}
```
AndrewBastin, line 394: The deletion work should absolutely move to a separate component.

Also, why are we fetching and calculating Teams list when the user is in this page itself ? This should be deferred to until the user actually opens the modal.

Hence, the finding owned teams and such should only happen when the modal is being summoned to prevent unnecessary work.

packages/hoppscotch-app/src/pages/profile.vue
```diff
@@ -396,61 +458,14 @@ const sendEmailVerification = () => {
...
-  () => adapterLoading.value && myShortcodes.value.length === 0
-)
-
 onLoggedIn(() => {
   adapter.initialize()
 })
 
 onAuthEvent((ev) => {
-  if (ev.event === "logout" && adapter.isInitialized()) {
+  if (ev.event === "logout") {
```
AndrewBastin, line 466: Why remove the initialized check ?

AndrewBastin (changes requested):
I am kinda disappointed this implementation broke two principles we have already established in our team discussions and frontend guidelines.

1. Keep components as small as possible and split off responsibility to separate components (Single Responsibility per component).
2. Do as less work as possible, if we want to speed up even, only do work which we expect to be common on that scenario.

Also, red flag to the frontend reviewers who let this slide. :/

packages/hoppscotch-app/src/pages/profile.vue
anwarulislam: The first UI approach was showing the team list on load. Not in the modal after clicking on Delete Account. That's why I had to calculate the team list on load. Still, it's my bad not to check after the UX changes by the reviewer.

I must separate the section for the user deletion feature into a component as I did with the `Shortcodes`.

AndrewBastin (changes requested):
I have a bit of issues still here.

a) The modal loading state shows the input field eventhough we are still loading the data.
b) The modal still doesn't defer the work of fetching teams for this operation until the modal opens. Instead, once logged in, the adapter is initialized and starts fetching the data.

AndrewBastin (changes requested):
Just one small correction.

@liyasthomas this can be merged following this correction without my review.

liyasthomas (approved):
LGTM

PR #2871, opened 2022-12-06
fix: circular watcher dependencies on invite.vue causing infinite loop

amk-dev:
**Before**

There is an infinite loop caused by two watchEffects modifying dependencies in a circular way.

1. WatchEffect#1 modifies a dependency of WatchEffect#2
2. WatchEffect#2 calls a function `expect` which modifies dependencies of WatchEffect#1
3. WatchEffect#1 runs again, and this becomes a loop

**After**

We removes this circular dependency by

1. Converting one `watchEffect` into a `watch` so we have explicit control over the reactivity, and remove tracking reactivity for `source` and `isPaused` in that watch
2. Remove `updateTicker` ref which is replaced by a function

### DIFF
--- a/packages/hoppscotch-common/src/composables/graphql.ts
+++ b/packages/hoppscotch-common/src/composables/graphql.ts
@@ -9,6 +9,7 @@ import {
   watchEffect,
   WatchStopHandle,
   watchSyncEffect,
+  watch,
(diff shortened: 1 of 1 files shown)


AndrewBastin (approved):
lgtm 💯

PR #2873, opened 2022-12-09
fix: cannot write to body when a request is loaded from history

amk-dev:
**Before**

After Loading a request from user's hoppscotch history, changes to the request body where not taking effect. this is happening, because we are calling `setRequest` with a computed property, which is readonly.

**After**

We pass the computed property as a normal object to `setRequest` which fixes the issue.

### DIFF
--- a/packages/hoppscotch-common/src/components/history/index.vue
+++ b/packages/hoppscotch-common/src/components/history/index.vue
@@ -99,7 +99,7 @@
           :show-more="showMore"
           @toggle-star="toggleStar(entry.entry)"
           @delete-entry="deleteHistory(entry.entry)"
-          @use-entry="useHistory(entry.entry)"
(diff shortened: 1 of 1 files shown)


packages/hoppscotch-common/src/components/history/index.vue
```diff
@@ -99,7 +99,7 @@
           :show-more="showMore"
           @toggle-star="toggleStar(entry.entry)"
           @delete-entry="deleteHistory(entry.entry)"
-          @use-entry="useHistory(entry.entry)"
+          @use-entry="useHistory(toRaw(entry.entry))"
```
liyasthomas, line 102: Forgot to import `toRaw()` from vue.

liyasthomas (approved):
LGTM

PR #2915, opened 2023-02-02
feat: active tab no longer resets after request

### LINKED ISSUE #2080: [feature]: Active tab (JSON, Raw, Headers, Test Results) should not reset after sending request

### Summary

When checking the test results I want to keep this tab active, also after sending the request.
Currently the active tab changes to JSON after sending the request.

The current active tab should stay active ;-)

Productivity increase ;-)

aitchnyu:
Closes #2080

### Description

When the `ResponseBodyRenderer` component renders for each request, the active tab (json, raw etc) is always set to the first one. Users prefer if same tab is persisted. So I moved tab preferences from `ResponseBodyRenderer` to parent so it loads with previously opened tab.

### Checks

### DIFF
--- a/packages/hoppscotch-common/src/components/http/Response.vue
+++ b/packages/hoppscotch-common/src/components/http/Response.vue
@@ -12,38 +12,26 @@
   </div>
 </template>
 
-<script lang="ts">
-import { computed, defineComponent, watch, ref } from "vue"
+<script setup lang="ts">
+import { computed, watch, ref } from "vue"
 import { startPageProgress, completePageProgress } from "@modules/loadingbar"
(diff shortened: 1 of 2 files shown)


anwarulislam (changes requested):
@aitchnyu since you have modified the component already. Can you convert both components to `setup` script? `<script lang="ts" setup>`. And change functionality accordingly.

anwarulislam:
@aitchnyu instead of sharing `selectedTab` value back and forth why not just store it in a `localPersistence`?

AndrewBastin:
Closing this PR as the changes from this are present in #2917.

PR #2921, opened 2023-02-13
fix: response time for requests via extension has incorrect value

### LINKED ISSUE #2897: [bug]: Time counter doesn't stop (ever) after a request was cancelled
This might be a simple issue to solve:

I use the web Hoppscotch app.
If my API has a bug and doesn't respond to a request, Hopp allows me to click the big "cancel" button to stop the listening.

After that, every other request I make happens to -- as fast as the response time is -- never clear the time counter again:

image

The solution is probably just checking the counter function and the cancel button function to see what's missing.

Thank you all for this amazing app, by the way. :)

aitchnyu:
Closes #2897

### Description

The reported bug was in the extension. If one request had failed to return a response, second request shows a huge response time and threw an error. We guessed some object whose reference was shared between requests. Calling the `cancelRequest()` on the extension fixed the response time problem and the error.

Fixed the typing interface for `window.__POSTWOMAN_EXTENSION_HOOK__` with correct [method name](https://github.com/hoppscotch/hoppscotch-extension/blob/main/src/index.ts#L265) . Exported the `window` typing, seems it wasn't previously doing anything.

### Checks

### DIFF
--- a/packages/hoppscotch-common/src/components/http/Request.vue
+++ b/packages/hoppscotch-common/src/components/http/Request.vue
@@ -274,6 +274,10 @@ import { createShortcode } from "~/helpers/backend/mutations/Shortcode"
 import { runMutation } from "~/helpers/backend/GQLClient"
 import { UpdateRequestDocument } from "~/helpers/backend/graphql"
 import { getPlatformSpecialKey as getSpecialKey } from "~/helpers/platformutils"
+import {
+  cancelRunningExtensionRequest,
+  hasExtensionInstalled,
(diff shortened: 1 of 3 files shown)


aitchnyu:
Wondering if we should fix this too. There are cases of this similar code in shims.d.ts, env.d.ts and vite-env.d.ts. I want to know the intended effect and how to verify that.

```
declare module '*.vue' {
  import type { DefineComponent } from 'vue'
  const component: DefineComponent<{}, {}, any>
  export default component
}
```

AndrewBastin (approved):
lgtm :100:

PR #2940, opened 2023-03-02
chore: move collections sync system to platform

amk-dev:
This PR extracts out the collections firebase syncing to hoppscotch-web, and defines the common interface to be used with other hoppscotch platforms.


AndrewBastin (approved):
lgtm 💯

PR #2944, opened 2023-03-07
refactor: optional variables to createEnvironment and fixing the order of initializing GqlClient

amk-dev:
**Before**

1. the `createEnvironment` store action did not support variables as parameters, we were using two actions `createEnvironment` and `updateEnvironment` just for creation.

2. there existed an edge case where the GQLClient was not initialized when `runGqlSubscription` was called, which caused the subscription to not run 

**After**

1. `createEnvironment` store action accepts optional variables parameter.

2. Updated the order of initializing GQLClient to prevent `runGqlSubscription` from not running subscription


AndrewBastin (approved):
lgtm 💯

PR #2947, opened 2023-03-13
refactor: store changes for collections

amk-dev:
**Before**

1. You needed the entire collection to call `editCollection`. This was a limitation when handling `UserCollectionEdited` subscription, because we don't have access to the entire collection. Also the only thing we updated was the collection name.

2. `navigateToFolderWithIndexPath` was not exported, but were required in the sync layer

**After**

1. `editCollection` store dispatcher takes a partial collection as an argument, instead of an entire collection
2. `navigateToFolderWithIndexPath` is exported


AndrewBastin (approved):
lgtm 💯

PR #2952, opened 2023-03-27
refactor: make editFolder, editCollection take Partial collection as parameter

amk-dev:
before:

editCollection, editFolder dispatchers needed the entire collection to make an edit. this was not ideal when we want to update small part of the collection.

after:

now editFolder and editCollection accepts a partial collection and folder object and merges it with the collection. this is also backward compatible, so no code changes to existing system are needed.


AndrewBastin (approved):
lgtm 💯

PR #2955, opened 2023-03-30
chore: add removeDuplicateEntry dispatcher to history store

amk-dev:
This PR introduces a new dispatcher to history store, that deletes a duplicate history entry given an id, which is used to delete duplicate insertions from subscription or store sync functions.


AndrewBastin (approved):
lgtm 💯

PR #2963, opened 2023-04-04
chore: load terms of service & privacy policy links from env variables

amk-dev:
**Before**

Terms Of Service and Privacy Policy links where hardcoded

**After**

Terms Of Service and Privacy Policy links are now loaded from env variables. they're optional. and if they're not provided, the TOS and Privacy Policy text are not rendered.


AndrewBastin (approved):
lgtm 💯

PR #2966, opened 2023-04-05
fix: tabhead and scrolling issue

anwarulislam:
### Description
This PR fixes tabhead issues:
- `closeVisibility` is a prop. It can be `"hover" | "always" | "never"`
- `suffix` is a slot. where we have added the indicator for the unsaved tab/request
- And vertical space was added to tabhead by increasing line-height


AndrewBastin (approved):
lgtm 💯

PR #2974, opened 2023-04-08
refactor: add optional ids to environments

amk-dev:
**Before**

The env store didn't have ids for environments, this created challenges when syncing these changes to the backend.

**After**

This PR adds ids to the environments store, so hoppscotch platform can use ids for syncing if needed. also this change is kept backward compatible.


AndrewBastin (approved):
lgtm :100:

PR #2975, opened 2023-04-09
Update README.md to add required step for local development

Nabeelshar:
add required step to run hoppscotch on local device

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,9 +292,10 @@ _Sample keys only work with the [production build](https://hoppscotch.io)._
 
 1. [Clone this repo](https://help.github.com/en/articles/cloning-a-repository) with git.
 2. Install pnpm using npm by running `npm install -g pnpm`.
-3. Install dependencies by running `pnpm install` within the directory that you cloned (probably `hoppscotch`).
-4. Start the development server with `pnpm run dev`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -292,9 +292,10 @@ _Sample keys only work with the [production build](https://hoppscotch.io)._
 
 1. [Clone this repo](https://help.github.com/en/articles/cloning-a-repository) with git.
 2. Install pnpm using npm by running `npm install -g pnpm`.
-3. Install dependencies by running `pnpm install` within the directory that you cloned (probably `hoppscotch`).
-4. Start the development server with `pnpm run dev`.
-5. Open the development site by going to [`http://localhost:3000`](http://localhost:3000) in your browser.
+3. Run `cp .env.example .env`(update them with your own info)
```
AndrewBastin, line 295: Isn't this already mentioned in Step 0 (Developing section).

Nabeelshar: There is difference step 0 is suggesting for production build, while sample keys work with local build.

AndrewBastin: That is an intentional decision. Although we do allow the keys to work by default with local build, we don't recommend building with it as if you make a misconfiguration or something you can end up with messed up user data in the production firebase.

AndrewBastin:
Closing this PR as this was an intended statement.

PR #2977, opened 2023-04-11
chore: changes to support id based syncing for collections

amk-dev:
This PR introduces changes to support ID based syncing for collections.

**Changes**

- `removeGraphqlCollection`, `removeGraphqlFolder`, `removeGraphqlRequest`, `removeRestCollection`, `removeRestFolder`, `removeRestRequest` all accepts an optional id. this id is not used in the store right now. but is used in the syncing system for removing these resources.

- Function for removing duplicates from the collections store


AndrewBastin (approved):
lgtm 💯

PR #3002, opened 2023-04-18
fix: selected env changed while sidebar collapsed

### LINKED ISSUE #2995: [bug]: when you close the sidebar and open it again, if you have env selected, it resets

### Current behavior

When you close the sidebar and open it again, if you have env selected, it resets:

image

### Steps to reproduce

1. Open any request
2. Select any env
3. Click on "Collapse sidebar"
4. Click on "Expand sidebar"

### Environment

Production

### Version

Cloud

nivedin:
Closes #2995 

### Description
This PR resolves the problem where, when a user selects a team environment and toggles the sidebar, the selected environment is reset.

### Checks


packages/hoppscotch-common/src/newstore/localpersistence.ts
```diff
@@ -210,35 +208,21 @@ function setupEnvironmentsPersistence() {
...
-    O.fromNullable(window.localStorage.getItem("selectedEnvIndex")),
-    O.map(parseInt), // If not null, parse to integer
-    O.chain(
-      O.fromPredicate(
-        Number.isInteger // Check if the number is proper int (not NaN)
-      )
-    ),
-    O.getOrElse(() => -1) // If all the above conditions pass, we are good, else set default value (-1)
+  const selectedEnvIndex = JSON.parse(
+    window.localStorage.getItem("selectedEnvIndex") || "null"
```
AndrewBastin, line 212: Use `??` instead of `||` when applicable

AndrewBastin (changes requested):

packages/hoppscotch-common/src/newstore/localpersistence.ts
```diff
@@ -210,35 +208,21 @@ function setupEnvironmentsPersistence() {
 }
 
 function setupSelectedEnvPersistence() {
```
amk-dev, line 210: when dealing with localstorage values, backward compatibility is important because different versions of the app may have different values.

in this case our old app can have selectedEnvIndex as -1, which is not a valid `selectedEnvIndex` in our new implementation. so we need to check for that handle it properly. otherwise our app can break.

there is a function `checkAndMigrateOldSettings`, where you can add the check. 

```suggestion
const selectedEnvIndex = localStorage.getItem("selectedEnvIndex")

  if (selectedEnvIndex == "-1") {
    localStorage.setItem(
      "selectedEnvIndex",
      JSON.stringify({
        type: "NO_ENV_SELECTED",
      })
    )
  }
```

packages/hoppscotch-common/src/components/environments/index.vue
```diff
@@ -408,7 +408,6 @@ const selectedEnv = computed(() => {
         teamEnvID: selectedEnvironmentIndex.value.teamEnvID,
       }
     } else {
-      selectedEnvironmentIndex.value = { type: "NO_ENV_SELECTED" }
       return { type: "NO_ENV_SELECTED" }
     }
   } else {
```
amk-dev, line 413: It is a bad practice to have side effects in computed properties. since we're here already,     can you also remove the `selectedEnvironmentIndex.value = { type: "NO_ENV_SELECTED" }` from the next else, if it doesnt serve any purpose. if it has purpose, we can move it to a watcher

amk-dev (changes requested):
look into the comments

amk-dev (approved):
LGTM 🚀

AndrewBastin (approved):

PR #3010, opened 2023-04-21
fix: request name not updating in the save request modal

nivedin:
### Description
This fixes the issue in the save request modal where the request name was not being updated while saving the current active tab request.

### Checks

### DIFF
--- a/packages/hoppscotch-common/src/components/collections/SaveRequest.vue
+++ b/packages/hoppscotch-common/src/components/collections/SaveRequest.vue
@@ -135,6 +135,15 @@ const requestName = ref(
   props.mode === "rest" ? restRequestName.value : gqlRequestName.value
 )
 
+watch(
+  () => [currentActiveTab.value.document.request.name, gqlRequestName.value],
+  () => {
+    if (props.mode === "rest")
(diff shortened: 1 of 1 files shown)


packages/hoppscotch-common/src/components/collections/SaveRequest.vue
```diff
@@ -126,12 +126,13 @@ const emit = defineEmits<{
...
-  () =>
-    props.mode === "rest"
-      ? currentActiveTab.value.document.request.name
-      : gqlRequestName.value
+const restRequestName = computedWithControl(
+  () => currentActiveTab.value,
+  () => currentActiveTab.value.document.request.name
+)
+
+const requestName = ref(
```
amk-dev, line 134: Here since the ref is initialized when the component is rendered first. it won't be updated if there is an update to the currentActiveTab.value, for eg: try renaming the request, and see the value in the save request model.

you can add a watcher on the restRequestName that will update the ref incase the name changes from the tabs side. 

```javascript
watch(restRequestName, (newValue) => {
  requestName.value = newValue
})
```

amk-dev (approved):
LGTM 🚀

PR #3012, opened 2023-04-21
fix: wrong tab selected after navigating from different route

anwarulislam:
### Description
Issue: If multiple tabs are opened, and the first tab is selected. after going to another page from the REST page and then coming back to REST, the selection goes to the next tab.

Previously, we were forcefully selecting the first possible `Window` forcefully after removing any active `Window`. And it was happening `onBeforeUnmounted`. But, now it will select the last `Window` automatically.

- [ ] Not Completed
- [x] Completed

### Checks


AndrewBastin (approved):
lgtm 💯

PR #3041, opened 2023-05-08
fix: invalid environment index can break the app

amk-dev:
Closes HFE-40

**Before**

If we're loading an environment index that doesnt exist from the localstorage, it could lead to unwanted behaviours.
the type for `currentEnvironment$` was not considering the enviornment could be undefined.

**After**

When setting the selectedEnvironmentIndex, we verify if the environment corresponding to the index exist in the store. Otherwise we set selectedEnvironmentIndex for 'NO_ENV_SELECTED'

Also we update the type of currentEnvironment$ to include a possible undefined

### DIFF
--- a/packages/hoppscotch-common/src/newstore/environments.ts
+++ b/packages/hoppscotch-common/src/newstore/environments.ts
@@ -42,21 +42,18 @@ const dispatchers = defineDispatchers({
       selectedEnvironmentIndex,
     }: { selectedEnvironmentIndex: SelectedEnvironmentIndex }
   ) {
-    let isValidIndex = true
-
-    if (selectedEnvironmentIndex.type == "MY_ENV") {
-      isValidIndex =
-        selectedEnvironmentIndex.type == "MY_ENV" &&
-        !!store.environments[selectedEnvironmentIndex.index]
+    if (
+      selectedEnvironmentIndex.type === "MY_ENV" &&
+      !!store.environments[selectedEnvironmentIndex.index]
+    ) {
+      return {
+        selectedEnvironmentIndex,
+      }
+    } else {
+      return {
+        type: "NO_ENV_SELECTED",
+      }
     }
-
-    return isValidIndex
-      ? {
-          selectedEnvironmentIndex,
-        }
-      : {
-          type: "NO_ENV_SELECTED",
-        }
   },
   appendEnvironments(
     { environments }: EnvironmentStore,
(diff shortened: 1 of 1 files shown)


packages/hoppscotch-common/src/newstore/environments.ts
```diff
@@ -37,14 +37,26 @@ type EnvironmentStore = typeof defaultEnvironmentsState
...
+    store: EnvironmentStore,
     {
       selectedEnvironmentIndex,
     }: { selectedEnvironmentIndex: SelectedEnvironmentIndex }
   ) {
-    return {
-      selectedEnvironmentIndex,
+    let isValidIndex = true
+
+    if (selectedEnvironmentIndex.type == "MY_ENV") {
```
AndrewBastin, line 47: === ?

packages/hoppscotch-common/src/newstore/environments.ts
```diff
@@ -37,14 +37,26 @@ type EnvironmentStore = typeof defaultEnvironmentsState
...
+      isValidIndex =
+        selectedEnvironmentIndex.type == "MY_ENV" &&
```
AndrewBastin, line 49: ===

AndrewBastin, line 49: Won't this statement be always true because of the if condition above ?

packages/hoppscotch-common/src/newstore/environments.ts
```diff
@@ -37,14 +37,26 @@ type EnvironmentStore = typeof defaultEnvironmentsState
...
+        !!store.environments[selectedEnvironmentIndex.index]
     }
+
+    return isValidIndex
+      ? {
+          selectedEnvironmentIndex,
+        }
+      : {
+          type: "NO_ENV_SELECTED",
+        }
```
AndrewBastin, line 59:
```suggestion
if (
      selectedEnvironmentIndex.type == "MY_ENV" &&
      !!store.environments[selectedEnvironmentIndex.index]
    ) {
      return {
        selectedEnvironmentIndex,
      }
    } else {
      return {
        type: "NO_ENV_SELECTED",
      }
    }
```

AndrewBastin (changes requested):

packages/hoppscotch-common/src/newstore/environments.ts
amk-dev: yes. i don't know what i was thinking while writing this.

AndrewBastin (approved):
lgtm

PR #3043, opened 2023-05-09
chore: add onCodemirrorInstanceMount hook to platform

amk-dev:
This PR adds an optional `onCodemirrorInstanceMount` hook to platform. this hook runs after the codemirror view is inited. which is helpful to run code after codemirror instances are mounted.


AndrewBastin (approved):
lgtm

PR #3052, opened 2023-05-11
chore: environment selector with new ux

liyasthomas:
Closes HFE-64

### Description
This is a continuation of reverting the environment selector component back to its original features with the new design language that we follow at Hoppscotch. Users can now switch between "My environments" and "Team environments" on a Tab UI within the Environment selector popover.

TLDR; allows users to switch to personal environments from team workspaces.

### Checks

### DIFF
--- a/packages/hoppscotch-common/src/components/environments/Selector.vue
+++ b/packages/hoppscotch-common/src/components/environments/Selector.vue
@@ -1,125 +1,128 @@
 <template>
-  <tippy
-    interactive
-    trigger="click"
-    theme="popover"
-    :on-shown="() => tippyActions!.focus()"
-  >
-    <span
-      v-tippy="{ theme: 'tooltip' }"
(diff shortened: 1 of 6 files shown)


anwarulislam:
@liyasthomas I know we are using Tab component here. But, what do you think about the UI? which one looks better? I prefer the second one.

1.

image


2. 

image

liyasthomas:
@anwarulislam, I think we should keep the Tab UIs consistent across the platform. So let's stick to the current UI: 1.

Moreover, I'm not a fan of representing active/selected states only with a light (or dark) shade of background color.

**Caveats:**

* Hardly distinguishable from disabled button state
* I personally prefer active states should represent the accent color
* Hover, focused states would be difficult to recognize and will be inconsistent with the other components

amk-dev (approved):
LGTM 🚀

PR #3055, opened 2023-05-15
use alpine image & save postgres data to host

smilence86:
Closes # 

### Checks


balub:
@smilence86 the prisma version we currently use does not work with `alpine` versions of `nodejs` and we will be adding optional docker volumes support to the db that users can uncomment and use should they close too.

PR #3063, opened 2023-05-18
fix: sync popup firing multiple times

nivedin:
Closes HFE-46

### Description
This PR fixes the issue where popup for workspace restoration was firing multiple times when navigating from other pages and returning back to the REST page and on some other occasions.

### Checks
- [X] My pull request adheres to the code style of this project
- [X] All the tests have passed

### DIFF
--- a/packages/hoppscotch-common/src/pages/index.vue
+++ b/packages/hoppscotch-common/src/pages/index.vue
@@ -123,6 +123,11 @@ import { useToast } from "~/composables/toast"
 import { PersistableRESTTabState } from "~/helpers/rest/tab"
 import { watchDebounced } from "@vueuse/core"
 import { oauthRedirect } from "~/helpers/oauth"
+import { useReadonlyStream } from "~/composables/stream"
+import {
+  changeCurrentSyncStatus,
+  currentSyncingStatus$,
+} from "~/newstore/syncing"
 
 const savingRequest = ref(false)
 const confirmingCloseForTabID = ref<string | null>(null)
@@ -134,7 +139,10 @@ const toast = useToast()
 
 const tabs = getActiveTabs()
 
-const confirmSync = ref(false)
+const confirmSync = useReadonlyStream(currentSyncingStatus$, {
+  initialSync: false,
+  sync: true,
+})
 const tabStateForSync = ref<PersistableRESTTabState | null>(null)
 
 function bindRequestToURLParams() {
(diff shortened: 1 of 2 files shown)


packages/hoppscotch-common/src/pages/index.vue
```diff
@@ -279,9 +298,12 @@ function setupTabStateSync() {
...
 
-      if (tabStateFromSync) {
+      if (tabStateFromSync && !confirmSync.value.initialSync) {
         tabStateForSync.value = tabStateFromSync
-        confirmSync.value = true
+        changeCurrentSyncStatus({
+          initialSync: true,
+          sync: false,
+        })
       }
```
amk-dev, line 307: From what i understand, Here the user is not actually choosing the initialSync, but i belive we're firing the `changeCurrentSyncStatus` to trigger the watcher that looks for `confirmSync.value.initialSync`, which shows the toast to the user. instead we can just move the content of that watcher to a function and call it.

If there's something i'm missing here, do let me know. 

```suggestion
      if (tabStateFromSync && !confirmSync.value.initialSync) {
        tabStateForSync.value = tabStateFromSync

        showSyncToast()
      }
```

packages/hoppscotch-common/src/pages/index.vue
```diff
@@ -203,28 +211,39 @@ const onSaveModalClose = () => {
...
           },
-        },
-      ],
-    })
+        ],
+      })
+    }
   }
-})
+)
```
amk-dev, line 246: We don't need to do it in a watcher, because the only place we're triggering this is in the `onLoggedIn` hook. ( see the comment on `onLoggedIn`  )

```suggestion
const showSyncToast = () => {
  toast.show(t("confirm.sync"), {
    duration: 0,
    action: [
      {
        text: `${t("action.yes")}`,
        onClick: (_, toastObject) => {
          syncTabState()
          changeCurrentSyncStatus({
            initialSync: true,
            sync: true,
          })
          toastObject.goAway(0)
        },
      },
      {
        text: `${t("action.no")}`,
        onClick: (_, toastObject) => {
          changeCurrentSyncStatus({
            initialSync: true,
            sync: false,
          })
          toastObject.goAway(0)
        },
      },
    ],
  })
}
```

amk-dev (approved):
LGTM 🚀

AndrewBastin (approved):
lgtm

PR #3081, opened 2023-05-26
chore: update wss url to ws in .env.example

amk-dev:
Closes hfe-58

**Description**

Our environment file currently includes a WSS URL (WebSockets Sequer URL). However, while testing on localhost, this will fail ( none of our subscriptions will work ) because you need to set up certificates and run a WSS server on your local machine. To prevent this, we recommend that the local server URL in the environment file default to using WebSockets instead of the WSS protocol.


liyasthomas (approved):
LGTM 💯

AndrewBastin (approved):
lgtm 💯

PR #3108, opened 2023-06-03
fix: stack order of tab inside environment selector

liyasthomas:
### Summary
🐛📏🗂️

Increased z-index of environment selector tabs to prevent them from being obscured by request editor. Fixed a UI bug in `Selector.vue`.

> _`z-0` to `z-10`_
> _tabs rise above the editor_
> _autumn bug is fixed_

### Walkthrough
*  Increase the z-index of the environment selector tabs to fix a bug where they would be hidden behind the request editor when scrolling ([link](https://github.com/hoppscotch/hoppscotch/pull/3108/files?diff=unified&w=0#diff-584d8291b417c80e6df02186ce776cf6bf6bb10c3a9627c2512fbb0ac34fa664L52-R52))

### DIFF
--- a/packages/hoppscotch-common/src/components/environments/Selector.vue
+++ b/packages/hoppscotch-common/src/components/environments/Selector.vue
@@ -49,7 +49,7 @@
         />
         <HoppSmartTabs
           v-model="selectedEnvTab"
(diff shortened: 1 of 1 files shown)


nivedin (approved):
LGTM 💯

PR #3130, opened 2023-06-09
fix: fix url getting overridden when query params are present

### LINKED ISSUE #3098: [bug]: When installed as Chrome PWA the first tab url is always overridden with the default 'https://echo.hoppscotch.io' request

### Bug behaviour

While using the installed PWA in Chrome (Version 114.0.5735.45 (Official Build) beta (arm64)) when I start or refresh the application, my first tab request is always replaced with `https://echo.hoppscotch.io`:

https://github.com/hoppscotch/hoppscotch/assets/2415410/b38201dc-7668-4fc8-b27c-6a40ff1b2d9f

### Expected behaviour

When using the app via browser, the tab url correctly persists between tab reloads and closing: 

https://github.com/hoppscotch/hoppscotch/assets/2415410/e97dad2d-83c5-4c43-bf9d-68ce95daefa8

### Steps to reproduce

1. Install Hoppscotch as a Chrome application
2. Open the application and change the default url from `https://echo.hoppscotch.io` to any other (e.g. `https://ifconfig.me/`)
3. Reload or reopen the application
4. Url is back to `https://echo.hoppscotch.io` instead of `https://ifconfig.me/`

### Environment

Production

### Version

Cloud

amk-dev:
fixes #3098
fixes HFE-71

**Before**

We support loading request properties from URL query params. we have a function `translateExtURLParams` that takes query params and converts it into Hoppscotch request properties, but this function was always adding these properties to the default request. so whenever a query param is present, this function is running, and it overwrites the existing request.

**After**

This function takes an optional `initialRequest` parameter that will be used instead of the defaultRequest if passed.

### DIFF
--- a/packages/hoppscotch-common/src/helpers/RESTExtURLParams.ts
+++ b/packages/hoppscotch-common/src/helpers/RESTExtURLParams.ts
@@ -6,14 +6,18 @@ import { isJSONContentType } from "./utils/contenttypes"
  * Handles translations for all the hopp.io REST Shareable URL params
  */
 export function translateExtURLParams(
-  urlParams: Record<string, any>
+  urlParams: Record<string, any>,
+  initialReq?: HoppRESTRequest
 ): HoppRESTRequest {
-  if (urlParams.v) return parseV1ExtURL(urlParams)
-  else return parseV0ExtURL(urlParams)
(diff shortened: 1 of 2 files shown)


AndrewBastin (approved):
lgtm :100:

PR #3143, opened 2023-06-16
refactor: modified shortcodes UI to maintain table layout in smaller screens

JoelJacobStephen:
### Issue Number
Closes HFE-104

### Description
This PR aims to address the UI inconsistencies in the Shortcodes table UI where the headers of the table disappear and the table collapses to a multiline container when screen size is adjusted. The new UI changes makes sure that the table layout holds even for smaller screen sizes.

image

### Checks


anwarulislam:
@JoelJacobStephen, on a smaller screen, the UI is still broken. This view is in width 350. I think for mobile view we should have list view instead of table view.

image

anwarulislam (changes requested):
Please check the comment

JoelJacobStephen:
> @JoelJacobStephen, on a smaller screen, the UI is still broken. This view is in width 350. I think for mobile view we should have list view instead of table view.
> 
> image
@anwarulislam 

The Shortcodes Table UI is currently scaling properly upto the tablet view in this fix but I did not focus on the mobile view since mobile view is not supported. This table will be completely rewritten using the HoppSmartTable component which is will be introduced in my next PR.

PR #3112, opened 2023-06-05
Support for unified docker using S6 from LSIO

webysther:
Closes #3111 #3093 #3092 #3091 #3069 #3060 #3036 #3009 #3007 #2999 #2990 #2581
And this PR answer multiple questions in discussions.

### Description

I published this docker to provide a glympse of what can be done, hope this help to provide a better framework for the community, best whishes.

Note¹: Is necessary include `pwgen` and `python3-pip` to base docker to this work, as `COPY .docker /`
Note²: Here have the images ready to use: https://hub.docker.com/repository/docker/webysther/hoppscotch

# Supported tags and respective `Dockerfile` links

# [hoppscotch](https://github.com/hoppscotch/hoppscotch/pull/3112)

Hoppscotch latest and base are based on PRs [#3107](https://github.com/hoppscotch/hoppscotch/pull/3107) and [#3112](https://github.com/hoppscotch/hoppscotch/pull/3112) for a version unified.

Based on LSIO Images featuring:

* easy user mappings (PGID, PUID)
* custom base image with s6 overlay
* weekly base OS updates with common layers across the entire LinuxServer.io ecosystem to minimise space usage, down time and bandwidth
* regular security updates

The architecture supported by this image is:

| Architecture | Available | Tag |
| :----: | :----: | ---- |
| x86-64 | ✅ | amd64-\<version tag\> |

## Version Tags

This image provides various versions that are available via tags. Please read the descriptions carefully and exercise caution when using unstable or development tags.

| Tag | Available | Description |
| :----: | :----: |--- |
| latest | ✅ | Stable releases from Hoppscotch |

## Application Setup

- Once running the URL will be `http://<host-ip>:80` and `http://<host-ip>:80/admin` (admin).

The application accepts a series of environment variables to further customize itself on boot.

Provided by the PRs [#3107](https://github.com/hoppscotch/hoppscotch/pull/3107) and [#3112](https://github.com/hoppscotch/hoppscotch/pull/3112):

| Parameter | Function |
| :---: | --- |
| `-e MODE=` | Default is **single** but can use **nouser** or **team**|
| `-e ENABLE_ADMIN=` | Default is false. Set to true to build and run admin|
| `-e ENABLE_API=` | Default is true. Build and run backend|
| `-e ENABLE_SMTP=` | Default is true. SMTP server that redirect all e-mail to stdout|
| `-e DOMAIN=` | Default is **locahost**. |
[...]

PR #3150, opened 2023-06-18
fix: Add healthcheck for hoppscotch-db in docker-compose, and more

### LINKED ISSUE #2999: [bug]: Docker Compose Backend container always restarting

### Current behavior

I have cloned the repository, copied the .env.example file to .env and changed with custom settings

Running docker compose up -ad the backend container is build but continous restarting with following error:

hoppscotch-backend  | node:internal/modules/cjs/loader:958
hoppscotch-backend  |   throw err;
hoppscotch-backend  |   ^
hoppscotch-backend  | 
hoppscotch-backend  | Error: Cannot find module '/usr/src/app/dist/main'
hoppscotch-backend  |     at Module._resolveFilename (node:internal/modules/cjs/loader:955:15)
hoppscotch-backend  |     at Module._load (node:internal/modules/cjs/loader:803:27)
hoppscotch-backend  |     at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
hoppscotch-backend  |     at node:internal/main/run_main_module:18:47 {
hoppscotch-backend  |   code: 'MODULE_NOT_FOUND',
hoppscotch-backend  |   requireStack: []
hoppscotch-backend  | }

### Steps to reproduce

Clone repository
Copy .env.example to .env
Execute docker compose up -d

### Environment

Production

### Version

Self-hosted

NicklasWallgren:
Run database migration on docker-compose up.
Add docker-compose healthcheck for hoppscotch-backend. 
Remove volume to fix `MODULE_NOT_FOUND` error.

Closes #2999

### Description
1. Applies the database migration after the database is ready, and before the `hoppscotch-backend` application starts. 
2. Removes unnecessary volumes which results in `MODULE_NOT_FOUND` and/or permission errors.

### Checks

- [X] My pull request adheres to the code style of this project

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -67,11 +67,13 @@ services:
       - "5432:5432"
     user: postgres
     environment:
+      # The default user defined by the docker image
+      POSTGRES_USER: postgres
       # NOTE: Please UPDATE THIS PASSWORD!
       POSTGRES_PASSWORD: testpass
       POSTGRES_DB: hoppscotch
     healthcheck:
-      test: [ "CMD-SHELL", "pg_isready" ]
+      test: ["CMD-SHELL", "sh -c 'pg_isready -U $${POSTGRES_USER} -d $${POSTGRES_DB}'"]
       interval: 5s
       timeout: 5s
       retries: 10
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -78,8 +78,8 @@ services:
       POSTGRES_DB: hoppscotch
     healthcheck:
       test: [ "CMD-SHELL", "pg_isready" ]
-      interval: 10s
+      interval: 5s
       timeout: 5s
-      retries: 5
+      retries: 10
 
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -27,12 +27,6 @@ services:
         condition: service_healthy
     ports:
       - "3170:3000"
-    command:
-      - /bin/sh
-      - -c
-      - |
-        pnpm exec prisma migrate deploy
-        pnpm run start:prod
 
   # The main hoppscotch app. This will be hosted at port 3000
   # NOTE: To do TLS or play around with how the app is hosted, you can look into the Caddyfile for
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,6 +19,8 @@ services:
       - DATABASE_URL=postgresql://postgres:testpass@hoppscotch-db:5432/hoppscotch?connect_timeout=300
       - PORT=3000
     volumes:
+      # Uncomment the line below when modifying code. Only applicable when using the "dev" target.
+      # - ./packages/hoppscotch-backend/:/usr/src/app
       - /usr/src/app/node_modules/
     depends_on:
       hoppscotch-db:
@@ -66,7 +68,7 @@ services:
   # you are using an external postgres instance
   # This will be exposed at port 5432
   hoppscotch-db:
-    image: postgres:15.3-alpine3.18
+    image: postgres:15
     ports:
       - "5432:5432"
     user: postgres
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,12 +19,18 @@ services:
       - DATABASE_URL=postgresql://postgres:testpass@hoppscotch-db:5432/hoppscotch?connect_timeout=300
       - PORT=3000
     volumes:
-      - ./packages/hoppscotch-backend/:/usr/src/app
       - /usr/src/app/node_modules/
     depends_on:
-      - hoppscotch-db
+      hoppscotch-db:
+        condition: service_healthy
     ports:
       - "3170:3000"
+    command:
+      - /bin/sh
+      - -c
+      - |
+        pnpm exec prisma migrate deploy
+        pnpm run start:prod
 
   # The main hoppscotch app. This will be hosted at port 3000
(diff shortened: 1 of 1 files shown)


balub:
@AndrewBastin This PR looks good to me I tested everything out works fine. However, the fact that we got rid of the volumes will be a problem when we are working on the BE codebase, now this isn't really a problem when all you care about is running the platform. My suggestion (@NicklasWallgren ) would be if we could comment out the volumes part along with a comment saying "uncomment when modifying code" or something.

NicklasWallgren:
> @AndrewBastin This PR looks good to me I tested everything out works fine. However, the fact that we got rid of the volumes will be a problem when we are working on the BE codebase, now this isn't really a problem when all you care about is running the platform. My suggestion (@NicklasWallgren ) would be if we could comment out the volumes part along with a comment saying "uncomment when modifying code" or something.

Fair point, I'll add a comment 👍

balub (approved):
Thank you. @NicklasWallgren for your contribution, everything LGTM 🚀.

balub:
@NicklasWallgren After discussion with the rest of the team, we kinda decided automatically running migrations is not something we want to advocate for and rather prefer to have the users do it manually. So could you remove that part from your PR, so that we can merge everything else.

NicklasWallgren:
Alright, that is unfortunate.

It's a fairly common practice to run db migration before or during boot, either via a embedded tool or via init containers in k8s. 

FYI, we're planing to add a init container to the unofficial helm chart which will apply the db migrations before the container has started. See https://github.com/WDaan/hoppscotch-helm/issues/6

balub:
@NicklasWallgren Thank you for your contribution anyway, I agree it's a common practice to run migrations automatically but we want our users to be more involved in all the steps for this initial version of our Self-host project.

balub:
@NicklasWallgren and @WDann I was hoping to if possible connect with you folks, our team does not have a lot of experience working with helm charts, k8's etc I was hoping maybe together we could help each other out and have some official helm charts or something along those lines?

AndrewBastin:
> So could you remove that part from your PR, so that we can merge everything else.

@NicklasWallgren can you follow up on this ? thanks.

NicklasWallgren:
> > So could you remove that part from your PR, so that we can merge everything else.
> 
> @NicklasWallgren can you follow up on this ? thanks.

I've updated the PR.

NicklasWallgren:
> @NicklasWallgren and @WDaan I was hoping to if possible connect with you folks, our team does not have a lot of experience working with helm charts, k8's etc I was hoping maybe together we could help each other out and have some official helm charts or something along those lines? Discord, email whatever works for you folks.

I'm happy to help. @WDaan has laid the foundation, but there is still much to be done before the Helm chart can be considered ready.

The biggest hurdle is probably the frontend docker images, which doesn't provide "dynamic" environment variables. I was thinking of creating custom docker images with support of dynamic env values, but maybe we could incorporate that into the official images.

WDaan:
What @NicklasWallgren said ^^

I'd be happy to transfer the repo to your org, might attract a bigger audience there (and make it a bit more official).

AndrewBastin:
I have a plan for dynamic frontend environment variables, we aren't a fan of just `sed`-ing through the code as what #3112 introduced and is looking for a solution more well oriented from the code side so that it just doesn't depend on the code structure for its functioning.

I am trying to get that going and should have a PR open in the coming weeks to get the implementation going for `2023.8.0` (again as I mentioned in #3112).

balub:
@NicklasWallgren could we possibly add more flags into the healthcheck test command like it is mentioned in this [answer](https://stackoverflow.com/a/73049146) ?

NicklasWallgren:
> @NicklasWallgren could we possibly add more flags into the healthcheck test command like it is mentioned in this [answer](https://stackoverflow.com/a/73049146) ?

Sure, but I don't see any benefits with adding additional flags :) 

**pg_isready** checks if the server (instance, cluster, ..) is ready to accept connections. It is not intended to check if a specific database (or user or table) inside that instance exists.

**Documentation**
"It is not necessary to supply user name, password, or database name values to obtain the server status; however, if incorrect values are provided, the server will log a failed connection attempt."

NicklasWallgren:
Are we ready to merge this one? 🙂

docker-compose.yml
```diff
@@ -60,12 +62,18 @@ services:
...
+    image: postgres:15
     ports:
       - "5432:5432"
+    user: postgres
     environment:
       # NOTE: Please UPDATE THIS PASSWORD!
       POSTGRES_PASSWORD: testpass
       POSTGRES_DB: hoppscotch
+    healthcheck:
+      test: [ "CMD-SHELL", "pg_isready" ]
```
ankitsridhar16, line 74: @NicklasWallgren we could use 
`test: ["CMD-SHELL", "pg_isready -U $${POSTGRES_USER} -d $${POSTGRES_DB}"]`

NicklasWallgren: But why? As stated by the documentation "It is not necessary to supply correct user name, password, or database name values to obtain the server status;"

The `pg_isready` will return 0 either way, if the database is accepting a connection.

```bash
postgres@baf25b68a6a1:/$ pg_isready -U wrong_user -d wrong_password
/var/run/postgresql:5432 - accepting connections
postgres@baf25b68a6a1:/$ echo $?
0
```

balub (approved):

ankitsridhar16 (approved):

PR #3152, opened 2023-06-19
fix: use --location param for url when parsing curl

### LINKED ISSUE #3109: Import CURL does not work
Hey, 

Trying to import a valid CURL and it does not work, it only sets the method type, and sometimes sets the headers and body, 
and the URL is set to https://echo.hoppscotch.io/

curl --location 'http://my-private-url.test.com' \
--header 'Accept: application/jsonx' \
--header 'Content-Type: application/jsonx' \
--data '{"a":1,"b":2,"c":3}'

the curl imports fine in Postman.

Can you please check and advise? thanks

amk-dev:
Fixes #3109 

### Url is not properly parsed when importing using curl 

**Before**

We were not respecting the `--location` parameter when parsing the curl request.

**After** 

Respect the `--location` parameter when parsing the curl request.

### Body And Headers are not loaded properly

**Before**

In component `RequestOptions.vue`, we have a ref `request` which is set to `props.modelValue` in the setup function. but when the component updates with new props, this is not updated. this caused child components that uses this `request` ref to not get updated reactively.

**After**

Added a watcher that updates the `request` ref, when the props change.

### DIFF
--- a/packages/hoppscotch-common/src/components/http/RequestOptions.vue
+++ b/packages/hoppscotch-common/src/components/http/RequestOptions.vue
@@ -72,6 +72,10 @@ const emit = defineEmits<{
 
 const request = ref(props.modelValue)
 
+watch(props, (updatedProps) => {
+  request.value = updatedProps.modelValue
+})
+
 watch(
   () => request.value,
   (newVal) => {
--- a/packages/hoppscotch-common/src/components/http/RequestOptions.vue
+++ b/packages/hoppscotch-common/src/components/http/RequestOptions.vue
@@ -54,7 +54,8 @@
 <script setup lang="ts">
 import { useI18n } from "@composables/i18n"
 import { HoppRESTRequest } from "@hoppscotch/data"
-import { computed, ref, watch } from "vue"
+import { useVModel } from "@vueuse/core"
+import { computed, ref } from "vue"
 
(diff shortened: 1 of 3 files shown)


nivedin (changes requested):
```
curl --location 'http://my-private-url.test.com/'
--header 'Accept: application/jsonx'
--header 'Content-Type: application/jsonx'
--data '{"a":1,"b":2,"c":3}'
```

when trying to import the curl the body is not set, and setting the header has also some bugs.

nivedin (approved):
LGTM 💯

packages/hoppscotch-common/src/components/http/RequestOptions.vue
```diff
@@ -72,6 +72,10 @@ const emit = defineEmits<{
 
 const request = ref(props.modelValue)
 
+watch(props, (updatedProps) => {
```
AndrewBastin, line 75: I think we have `request` just representing the modelValue right ?
In that case, why not just define `request` using `useVModel` ?

Will something break ?
```ts
import { useVModel } from "@vueuse/core"

const request = useVModel(props, "modelValue", emit)
```

AndrewBastin (changes requested):

AndrewBastin (approved):
lgtm

PR #3153, opened 2023-06-19
Fix typos

omerbaflah:
### Description
This PR fixes typos

### Checks

### DIFF
--- a/packages/hoppscotch-backend/src/admin/admin.service.ts
+++ b/packages/hoppscotch-backend/src/admin/admin.service.ts
@@ -51,8 +51,8 @@ export class AdminService {
 
   /**
    * Invite a user to join the infra.
-   * @param adminUID Admin's UID
-   * @param adminEmail Admin's email
+   * @param adminUID Admins UID
+   * @param adminEmail Admins email
    * @param inviteeEmail Invitee's email
    * @returns an Either of `InvitedUser` object or error
    */
--- a/packages/hoppscotch-backend/src/admin/admin.service.ts
+++ b/packages/hoppscotch-backend/src/admin/admin.service.ts
@@ -51,8 +51,8 @@ export class AdminService {
 
   /**
    * Invite a user to join the infra.
-   * @param adminUID Admins UID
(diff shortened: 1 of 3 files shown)


balub (dismissed):

packages/hoppscotch-backend/src/admin/admin.service.ts
```diff
@@ -51,8 +51,8 @@ export class AdminService {
 
   /**
    * Invite a user to join the infra.
-   * @param adminUID Admin's UID
-   * @param adminEmail Admin's email
+   * @param adminUID Admins UID
+   * @param adminEmail Admins email
```
AndrewBastin, line 55: This correction is wrong if you are talking grammatically.
`adminUID` and `adminEmail` represent UID and the Email of the admin (i.e Admin's UID and Admin's Email).

This should be reverted.

packages/hoppscotch-backend/src/types/AuthError.ts
```diff
@@ -2,7 +2,7 @@ import { HttpStatus } from '@nestjs/common';
 
 /**
  ** Custom interface to handle errors specific to Auth module
- ** Since its REST we need to return HTTP status code along with error message
+ ** Since it's REST we need to return the HTTP status code along with the error message
```
AndrewBastin, line 5:
```suggestion
 ** Since its REST we need to return the HTTP status code along with the error message
```

AndrewBastin (changes requested):

packages/hoppscotch-backend/src/admin/admin.service.ts
omerbaflah: Thanks, Andrew for the clarification, I'll revert this change.

packages/hoppscotch-backend/src/types/AuthError.ts
omerbaflah: noted and fixed.

AndrewBastin (approved):
lgtm

PR #3157, opened 2023-06-22
fix: remove existing team invitation for an invitee when adding invitee to team by admin (HBE-229)

ankitsridhar16:
Closes # 
HBE- 229

### Description
When an admin adds an invitee to team by email and team ID the pending team invitation is not revoked. Because of this if this invitee was invited by a team member there would be a team pending team invitation for the invitee which is not required and is invalid hence this PR solves the issue by revoking the invitation.

### Checks

### Additional Information
Testing Instructions:
- Create a team and add a user to a team by sending an invitation, an entry would be made to the teamInvitation model.
- Now add the user by using admin privilege and mutation , the user would be added to the team directly and the existing invitation will be revoked and the entry will be deleted.


packages/hoppscotch-backend/src/admin/admin.service.ts
```diff
@@ -240,14 +240,26 @@ export class AdminService {
       teamID,
       user.value.uid,
     )();
-    if (E.left(isUserAlreadyMember)) {
+    if (E.isLeft(isUserAlreadyMember)) {
```
mirarifhasan, line 243:
```suggestion
    if (E.isLeft(teamMember)) {
```
As it is not a boolean type value, we can change the variable name.

packages/hoppscotch-backend/src/admin/admin.service.ts
```diff
@@ -240,14 +240,26 @@ export class AdminService {
...
       const addedUser = await this.teamService.addMemberToTeamWithEmail(
         teamID,
         userEmail,
         role,
       );
       if (E.isLeft(addedUser)) return E.left(addedUser.left);
 
+      const isUserInvited =
```
mirarifhasan, line 251:
```suggestion
      const teamInvitation =
```
As it is not a boolean type value, we can change the variable name.

packages/hoppscotch-backend/src/team-invitation/team-invitation.service.ts
```diff
@@ -63,6 +66,32 @@ export class TeamInvitationService {
...
 
+  /**
+   * Get the team invite for an invitee with email and teamID.
+   * @param inviteeEmail invitee email
+   * @param teamID team id
+   * @returns an Either of team invitation for the invitee or error
+   */
+  async getTeamInviteByEmailAndTeamID(inviteeEmail: string, teamID: string) {
+    const isEmailValid = validateEmail(inviteeEmail);
+    if (!isEmailValid) return E.left(INVALID_EMAIL);
```
mirarifhasan, line 77: This callee `getTeamInviteByEmailAndTeamID` function validates the new email.
The caller `AdminService.addUserToTeam` function also validates the same email.

So we can keep one and eliminate the other one.

mirarifhasan (changes requested):

packages/hoppscotch-backend/src/team-invitation/team-invitation.service.ts
ankitsridhar16: `getTeamInviteByEmailAndTeamID` is a service method inside team invitation service written with the motivation so that we could use it later for similar operation if required I think we can keep it.

packages/hoppscotch-backend/src/admin/admin.service.ts
ankitsridhar16: cool

packages/hoppscotch-backend/src/admin/admin.service.ts
ankitsridhar16: okay

balub (approved):
Please resolve all the unresolved conversations @ankitsridhar16 .

mirarifhasan (approved):

PR #3174, opened 2023-07-06
fix: email not being checked case insensitive on team invitation acceptance

AndrewBastin:
Closes HBE-238

### Description
This PR deals with a tiny bug in the team invitation acceptance logic and `TeamInviteeGuard` on how they compare if two emails are equal.

In the places fixed by this PR, they were compared in a case-sensitive way, but generally emails should be compared in case-insensitive as many email servers do normalise it to lowercase (even though it is not specified as such in the Email Spec).

### Checks

### Additional Information
N/A


balub (approved):
Tested, LGTM. 🚀

PR #3162, opened 2023-06-27
[feat] : Allow admins to revoke a team invite (HBE-230)

ankitsridhar16:
Closes # HBE-233

### Description
Currently Admins can view pending invitations for a team, but they can't revoke an invitation. This PR introduces a mutation to revoke a team invite by invite ID that will be used from the admin dashboard.

### Checks

- [x] My pull request adheres to the code style of this project
- [ ] My code requires changes to the documentation
- [ ] I have updated the documentation as required
- [x] All the tests have passed

### Additional Information

Testing Instructions:
- Create a team and invite users by email
- get pending invitations for the team by team ID
- use `revokeTeamInviteByAdmin` mutation with a invite ID

PR #3176, opened 2023-07-07
Fixed team invite case sensitive and admin can't delete invites

rhehl:
I had the problem that by using MS as an authentication provider, users ended up in the database with email addresses in which the first letter of the first and last name was capitalized, which is basically not a problem. However, this meant that if the team invite was not spelled exactly the same way, it was not possible to accept it.

The second problem was that if I hadn't created the invite or the team, as an admin I couldn't delete the broken invite because it only checked whether I had the necessary role in the team.

### DIFF
--- a/packages/hoppscotch-backend/src/team-invitation/team-invitation.service.ts
+++ b/packages/hoppscotch-backend/src/team-invitation/team-invitation.service.ts
@@ -256,7 +256,7 @@ export class TeamInvitationService {
           pipe(
             undefined,
             TE.fromPredicate(
-              (a) => acceptedBy.email === invitation.inviteeEmail,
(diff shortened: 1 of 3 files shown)


balub:
@rhehl A PR already exists to solve this exact issue [here](https://github.com/hoppscotch/hoppscotch/pull/3174), it will be merged and released in the next point release. 

Still thank you for your contribution 😄

rhehl:
@balub ok i missed that, what about the packages/hoppscotch-backend/src/team-invitation/team-invite-team-owner.guard.ts changes?

balub:
We have another [PR](https://github.com/hoppscotch/hoppscotch/pull/3162) for that too.

PR #3178, opened 2023-07-09
feat: introducing a new smart table hoppscotch ui component

JoelJacobStephen:
### Issue Number
Closes HFE-130

### Description
This PR introduces a new Smart Table component in `hoppscotch-ui`. This component can be used to create a table with any number of rows and columns. Parent components can control various aspects of the table including content, border, styles and can provide their own implementation for the action button at the end of each row of the table. It also provides the ability to insert badges and subtitles in any row and column of the table. 

This PR replaces all the tables used in the `hoppscotch-sh-admin` repo with the `HoppSmartTable` component

### Checks


packages/hoppscotch-ui/src/components/smart/Table.vue
```diff
@@ -0,0 +1,129 @@
...
+  /** The headings of the table */
+  headings: string[]
+  /** The styles to be applied to the table cells */
+  cellStyles?: string
+  /** The name of the badge */
+  badgeName?: string
+  /** The indices of the rows that needs to have a badge */
+  badgeRowIndices?: (number | undefined)[]
+  /** The index of the column that needs to have a badge */
+  badgeColName?: string
```
anwarulislam, line 107: Since it's a genertic table component. badge related props shouldn't be there. All table are not gonna have badge for its specific cell

packages/hoppscotch-ui/src/components/smart/Table.vue
```diff
@@ -0,0 +1,129 @@
...
+}>()
+
+// Returns all the columns that needs to have a subtitle
+const subtitleColumns = computed(() =>
+  props.subtitles ? props.subtitles.map((item) => item.colName) : []
+)
+
+// Returns the subtitle for the given column and row
+const itemSubtitle = (subtitle: string | string[], rowIndex: number) =>
+  typeof subtitle === "object" ? subtitle[rowIndex] : subtitle
```
anwarulislam, line 128: We should be able to access content from the parent component itself. We shouldn't have subtitles in the table comopnent. Make the table component generic. Just to render tr/td. We should be able to access row data from parent component like this `<template #row="item">`

anwarulislam (changes requested):
Please look at these issues

packages/hoppscotch-ui/src/components/smart/Table.vue
nivedin, line 1: can we make the Table more generic, like using slots for the head and body so that we can use custom styles for `<td> <th> <tr>` rather than passing styles.

Something like this
```
<HoppTable :list="list">
    <template #head>
      <tr>
        <th>Name</th>
        <th>Age</th>
        <th>Email</th>
        <th>Address</th>
      </tr>
    </template>
    <template #body="{rows}">
      <tr v-for="row in rows" :key="row.id">
        <td>{{ row.name }}</td>
        <td>{{ row.age }}</td>
        <td>{{ row.email }}</td>
        <td>
          {{ row.address.street }},
          {{ row.address.city }}
          {{ row.address.state }}
        </td>
      </tr>
    </template>
  </HoppTable>
```

nivedin (changes requested):
please look into these comments

packages/hoppscotch-ui/src/components/smart/Table.vue
liyasthomas: +1 for this.

packages/hoppscotch-sh-admin/src/pages/users/index.vue
```diff
@@ -31,16 +31,100 @@
...
-          v-else-if="usersList.length >= 1"
-          :usersList="usersList"
-          :fetching="fetching"
-          :error="error"
-          @goToUserDetails="goToUserDetails"
-          @makeUserAdmin="makeUserAdmin"
-          @makeAdminToUser="makeAdminToUser"
-          @deleteUser="deleteUser"
-        />
+        <div v-else-if="usersList.length >= 1" class="m-5">
```
liyasthomas, line 34: Instead of the `>= 1` condition, let's use the `>0` check, hence it'll be consistent across the platform.

liyasthomas (changes requested):

packages/hoppscotch-sh-admin/src/pages/teams/index.vue
```diff
@@ -13,24 +13,57 @@
...
-          @deleteTeam="deleteTeam"
-          class=""
-        />
+        <div v-else-if="teamsList.length == 0" class="px-2">
+          <p class="text-lg">
+            {{ t('teams.no_teams') }}
+          </p>
+        </div>
+
+        <div v-else class="m-5">
```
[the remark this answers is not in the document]
JoelJacobStephen: I am using a margin of 5 around the table because I want a bit of space on all sides of the table and I don't want it to reach the ends of the border. I feel it helps the table to stand out instead of spreading across the border of the frame.

nivedin (changes requested):
please look into these comments

packages/hoppscotch-sh-admin/src/pages/users/index.vue
```diff
@@ -31,16 +31,99 @@
...
+                          @click="deleteUser(item)"
+                        />
+                      </div>
+                    </template>
+                  </tippy>
+                </span>
+              </div>
+            </template>
+          </HoppSmartTable>
+        </div>
```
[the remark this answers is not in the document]
nivedin: we can now use the body slot data and iterate over the row. the above code is just a sample table we can work over it :)

packages/hoppscotch-sh-admin/src/pages/teams/index.vue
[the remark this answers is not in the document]
nivedin: I don't think the table will hit the border since there is horizontal padding on both sides by the parent div. also its better to align with the button on top
image

packages/hoppscotch-sh-admin/src/pages/teams/index.vue
```diff
@@ -13,24 +13,107 @@
...
+                </td>
+
+                <td @click="goToTeamDetails(team.id)" class="py-4 px-3">
+                  <span class="ml-7">
+                    {{ team.members?.length }}
+                  </span>
+                </td>
+
+                <td>
+                  <div class="relative">
```
nivedin, line 81:
```suggestion
              <tr
                v-for="team in list"
                :key="team.id"
                class="text-secondaryDark hover:bg-divider hover:cursor-pointer rounded-xl"
                @click="goToTeamDetails(team.id)"
              >
                <td class="py-4 px-3 max-w-50">
                  <div class="flex">
                    <span class="truncate">
                      {{ team.id }}
                    </span>
                  </div>
                </td>

                <td class="py-4 px-3 min-w-80">
                  <span
                    v-if="team.name"
                    class="flex items-center ml-4 truncate"
                  >
                    {{ team.name }}
                  </span>
                  <span v-else class="flex items-center ml-4">
                    {{ t('teams.unnamed') }}
                  </span>
                </td>

                <td class="py-4 px-3">
                  <span class="ml-7">
                    {{ team.members?.length }}
                  </span>
                </td>

                <td @click.stop>
                  <div class="relative">
```

we can use @click on the row instead of all <td> and can stop event propagation on the tippy using @click.stop

packages/hoppscotch-sh-admin/src/pages/users/index.vue
```diff
@@ -31,16 +31,141 @@
...
+                      {{ getCreatedDate(user.createdOn) }}
+                      <div class="text-gray-400 text-tiny">
+                        {{ getCreatedTime(user.createdOn) }}
+                      </div>
+                    </div>
+                  </div>
+                </td>
+
+                <td>
+                  <div class="relative">
```
nivedin, line 109: same thing mentioned for team table

nivedin (changes requested):
please look into these comments

nivedin (approved):
LGTM 💯

packages/hoppscotch-sh-admin/src/pages/teams/index.vue
```diff
@@ -13,24 +13,102 @@
...
+                <td class="py-4 px-3 min-w-80">
+                  <span
+                    v-if="team.name"
+                    class="flex items-center ml-4 truncate"
+                  >
+                    {{ team.name }}
+                  </span>
+                  <span v-else class="flex items-center ml-4">
+                    {{ t('teams.unnamed') }}
+                  </span>
```
jamesgeorge007, line 66:
```suggestion
                  <span
                    class="flex items-center ml-4 truncate"
                    :class="{ 'truncate' : team.name }"
                  >
                    {{ team.name ?? t('teams.unnamed') }}
                  </span>
```

How about having a single `span` element in the markup and conditionally rendering the contents?

packages/hoppscotch-sh-admin/src/pages/users/index.vue
```diff
@@ -31,16 +31,139 @@
...
+                class="text-secondaryDark hover:bg-divider hover:cursor-pointer rounded-xl"
+                @click="goToUserDetails(user.uid)"
+              >
+                <td class="py-2 px-3 max-w-30">
+                  <div class="flex">
+                    <span class="truncate">
+                      {{ user.uid }}
+                    </span>
+                  </div>
+                </td>
```
jamesgeorge007, line 61: Would it be possible to get rid of the `div` element and apply the `flex` class to the `td` element?

packages/hoppscotch-sh-admin/src/pages/users/index.vue
```diff
@@ -31,16 +31,139 @@
...
+          <HoppSmartTable :list="usersList">
+            <template #head>
+              <tr
+                class="text-secondary border-b border-dividerDark text-sm text-left bg-primaryLight"
+              >
+                <th class="px-6 py-2">{{ t('users.id') }}</th>
+                <th class="px-6 py-2">{{ t('users.name') }}</th>
+                <th class="px-6 py-2">{{ t('users.email') }}</th>
+                <th class="px-6 py-2">{{ t('users.date') }}</th>
+                <th class="px-6 py-2"></th>
```
jamesgeorge007, line 44: Is the empty table cell intentional?

packages/hoppscotch-sh-admin/src/pages/users/index.vue
```diff
@@ -31,16 +31,139 @@
...
-          v-else-if="usersList.length >= 1"
-          :usersList="usersList"
-          :fetching="fetching"
-          :error="error"
-          @goToUserDetails="goToUserDetails"
-          @makeUserAdmin="makeUserAdmin"
-          @makeAdminToUser="makeAdminToUser"
-          @deleteUser="deleteUser"
-        />
+        <div v-else-if="usersList.length > 0">
```
jamesgeorge007, line 34:
```suggestion
        <div v-else-if="usersList.length">
```

Won't a truthy check suffice?

packages/hoppscotch-sh-admin/src/pages/users/index.vue
```diff
@@ -31,16 +31,139 @@
...
+                    <span class="truncate">
+                      {{ user.uid }}
+                    </span>
+                  </div>
+                </td>
+
+                <td class="py-2 px-3">
+                  <div
+                    v-if="user.displayName"
+                    class="flex items-center space-x-3"
```
jamesgeorge007, line 66: Is it a convention to use space utilities with `flex` in Windi CSS? They have some limitations in Tailwind, as outlined in the [docs](https://tailwindcss.com/docs/space#limitations).

packages/hoppscotch-sh-admin/src/pages/users/index.vue
```diff
@@ -31,16 +31,139 @@
...
+
+                <td class="py-2 px-3">
+                  <div class="flex items-center">
+                    <div class="flex flex-col">
+                      {{ getCreatedDate(user.createdOn) }}
+                      <div class="text-gray-400 text-tiny">
+                        {{ getCreatedTime(user.createdOn) }}
+                      </div>
+                    </div>
+                  </div>
```
jamesgeorge007, line 103: Is the outermost `div` intentional? Can the classes be applied to the nearest child with `flex flex-col`?

packages/hoppscotch-sh-admin/src/pages/users/invited.vue
```diff
@@ -16,64 +16,36 @@
           <div v-if="fetching" class="flex justify-center">
             <HoppSmartSpinner />
           </div>
-          <div v-else-if="error || invitedUsers === undefined">
+          <div
+            v-else-if="
+              error || invitedUsers === undefined || invitedUsers.length === 0
+            "
+          >
             <p class="text-xl">{{ t('users.no_invite') }}</p>
```
jamesgeorge007, line 24: Couldn't find a locale entry corresponding to `users.no_invite`. Could you confirm if it's the intended usage? Also, shouldn't the case of error be handled separately with a different message?

packages/hoppscotch-sh-admin/src/pages/users/invited.vue
```diff
@@ -16,64 +16,36 @@
...
-                    </span>
-                  </div>
-                </td>
-                <td class="py-2 px-3">
-                  <span class="flex items-center">
-                    {{ user?.adminEmail }}
+          <div v-else>
+            <HoppSmartTable :list="newInvitedUsersList" :headings="headings">
+              <template #invitedOn="{ item }">
+                <div class="flex flex-col truncate">
```
jamesgeorge007, line 30: Is `flex flex-col` necessary? Isn't a single column the default layout?

packages/hoppscotch-sh-admin/src/pages/users/invited.vue
```diff
@@ -16,64 +16,36 @@
...
+
+                  <div>
+                    <div class="text-gray-400 text-tiny">
+                      <span>
+                        <span>
+                          {{ getCreatedTime(item.invitedOn) }}
+                        </span>
+                      </span>
                     </div>
                   </div>
```
jamesgeorge007, line 44: Is the wrapper `div` necessary?

packages/hoppscotch-sh-admin/src/pages/users/invited.vue
```diff
@@ -100,4 +72,24 @@ const getCreatedTime = (date: string) => format(new Date(date), 'hh:mm a');
...
+// The new invited users list that is used in the table
+const newInvitedUsersList = computed(() => {
+  return invitedUsers.value?.map((user) => {
+    return {
+      adminUid: user.adminUid,
+      adminEmail: user.adminEmail,
+      inviteeEmail: user.inviteeEmail,
+      invitedOn: user.invitedOn,
+    };
+  });
```
jamesgeorge007, line 85:
```suggestion
  return invitedUsers.value?.map(({ adminUid, adminEmail, inviteeEmail, invitedOn }) => {
    return {
      adminUid,
      adminEmail,
      inviteeEmail,
      invitedOn,
    };
  });
```

packages/hoppscotch-ui/src/components/smart/Table.vue
```diff
@@ -0,0 +1,68 @@
+<template>
+  <div class="overflow-auto rounded-md border border-dividerDark shadow-md">
+    <table class="w-full">
+      <thead>
+        <slot name="head">
+          <tr
+            class="text-secondary border-b border-dividerDark text-sm text-left bg-primaryLight"
+          >
```
jamesgeorge007, line 8: It seems the current instances of usage have the `tr` element with the same set of classes. Should we move it out of the `slot`?

packages/hoppscotch-ui/src/components/smart/Table.vue
```diff
@@ -0,0 +1,68 @@
...
+        </slot>
+      </thead>
+
+      <tbody class="divide-y divide-divider">
+        <slot name="body" :list="list">
+          <tr
+            v-for="(rowData, rowIndex) in list"
+            :key="rowIndex"
+            class="text-secondaryDark hover:bg-divider hover:cursor-pointer rounded-xl"
+            :class="yBorder ? 'divide-x divide-divider' : ''"
```
jamesgeorge007, line 22:
```suggestion
            :class="{ 'divide-x divide-divider' : yBorder }"
```

packages/hoppscotch-ui/src/components/smart/Table.vue
```diff
@@ -0,0 +1,68 @@
...
+          >
+            <th v-for="th in headings" scope="col" class="px-6 py-3">
+              {{ th.label ?? th.key }}
+            </th>
+          </tr>
+        </slot>
+      </thead>
+
+      <tbody class="divide-y divide-divider">
+        <slot name="body" :list="list">
```
jamesgeorge007, line 17: Is it necessary to supply the `list` as a slot prop? Isn't that something the component receives as a prop, and the parent context has access to it?

packages/hoppscotch-ui/src/stories/Table.story.vue
```diff
@@ -0,0 +1,58 @@
...
+    <Variant title="General">
+      <HoppSmartTable :list="list" :headings="headings" />
+    </Variant>
+    <Variant title="Custom">
+      <HoppSmartTable>
+        <template #head>
+          <th>ID</th>
+          <th>Name</th>
+          <th>Members</th>
+          <th>Role</th>
```
jamesgeorge007, line 12:
```suggestion
          <tr
            class="text-secondary border-b border-dividerDark text-sm text-left bg-primaryLight"
          >
          <th v-for="heading in headings" :key="heading.key" scope="col" class="px-6 py-3">{{ heading.label }}</th>
          </tr>
```

jamesgeorge007: Is the `scope` attribute necessary here?

packages/hoppscotch-ui/src/stories/Table.story.vue
```diff
@@ -0,0 +1,58 @@
...
+            <td>Joel</td>
+            <td>10</td>
+            <td>Frontend Engineer</td>
+          </tr>
+          <tr>
+            <td>123456</td>
+            <td>Anwar</td>
+            <td>12</td>
+            <td>Frontend Engineer</td>
+          </tr>
```
jamesgeorge007, line 26:
```suggestion
          <tr v-for="item in list" :key="item.id" class="text-secondaryDark hover:bg-divider hover:cursor-pointer rounded-xl">
            <td v-for="cellHeading in headings" :key="cellHeading.key" class="max-w-40 pl-6 py-1"> {{ item[cellHeading.key] ?? '-' }} </td>
          </tr>
```

packages/hoppscotch-ui/src/stories/Table.story.vue
```diff
@@ -0,0 +1,58 @@
...
+      </HoppSmartTable>
+    </Variant>
+  </Story>
+</template>
+
+<script setup lang="ts">
+import { HoppSmartTable } from "../components/smart"
+
+// Table Headings
+const headings = [
```
jamesgeorge007, line 37:
```suggestion
const headings: CellHeading[] = [
```

Requires importing the type from the component source.

packages/hoppscotch-ui/src/stories/Table.story.vue
```diff
@@ -0,0 +1,58 @@
...
+  { key: "id", label: "ID" },
+  { key: "name", label: "Name" },
+  { key: "members", label: "Members" },
+  { key: "role", label: "Role" },
+]
+
+const list = [
```
jamesgeorge007, line 44: Would it be possible to type this such that the keys are inferred from the value of `key` from the `headings` array above?

packages/hoppscotch-sh-admin/src/pages/teams/index.vue
```diff
@@ -13,24 +13,102 @@
...
+        <div v-else>
+          <HoppSmartTable :list="teamsList">
+            <template #head>
+              <tr
+                class="text-secondary border-b border-dividerDark text-sm text-left bg-primaryLight"
+              >
+                <th class="px-6 py-2">{{ t('teams.id') }}</th>
+                <th class="px-6 py-3">{{ t('teams.name') }}</th>
+                <th class="px-6 py-2">{{ t('teams.members') }}</th>
+                <th class="px-6 py-2"></th>
```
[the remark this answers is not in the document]
JoelJacobStephen: Yes. This is kept empty as the last cell of the row is dedicated to the action menu.

packages/hoppscotch-sh-admin/src/pages/teams/index.vue
```diff
@@ -13,24 +13,102 @@
...
-          @deleteTeam="deleteTeam"
-          class=""
-        />
+        <div v-else-if="teamsList.length == 0" class="px-2">
+          <p class="text-lg">
+            {{ t('teams.no_teams') }}
+          </p>
+        </div>
+
+        <div v-else>
```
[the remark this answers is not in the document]
JoelJacobStephen: We need to also check the case where length of team can be 0. This becomes true in this case.

packages/hoppscotch-sh-admin/src/pages/users/index.vue
JoelJacobStephen: Applying flex to the td element causes few issues. Also the truncation in the span doesn't work as expected when the div is absent.

packages/hoppscotch-sh-admin/src/pages/users/index.vue
JoelJacobStephen: Yes. The last cell is reserved for the action button. Hence a blank cell is given for the header

packages/hoppscotch-sh-admin/src/pages/users/invited.vue
JoelJacobStephen: The entry is present in en.json file in sh-admin repo. I have now created a separate check for just the error case.

packages/hoppscotch-ui/src/components/smart/Table.vue
JoelJacobStephen: In future, we might need to have custom classes for the tr. Hence the tr with the default classes is inside the slot with the scope that it can be overriden in the parent.

packages/hoppscotch-ui/src/components/smart/Table.vue
JoelJacobStephen: We are using it in the parent to render individual items in the list. We are sending the list as a prop and also using it as a slot prop because in future, we might implement features such as filtering and sorting

packages/hoppscotch-ui/src/components/smart/Table.vue
```diff
@@ -0,0 +1,69 @@
...
+<script lang="ts" setup generic="Item extends Record<string, unknown>">
+export type CellHeading = {
+  key: string
+  label?: string
+  preventClick?: boolean
+}
+
+defineProps<{
+  /** Whether to show the vertical border between columns */
+  yBorder?: boolean
```
jamesgeorge007, line 57: Do we follow a convention for naming boolean props? Something like prefixing it with `include/show`?

packages/hoppscotch-sh-admin/src/pages/users/invited.vue
```diff
@@ -100,4 +69,26 @@ const getCreatedTime = (date: string) => format(new Date(date), 'hh:mm a');
 // Get Invited Users
 const { fetching, error, data } = useQuery({ query: InvitedUsersDocument });
 const invitedUsers = computed(() => data?.value?.admin.invitedUsers);
+
+// The new invited users list that is used in the table
+const newInvitedUsersList = computed(() => {
```
jamesgeorge007, line 74: We could perhaps rename it to better suit the purpose. Something along the lines of `invitedUsersListToRender`.

packages/hoppscotch-sh-admin/src/pages/users/invited.vue
```diff
@@ -16,64 +16,33 @@
...
-                  <div>
-                    <span class="truncate">
-                      {{ user?.adminUid }}
-                    </span>
-                  </div>
-                </td>
-                <td class="py-2 px-3">
-                  <span class="flex items-center">
-                    {{ user?.adminEmail }}
+          <div v-else>
```
jamesgeorge007, line 30: We can probably swap the order of these elements so the truthy check comes above, and the empty state goes as a fallback. Also, preferably let's use the same property instead of referring to both `invitedUsers` and `newInvitedUsers` (also a good candidate for renaming).


```vue
<div v-if="newInvitedUsersList?.length">
  <HoppSmartTable ...>
  ...
  </HoppSmartTable>
</div>

<!-- Empty state  -->
<div v-else>
</div>
```

packages/hoppscotch-sh-admin/src/pages/users/invited.vue
```diff
@@ -100,4 +69,26 @@ const getCreatedTime = (date: string) => format(new Date(date), 'hh:mm a');
...
+  return invitedUsers.value?.map(
+    ({ adminUid, adminEmail, inviteeEmail, invitedOn }) => {
+      return {
+        adminUid,
+        adminEmail,
+        inviteeEmail,
+        invitedOn,
+      };
+    }
+  );
```
jamesgeorge007, line 84: Shouldn't there be a fallback to `[]` in case `invitedUsers` itself is falsy?

packages/hoppscotch-sh-admin/src/pages/users/invited.vue
jamesgeorge007: On second thought, do we need this computed property? Can't we supply `invitedUsersList` straightaway?

packages/hoppscotch-sh-admin/src/pages/teams/index.vue
```diff
@@ -12,25 +12,92 @@
...
+                <span
+                  class="flex items-center ml-4 truncate"
+                  :class="{ truncate: team.name }"
+                >
+                  {{ team.name ?? t('teams.unnamed') }}
+                </span>
+              </td>
+
+              <td class="py-4 px-3">
+                <span class="ml-7">
```
jamesgeorge007, line 56: Instead of the custom margin on the left for the span elements, would it be possible to adjust it via padding on the left from the parent `td`?

packages/hoppscotch-sh-admin/src/pages/users/index.vue
```diff
@@ -22,27 +22,141 @@
...
-          :usersList="usersList"
-          :fetching="fetching"
-          :error="error"
-          @goToUserDetails="goToUserDetails"
-          @makeUserAdmin="makeUserAdmin"
-          @makeAdminToUser="makeAdminToUser"
-          @deleteUser="deleteUser"
-        />
+        <div v-else-if="usersList.length">
+          <HoppSmartTable :list="usersList">
```
jamesgeorge007, line 32: Unnecessary `div` wrapper; can move the check to the component.

packages/hoppscotch-sh-admin/src/pages/users/index.vue
```diff
@@ -22,27 +22,141 @@
...
+                  </div>
+                  <div v-else class="flex items-center space-x-3">
+                    <span> {{ t('users.unnamed') }} </span>
+                    <span
+                      v-if="user.isAdmin"
+                      class="text-xs font-medium px-3 py-0.5 rounded-full bg-green-900 text-green-300"
+                    >
+                      {{ t('users.admin') }}
+                    </span>
+                  </div>
```
jamesgeorge007, line 81:
```suggestion
                  <div class="flex items-center" :class="user.displayName ? 'space-x-2' : 'space-x-3'">
                    <span>
                      {{ user.displayName ?? t('users.unnamed') }}
                    </span>

                    <span
                      v-if="user.isAdmin"
                      class="text-xs font-medium px-3 py-0.5 rounded-full bg-green-900 text-green-300"
                    >
                      {{ t('users.admin') }}
                    </span>
                  </div>
```

jamesgeorge007: I noticed there was different spacing applied based on the availability of `user.displayName`. It'd be better to keep it consistent and stick with a single-spacing utility class. Please try it and see which one fits in better.

packages/hoppscotch-sh-admin/src/pages/users/index.vue
```diff
@@ -22,27 +22,141 @@
...
+                </td>
+
+                <td class="py-2 px-3">
+                  {{ user.email }}
+                </td>
 
-        <div v-else class="flex justify-center">{{ t('users.no_users') }}</div>
+                <td class="py-2 px-3">
+                  <div class="flex flex-col">
```
jamesgeorge007, line 89:
```suggestion
                  <div>
```

Isn't a single column the default layout? Could you verify it the behavior remains the same without the classes applied?

packages/hoppscotch-sh-admin/src/pages/users/index.vue
```diff
@@ -22,27 +22,141 @@
...
+                            :icon="IconUserMinus"
+                            :label="t('users.remove_admin_status')"
+                            class="!hover:bg-emerald-600"
+                            @click="
+                              () => {
+                                makeAdminToUser(user.uid);
+                                hide();
+                              }
+                            "
+                          />
```
jamesgeorge007, line 134:
```suggestion
                          <HoppSmartItem
                            :icon="user.isAdmin ? IconUserMinus : IconUserCheck"
                            :label="user.isAdmin ? t('users.remove_admin_status') : t('users.make_admin')"
                            class="!hover:bg-emerald-600"
                            @click="
                              () => {
                               user.isAdmin ? makeAdminToUser(user.uid) : makeUserAdmin(user.uid);
                                hide();
                              }
                            "
                          />
```

packages/hoppscotch-sh-admin/src/pages/users/index.vue
```diff
@@ -22,27 +22,122 @@
...
-        <div v-else class="flex justify-center">{{ t('users.no_users') }}</div>
+              <td class="py-2 px-7">
+                {{ user.email }}
+              </td>
+
+              <td class="py-2 px-7">
+                {{ getCreatedDate(user.createdOn) }}
+                <div class="text-gray-400 text-tiny">
+                  {{ getCreatedTime(user.createdOn) }}
+                </div>
```
jamesgeorge007, line 83:
```suggestion
                <span class="text-gray-400 text-tiny">
                  {{ getCreatedTime(user.createdOn) }}
                </span>
```

Will a `span` work in this case? The table cell contents only need to be an inline element, right?

packages/hoppscotch-sh-admin/src/pages/users/invited.vue
```diff
@@ -12,69 +12,34 @@
...
+        >
+          <template #invitedOn="{ item }">
+            <div v-if="item" class="pr-2 truncate">
+              <span class="truncate">
+                {{ getCreatedDate(item.invitedOn) }}
+              </span>
+
+              <div class="text-gray-400 text-tiny">
+                {{ getCreatedTime(item.invitedOn) }}
+              </div>
```
jamesgeorge007, line 36: Can we make the above element `span` to be consistent with the sibling?

```suggestion
              <span class="text-gray-400 text-tiny">
                {{ getCreatedTime(item.invitedOn) }}
              </span>
```

packages/hoppscotch-sh-admin/src/pages/users/invited.vue
```diff
@@ -12,69 +12,34 @@
...
+            </div>
+            <span v-else> - </span>
+          </template>
+        </HoppSmartTable>
+
+        <p v-else class="text-lg">{{ t('users.no_invite') }}</p>
```
jamesgeorge007, line 42:
```suggestion
        <div v-else class="text-lg">{{ t('users.no_invite') }}</div>
```

Since it's a short message, I think we can keep using `div`; also because it has been the convention followed.

packages/hoppscotch-sh-admin/src/pages/teams/index.vue
```diff
@@ -12,25 +12,90 @@
...
+                  </tippy>
+                </div>
+              </td>
+            </tr>
+          </template>
+        </HoppSmartTable>
+
+        <p v-else class="px-2 text-lg">
+          {{ t('teams.no_teams') }}
+        </p>
```
jamesgeorge007, line 95:
```suggestion
        <div v-else class="px-2 text-lg">
          {{ t('teams.no_teams') }}
        </div>
```

Since it's a short message, I think we can keep using `div`; also because it has been the convention followed.

packages/hoppscotch-sh-admin/src/pages/users/invited.vue
```diff
@@ -12,69 +12,34 @@
...
-              </tr>
-            </tbody>
-          </table>
+        <div v-if="fetching" class="flex justify-center">
+          <HoppSmartSpinner />
         </div>
+
+        <p v-else-if="error" class="text-xl">
+          {{ t('users.invite_load_list_error') }}
+        </p>
```
jamesgeorge007, line 21:
```suggestion
        <div v-else-if="error" class="text-xl">
          {{ t('users.invite_load_list_error') }}
        </div>
```

Since it's a short message, I think we can keep using `div`; also because it has been the convention followed.

jamesgeorge007 (commented):
Looks good. A few comments and resolving the conflicts and then we can :shipit:

packages/hoppscotch-sh-admin/src/pages/users/index.vue
JoelJacobStephen: The created time should be in a new line. Replacing it with a span will result in both the date and time to be in the same line.

packages/hoppscotch-sh-admin/src/pages/users/invited.vue
JoelJacobStephen: The created time should be in a new line. Replacing it with a span will result in both the date and time to be in the same line.

jamesgeorge007 (approved):

packages/hoppscotch-sh-admin/src/pages/users/index.vue
```diff
@@ -22,27 +22,122 @@
...
+              :key="user.uid"
+              class="text-secondaryDark hover:bg-divider hover:cursor-pointer rounded-xl"
+              @click="goToUserDetails(user.uid)"
+            >
+              <td class="py-2 px-7 max-w-30">
+                <div class="flex">
+                  <span class="truncate">
+                    {{ user.uid }}
+                  </span>
+                </div>
```
jamesgeorge007, line 57: Is the `div` wrapper necessary here?

packages/hoppscotch-sh-admin/src/pages/users/index.vue
```diff
@@ -22,27 +22,122 @@
...
+                    {{ user.displayName ?? t('users.unnamed') }}
+                  </span>
+
+                  <span
+                    v-if="user.isAdmin"
+                    class="text-xs font-medium px-3 py-0.5 rounded-full bg-green-900 text-green-300"
+                  >
+                    {{ t('users.admin') }}
+                  </span>
+                </div>
```
jamesgeorge007, line 72: The `div` here is intentional, right? Came across [this](https://stackoverflow.com/questions/1110915/is-a-div-inside-a-td-a-bad-idea#:~:text=Using%20a%20div%20instide%20a,the%20elements%20will%20be%20sized.) stack overflow thread about using `div`s in `td`.

packages/hoppscotch-sh-admin/src/pages/users/invited.vue
```diff
@@ -12,71 +12,34 @@
...
-                </td>
-              </tr>
-            </tbody>
-          </table>
+        <div v-else-if="error" class="text-xl">
+          {{ t('users.invite_load_list_error') }}
         </div>
+
+        <HoppSmartTable
+          v-if="invitedUsers?.length"
```
jamesgeorge007, line 24:
```suggestion
          v-else-if="invitedUsers?.length"
```

Shouldn't this be `v-else-if`. `invitedUsers` won't yield anything during the loading state or on the course of an error, right?

packages/hoppscotch-sh-admin/src/pages/users/index.vue
JoelJacobStephen: Yes. It is used to contain the name and the admin badge together and center it vertically. Can you suggest a way of implementing this without a parent div while keep the style intact? I tried doing it without the div but not able to get the same result.

jamesgeorge007: My point was about using a block-level element within `td`. But since `td` itself is a table cell that is sort of an `inline-block` element, and since it fits our use case, let's keep it as it is.

jamesgeorge007 (approved):

PR #3182, opened 2023-07-13
chore: update generateSitemap usage

amk-dev:
**Before**

We were using an workaround with the generateSitemap plugin.
see https://github.com/jbaubree/vite-plugin-pages-sitemap/issues/173 for context.

**After**

thanks to the package authors fixing the need for the workaround in v1.4.5, we have removed the workaround


AndrewBastin (approved):

liyasthomas (approved):
LGTM 💯

PR #3193, opened 2023-07-17
feat: container registry friendly docker images and all-in-one container

AndrewBastin:
### Introduction
This PR intends to do the following things at a high level:
1. Update the frontends (`sh-admin` and `selfhost-web`) to have environment variables definable/updatable after build.
2. Introduce more efficient build process for the Self-hostable containers.
3. Introduce the all-in-one (AIO) container that expose the `selfhost-web`, `sh-admin` and `backend` apps in different ports.

The high-level end goal of this PR is to have solid container registry friendly containers of Hoppscotch.

### Things changed
- Added root `.dockerignore` to ignore all `node_modules` folders recursively as well.
- Updated frontend environment variables to use `APP_` prefix instead of `VITE_`.
- Introduce `@import-meta-env/cli`, `@import-meta-env/unplugin` as dev dependencies to `sh-admin` and `selfhost-web`.
- Use `@import-meta-env/unplugin` vite plugin to do environment variable processing.
- Added placeholder script in entry point `index.html` files in `sh-admin` and `selfhost-web` for `@import-meta-env/cli` to replace environment variables after build.
- Introducing `prod.Dockerfile` that builds production (container registry friendly) docker containers for the app.
- `prod-Dockerfile` defines 4 final target images, `app`, `backend`, `sh_admin` and `aio`.
- Introduce `prod_run.mjs` for `sh-admin` and `selfhost-web` that will be executed on runtime to set environment variables during the run of `app` and `sh_admin` containers.
- Introduce `aio_run.mjs` for `aio` container to execute during its runtime.

### Why is this still a draft ?
There are a couple of things here that I do feel we can still improve upon after I get initial feedback and this PR is fully open:
- Improvements in the dev builds which we can derive from the prod build process.
- Reducing the container size particularly for the `backend` and `aio` containers.
	- ~There should be a PR eventually to bump Prisma version which will allow us to move into `node:18-alpine` instead of `node:18-bookworm`.~ [Done]
	- We need to look into using `pnpm deploy` to extract the backend project out so the `backend` container doesn't need to have the entire monorepo (along with the deps) in its image.
- There needs to be more documentation on the Dockerfile and the new scripts to make it clearer how this works.
- ~The AIO container doesn't deal with termination really well along with situations when either the caddy server or the backend crashes.~ [This has been fixed]
- There is a weird behaviour with `prod_run.mjs` (both of them) and `aio_run.mjs` where the parsing of the placeholder messes up due to quotes in the .env file. If you run into this issue, just try removing the quotes in your .env entries, but I think we need to solve this a bit more gracefully.

### Things that may break/change
- ~Environment variable defined with the `VITE_` will definitely be broken. We can look into impact assumptions, mitigations or to reverse this decision before opening this PR from draft status.~ [This has been reverted]

### How to test this ?
- Build the target from the `prod.Dockerfile` dockerfile
```bash
# assuming CWD is the root of the repo
# target can be "app", "sh_admin", "backend"

docker build -t <target> --target <target> -f prod.Dockerfile .
```
- Start the container with `docker run` (Example, here is the command for `aio`)
```bash
# Look into prod.Dockerfile for what ports are required
docker run -p 3000:3000 -p 3100:3100 -p 3170:3170 --env-file .env aio
```

### Checks


Brulemat:
In companies it is common to have a specific CA certificate and a nexus for the registry. it is interesting to be able to overload these points by adding in the dockerfile for example


```
ENV ENTERPRISE_REGISTRY=https://registry.npmjs.org/
ENV NODE_EXTRA_CA_CERTS="/usr/local/share/ca-certificates/internal-ca.crt"
COPY ENTERPRISE-CA.crt /usr/local/share/ca-certificates/internal-ca.crt
RUN update-ca-certificates
RUN npm set registry $ENTERPRISE_REGISTRY
```

AndrewBastin:
@Brulemat once the containers are built (we will be providing those in a container registry), there will be no requirement to fetch additional stuff from npm or anything. So I don't exactly see a point here, unless you are building the container on premise or something like that.

ft6502:
@AndrewBastin : The connection to the OAuth-Provider or the database would be an example. If these are running on company servers, then usually using internal certificates.

ft6502:
@AndrewBastin : It looks like you already put a lot of work into this. Cool! May I suggest to use https://github.com/GoogleContainerTools/distroless containers for the following reasons:

-  You can use a "non root" image. A real security benefit from my point of view
- They are small
- They don't contain things like shells/deployment-tools etc. So if an attacker breaks out, his life is a bit harder

I'm an JS-novice but quite good at docker/k8s/shell. If you are interested I could try to help

webysther:
Will be nice make a better version than mine that don't need build on-the-fly, now with less than 4GB is possible to use the free actions which in the past with 16GB was impossible.
After all, congratulations, I hope reduce the curve to put this project to work.

webysther:
I also suggest normalize the names to use `app`, `api` and `admin`, makes more sense.

webysther:
Make mostly env vars not required and build on-the-fly using base values defined in few VARs like I do will simplify the addoption also.

webysther:
> @AndrewBastin : It looks like you already put a lot of work into this. Cool! May I suggest to use https://github.com/GoogleContainerTools/distroless containers for the following reasons:
> 
> * You can use a "non root" image. A real security benefit from my point of view
> * They are small
> * They don't contain things like shells/deployment-tools etc. So if an attacker breaks out, his life is a bit harder
> 
> I'm an JS-novice but quite good at docker/k8s/shell. If you are interested I could try to help

Yes, rootless is a thing for containers. But I think healthcheck need to be put in the radar to make sure the start without problem.

webysther:
> In companies it is common to have a specific CA certificate and a nexus for the registry. it is interesting to be able to overload these points by adding in the dockerfile for example
> 
> ```
> ENV ENTERPRISE_REGISTRY=https://registry.npmjs.org/
> ENV NODE_EXTRA_CA_CERTS="/usr/local/share/ca-certificates/internal-ca.crt"
> COPY ENTERPRISE-CA.crt /usr/local/share/ca-certificates/internal-ca.crt
> RUN update-ca-certificates
> RUN npm set registry $ENTERPRISE_REGISTRY
> ```

Maybe creating a self-signed on-the-fly in a folder and if you need a cert mount using bind with you certs.

webysther:
Another tip, lots of security problems with lib used on NPM, to reduce the surface copy from another cnotainer is a gold standard now (this mean I'm stupid lazy to do this in my image), but consider check the image before make public available:

image

prod.Dockerfile
```diff
@@ -0,0 +1,57 @@
...
+EXPOSE 8080
+CMD ["/bin/sh", "-c", "node /usr/prod_run.mjs && caddy run --config /etc/caddy/Caddyfile --adapter caddyfile"]
+
+FROM backend as aio
+RUN apt-get update
+RUN apt-get install -y caddy
+RUN npm install -g @import-meta-env/cli
+COPY --from=fe_builder /usr/src/app/packages/hoppscotch-selfhost-web/dist /site/selfhost-web
+COPY --from=sh_admin_builder /usr/src/app/packages/hoppscotch-sh-admin/dist /site/sh-admin
+COPY aio.Caddyfile /etc/caddy/Caddyfile
```
balub, line 53: Maybe we can add a healthcheck before `CMD` to check if all our services are functioning properly. We could have a script we run that checks everything and call that script in the `HEALTHCHECK` stage.

balub:
image
We will have to look into the quotes issue, the magic link logic fails currently because of it.

balub:
Screenshot 2023-08-17 at 12 18 56 PM
I think we need to explicitly state the `PORT` and "PRODUCTION` env variables because this goes away if you do.

balub (approved):

PR #3227, opened 2023-08-02
refactor: fonts are now bundled with packages

liyasthomas:
### Major changes

1. Migrated to [unplugin-fonts](https://github.com/cssninjaStudio/unplugin-fonts)

	> vite-plugin-fonts package has been deprecated in favor of unplugin-fonts.

2. Migrated to [Fontsource](https://fontsource.org/docs/getting-started/introduction)

	> Fontsource is a collection of open-source fonts that are packaged into individual NPM packages for self-hosting in your web applications.

3. Migrated all fonts to [variable fonts](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_fonts/Variable_fonts_guide)

	> Variable fonts allow developers to use a single font file that contains multiple variations of the same font, such as different weights, widths, and styles. This can result in smaller file sizes and more flexible design options.

4. Migrated to [Material Symbols](https://fonts.google.com/icons)

	> Material Symbols is the successor to Material Icons in a single font file with a wide range of design variants.

### Summary

This pull request adds support for variable fonts in the Hoppscotch web app and its packages. It replaces the `vite-plugin-fonts` package with the `unplugin-fonts` package and adds font source variable packages as dependencies. It also updates the SCSS files to use variable fonts instead of static fonts and imports the font source variables in the relevant TypeScript files.


packages/hoppscotch-ui/src/plugin.ts
```diff
@@ -1,8 +1,5 @@
 import type { Plugin, App } from "vue"
```
anwarulislam, line 2: We need this `virtual:windi` and `style.scss` imported in `plugin.ts`

```
import "./assets/scss/styles.scss"
import "virtual:windi.css"
```

anwarulislam (changes requested):

anwarulislam (approved):
LGTM 🚀

JoelJacobStephen (approved):
LGTM!!

PR #3276, opened 2023-08-23
fix: vertical scroll not working on codemirror instance when lines are not wrapped

amk-dev:
fixs HFE-149

**Before**

Vertical scroll was not working when line wrap is disabled on codemirror instances.

**After**

Now the scroll works as intented. recently we introduced a `overscroll-bahaviour: none` to the wildcard selector ( * ). this change was causing the trouble with the codemirror instances. so we override this rule for the `.cm-scroller`. this way the codemirror instances are not affected by the `overscroll-bahaviour: none` anymore.


liyasthomas (approved):
LGTM 💯

AndrewBastin (approved):

PR #3280, opened 2023-08-23
fix: fix issues in displaying the suggestions menu on EnvInput

amk-dev:
fixs HFE-148

**Before**

Suggestions menu was not displayed when the EnvInput was focused. 

We had an `onClickOutside` handler defined for suggestionsMenu, it hides the suggestions menu. 
But the input div was outside of this suggestions menu, causing the onClickOutside handler to run everytime someone clicked on the input div, which infact hid the suggestions menu.
 
**After**

Suggestions menu is displayed properly. moved the onClickOutside to target the `autoCompleteWrapper` instead of the suggestions menu.


nivedin (approved):
LGTM 💯

PR #3293, opened 2023-08-25
fix: fix issues with codegen

amk-dev:
Closes HFE-150

**Before**

Generating code for requests were breaking. the package we were using `httpsnippet` was depending on `ajv@6`, but they didnt have it on their peerDependencies, it was picking up ajv@8 which are added by some other packages, which breaks `httpsnippet`.

**After**

Fixed it for now with pnpm's `pnpm.packageExtensions` which patches httpsnippets package.json to include the ajv@6 in peerDeps.

After the fix, codegeneration worked normally except for when there are no body. we set `HAR.Request.postData` as optional. but `httpsnippet` was considering it mandatory. to fix this, we use `x-unknown` as the `postData.mimeType` when the contentType is unknown.


liyasthomas (approved):
LGTM 💯

PR #3296, opened 2023-08-27
feat: smaller container images

AndrewBastin:
This PR intends to reduce the size of the container images (mostly the backend container and because of that, the new AIO container). In my builds locally, the container image sizes have around 90% size reduction.

### Description
This PR uses the [`pnpm deploy`](https://pnpm.io/cli/deploy) command to isolate the backend build into a full independent package which place in a separate fresh build stage to run.

A big thing to note here is to reduce size, the new backend container (and hence the AIO container) does not have pnpm, so you will need to use `npm` and `npx` for any package relating stuff (for example, DB migration runs).

### Checks
- [ ]


prod.Dockerfile
```diff
@@ -12,17 +12,25 @@ COPY . .
...
 
 
-FROM base_builder as backend
+FROM base_builder as backend_builder
 WORKDIR /usr/src/app/packages/hoppscotch-backend
-RUN pnpm exec prisma generate
-RUN pnpm run build
+RUN pnpm --filter=hoppscotch-backend deploy /usr/build/backend
+
+FROM node:18-alpine3.16 as backend
```
balub, line 19:
```suggestion
FROM node:18-alpine3.19 as backend
```

prod.Dockerfile
```diff
@@ -12,17 +12,25 @@ COPY . .
...
 # Remove the env file to avoid backend copying it in and using it
-RUN rm "../../.env"
+RUN rm ".env"
 ENV PRODUCTION="true"
 ENV PORT=3170
 ENV APP_PORT=${PORT}
 ENV DB_URL=${DATABASE_URL}
-CMD ["pnpm", "run", "start:prod"]
+CMD ["node", "dist/main.js"]
 EXPOSE 3170
```
balub, line 34: Needs to be changed to accommodate changed introduced in subpath.

balub (changes requested):

prod.Dockerfile
AndrewBastin: updated as of latest commit

PR #3297, opened 2023-08-28
fix: graphql operation highlight issue on focus changed

anwarulislam:
Closes HFE-191

### Description
This PR fixes the issue of highlighting GraphQL Operation.
When multiple operations are in the same Tab we can highlight the operation with a cursor to run. If we run the request we basically defocused the cursor from the query editor. And after focusing on the query editor it doesn't update the operation name for the button. This PR fixed that issue.

- [ ] Not Completed
- [x] Completed

### Checks


JoelJacobStephen (approved):
LGTM!!

PR #3318, opened 2023-09-05
chore(fix/devcontainer): make devcontainer copy .env.example

SamJakob:
### Description
Make devcontainer copy `.env.example` instead of moving it.

This prevents the example from getting destroyed when the container is booted.

### Checks

- [x] My code requires **no** changes to the documentation
- [x] I have updated the documentation as required (n/a)

### Additional Information
This is useful when the devcontainer is used in a workspace or GitHub Codespace as it means that, when a commit is subsequently made, it does not destroy the `.env.example` file.


AndrewBastin (approved):
lgtm

Issue #3301, opened 2023-08-30
[bug]: Problems when importing OpenAPI/Swagger documents


JoeMarkov:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When I import a new collection from an existing swagger/openapi document, the host and basepath seem to be ignored. 

My dream is be able to import the following swagger document below and then be directly able to execute queries against it. 

As a new user, I found it very hard to figure out how to manually set the baseurl, I would have hoped that it used the host in the document as a default "baseurl". 

then the basepath is also ignored, so the URL it tries to access is wrong. 

I am doing this in a teacher/trainer setup.

The document:

```
swagger: '2.0'
info:
  version: '1.0'
  title: Customer API
  description: API for the www.myserver.com customer API
basePath: "/api"
host: www.myserver.com
schemes:
- http
consumes:
- application/json
produces:
- application/json
paths:
  "/Customers":
    get:
      operationId: getAllCustomers
      summary: Access all the customers in the database
      responses:
        '200':
          description: Successful Response
          schema:
            type: array
            items:
              "$ref": "#/definitions/customer"
          examples:
            application/json:
            - Id: 1
              FirstName: Muriel
              LastName: Cook
              StreetAddress: 22523 Willison Street
              City: Minneapolis
              ZipCode: '55415'
            - Id: 2
              FirstName: Kenny
              LastName: Taylor
              StreetAddress: 493 Sarah Drive
              City: Lake Charles
              ZipCode: '70601'
            - Id: 3
              FirstName: Patrick
              LastName: Masi
              StreetAddress: 932 White Lane
              City: Macon
              ZipCode: '31201'
  "/Customers/{customerid}":
    get:
      operationId: getCustomerById
      summary: Find Customer by ID
      description: The ID of the Customer
      parameters:
      - "$ref": "#/parameters/customerid"
      responses:
        '200':
          description: Successful Response
          schema:
            type: object
            items:
              "$ref": "#/definitions/customer"
          examples:
            application/json:
              Id: 1
              FirstName: Muriel
              LastName: Cook
              StreetAddress: 22523 Willison Street
              City: Minneapolis
              ZipCode: '55415'
definitions:
  customer:
    title: customer
    type: object
    properties:
      Id:
        type: integer
        description: Customer ID
        format: int32
      FirstName:
        type: string
        example: Tore
        description: The first name of the customer
      LastName:
        type: string
        example: Nestenius
        description: The last name of the customer
[...]

PR #3323, opened 2023-09-06
fix: missing baseurl on import openapi

anwarulislam:
Closes HFE-205 #3301 

### Description
When a collection imported from OpenAPI/Swagger server URL is missing in every request. This PR fixes that issue.

- [ ] Not Completed
- [x] Completed

### Checks

### DIFF
--- a/packages/hoppscotch-common/src/helpers/import-export/import/openapi.ts
+++ b/packages/hoppscotch-common/src/helpers/import-export/import/openapi.ts
@@ -195,7 +195,7 @@ const parseOpenAPIV3Body = (
   // We only take the first definition
   const [contentType, media]: [
     string,
-    OpenAPIV3.MediaTypeObject | OpenAPIV31.MediaTypeObject
+    OpenAPIV3.MediaTypeObject | OpenAPIV31.MediaTypeObject,
   ] = objs[0]
 
   return contentType in knownContentTypes
@@ -533,7 +533,7 @@ const parseOpenAPIUrl = (
    * Relevant v3 reference: https://swagger.io/specification/#server-object
    **/
   if (objectHasProperty(doc, "servers")) {
-    return doc.servers?.[0].url ?? ""
+    return doc.servers?.[0].url ?? "<<baseUrl>>"
   }
 
   // If the document is neither v2 nor v3 then return a env variable as placeholder


packages/hoppscotch-common/src/helpers/import-export/import/openapi.ts
```diff
@@ -514,6 +514,17 @@ const parseOpenAPIAuth = (
...
+  let url = ""
+  if (objectHasProperty(doc, "host")) {
+    url += doc.host
+  }
+  if (objectHasProperty(doc, "basePath")) {
+    url += doc.basePath
+  }
+  return url
+}
+
```
AndrewBastin, line 527: There are differences to how OpenAPI v3 and OpenAPI v2 (Swagger) deals with the URL stuff.

Please see the differences here:
- Relevant v3 reference: https://swagger.io/specification/#server-object
- Relevant v2 reference: https://swagger.io/specification/v2/#:~:text=to%20be%20obscured.-,Schema,-Swagger%20Object

This implementation only seems to be handling the V2 case and then blanking out the base URL part in v3.

AndrewBastin (changes requested):

packages/hoppscotch-common/src/helpers/import-export/import/openapi.ts
```diff
@@ -514,6 +514,32 @@ const parseOpenAPIAuth = (
...
+    return `${doc.host}${doc.basePath}`
+  }
+
+  /**
+   * OpenAPI V3 has version as a string in the document's openapi property.
+   * And host and basePath are in the document's servers property.
+   * Relevant v3 reference: https://swagger.io/specification/#server-object
+   **/
+  if (objectHasProperty(doc, "servers")) {
+    return doc.servers?.[0].url ?? ""
```
AndrewBastin, line 536: Revert this to `<<baseUrl>>` as well since we don't know the base url if this is not given.

AndrewBastin (changes requested):

packages/hoppscotch-common/src/helpers/import-export/import/openapi.ts
AndrewBastin: Resolved in 3d183dc

AndrewBastin (approved):

PR #3314, opened 2023-09-05
fix: keyboard shortcut for help and support

liyasthomas:
Changed the keyboard shortcut for opening the support modal to `shift-?` in `keybindings.ts`.

PR #3332, opened 2023-09-08
refactor: keyboard shortcuts now supports different keyboard layouts including Dvorak

### LINKED ISSUE #3322: [bug]: CMD-C on Dvorak keyboard causes Reset Request

### Current behavior

The "c" key in the Dvorak keyboard layout is in the position of the "i" key in the QWERTY layout. For some reason when I use a Dvorak layout and press CMD-C, it sends a CMD-I shortcut to Hoppscotch and resets the request. Other shortcuts work as expected.

Maybe this should be filed as a separate bug, but resetting a request cannot be undone, meaning that I lose all my work if I forget about this bug and try to copy some text. If it is difficult to implement undo for this action, I suggest this shortcut be removed since it's such a footgun.

### Steps to reproduce

1. Use a Dvorak layout (on Mac OS)
2. Select some text in the body of a request
3. Press CMD-C

 I've tested this in Firefox, Chrome, and the PWA, but not in other operating systems.

### Environment

Production

### Version

Cloud

JoelJacobStephen:
### Ticket

- Closes HFE-219
- Fixes #3322
- Suppressed #3314

### Description
This PR focuses on allowing the keyboard shortcuts in the app to support different keyboard layouts by using the unique key code parameter. This has been tested with QWERTY and DVORAK Keyboard Layouts and it works as expected.

### Objectives
- [x] Keyboard shortcuts should support DVORAK layout 
- Example: `CMD/CTRL - C` used to reset the request when using DVORAK Layout which has now been resolved. Only `CMD/CTRL - I` resets the request connection on both QWERTY and DVORAK Layouts

### Checks

### DIFF
--- a/packages/hoppscotch-common/src/helpers/keybindings.ts
+++ b/packages/hoppscotch-common/src/helpers/keybindings.ts
@@ -56,7 +56,7 @@ export const bindings: {
   "alt-x": "request.method.delete",
   "ctrl-k": "modals.search.toggle",
   "ctrl-/": "flyouts.keybinds.toggle",
-  "?": "modals.support.toggle",
+  "shift-/": "modals.support.toggle",
   "ctrl-m": "modals.share.toggle",
   "alt-r": "navigation.jump.rest",
   "alt-q": "navigation.jump.graphql",
@@ -120,7 +120,8 @@ function generateKeybindingString(ev: KeyboardEvent): ShortcutKey | null {
 }
 
 function getPressedKey(ev: KeyboardEvent): Key | null {
-  const val = ev.key.toLowerCase()
+  const val = ev.code.toLowerCase()
+
   // Check arrow keys
   if (val === "arrowup") return "up"
   else if (val === "arrowdown") return "down"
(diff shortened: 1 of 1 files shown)


liyasthomas (approved):
LGTM 💯

packages/hoppscotch-common/src/helpers/keybindings.ts
```diff
@@ -132,17 +133,16 @@ function getPressedKey(ev: KeyboardEvent): Key | null {
   if (isLetter) return ev.code.toLowerCase().substring(3) as Key
 
   // Check if number keys
-  if (val.length === 1 && !isNaN(val as any)) return val as Key
-
-  // Check if question mark
-  if (val === "?") return "?"
+  const isDigit = ev.code.toLowerCase().startsWith("digit")
+  if (isDigit) return ev.code.toLowerCase().substring(5) as Key
```
jamesgeorge007, line 137:
```suggestion
  const isDigit = val.startsWith("digit")
  if (isDigit) return val.substring(5) as Key
```

packages/hoppscotch-common/src/helpers/keybindings.ts
```diff
@@ -132,17 +133,16 @@ function getPressedKey(ev: KeyboardEvent): Key | null {
   if (isLetter) return ev.code.toLowerCase().substring(3) as Key
```
jamesgeorge007, line 133:
```suggestion
  if (isLetter) return val.substring(3) as Key
```

jamesgeorge007: Can't add the suggestion for the line above `132`.

jamesgeorge007 (commented):
Can't we use the `val` variable initialized at the top at all places?

packages/hoppscotch-common/src/helpers/keybindings.ts
JoelJacobStephen: @jamesgeorge007  I have made the changes.

AndrewBastin (approved):
lgtm 💯

PR #3351, opened 2023-09-13
fix: sticky searchbar hidden in codemirror

### LINKED ISSUE #3240: [feature]: search bar on request response always present when scrolling

### Summary

Currently, when trying to search for something, you press CTRL + F, and it works fine, but the problem is that the search bar stays at the top of the request response, and does not follow the scroll, so it's a bit annoying, because I can't use the search controls at the bottom of the request response.

It would be nice if even if you scroll the search bar is always visible.

https://github.com/hoppscotch/hoppscotch/assets/56084970/c81349b0-8b2e-46a0-9e80-9583ec3def1e

For better UX, as there are cases where we need to look for something in what is far below the request response.

nivedin:
Closes #3240 
Closes HFE-211

### Description
This PR fixes the issue in code mirror instances where the search panel gets hidden when scrolling because of the sticky nature of the components above it.

### Checks


liyasthomas (approved):
LGTM 💯

packages/hoppscotch-common/src/components/lenses/renderers/JSONLensRenderer.vue
```diff
@@ -86,7 +86,14 @@
...
       </div>
     </div>
-    <div ref="jsonResponse" class="flex flex-col flex-1 h-auto h-full"></div>
+    <div
+      ref="jsonResponse"
+      class="flex flex-col flex-1 h-auto h-full"
+      :class="{
+        responseToggleOff: !toggleFilter,
+        responseToggleOn: toggleFilter,
+      }"
```
jamesgeorge007, line 95:
```suggestion
      :class="toggleFilter ? 'responseToggleOn' : 'responseToggleOff'"
```

How about leveraging ternary here?

packages/hoppscotch-common/src/composables/codemirror.ts
```diff
@@ -221,6 +221,7 @@ export function useCodemirror(
     if (selection) {
       const from = selection.from
       const to = selection.to
```
jamesgeorge007, line 223:
```suggestion
      const { from, to } = selection
```

PR #3367, opened 2023-09-18
feat: tab service added

anwarulislam:
Closes HFE-207 HFE-208

### Description
- Refactored Tab system with new service architecture. Replaced previous `tab.ts` with tab service
- The selected tab in a REST/GQL page RequestOptions component can be stored in the tab info so that they can be updated.

### Checks


nivedin (changes requested):
https://github.com/hoppscotch/hoppscotch/assets/53208152/ab5a159f-ee27-4a7f-8759-c10c4d3bb7f3

getting this error when switching between tabs with a tab with empty body

packages/hoppscotch-common/src/services/tab/graphql.ts
```diff
@@ -0,0 +1,50 @@
...
+      },
+    })
+
+    this.watchCurrentTabID()
+  }
+
+  public getTabRefWithSaveContext(ctx: HoppGQLSaveContext) {
+    for (const tab of this.tabMap.values()) {
+      // For `team-collection` request id can be considered unique
+      if (ctx && ctx.originLocation === "team-collection") {
```
jamesgeorge007, line 27:
```suggestion
      if (ctx?.originLocation === "team-collection") {
```

packages/hoppscotch-common/src/services/tab/rest.ts
```diff
@@ -0,0 +1,51 @@
...
+      },
+    })
+
+    this.watchCurrentTabID()
+  }
+
+  public getTabRefWithSaveContext(ctx: HoppRESTSaveContext) {
+    for (const tab of this.tabMap.values()) {
+      // For `team-collection` request id can be considered unique
+      if (ctx && ctx.originLocation === "team-collection") {
```
jamesgeorge007, line 27:
```suggestion
      if (ctx?.originLocation === "team-collection") {
```

packages/hoppscotch-common/src/services/tab/rest.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  public getDirtyTabsCount() {
+    let count = 0
+
+    for (const tab of this.tabMap.values()) {
+      if (tab.document.isDirty) count++
+    }
+
+    return count
+  }
```
jamesgeorge007, line 50: Is it possible to abstract the common parts into helpers or something along those lines since `GraphQL` also uses a similar implementation?

Applicable to the `getTabRefWithSaveContext` method as well.

packages/hoppscotch-common/src/services/tab/tab.ts
```diff
@@ -0,0 +1,210 @@
...
+
+  public getTabs(): HoppTab<Doc>[] {
+    return Array.from(this.tabMap.values())
+  }
+
+  public getActiveTab(): HoppTab<Doc> | null {
+    if (!this.currentTabID.value) {
+      return null
+    }
+    return this.tabMap.get(this.currentTabID.value) || null
```
jamesgeorge007, line 81:
```suggestion
    return this.tabMap.get(this.currentTabID.value) || null
```

Is the early return necessary here? Won't failing to retrieve a value yield `undefined` and the expression falls back to `null`?

packages/hoppscotch-common/src/services/tab/tab.ts
```diff
@@ -0,0 +1,210 @@
...
+      .map((tab) => this.getTabRef(tab.id))
+  }
+
+  private generateNewTabID() {
+    while (true) {
+      const id = uuidV4()
+
+      if (!this.tabMap.has(id)) return id
+    }
+  }
```
jamesgeorge007, line 209: Is the `while` loop intentional here?

jamesgeorge007 (changes requested):
I encountered a few errors while navigating between tabs and performing a few actions. Can you look into it?

https://github.com/hoppscotch/hoppscotch/assets/25279263/61c34adc-2527-42e2-ab07-0bf489e72a50

@JoelJacobStephen can give you more context.

packages/hoppscotch-common/src/helpers/collection/collection.ts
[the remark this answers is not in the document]
anwarulislam: Hi, @jamesgeorge007!
@AndrewBastin suggested to use the `getService` in a function for now. That `getService` function is deprecated though. We will eventually convert everything to service architecture.

packages/hoppscotch-common/src/components/lenses/ResponseBodyRenderer.vue
```diff
@@ -107,11 +104,13 @@ watch(
       "results",
     ]
 
+    const { responseTabPreference } = doc.value
+
     if (
-      selectedTabPreference.value &&
-      validRenderers.includes(selectedTabPreference.value)
+      responseTabPreference &&
+      validRenderers.includes(responseTabPreference)
```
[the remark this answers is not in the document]
anwarulislam: `responseTabPreference` can be undefined so that is a type error if we add `undefined` value in includes method

packages/hoppscotch-common/src/services/tab/rest.ts
anwarulislam: `saveContext` is different for each `TabService`. so we need to keep that out from the abstract `TabService`

packages/hoppscotch-common/src/services/tab/tab.ts
anwarulislam: Yes, intentional. Because there is a possibility of having duplicate `uuid` for tab if we don't use while loop

packages/hoppscotch-common/src/services/tab/tab.ts
anwarulislam: It's because of a type error.

packages/hoppscotch-common/src/services/tab/rest.ts
anwarulislam: Each tab service has its own `dirty` mechanism. Also, this is inside a tab `document`. We don't know the document's type.

packages/hoppscotch-common/src/components/lenses/ResponseBodyRenderer.vue
```diff
@@ -29,18 +29,18 @@
...
-          tab.testResults.envDiff.selected.updations.length ||
-          tab.testResults.envDiff.global.updations.length)
+        doc.testResults &&
+        (doc.testResults.expectResults.length ||
+          doc.testResults.tests.length ||
+          doc.testResults.envDiff.selected.additions.length ||
+          doc.testResults.envDiff.selected.updations.length ||
+          doc.testResults.envDiff.global.updations.length)
           ? true
           : false
```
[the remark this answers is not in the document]
jamesgeorge007: Good catch. Can we make it a computed property and possibly use the `Boolean` constructor to convert it to a boolean?

packages/hoppscotch-common/src/components/lenses/ResponseBodyRenderer.vue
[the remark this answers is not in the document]
jamesgeorge007: Got it. Is it beneficial to explicitly cast it in cases like the above? Is it something that we usually do?

packages/hoppscotch-common/src/services/tab/tab.ts
jamesgeorge007: Could you try it again? I wasn't getting a type error when I gave it a shot.

packages/hoppscotch-common/src/services/tab/tab.ts
jamesgeorge007: Would it be possible to check against the current set of IDs associated with other tabs and re-generate if it's found to be a duplicate?

anwarulislam: I think the while loop doing exactly the same. When we return from the loop the loop invalidated, right?

packages/hoppscotch-common/src/components/lenses/ResponseBodyRenderer.vue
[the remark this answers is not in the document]
anwarulislam: Yeah, we usually keep typesafe always.

packages/hoppscotch-common/src/services/tab/tab.ts
jamesgeorge007: Hmm, yes. We'll need a way to repeatedly re-generate UUIds. @AndrewBastin, thoughts?

jamesgeorge007: Is it a good idea to have a recursive call?

jamesgeorge007 (commented):
Could you look into the failing tests?

jamesgeorge007 (commented):
Since the logic is moved under services can we remove [helpers/rest/tab.ts](https://github.com/hoppscotch/hoppscotch/blob/main/packages/hoppscotch-common/src/helpers/rest/tab.ts) and [helpers/graphql/tab.ts](https://github.com/hoppscotch/hoppscotch/blob/main/packages/hoppscotch-common/src/helpers/graphql/tab.ts)?

jamesgeorge007 (approved):
LGTM. Unrelated, but if possible could you see if [helpers/graphql/eq.ts](https://github.com/hoppscotch/hoppscotch/blob/main/packages/hoppscotch-common/src/helpers/graphql/eq.ts) can be removed as well?

PR #3374, opened 2023-09-19
fix: tooltip position in editor instance

liyasthomas:
### <samp>🤖 Generated by Copilot at e8a380a</samp>

This pull request refactors and fixes the tooltip and popover styles across different packages. It moves the common styles to the `hoppscotch-common` package, removes the redundant and conflicting styles from the `hoppscotch-sh-admin` and `hoppscotch-ui` packages, and adjusts the alignment and shadow of the CodeMirror tooltip.

### DIFF
--- a/packages/hoppscotch-common/assets/scss/styles.scss
+++ b/packages/hoppscotch-common/assets/scss/styles.scss
@@ -137,10 +137,10 @@ a {
 
 .cm-tooltip {
   .tippy-box {
-    @apply shadow-none;
+    @apply shadow-none #{!important};
     @apply fixed;
     @apply inline-flex;
-    @apply -mt-8;
+    @apply -mt-7.5;
   }
 }
 
--- a/packages/hoppscotch-sh-admin/assets/scss/styles.scss
+++ b/packages/hoppscotch-sh-admin/assets/scss/styles.scss
@@ -137,14 +137,14 @@ a {
 
 .cm-tooltip {
   .tippy-box {
-    @apply shadow-none;
+    @apply shadow-none #{!important};
     @apply fixed;
     @apply inline-flex;
-    @apply -mt-8;
+    @apply -mt-7.5;
   }
 }
 
-.tippy-box[data-theme~="tooltip"] {
+.tippy-box[data-theme~='tooltip'] {
   @apply bg-tooltip;
   @apply border-solid border-tooltip;
   @apply rounded;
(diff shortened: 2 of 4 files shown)


packages/hoppscotch-common/assets/scss/styles.scss
```diff
@@ -137,10 +137,10 @@ a {
 
 .cm-tooltip {
   .tippy-box {
-    @apply shadow-none;
+    @apply shadow-none #{!important};
     @apply fixed;
     @apply inline-flex;
-    @apply -mt-8;
+    @apply -mt-7.5;
```
JoelJacobStephen, line 143: It is better to set the margin to either 7 or 8 since Tailwind doesn't support 7.5. We are moving back to Tailwind in the next point release and this style won't be applied.

packages/hoppscotch-sh-admin/assets/scss/styles.scss
```diff
@@ -137,14 +137,14 @@ a {
 
 .cm-tooltip {
   .tippy-box {
-    @apply shadow-none;
+    @apply shadow-none #{!important};
     @apply fixed;
     @apply inline-flex;
-    @apply -mt-8;
+    @apply -mt-7.5;
```
JoelJacobStephen, line 143: Margin of 7.5 isn't supported in Tailwind and we are moving to tailwind in the next release. Hence this might not be applied.

packages/hoppscotch-ui/src/assets/scss/styles.scss
```diff
@@ -137,10 +137,10 @@ a {
 
 .cm-tooltip {
   .tippy-box {
-    @apply shadow-none;
+    @apply shadow-none #{!important};
     @apply fixed;
     @apply inline-flex;
-    @apply -mt-8;
+    @apply -mt-7.5;
```
JoelJacobStephen, line 143: Margin of 7.5 isn't supported in Tailwind and we are moving to tailwind in the next release. Hence this style might not be applied.

JoelJacobStephen (changes requested):

packages/hoppscotch-common/assets/scss/styles.scss
liyasthomas: Neither 7 or 8 suits the desired outcome. Hence 7.5. Tailwind do supports custom values like m-[7.5rem], hence migrating to Tailwind won't be an issue.

anwarulislam (approved):
LGTM 🚀

JoelJacobStephen (approved):
LGTM!!

PR #3385, opened 2023-09-22
fix: blank screen in admin dashboard when failing to refresh tokens or when cookies are expired and user not fully logged out in such cases

JoelJacobStephen:
### Ticket
Closes HFE-222

### Description
This PR focuses on fixing an issue in the Admin Dashboard where the user isn't fully logged out ( only the frontend handles the situation without sending the appropriate calls to the backend ) and is either shown the login page or a blank screen when any of the following conditions occur:

- User is Unauthorised
- Cookies are expired
- Attempting to refresh the token throws a 401 error

### Objectives
- [x] Fix the issue where a blank screen is shown when cookies are expired
- [x] Fix the auth issue where user is not logged out when unauthorised or when cookies have expired
- [x] Fix the issue where attempting to refresh the tokens throws a 401 error and results in a blank screen 

### Checks


nivedin (changes requested):
https://github.com/hoppscotch/hoppscotch/blob/6482751986cc13804d1e6b35cdd1f8a3e3db414f/packages/hoppscotch-sh-admin/src/helpers/auth.ts#L152

i think we can remove this signout method from here because it can lead to infinite loop on page refresh when loging out if a token is expired or invalid

nivedin (approved):
LGTM 💯

Issue #3303, opened 2023-08-30
[bug]: Import/Export Environments does not have an option to import from Postman


MustansirM:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

Clicking on the Import/Export button on Collections works as expected, where import options are listed as Hoppscotch, Postman, OpenAPI, among others. If Postman is selected, a further select file section is shown, which allows for selecting a local json file.

However clicking on the corresponding Import/Export button in the Environments sections only lists Import from Hoppscotch. On clicking the option, a local json file can be selected. If the json file has been generated from Postman, it works correctly.

This seems to be a UI/messaging issue, where despite a Postman exported environment file working correctly, it is not listed as an option. Is this meant to be intentional or are there edge cases where a Postman environment json does not work as expected and hence the option is not shown to the user?

If Postman exported files are supported and only the UI is misleading, I can take up that fix.

### Steps to reproduce

1. Go to the Environments section
2. Click on the Import/Export option

### Environment

Production

### Version

Cloud

JoelJacobStephen:
@MustansirM This issue is solved. Can you confirm once on your end so that I can close this issue?

PR #3425, opened 2023-10-05
refactor: revamp the importers & exporters systems to be reused 

amk-dev:
Closes HFE-237,HFE-204,HFE-217
Fixs #3303 

**Changes**

1. ### **revamp the importers and exporters systems to be reused**

   Previously, our import logic was shattered around a lot of places, and duplicated a lot of code. this PR adds a simple consistant way to add importers without leaking the implementation to different places.

   Here are some details to help navigate the PR better.

- We remove duplicating code by using `Import Sources`

  Import sources are the different ways we can import collections,environments etc.

  For example, we can import from a file, from a url etc. the flow will look like this,

  `FileSource/UrlSource/GistSource gets the data from the file/url/gist` -> `and that data is passed to hoppscotchImporter/insomniaImporter/openAPIImporter` etc.

  Importers can say which sources they support, and our UI will render those options for the user to select.

  Previously we were duplicating logic to import from a file / url / gist etc across different files. but now we have Import Sources abstractions that prevents duplication.

  - `FileSource`

    - a function that creates a step with the FileImport.vue component. this component renders a file input and emits the file content as a string. the logic to get the file content is also moved to this component. previously we were duplicating this logic across different places.

  - `UrlSource`

    - a function that creates a step with the UrlImport.vue component. this component renders a url input and emits the url content as a string. the logic to get the url content is also moved to this component. previously we were duplicating this logic across different places.

  - `GistSource`

    - This step uses the same component as UrlSource. but it has logic to fetch the gist content using github api instead of a just fetching the url. previously we were duplicating this logic across different places.

- Making Step By Step UI more flexible and maintainable with `useSteps` composable.

  Previously imports had 2 steps. one to select the importer and then to Select File/Url etc. this was implemented very imperatively/hardcoded with if statements. but now our requirements are more complex. for example, we need to support multiple sources for a single importer, which adds the need for 3 steps. doing it using the previous implementation was not an ideal solution. so we added a `useSteps` composable that allows us to define steps and navigate between them, easily.

  Previously, we were duplicating the step by step UI logic across different places. eg: `components/importexport/ImportExport.vue` and `components/collections/ImportExport.vue` had the same logic. but they were duplicated. this PR adds a `defineStep` function that allows us to define a step and reuse it across different places. this makes the code more maintainable and flexible.

  a new function `defineStep` is used to create a step that can be used in the step ui. a step is nothing but an component + its props, which can be rendered when needed. for example a simple flow looks like this.

  - Step 1 - `component: ImportExportList, props: { importers, exporters }`, renders a list of importers and exporters
  - Step 2 - `component: ImportExportSourcesList, props: { sources }`, renders a list of sources for the selected importer/exporter
  - Step 3 - `component: FileImport, props: { acceptedFileTypes, caption, onImportFromFile }`, renders a file input and emits the file content as a string. the selected importer will be executed with the file content.

  `defineStep` function takes an ID, the component to render for the Step. and its props/emits as parameters and it will create a step for you. the props/emits also gets nice typescript autocomplete as you would expect.

- Extract duplicated UI logic to `components/importexport/ImporterExport.vue`

  - We have 3 ImportExport UIs, all of them had the same logic. but they were duplicated and had inconsistant UIs. this PR adds a `components/importexport/ImportExport.vue` component, that act as the base component with all the common UI, it takes `importerModules` and `exporterModules` as props and renders the UI, also implements the step logic and handles navigating between the steps.

- How Importers for Collections, Environments are implemented using the common ImportExport component

  - We create a simple wrapper for the common `ImporterExport.vue` component.

  - This wrapper defines the importerModules and exporterModules. and passes it to the `components/importexport/ImporterExport.vue` component.

  - The importers and exporters will have type `ImporterOrExporter`,

    - See `ImporterOrExporter` type.

      - It has `supported_sources` for adding an array of supported sources. eg: FileSource, UrlSource ( eg: OpenAPI Importer supports UrlImport + File Import )

      - It has `component` for adding a single source/any component. eg: FileSource ( eg: All other importers other than OpenAPI uses FileSource only )

      - It has `action` for adding a function to execute without showing any UI. ( eg: Export as JSON uses this ).

    - When the user selects an importer, if it has a `component`, we show the component UI, else we render the list of sources. it is also possible to just execute a function without showing any UI using `action`. ( eg: export as json use this.)

  - Here is an example importer defined,

    - ```typescript
      const HoppEnvironmentsImport: ImporterOrExporter = {
        metadata: {
          // defines the metadata
        },
        // give the source
        component: FileSource({
          acceptedFileTypes: "application/json",
          caption: "import.hoppscotch_environment_description",
          onImportFromFile: async (environments) => {
            // executes the importer function
            // you can import importer functions from '~/helpers/import', they are just regular functions. Since we are using them directly instead of in a loop of importer functions, they don't have to follow any specific type. this adds a lot of flexibility + speed of iteration. but takes away a little convenience.
            const res = await hoppEnvImporter(environments)();

            // do component specific stuff
            if (E.isLeft(res)) {
              failedImport();
              return;
            } else {
              importSuccessful(res.right);
            }

            emit("hide-modal");
          },
        }),
      };
      ```

2. ### **Use the same system for collections, envs, graphql import&exports**

   Previously we had seperate components with inconsistant UIs for collections, envs, and gql imports/exports. since we use the same base components, now we have one consistant import/export UI all across the product.

3. ### **Implement `hoppEnv` + `postmanEnv` imports with zod validation**

   Previously we didn't have a seperate entry for importing hopp environment and postman environment. but the code was kinda handling both of these platforms. this PR adds seperate entries in the Import/Export list. also these importers have zod validation. ( the schema for postman is created from the previous implementation. this can be improved in the future. )

4. ### **OpenAPI imports from URL + File**

   Previously we only had support for importing openAPI from a file. but now we have support for importing from a url too. ( note: this can be enabled for any of our importers, but we only enabled it for openAPI for now. )


packages/hoppscotch-common/src/components/environments/ImportExport.vue
```diff
@@ -158,251 +60,209 @@ const environmentJson = computed(() => {
...
-      type: "HOPP_EXPORT_ENVIRONMENT",
-      platform: "rest",
-    })
 
-    window.open(res.data.html_url)
-  } catch (e) {
-    toast.error(t("error.something_went_wrong").toString())
-    console.error(e)
-  }
+      emit("hide-modal")
```
jamesgeorge007, line 91: Do we don't want to hide the modal on failure? Also, isn't the else redundant?

packages/hoppscotch-common/src/components/environments/ImportExport.vue
```diff
@@ -158,251 +60,209 @@ const environmentJson = computed(() => {
...
+    caption: "import.postman_environment_description",
+    onImportFromFile: async (environments) => {
+      const res = await postmanEnvImporter(environments)()
+
+      if (E.isLeft(res)) {
+        failedImport()
+        return
+      } else {
+        importSuccessful(res.right)
+      }
```
jamesgeorge007, line 116: Redundant else.

packages/hoppscotch-common/src/components/environments/ImportExport.vue
```diff
@@ -158,251 +60,209 @@ const environmentJson = computed(() => {
...
+    caption: "import.environments_from_gist_description",
+    onImportFromGist: async (environments) => {
+      const res = await hoppEnvImporter(environments)()
+
+      if (E.isLeft(res)) {
+        failedImport()
+        return
+      } else {
+        importSuccessful(res.right)
       }
```
jamesgeorge007, line 142: Redundant else.

packages/hoppscotch-common/src/components/environments/ImportExport.vue
```diff
@@ -158,251 +60,209 @@ const environmentJson = computed(() => {
...
-          () => {
-            loading.value = false
-            hideModal()
-            fileImported()
-          }
-        )
-      )()
+    if (E.isRight(message)) {
+      toast.success(message.right)
     } else {
```
jamesgeorge007, line 166: Redundant else.

packages/hoppscotch-common/src/components/environments/ImportExport.vue
```diff
@@ -158,251 +60,209 @@ const environmentJson = computed(() => {
...
-          (err: GQLError<string>) => {
-            console.error(err)
-            toast.error(`${getErrorMessage(err)}`)
-          },
-          () => {
-            // wait for all the environments to be created then fire the toast
-          }
-        )
-      )()
+      toast.error(t("export.failed").toString())
```
jamesgeorge007, line 167:
```suggestion
      toast.error(t("export.failed"))
```

packages/hoppscotch-common/src/components/environments/ImportExport.vue
```diff
@@ -158,251 +60,209 @@ const environmentJson = computed(() => {
...
+    disabled: !currentUser.value
+      ? true
+      : currentUser.value.provider !== "github.com"
+      ? true
+      : false,
+    applicableTo: ["personal-workspace", "team-workspace"],
+  },
+  action: async () => {
+    if (!currentUser.value) {
+      toast.error(t("profile.no_permission").toString())
```
jamesgeorge007, line 193:
```suggestion
      toast.error(t("profile.no_permission"))
```

packages/hoppscotch-common/src/components/environments/ImportExport.vue
```diff
@@ -158,251 +60,209 @@ const environmentJson = computed(() => {
...
+      // @ts-ignore
+      currentUser.provider !== "github.com"
+      ? `export.require_github`
+      : "export.create_secret_gist",
+    icon: IconUser,
+    disabled: !currentUser.value
+      ? true
+      : currentUser.value.provider !== "github.com"
+      ? true
+      : false,
```
jamesgeorge007, line 188: Can we simplify this similar to the above?

packages/hoppscotch-common/src/components/environments/ImportExport.vue
```diff
@@ -158,251 +60,209 @@ const environmentJson = computed(() => {
...
+  metadata: {
+    id: "export.as_gist",
+    name: "export.create_secret_gist",
+    title: !currentUser
+      ? "export.require_github"
+      : // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      currentUser.provider !== "github.com"
+      ? `export.require_github`
+      : "export.create_secret_gist",
```
jamesgeorge007, line 182: Instead of a negated check, let's check for the positive condition.

packages/hoppscotch-common/src/components/environments/ImportExport.vue
```diff
@@ -158,251 +60,209 @@ const environmentJson = computed(() => {
...
-    if (!content) {
-      toast.show(t("action.choose_file").toString())
+const HoppEnvironmentsGistExporter: ImporterOrExporter = {
+  metadata: {
+    id: "export.as_gist",
+    name: "export.create_secret_gist",
+    title: !currentUser
+      ? "export.require_github"
+      : // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
```
jamesgeorge007, line 179: Do we know the root cause of this type error?

packages/hoppscotch-common/src/components/environments/ImportExport.vue
```diff
@@ -158,251 +60,209 @@ const environmentJson = computed(() => {
...
+    if (accessToken) {
+      const res = await environmentsGistExporter(
+        JSON.stringify(environmentJson.value),
+        accessToken
+      )
+
+      if (E.isLeft(res)) {
+        toast.error(t("export.failed").toString())
+      } else {
+        toast.success(t("export.success").toString())
```
jamesgeorge007, line 208:
```suggestion
        toast.error(t("export.failed"))
      } else {
        toast.success(t("export.success"))
```

packages/hoppscotch-common/src/components/environments/ImportExport.vue
```diff
@@ -158,251 +60,209 @@ const environmentJson = computed(() => {
...
+const failedImport = () => {
+  toast.error(t("import.failed").toString())
 }
 
-const importFromHoppscotch = (environments: Environment[]) => {
+const importSuccessful = async (environments: Environment[]) => {
   if (props.environmentType === "MY_ENV") {
     appendEnvironments(environments)
-    fileImported()
+    toast.success(t("import.success").toString())
```
jamesgeorge007, line 230:
```suggestion
    toast.success(t("import.success"))
```

packages/hoppscotch-common/src/components/environments/ImportExport.vue
```diff
@@ -158,251 +60,209 @@ const environmentJson = computed(() => {
...
+  if (failedImports) {
+    toast.error(t("import.failed").toString())
   } else {
-    switch (err.error) {
-      case "team_environment/not_found":
-        return t("team_environment.not_found")
-      default:
-        return t("error.something_went_wrong")
-    }
+    toast.success(t("import.success").toString())
```
jamesgeorge007, line 258: Isn't `toString` unnecessary?

packages/hoppscotch-common/src/components/environments/ImportExport.vue
```diff
@@ -158,251 +60,209 @@ const environmentJson = computed(() => {
...
   }
 }
+
+const emit = defineEmits<{
+  (e: "hide-modal"): () => void
+}>()
+
+// looking into implementing the collections export for environments
+// check here /Users/akash/hoppscotch/packages/hoppscotch-common/src/helpers/import-export/export/gistExport.ts
```
jamesgeorge007, line 267: Might want to remove this comment later.

packages/hoppscotch-common/src/components/importexport/ImportExport.vue
```diff
@@ -0,0 +1,166 @@
...
+        :title="t('action.go_back')"
+        :icon="IconArrowLeft"
+        @click="goToPreviousStep"
+      />
+    </template>
+    <template #body>
+      <component
+        :is="currentStep.component"
+        v-bind="currentStep.props()"
+      ></component>
```
jamesgeorge007, line 21:
```suggestion
     />
```

packages/hoppscotch-common/src/components/importexport/ImportExportList.vue
```diff
@@ -0,0 +1,76 @@
...
+            class="flex"
+          >
+            <HoppSmartItem
+              v-tippy="{ theme: 'tooltip' }"
+              :icon="exporter.icon"
+              :label="t(`${exporter.name}`)"
+              :disabled="exporter.disabled"
+              :loading="exporter.loading"
+              @click="emit('exporter-selected', exporter.id)"
+            ></HoppSmartItem>
```
jamesgeorge007, line 31:
```suggestion
            />
```

packages/hoppscotch-common/src/components/importexport/ImportExportList.vue
```diff
@@ -0,0 +1,76 @@
...
+          :label="t(`${importer.name}`)"
+          @click="emit('importer-selected', importer.id)"
+        />
+      </template>
+    </HoppSmartExpand>
+    <hr />
+    <div class="flex flex-col space-y-2">
+      <template v-for="exporter in exporters" :key="exporter.id">
+        <!-- adding the title to a span if the item is visible, otherwise the title won't be shown -->
+        <template v-if="exporter.disabled && exporter.title">
```
jamesgeorge007, line 18: Do we need the `template` wrappers? Can't the `v-if/else` directives be applied straightaway to the direct children?

packages/hoppscotch-common/src/components/importexport/ImportExportList.vue
```diff
@@ -0,0 +1,76 @@
...
+        <template v-else>
+          <HoppSmartItem
+            v-tippy="{ theme: 'tooltip' }"
+            :icon="exporter.icon"
+            :title="t(`${exporter.title}`)"
+            :label="t(`${exporter.name}`)"
+            :loading="exporter.loading"
+            :disabled="exporter.disabled"
+            @click="emit('exporter-selected', exporter.id)"
+          ></HoppSmartItem>
```
jamesgeorge007, line 44:
```suggestion
          />
```

packages/hoppscotch-common/src/components/importexport/ImportExportSteps/FileImport.vue
```diff
@@ -0,0 +1,97 @@
...
+    </p>
+
+    <p>
+      <HoppButtonPrimary
+        class="w-full"
+        :label="t('import.title')"
+        :disabled="!hasFile"
+        @click="emit('importFromFile', fileContent)"
+      />
+    </p>
```
jamesgeorge007, line 37: Do we need to wrap the button in a para?

packages/hoppscotch-common/src/components/importexport/ImportExportSteps/FileImport.vue
```diff
@@ -0,0 +1,97 @@
...
+
+    <!-- :loading="importingMyCollections" -->
```
jamesgeorge007, line 39: Comment to be removed?

packages/hoppscotch-common/src/components/importexport/ImportExportSteps/FileImport.vue
```diff
@@ -0,0 +1,97 @@
...
+    const content = target!.result as string | null
+    if (!content) {
+      hasFile.value = false
+      toast.show(t("action.choose_file").toString())
+      return
+    }
+
+    fileContent.value = content
+
+    hasFile.value = !!content?.length
```
jamesgeorge007, line 92:
```suggestion
    hasFile.value = Boolean(content?.length)
```

packages/hoppscotch-common/src/components/importexport/ImportExportSteps/MyCollectionImport.vue
```diff
@@ -0,0 +1,64 @@
+<template>
+  <div class="select-wrapper">
+    <select
+      v-model="mySelectedCollectionID"
+      autocomplete="off"
+      class="select"
+      autofocus
+    >
+      <option :key="undefined" :value="undefined" disabled selected>
```
jamesgeorge007, line 9: Is the `value` set to `undefined` on purpose?

packages/hoppscotch-common/src/components/importexport/ImportExportSteps/MyCollectionImport.vue
```diff
@@ -0,0 +1,64 @@
...
+  </div>
+
+  <p class="my-4">
+    <HoppButtonPrimary
+      class="w-full"
+      :label="t('import.title')"
+      :disabled="!hasSelectedCollectionID"
+      @click="fetchCollectionFromMyCollections"
+    />
+  </p>
```
jamesgeorge007, line 30: Is it necessary to wrap the button in a para?

packages/hoppscotch-common/src/components/importexport/ImportExportSteps/MyCollectionImport.vue
```diff
@@ -0,0 +1,64 @@
...
+        {{ collection.name }}
+      </option>
+    </select>
+  </div>
+
+  <p class="my-4">
+    <HoppButtonPrimary
+      class="w-full"
+      :label="t('import.title')"
+      :disabled="!hasSelectedCollectionID"
```
jamesgeorge007, line 27:
```suggestion
      :label="t('import.title')"
      :disabled="!hasSelectedCollectionID"
      class="w-full"
```

Let's keep `class` below other props, [Ref](https://eslint.vuejs.org/rules/attributes-order.html).

packages/hoppscotch-common/src/components/importexport/ImportExportSteps/UrlImport.vue
```diff
@@ -0,0 +1,100 @@
+<template>
+  <!-- v-else-if="step.name === 'URL_IMPORT'" -->
```
jamesgeorge007, line 2: Comment to be removed?

packages/hoppscotch-common/src/components/importexport/ImportExportSteps/UrlImport.vue
```diff
@@ -0,0 +1,100 @@
...
+
+    <p>
+      <HoppButtonPrimary
+        class="w-full"
+        :label="t('import.title')"
+        :disabled="!hasURL"
+        :loading="isFetchingUrl"
+        @click="fetchUrlData"
+      />
+    </p>
```
jamesgeorge007, line 34: Is it necessary to wrap the button in a para?

packages/hoppscotch-common/src/components/importexport/ImportExportSteps/UrlImport.vue
```diff
@@ -0,0 +1,100 @@
...
+        :placeholder="`${t('import.from_url')}`"
+      />
+    </p>
+
+    <p>
+      <HoppButtonPrimary
+        class="w-full"
+        :label="t('import.title')"
+        :disabled="!hasURL"
+        :loading="isFetchingUrl"
```
jamesgeorge007, line 31:
```suggestion
        :label="t('import.title')"
        :disabled="!hasURL"
        :loading="isFetchingUrl"
        class="w-full"
```

packages/hoppscotch-common/src/components/importexport/ImportExportSteps/UrlImport.vue
```diff
@@ -0,0 +1,100 @@
...
+const hasURL = ref(false)
+
+const isFetchingUrl = ref(false)
+
+watch(inputChooseGistToImportFrom, (url) => {
+  if (url === "") {
+    hasURL.value = false
+  } else {
+    hasURL.value = true
+  }
```
jamesgeorge007, line 67:
```suggestion
  hasURL.value = Boolean(url)
```

Will the above work?

jamesgeorge007: Can `url` be other falsy values like `null` or `undefined`? Do we want hasURL` to be `true` in those cases?

packages/hoppscotch-common/src/components/environments/ImportExport.vue
amk-dev: Our central repo uses Firebase User Types, but selfhost-web uses HoppUser type. only the Firebase type has a provider attached to it.

We could make an attempt to fix both for a common interface. but not taking that effort now. also the gist exporter is disabled for selfhost-web.

amk-dev: also most of these weired ternary stuff came from our old code while i copied them to the new system.

packages/hoppscotch-common/src/components/environments/ImportExport.vue
amk-dev: we're not returning from the if. we can make it a ternery for aesthetic purposes.

packages/hoppscotch-common/src/components/environments/ImportExport.vue
amk-dev: When the import fails, the user might want to try again. so probably better to leave to leave the dialog open.

packages/hoppscotch-common/src/components/importexport/ImportExportSteps/FileImport.vue
amk-dev: Isn't both the same, and comes down to preference ?

packages/hoppscotch-common/src/components/importexport/ImportExportSteps/MyCollectionImport.vue
amk-dev: according to the rule, we should keep the globals (category `class` belongs to) above @click only. which its already in. the order of the bound variables & class does not seem to matter, because they're in the same priortity.

packages/hoppscotch-common/src/components/importexport/ImportExportSteps/MyCollectionImport.vue
amk-dev: i don't think so. old code.

packages/hoppscotch-common/src/components/importexport/ImportExportSteps/UrlImport.vue
amk-dev: same situation as the last one.

packages/hoppscotch-common/src/components/importexport/ImportExportSteps/UrlImport.vue
amk-dev: Url is the value of `inputChooseGistToImportFrom` which is expected to be string, initialized as an empty string. so it on normal cases won't have any other falsy values.

amk-dev: also the hasURL.value = Boolean(url) should work here.

packages/hoppscotch-common/locales/en.json
```diff
@@ -298,14 +302,22 @@
...
     "json_description": "Import collections from a Hoppscotch Collections JSON file",
-    "title": "Import"
+    "title": "Import",
+    "hoppscotch_environment": "Hoppscotch Environment",
+    "hoppscotch_environment_description": "Import Hoppscotch Environment JSON file",
+    "postman_environment": "Postman Environment",
+    "postman_environment_description": "Import Postman Environment JSON file",
+    "environments_from_gist": "Import From GIST",
+    "environments_from_gist_description": "Import Hoppscotch Environments From GIST",
+    "gql_collections_from_gist_description": "Import GraphQL Collections From GIST"
```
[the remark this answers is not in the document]
amk-dev: We seems to use Gist instead of GIST

packages/hoppscotch-common/src/components/collections/ImportExport.vue
```diff
@@ -1,361 +1,430 @@
...
+      props.collectionsType.selectedTeam
+    ) {
+      const res = await teamCollectionsExporter(
+        props.collectionsType.selectedTeam.id
+      )
+
+      if (E.isRight(res)) {
+        initializeDownloadCollection(
+          res.right.exportCollectionsToJSON,
+          "team-collections.json"
```
nivedin, line 328:
```suggestion
          "team-collections"
```
we do not need the .json here since its already set in the `initializeDownloadCollection` function.

packages/hoppscotch-common/src/components/collections/graphql/ImportExport.vue
```diff
@@ -1,284 +1,177 @@
...
+    name: "export.as_json",
+    title: "action.download_file",
+    icon: IconUser,
+    disabled: false,
+    applicableTo: ["personal-workspace", "team-workspace"],
+  },
+  action: async () => {
+    const message = initializeDownloadCollection(
+      gqlCollectionsExporter(gqlCollections.value),
+      "GQLCollections.json"
```
nivedin, line 109:
```suggestion
          "GQLCollections"
```
we do not need the .json here since its already set in the `initializeDownloadCollection` function.

packages/hoppscotch-common/src/components/collections/ImportExport.vue
```diff
@@ -1,361 +1,430 @@
...
 
-  if (!inputFileToImport.files || inputFileToImport.files.length === 0) {
-    inputChooseFileToImportFrom.value[0].value = ""
-    hasFile.value = false
-    toast.show(t("action.choose_file").toString())
-    return
-  }
+    const message = initializeDownloadCollection(
+      myCollectionsExporter(myCollections.value),
+      "Collections.json"
```
nivedin, line 293:
```suggestion
      "Collections"
```
we do not need the .json here since its already set in the `initializeDownloadCollection` function.

packages/hoppscotch-common/src/components/environments/ImportExport.vue
```diff
@@ -158,251 +60,199 @@ const environmentJson = computed(() => {
...
+    const message = initializeDownloadCollection(
+      environmentsExporter(environmentJson.value),
+      "Environments"
+    )
 
-const openDialogChooseFileToImportFrom = () => {
-  if (inputChooseFileToImportFrom.value)
-    inputChooseFileToImportFrom.value.click()
+    E.isRight(message)
+      ? toast.success(message.right)
```
nivedin, line 163:
```suggestion
      ? toast.success(t(message.right))
```

packages/hoppscotch-common/src/components/environments/ImportExport.vue
```diff
@@ -158,251 +60,199 @@ const environmentJson = computed(() => {
...
+const failedImport = () => {
+  toast.error(t("import.failed").toString())
 }
 
-const importFromHoppscotch = (environments: Environment[]) => {
+const importSuccessful = async (environments: Environment[]) => {
   if (props.environmentType === "MY_ENV") {
     appendEnvironments(environments)
-    fileImported()
+    toast.success(t("import.success"))
```
nivedin, line 223: `import.success` is not defined

packages/hoppscotch-common/src/components/collections/ImportExport.vue
```diff
@@ -1,361 +1,430 @@
...
+          } else {
+            showImportFailedError()
+          }
+        },
+      }),
+    },
+    {
+      id: "url_import",
+      name: "import.from_url",
+      icon: IconFile,
```
nivedin, line 190:
```suggestion
      icon: IconLink,
```
we can use a link icon rather than a file here

nivedin (changes requested):
Please look into these comments

packages/hoppscotch-common/src/components/environments/ImportExport.vue
jamesgeorge007: Good catch. Ternary works 👍

packages/hoppscotch-common/src/components/importexport/ImportExportSteps/FileImport.vue
jamesgeorge007: Yup, but isn't the `Boolean` constructor much more readable?

packages/hoppscotch-common/src/components/importexport/ImportExportSteps/MyCollectionImport.vue
jamesgeorge007: Yup, it's just a convention that I've been following.

> but in this case doesnt it make it more pleasent to read, if the class is not sandwiched between vue attributes

In the suggested approach, it appears below all component props. Hence, there's a clear distinction. But yeah, it's just a matter of preference.

packages/hoppscotch-common/src/components/collections/ImportExport.vue
```diff
@@ -1,361 +1,475 @@
...
-          </span>
-        </div>
-      </div>
-    </template>
-  </HoppSmartModal>
+  <ImportExportBase
+    ref="collections-import-export"
+    modal-title="modal.collections"
+    :importer-modules="importerModules"
+    :exporter-modules="exporterModules"
```
jamesgeorge007, line 6:
```suggestion
    :importer-modules="importerModules"
    :exporter-modules="exporterModules"
    modal-title="modal.collections"
```

Shouldn't static props appear below dynamic ones?

packages/hoppscotch-common/src/components/collections/ImportExport.vue
```diff
@@ -1,361 +1,475 @@
...
-  stepResults.value = []
-  inputChooseFileToImportFrom.value = ""
-  inputChooseGistToImportFrom.value = ""
-  mySelectedCollectionID.value = undefined
+  const isTeamCollection = collectionsType.type === "team-collections"
+
+  return isTeamCollection
+    ? collectionsType.selectedTeam?.myRole == "EDITOR" ||
+        collectionsType.selectedTeam?.myRole == "OWNER"
+    : false
```
jamesgeorge007, line 442: I think in this case, since there is a condition chain, `if` would be appropriate. Return early if `isTeamCollection` is `false`; else, return the value yield from the role check. If you prefer to keep the ternary version, maybe using `Array.includes` for the role check might suit here.

packages/hoppscotch-common/src/components/collections/ImportExport.vue
```diff
@@ -1,361 +1,475 @@
...
+    )
+
+    return E.isRight(res)
+      ? E.right(res.right.exportCollectionsToJSON)
+      : E.left(res.left)
+  } else {
+    collections = JSON.stringify(myCollections.value, null, 2)
+
+    return E.right(collections)
+  }
```
jamesgeorge007, line 473: Can there be a case where `selectedTeam.id` doesn't yield a value where it goes to the `else` block despite being in the team workspace? In that case, don't we require a fallback instead of the control going to the `my-collections` block?
Also, the `else` block is redundant since we're returning early. We can inline the JSON stringified collections object instead of assigning it to a variable since there aren't multiple references.

packages/hoppscotch-common/src/components/importExport/Base.vue
```diff
@@ -0,0 +1,163 @@
+<template>
+  <HoppSmartModal
+    dialog
+    :title="t(modalTitle)"
+    styles="sm:max-w-md"
```
jamesgeorge007, line 5:
```suggestion
    :title="t(modalTitle)"
    styles="sm:max-w-md"
    dialog
```

The [style guide](https://eslint.vuejs.org/rules/attributes-order.html#rule-details) recommends to keep boolean props below static and dynamic props.

packages/hoppscotch-common/src/components/collections/graphql/ImportExport.vue
```diff
@@ -1,284 +1,177 @@
...
-          @click="exportJSON"
-        />
-      </div>
-    </template>
-  </HoppSmartModal>
+  <ImportExportBase
+    ref="collections-import-export"
+    modal-title="graphql_collections.title"
+    :importer-modules="importerModules"
+    :exporter-modules="exporterModules"
```
jamesgeorge007, line 6:
```suggestion
    :importer-modules="importerModules"
    :exporter-modules="exporterModules"
    modal-title="graphql_collections.title"
```

packages/hoppscotch-common/src/components/environments/ImportExport.vue
```diff
@@ -1,155 +1,57 @@
...
-          @click="exportJSON"
-        />
-      </div>
-    </template>
-  </HoppSmartModal>
+  <ImportExportBase
+    ref="collections-import-export"
+    modal-title="environment.title"
+    :importer-modules="importerModules"
+    :exporter-modules="exporterModules"
```
jamesgeorge007, line 6:
```suggestion
    :importer-modules="importerModules"
    :exporter-modules="exporterModules"
    modal-title="environment.title"
```

packages/hoppscotch-common/src/components/environments/ImportExport.vue
```diff
@@ -158,251 +60,199 @@ const environmentJson = computed(() => {
...
-        )
-      )()
+const HoppEnvironmentsGistExporter: ImporterOrExporter = {
+  metadata: {
+    id: "export.as_gist",
+    name: "export.create_secret_gist",
+    title:
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      currentUser && currentUser.provider == "github.com"
```
jamesgeorge007, line 175:
```suggestion
      currentUser?.provider === "github.com"
```

packages/hoppscotch-common/src/components/importExport/ImportExportSteps/FileImport.vue
```diff
@@ -0,0 +1,95 @@
...
+      <input
+        id="inputChooseFileToImportFrom"
+        ref="inputChooseFileToImportFrom"
+        name="inputChooseFileToImportFrom"
+        type="file"
+        class="p-4 cursor-pointer transition file:transition file:cursor-pointer text-secondary hover:text-secondaryDark file:mr-2 file:py-2 file:px-4 file:rounded file:border-0 file:text-secondary hover:file:text-secondaryDark file:bg-primaryLight hover:file:bg-primaryDark"
+        :accept="acceptedFileTypes"
+        @change="onFileChange"
+      />
+    </p>
```
jamesgeorge007, line 28: Isn't it better to wrap the input element in a `div` instead?

packages/hoppscotch-common/src/components/importExport/ImportExportSteps/FileImport.vue
```diff
@@ -0,0 +1,95 @@
...
+
+    <div>
+      <HoppButtonPrimary
+        class="w-full"
+        :label="t('import.title')"
+        :disabled="!hasFile"
+        @click="emit('importFromFile', fileContent)"
+      />
+    </div>
```
jamesgeorge007, line 37: Am I right in assuming the role of the `div` wrapper here is to provide a block-level element behavior?

packages/hoppscotch-common/src/components/importExport/ImportExportSteps/MyCollectionImport.vue
```diff
@@ -0,0 +1,65 @@
+<template>
+  <div class="select-wrapper">
+    <select
+      v-model="mySelectedCollectionID"
+      autocomplete="off"
+      class="select"
+      autofocus
+    >
+      <option :key="undefined" :value="undefined" disabled selected>
```
jamesgeorge007, line 9: What's the role of this option element? Does it serve as the placeholder option?

jamesgeorge007 (commented):
Could you add relevant screenshots to the PR description?

packages/hoppscotch-common/src/components/collections/ImportExport.vue
```diff
@@ -1,361 +1,475 @@
...
+    disabled: false,
+    applicableTo: ["team-workspace"],
+    isLoading: isHoppTeamCollectionExporterInProgress,
+  },
+  action: async () => {
+    isHoppTeamCollectionExporterInProgress.value = true
 
-  const inputFileToImport = inputChooseFileToImportFrom.value[0]
+    if (
+      props.collectionsType.type == "team-collections" &&
```
jamesgeorge007, line 344:
```suggestion
      props.collectionsType.type === "team-collections" &&
```

packages/hoppscotch-common/src/components/importExport/ImportExportSteps/MyCollectionImport.vue
amk-dev: Yes.

jamesgeorge007 (changes requested):
- The `import from Postman` environment flow resulted in an error toast saying unrecognized format.
  
  image

- Import collections/environments from Gist doesn't work as expected. The modal went to the loading state, and there was no further feedback. The relevant files used for import can be found [here](https://gist.github.com/jamesgeorge007/082e49a88b384001181cb5e044f4f959).

  https://github.com/hoppscotch/hoppscotch/assets/25279263/ce2946c3-7d5c-453c-a514-ce9340b3ef85

- The `Create secret Gist` option was disabled even after logging in with GitHub. Discussed separately; this is expected in `selfhost` since login provider info isn't implemented yet. It'll work on `central`. We can maybe disable it on `selfhost` and make it available only on `central`, till all the dependents are in place.

  image

- The Open API import from step doesn't permit `yaml` files.

- Downloading empty collections and environments is permitted. Also, the empty state updates aren't reflected. This seems like an issue with the branch not having the latest changes.

jamesgeorge007 (dismissed):

amk-dev (commented):
eslint changes LGTM 👍🏼

PR #3415, opened 2023-10-04
HBE-270 Test-Case timestamp issue fix in backend

mirarifhasan:
Closes HBE-270

### Description

In this PR, we fix the broken test cases regarding timestamp mismatch. We encounterd this issue, when GitHub action runs.

Failure GitHub Action: [https://github.com/hoppscotch/hoppscotch/actions/runs/6312305505/job/17138088088](https://github.com/hoppscotch/hoppscotch/actions/runs/6312305505/job/17138088088)

Screenshot 2023-10-04 at 4 06 15 PM

### Checks

- [x] My pull request adheres to the code style of this project
- [ ] My code requires changes to the documentation
- [ ] I have updated the documentation as required
- [x] All the tests have passed

### Additional Information

Nil

PR #3445, opened 2023-10-11
HBE-296 feat: introducing 'infra' type and splitting model properties between 'admin' and 'infra'

mirarifhasan:
### Issue Number
Closes HBE-296

### Description

Previously all `admin-dashboard` related graphql query was done under `admin` type. But that ware not sound like admin properties. 
(i.e.`usersCount`, `teamCollectionsCount` does not sound like admin property)

For that, we introduce a new type `infra`. This (infra) type will contain all the things related to the matrix and other stuff (example given below in ref-2). 

`admin` type will be there as well with admin-specific properties.
 
So, now the `admin` and `infra` query and their properties looks like below:

```ts
// ref-1
query {
  admin {
    uid
    displayName
    email
    photoURL
    createdOn
  }
}

// ref-2
query {
  infra {
    userList{}
    userInfo(){}
    invitedUsers(){}
    teamList(){}
    teamInfo(){}
    // matrix
    usersCount
    teamsCount
    teamCollectionsCount
    teamRequestsCount
  }
}
```

### Checks

### Additional Information

Merge #3415 PR to avoid build failure in Test action.

### How to test
1. Test `admin` query
2. Test `infra` query


packages/hoppscotch-backend/src/admin/infra.resolver.ts
```diff
@@ -0,0 +1,223 @@
...
+@UseGuards(GqlThrottlerGuard)
+@Resolver(() => Infra)
+export class InfraResolver {
+  constructor(private adminService: AdminService) {}
+
+  @Query(() => Infra, {
+    description: 'Gives details of the admin executing this query',
+  })
+  @UseGuards(GqlAuthGuard, GqlAdminGuard)
+  infra(@GqlAdmin() admin: Admin) {
```
balub, line 35:
```suggestion
  ExecutedBy(@GqlAdmin() admin: Admin) {
```

packages/hoppscotch-backend/src/admin/infra.resolver.ts
```diff
@@ -0,0 +1,223 @@
...
+    if (E.isLeft(user)) throwErr(user.left);
+    return user.right;
+  }
+
+  @ResolveField(() => [User], {
+    description: 'Returns a list of all the users in infra',
+  })
+  @UseGuards(GqlAuthGuard, GqlAdminGuard)
+  async allUsers(
+    @Parent() infra: Infra,
```
balub, line 70: The parameter is not being used, we can remove it from the signature.

packages/hoppscotch-backend/src/admin/infra.resolver.ts
```diff
@@ -0,0 +1,223 @@
...
+    @Args() args: PaginationArgs,
+  ): Promise<AuthUser[]> {
+    const users = await this.adminService.fetchUsers(args.cursor, args.take);
+    return users;
+  }
+
+  @ResolveField(() => [InvitedUser], {
+    description: 'Returns a list of all the invited users',
+  })
+  async invitedUsers(@Parent() infra: Infra): Promise<InvitedUser[]> {
```
balub, line 80:
```suggestion
  async invitedUsers(): Promise<InvitedUser[]> {
```
The parameter is not being used, we can remove it from the signature.

packages/hoppscotch-backend/src/admin/infra.resolver.ts
```diff
@@ -0,0 +1,223 @@
...
+    const users = await this.adminService.fetchInvitedUsers();
+    return users;
+  }
+
+  @ResolveField(() => [Team], {
+    description: 'Returns a list of all the teams in the infra',
+  })
+  async allTeams(
+    @Parent() infra: Infra,
```
balub, line 89: The parameter is not being used, we can remove it from the signature.

packages/hoppscotch-backend/src/admin/infra.resolver.ts
```diff
@@ -0,0 +1,223 @@
...
+  ): Promise<Team[]> {
+    const teams = await this.adminService.fetchAllTeams(args.cursor, args.take);
+    return teams;
+  }
+
+  @ResolveField(() => Team, {
+    description: 'Returns a team info by ID when requested by Admin',
+  })
+  async teamInfo(
+    @Parent() infra: Infra,
```
balub, line 100: The parameter is not being used, we can remove it from the signature.

packages/hoppscotch-backend/src/admin/infra.resolver.ts
```diff
@@ -0,0 +1,223 @@
...
+    const team = await this.adminService.getTeamInfo(teamID);
+    if (E.isLeft(team)) throwErr(team.left);
+    return team.right;
+  }
+
+  @ResolveField(() => Number, {
+    description: 'Return count of all the members in a team',
+  })
+  async membersCountInTeam(
+    @Parent() infra: Infra,
```
balub, line 117: The parameter is not being used, we can remove it from the signature.

packages/hoppscotch-backend/src/admin/infra.resolver.ts
```diff
@@ -0,0 +1,223 @@
...
+  ): Promise<number> {
+    const teamMembersCount = await this.adminService.membersCountInTeam(teamID);
+    return teamMembersCount;
+  }
+
+  @ResolveField(() => Number, {
+    description: 'Return count of all the stored collections in a team',
+  })
+  async collectionCountInTeam(
+    @Parent() infra: Infra,
```
balub, line 134: The parameter is not being used, we can remove it from the signature.

packages/hoppscotch-backend/src/admin/infra.resolver.ts
```diff
@@ -0,0 +1,223 @@
...
+  ): Promise<number> {
+    const teamCollCount = await this.adminService.collectionCountInTeam(teamID);
+    return teamCollCount;
+  }
+
+  @ResolveField(() => Number, {
+    description: 'Return count of all the stored requests in a team',
+  })
+  async requestCountInTeam(
+    @Parent() infra: Infra,
```
balub, line 150: The parameter is not being used, we can remove it from the signature.

packages/hoppscotch-backend/src/admin/infra.resolver.ts
```diff
@@ -0,0 +1,223 @@
...
+  ): Promise<number> {
+    const teamReqCount = await this.adminService.requestCountInTeam(teamID);
+    return teamReqCount;
+  }
+
+  @ResolveField(() => Number, {
+    description: 'Return count of all the stored environments in a team',
+  })
+  async environmentCountInTeam(
+    @Parent() infra: Infra,
```
balub, line 166: The parameter is not being used, we can remove it from the signature.

packages/hoppscotch-backend/src/admin/infra.resolver.ts
```diff
@@ -0,0 +1,223 @@
...
+  ): Promise<number> {
+    const envsCount = await this.adminService.environmentCountInTeam(teamID);
+    return envsCount;
+  }
+
+  @ResolveField(() => [TeamInvitation], {
+    description: 'Return all the pending invitations in a team',
+  })
+  async pendingInvitationCountInTeam(
+    @Parent() infra: Infra,
```
balub, line 182: The parameter is not being used, we can remove it from the signature.

balub (changes requested):
In SHC, the `Admin` type does not have any admin specific fields yet, so for now they can just return `User` info.

packages/hoppscotch-backend/src/admin/admin.resolver.ts
```diff
@@ -41,195 +29,13 @@ export class AdminResolver {
...
-  async teamRequestsCount() {
-    return this.adminService.getTeamRequestsCount();
-  }
+  // @Query(() => Admin, {
+  //   description: 'Gives details of the admin executing this query',
+  // })
+  // @UseGuards(GqlAuthGuard, GqlAdminGuard)
+  // admin(@GqlAdmin() admin: Admin) {
+  //   return admin;
+  // }
```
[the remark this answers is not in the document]
mirarifhasan: Since this query is not called by sh-admin, that's why I made a comment. 
But, I'll turn on this query to get user-type data at-least.

packages/hoppscotch-backend/src/admin/infra.resolver.ts
```diff
@@ -0,0 +1,211 @@
...
+export class InfraResolver {
+  constructor(private adminService: AdminService) {}
+
+  @Query(() => Infra, {
+    description: 'Gives details of the infra executing this query',
+  })
+  @UseGuards(GqlAuthGuard, GqlAdminGuard)
+  infra(@GqlAdmin() admin: Admin) {
+    return admin;
+  }
```
balub, line 37: This query should not return an `admin` object, it should only return all the stuff we use in the field resolvers.

balub (changes requested):

packages/hoppscotch-backend/src/admin/infra.resolver.ts
```diff
@@ -0,0 +1,211 @@
...
+import { TeamInvitation } from 'src/team-invitation/team-invitation.model';
+import { GqlAdmin } from './decorators/gql-admin.decorator';
+
+@UseGuards(GqlThrottlerGuard)
+@Resolver(() => Infra)
+export class InfraResolver {
+  constructor(private adminService: AdminService) {}
+
+  @Query(() => Infra, {
+    description: 'Gives details of the infra executing this query',
```
balub, line 32:
```suggestion
    description: 'Fetch details of the Infrastructure',
```

balub (changes requested):

packages/hoppscotch-backend/src/admin/admin.model.ts
```diff
@@ -1,4 +1,5 @@
 import { ObjectType } from '@nestjs/graphql';
+import { User } from 'src/user/user.model';
 
 @ObjectType()
-export class Admin {}
+export class Admin extends User {}
```
balub, line 5:
```suggestion
export class Admin {}
```
extending `User` results in errors when trying to run queries to fetch user-settings, user-environments etc.

balub: image

balub (changes requested):

packages/hoppscotch-backend/src/admin/admin.model.ts
mirarifhasan:
```suggestion
export class Admin extends PartialType(User) {}
```
@balub, I'm planning to use `PartialType`.  

After using `PartialType`, from `admin` query:
There will be only 
> `uid, displayName, email, photoURL, isAdmin, createdOn, currentRESTSession, currentGQLSession`. 

And there won't be 
> `settings, environments, globalEnvironments, RESTHistory, GQLHistory`

Screenshot 2023-10-18 at 5 40 19 PM

I think, `settings, environments, etc` are SH-App-related fields instead of SH-Admin. 

Along with that, if in the future (assume), admin needs to know the `settings/environments/etc` of a particular user or himself (from admin dashboard), then, `infra.userInfo(uid)` is there to serve purpose.

packages/hoppscotch-backend/src/admin/admin.model.ts
```diff
@@ -1,4 +1,5 @@
-import { ObjectType } from '@nestjs/graphql';
+import { ObjectType, PartialType } from '@nestjs/graphql';
+import { User } from 'src/user/user.model';
 
 @ObjectType()
-export class Admin {}
+export class Admin extends PartialType(User) {}
```
balub, line 5: Using the `PartialType` here is technically incorrect right, because we want the `Admin` type to inherit all the properties of the `User` object. From the image below we see that some of the properties from the `User` type arent listed in the `Admin` type at all.

image

balub (changes requested):

packages/hoppscotch-backend/src/admin/admin.model.ts
balub: Okay. It makes sense.

balub: If this is the case we can remove `currentRESTSession` and `currentGQLSession` properties too.

mirarifhasan: Okay, I'll omit that 3 fields.

balub (approved):
LTGM 🚀

PR #3457, opened 2023-10-16
refactor: versioning and migration mechanism for public data structures

AndrewBastin:
Closes HFE-276

This PR intends to introduce Zod schemas and [Verzod](https://github.com/AndrewBastin/verzod) into `@hoppscotch/data` to power updated migrations.

### Description
- chore add `verzod` and `zod` as dependencies
- bump typescript version
- update REST Request, GraphQL Request and Environment systems to use Verzod schemas
- add `ValidContentTypesList` to `contentTypes` module in `@hoppscotch/data`
- deprecated couple of functions in `@hoppscotch/data`.

### Checks


packages/hoppscotch-data/src/rest/index.ts
```diff
@@ -1,66 +1,54 @@
...
+      && !Number.isNaN(parseInt(data.v))
+    ) {
+      return parseInt(data.v)
     }
 
-export interface HoppRESTRequest {
-  v: string
-  id?: string // Firebase Firestore ID
+    // For V0 we have to check the schema
+    const result = V0_SCHEMA.safeParse(data)
```
jamesgeorge007, line 43:
```suggestion
    const result = V0_VERSION.schema.safeParse(data)
```

Can we refer to the schema from the default export?

packages/hoppscotch-data/src/rest/index.ts
```diff
@@ -137,105 +150,55 @@ export function makeRESTRequest(
...
-  if (x.auth === "Bearer Token")
-    return {
-      authType: "bearer",
-      authActive: true,
-      token: x.bearerToken,
-    }
+  if (result.type === "ok") return result.value
 
-  return { authType: "none", authActive: true }
+  return getDefaultRESTRequest()
```
jamesgeorge007, line 203:
```ts
  const result = HoppRESTRequest.safeParse(x)

  return result.type === "ok" ? result.value : getDefaultRESTRequest()
```

packages/hoppscotch-data/src/rest/v/1.ts
```diff
@@ -0,0 +1,219 @@
+import { defineVersion } from "verzod"
+import { z } from "zod"
+import { V0_SCHEMA } from "./0"
```
jamesgeorge007, line 3:
```suggestion

import { V0_SCHEMA } from "./0"
```

packages/hoppscotch-data/src/rest/v/1.ts
```diff
@@ -0,0 +1,219 @@
...
+}).and(
+  z.union([
+    z.object({
+      isFile: z.literal(true),
+      value: z.array(z.instanceof(Blob))
+    }),
+    z.object({
+      isFile: z.literal(false),
+      value: z.string()
+    })
```
jamesgeorge007, line 17: Is this a good candidate for `discriminatedUnion` since there's a 
shared key` isFile`?

packages/hoppscotch-data/src/rest/v/1.ts
```diff
@@ -0,0 +1,219 @@
...
+      z.literal("application/ld+json"),
+      z.literal("application/hal+json"),
+      z.literal("application/vnd.api+json"),
+      z.literal("application/xml"),
+      z.literal("application/x-www-form-urlencoded"),
+      z.literal("text/html"),
+      z.literal("text/plain"),
+    ]),
+    body: z.string()
+  })
```
jamesgeorge007, line 51: Is this a good candidate for `discriminatedUnion` since there's a 
shared key` contentType`?

packages/hoppscotch-data/src/rest/v/1.ts
```diff
@@ -0,0 +1,219 @@
...
+    return {
+      contentType: "application/json",
+      body: x.rawParams ?? "",
+    }
+  }
+
+  return {
+    contentType: "application/json",
+    body: "",
+  }
```
jamesgeorge007, line 160:
```suggestion
  return {
    contentType: "application/json",
    body: x.contentType === "application/json" ? x.rawParams ?? "" : "",
  }
```

packages/hoppscotch-data/src/rest/v/1.ts
```diff
@@ -0,0 +1,219 @@
...
+  initial: false,
+  schema: V1_SCHEMA,
+  up(old: z.infer<typeof V0_SCHEMA>) {
+    const endpoint = `${old.url}${old.path}`
+    const headers = old.headers
+    const params = old.params
+    const name = old.name
+    const method = old.method
+    const preRequestScript = old.preRequestScript
+    const testScript = old.testScript
```
jamesgeorge007, line 198:
```suggestion
    const { url, path, headers, params, name, method, preRequestScript, testScript } = old
    
    const endpoint = `${url}${path}`
```

packages/hoppscotch-data/src/graphql/v/1.ts
jamesgeorge007, line 1: Shouldn't we follow the same versioning convention everywhere? For environments and Rest-based collections, it started with `v0`, while it's `v1` for GQL.

packages/hoppscotch-data/src/graphql/index.ts
```diff
@@ -1,51 +1,76 @@
...
+    1: V1_VERSION,
+    2: V2_VERSION
+  },
+  getVersion(x) {
+    return typeof x === "object"
+      && x !== null
+      && "v" in x
+      && typeof x.v === "number"
+        ? x.v
+        : 0
```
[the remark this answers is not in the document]
jamesgeorge007: It seems like we'd need the verbose version since `x` is inferred as `unknown`.

packages/hoppscotch-data/src/environment/index.ts
```diff
@@ -1,14 +1,22 @@
...
+    0: V0_VERSION
+  },
+  getVersion(x) {
+    return V0_SCHEMA.safeParse(x).success
+      ? 0
+      : null
+  }
+})
+
+export type Environment = InferredEntity<typeof Environment>
```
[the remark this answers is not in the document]
jamesgeorge007: Noticed the convention is followed at multiple places. Let's keep it as it is.

jamesgeorge007:
The failing tests need investigation.

packages/hoppscotch-data/src/rest/v/1.ts
jamesgeorge007: It isn't possible to use `discriminatedUnion` here since the first case has values set to `null` and the discriminant doesn't have a meaningful value.

jamesgeorge007 (changes requested):
Visiting `/graphql` results in the following error. Strangely there aren't any console errors coming up.

image

Also, failing tests need investigation.

jamesgeorge007:
The above error relating to the GraphQL page crashing was due to some underlying dependency bumps. I've reverted updates to the lock file in https://github.com/hoppscotch/hoppscotch/pull/3457/commits/9247fb562842492e56d8e6fde1ec385a31932fe0

packages/hoppscotch-data/src/graphql/v/1.ts
jamesgeorge007: @AndrewBastin, was this intentional?

jamesgeorge007:
@AndrewBastin, tests were failing since `js-sandbox` had `node` specified as the `testEnvironment` and it failed to recognize `Blob` since it wasn't getting imported. We can go with the polyfill path, but an easier fix will be to specify the environment as the browser via `jsdom`. Lmk your thoughts.
cc @amk-dev

packages/hoppscotch-data/src/graphql/index.ts
```diff
@@ -1,51 +1,73 @@
...
+  }
```
amk-dev, line 30: anyway we're using zod in `hoppscotch-data`. so why don't we use it here also ? makes it more readable and explicit. 

```suggestion
  getVersion(x) {
    const versionedObject = z.object({
      v: z.number(),
    })

    const result = versionedObject.safeParse(x)

    return result.success ? result.data.v : 0
  },
```

packages/hoppscotch-data/src/rest/index.ts
```diff
@@ -1,66 +1,53 @@
...
 
-  name: string
-  method: string
-  endpoint: string
-  params: HoppRESTParam[]
-  headers: HoppRESTHeader[]
-  preRequestScript: string
-  testScript: string
+    return result.success ? 0 : null
+  },
```
amk-dev, line 45: If we want we could write this in a zod native way. but it'll feel a little verbose. something along the lines of,

```typescript
getVersion(data) {
    const v1AndAbove = z.object({
      // For V1 onwards we have the v string storing the number
      v: z.string().transform((versionString, ctx) => {
        const v = parseInt(versionString)

        if (isNaN(v)) {
          ctx.addIssue({
            code: z.ZodIssueCode.custom,
            message: "Cannot parse valid number from version string",
          })

          return z.NEVER
        }

        return v
      }),
    })

    const allVersionsSchema = z.union([
      v1AndAbove,
      // for v0, we don't have a v key, so we have to check if the schema matches
      V0_VERSION.schema.transform((schema) => ({
        v: 0,
      })),
    ])

    const result = allVersionsSchema.safeParse(data)
    return result.success ? result.data.v : null
  },
```

packages/hoppscotch-data/package.json
```diff
@@ -33,13 +35,15 @@
...
     "vite": "^3.2.3"
   },
   "dependencies": {
     "fp-ts": "^2.11.10",
     "io-ts": "^2.2.16",
     "lodash": "^4.17.21",
-    "parser-ts": "^0.6.16"
+    "parser-ts": "^0.6.16",
+    "verzod": "^0.1.1",
+    "zod": "^3.22.2"
```
amk-dev, line 47: Not connected to this PR, but i see that the verzod package has zod as a dependency. but probably we can move it to a peer dependency ?

AndrewBastin: Will look into that. But you are right. Zod can be a peer dependency.

jamesgeorge007 (approved):
The GraphQL page crashing and Codemirrot syntax highlighting issues are to be handled separately.

amk-dev (approved):

PR #3468, opened 2023-10-23
docs: add one-click deploy to readme

jbgh:
### Description
Adds [one-click deploy](https://app.trydome.io/signup?package=hoppscotch) with Dome to readme

### Checks
- [X ] My pull request adheres to the code style of this project
- [ X] My code requires changes to the documentation
- [ X] I have updated the documentation as required
- [ X] All the tests have passed


jbgh:
@liyasthomas bumping for review. Let me know if you have any thoughts/feedback. Thanks!

liyasthomas:
I don't get what is the point of this integration in the context of our application. I mean, there is no issue with this, but I don't see any point in having this either.

Since we’ve chosen our [Official Docker images](https://hub.docker.com/u/hoppscotch) for distribution, I’m not endorsing any other platforms for deployment.

jbgh:
I'm sorry for not providing context initially! I had sent an email earlier, but I'll add context here as well. 

We are exploring ways to simplify the self-hosting process for users, especially for those who may not have experience in setting up their own infrastructure. 

This link allows users to who may not have the experience standing up their own infra to deploy it in one-click with having to mess with configuration. Might not be the necessary for your more technical audience, but could be more applicable for QA and PMs.

The PR introduces a button and a link for the Hopscotch one-click deploy. When clicked the user is prompted to create an account, then they can deploy the package to the cloud.

After deployment, users receive a link like this, which they can then CNAME to any custom domain:
https://b17747d81a2429e0c59de820a21be691859b90c7.dome.tools/

Feel free to test it. We're more than willing to maintain this if find it valuable for your users

PR #3474, opened 2023-10-25
feat: new banner service and added ability to bind additional services from other platforms

JoelJacobStephen:
### Ticket
Closes HFE-254

### Description
This PR introduces a new banner service to the service architecture and a new banner component that uses the banner service to render the banner. It also introduces support for additional services that are used in other platforms and can be bounded to the service container. 

### Banner Information
A banner contains:

1. Text
2. Alternate text [optional] ( can be used to display a shorter text when the screen size is small ) 
3. Banner type - ( info, warning and error )

### Types of Banner

1. Info Banner - Used to display any important information
image

2. Warning Banner - Used to display a warning message
image

3. Error Banner - Used to display an error message
image

### Objectives
- [x] Introduce a new Banner Service to the service architecture.
- [x] A new Banner component which handles the rendering aspect of the banner
- [x] Allow service container to bind services from other platforms 

### Checks


packages/hoppscotch-common/src/platform/index.ts
```diff
@@ -9,10 +9,12 @@ import { AnalyticsPlatformDef } from "./analytics"
 import { InterceptorsPlatformDef } from "./interceptors"
 import { HoppModule } from "~/modules"
 import { InspectorsPlatformDef } from "./inspectors"
+import { Service } from "dioc"
 
 export type PlatformDef = {
   ui?: UIPlatformDef
   addedHoppModules?: HoppModule[]
+  addedServices?: Array<typeof Service<unknown>>
```
amk-dev, line 17: Update this to fix the type issue in dioc module

```suggestion
  addedServices?: Array<typeof Service<unknown> & { ID: string }>
```

packages/hoppscotch-common/src/components/app/Banner.vue
```diff
@@ -0,0 +1,35 @@
+<template>
+  <div
+    v-if="banner"
```
amk-dev, line 3: do not use the v-if on the root component, it is the responsibility of the parent component to decide when to show this or not show this, the v-if should be used there

packages/hoppscotch-common/src/components/app/Banner.vue
```diff
@@ -0,0 +1,35 @@
...
+defineProps<{
+  banner: BannerContent
+}>()
+
+const getBannerColor = (type: BannerType): string =>
+  type === "error"
+    ? "bg-red-700"
+    : type === "warning"
+    ? "bg-yellow-700"
+    : "bg-error"
```
amk-dev, line 34: this reads better

```suggestion
const getBannerColor = (type: BannerType): string => {
  const bgColors: Record<BannerType, string> = {
    error: "bg-red-700",
    warning: "bg-yellow-700",
    info: "bg-info",
  }

  return bgColors[type]
}
```

packages/hoppscotch-common/src/services/banner.service.ts
```diff
@@ -0,0 +1,21 @@
+import { Service } from "dioc"
+import { ref } from "vue"
+import { getI18n } from "~/modules/i18n"
+
+export type BannerType = "info" | "warning" | "error"
+
+export type BannerContent = {
+  type: BannerType
+  text: (t: ReturnType<typeof getI18n>) => string
```
amk-dev, line 9: i see now we're enforcing the use of i18n for the text, but that might not be the best idea, because as far as the banner is concerned, it's just a string. it is the responsibility of the component using this banner to decide whether to use i18n or not.

i would love to hear your points if you think using i18n here is better, 

```suggestion
  text: string
```

packages/hoppscotch-common/src/services/banner.service.ts
```diff
@@ -0,0 +1,21 @@
...
+  // Can be used to display an alternate text when display size is small
+  alternateText?: (t: ReturnType<typeof getI18n>) => string
```
amk-dev, line 11: same as the last comment

```suggestion
  alternateText?: string
```

packages/hoppscotch-common/src/components/app/Banner.vue
```diff
@@ -0,0 +1,35 @@
...
+    class="relative flex items-center px-4 py-2 transition text-tiny group"
+    :class="getBannerColor(banner.type)"
+  >
+    <icon-lucide-info class="mr-2 text-white" />
+    <span class="text-secondaryDark">
+      <span v-if="banner.alternateText" class="md:hidden">
+        {{ banner.alternateText(t) }}
+      </span>
+      <span class="<md:hidden">
+        {{ banner.text(t) }}
```
amk-dev, line 13: As far as the banner is concerned, it does not need to know about the i18n, its job is to take the text and display it, it is who ever is using the banner, needs to care about whether the text should be internationalized.

i have suggested changes on the banner.service.ts also. let me know what you think. 

```suggestion
        {{ banner.text }}
```

packages/hoppscotch-common/src/components/app/Banner.vue
```diff
@@ -0,0 +1,35 @@
+<template>
+  <div
+    v-if="banner"
+    class="relative flex items-center px-4 py-2 transition text-tiny group"
+    :class="getBannerColor(banner.type)"
+  >
+    <icon-lucide-info class="mr-2 text-white" />
+    <span class="text-secondaryDark">
+      <span v-if="banner.alternateText" class="md:hidden">
+        {{ banner.alternateText(t) }}
```
amk-dev, line 10: for the reasons mentioned in the next comment

```suggestion
        {{ banner.alternateText }}
```

packages/hoppscotch-common/src/components/app/Header.vue
```diff
@@ -281,8 +282,22 @@ const showTeamsModal = ref(false)
 const breakpoints = useBreakpoints(breakpointsTailwind)
 const mdAndLarger = breakpoints.greater("md")
 
+const bannerService = useService(BannerService)
+const banner = bannerService.content
```
amk-dev, line 286: if you like destructuring, save some lines.

```suggestion
const { content: banner } = useService(BannerService)
```

packages/hoppscotch-common/src/services/banner.service.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+export class BannerService extends Service {
+  public static readonly ID = "BANNER_SERVICE"
+  constructor() {
+    super()
+  }
+
+  public content = ref<BannerContent>()
```
amk-dev, line 20: be more explicit about content can be null. ( using null here instead of the previous undefined, because it makes it a little bit more explicit that there might be no value )

```suggestion
  public content = ref<BannerContent | null>(null)
```

packages/hoppscotch-common/src/components/app/Header.vue
```diff
@@ -281,8 +282,22 @@ const showTeamsModal = ref(false)
...
+watch(
+  network,
+  () =>
+    (banner.value = network.isOnline
+      ? undefined
+      : {
+          type: "info",
+          text: () => t("helpers.offline"),
+          alternateText: () => t("helpers.offline_short"),
+        })
```
amk-dev, line 298: if we're making the changes mentioned in, `banner.service.ts` and `Banner.vue` 
```suggestion
          text: t("helpers.offline"),
          alternateText: t("helpers.offline_short"),
```

packages/hoppscotch-common/src/components/app/Header.vue
```diff
@@ -281,8 +282,22 @@ const showTeamsModal = ref(false)
...
 const network = reactive(useNetwork())
 
+watch(
+  network,
+  () =>
+    (banner.value = network.isOnline
+      ? undefined
+      : {
+          type: "info",
+          text: () => t("helpers.offline"),
```
amk-dev, line 296: if we're making the changes mentioned in `banner.service.ts` and `Banner.vue`

```suggestion
          text: t("helpers.offline"),
```

packages/hoppscotch-common/src/components/app/Header.vue
```diff
@@ -281,8 +282,22 @@ const showTeamsModal = ref(false)
...
+          alternateText: () => t("helpers.offline_short"),
```
amk-dev, line 297: same as the last comment.

```suggestion
          alternateText: t("helpers.offline_short"),
```

amk-dev (changes requested):
Look into the comments.

packages/hoppscotch-common/src/components/app/Banner.vue
jamesgeorge007: Since `banner` is received as a prop, this can be made a computed property, right?

packages/hoppscotch-common/src/services/banner.service.ts
JoelJacobStephen: I think the suggested approach is better.

packages/hoppscotch-common/src/components/app/Banner.vue
```diff
@@ -0,0 +1,34 @@
+<template>
+  <div
+    class="relative flex items-center px-4 py-2 transition text-tiny group"
+    :class="getBannerColor"
```
jamesgeorge007, line 4:
```suggestion
    :class="bannerColor"
```

Let's avoid the `get` prefix since by default computed properties are getter only. If it was a method we could have kept the `get` prefix.

packages/hoppscotch-common/src/components/app/Banner.vue
```diff
@@ -0,0 +1,34 @@
...
+
+<script setup lang="ts">
+import { computed } from "vue"
+import { BannerContent, BannerType } from "~/services/banner.service"
+
+const props = defineProps<{
+  banner: BannerContent
+}>()
+
+const getBannerColor = computed(() => {
```
jamesgeorge007, line 26:
```suggestion
const bannerColor = computed(() => {
```

packages/hoppscotch-common/src/modules/dioc.ts
```diff
@@ -34,5 +35,8 @@ export default <HoppModule>{
     app.use(diocPlugin, {
       container: serviceContainer,
     })
+    for (const service of platform.addedServices ?? []) {
```
jamesgeorge007, line 38: Could you elaborate on the purpose of this new field?

JoelJacobStephen: There can be additional services in specific platforms that are not present in common. Hence such additional services can be bound to the service container.

Eg. Licensing service is unique to selfhost-web and is not present in common. That can be bound using this.

packages/hoppscotch-common/src/components/app/Banner.vue
```diff
@@ -0,0 +1,34 @@
+<template>
+  <div
+    class="relative flex items-center px-4 py-2 transition text-tiny group"
```
jamesgeorge007, line 3: The `group` class doesn't seem to serve a purpose here. Also, could you verify if `relative ` and `transition` classes are necessary?

packages/hoppscotch-common/src/components/app/Banner.vue
```diff
@@ -0,0 +1,34 @@
...
+    :class="bannerColor"
+  >
+    <icon-lucide-info class="mr-2 text-white" />
```
jamesgeorge007, line 6: Shouldn't we use a different icon based on the banner type? We might need to determine a color that fits in well with the background.

packages/hoppscotch-common/src/components/app/Banner.vue
```diff
@@ -0,0 +1,34 @@
...
+
+const props = defineProps<{
+  banner: BannerContent
+}>()
+
+const bannerColor = computed(() => {
+  const bgColors: Record<BannerType, string> = {
+    error: "bg-red-700",
+    warning: "bg-yellow-700",
+    info: "bg-error",
```
jamesgeorge007, line 30: How were the background colors chosen? Is `bg-error` the intended color for the `info` variant?

packages/hoppscotch-common/src/components/app/Banner.vue
jamesgeorge007, line 1: I noticed the root element `div` for the `Announcement` component had a `role` attribute set to `alert`. How are we planning to treat accessibility in these cases? Since there are multiple variants, the role` of `alert` alone might not be sufficient. Or, it could be added at the places of usage via fallthrough attributes.

cc @liyasthomas

jamesgeorge007 (changes requested):

packages/hoppscotch-common/src/components/app/Banner.vue
JoelJacobStephen: According to the PRD, we show red for the error type, yellow for warning and for the info type, I used the previously present colour that was used in the Announcement component. I will now replace bg-error with bg-stone-800. Any suggestions for the type based color scheme are welcome.

packages/hoppscotch-common/src/components/app/Banner.vue
JoelJacobStephen: Do we need to have a different icons for different types of the banner ( error, warning, info )? In my opinion, I feel the info icon conveys the meaning as a banner is typically used to show important information irrespective of what the banner is being used for. However, I am open to your idea if its better and I can maybe add a new property to the banner where we can define custom icons or specific icons which map to the respective type of the banner.

JoelJacobStephen: I have selected darker colours for the background of the banner, so that the 	text-white	 color for the icon is displayed perfectly. Open to suggestions..

packages/hoppscotch-common/src/components/app/Banner.vue
JoelJacobStephen: I am open to this idea. I can maybe add a role attribute to the banner and assign specific roles based on the type of banner selected.

packages/hoppscotch-common/src/components/app/Banner.vue
jamesgeorge007: It can be subjective, but I've seen the `info` icon often associated with informational messages. For instance, [Vuetify](https://vuetifyjs.com/en/components/alerts/#type), [Material UI](https://mui.com/material-ui/react-alert/#description).

I'd love to hear what others think. Let's maybe revisit it when the use case comes in the future.

packages/hoppscotch-common/src/components/app/Banner.vue
jamesgeorge007: `bg-stone-800` doesn't go well with `text-secondaryDark`. Also, shouldn't the icon and the text color be the same?

info

packages/hoppscotch-common/src/components/app/Banner.vue
jamesgeorge007: Something along the lines of choosing `alert` for the `error` type` and `status` for the `warning` and `info` types. I'd prefer we go ahead with this after getting a consensus from the team.

packages/hoppscotch-common/src/components/app/Banner.vue
JoelJacobStephen: It will look like the attached screenshot. I will make both the icon and text to `text-secondaryDark`

image

packages/hoppscotch-common/src/components/app/Banner.vue
JoelJacobStephen: Yes, this is a good idea. I will implement this as such...

liyasthomas: 💯

packages/hoppscotch-common/src/components/app/Banner.vue
JoelJacobStephen: What icons would you prefer for error and warning banner type if we might go ahead with the suggestion?

liyasthomas: We use `info` icon. https://lucide.dev/icons/info

liyasthomas: @jamesgeorge007 @JoelJacobStephen

JoelJacobStephen: Okay. I will use these icons for the specific icon types.

JoelJacobStephen: @liyasthomas @jamesgeorge007 

These are the colors and icons used for each of the banner types:

1. Info Banner
image

2. Warning Banner
image

3. Error Banner
image

liyasthomas: Looks good to me.

packages/hoppscotch-common/src/components/app/Banner.vue
```diff
@@ -0,0 +1,44 @@
...
+}>()
+
+const bannerColor = computed(() => {
+  const bgColors: Record<BannerType, string> = {
+    error: "bg-red-700",
+    warning: "bg-yellow-700",
+    info: "bg-stone-800",
+  }
+  return bgColors[props.banner.type]
+})
```
AndrewBastin, line 34: Since `bgColors` is always the same, we should move this out of the `computed` block to prevent this from being recreated for all updates.

```suggestion
const bgColors: Record<BannerType, string> = {
    error: "bg-red-700",
    warning: "bg-yellow-700",
    info: "bg-stone-800",
  }
  
const bannerColor = computed(() => bgColors[props.banner.type])
```

You could even move `bgColors` out into a separate `script` (not setup) block and we can share the same reference across component instances, but since the component is only used once, there is not much difference.

P.S this also applies to `bannerRole` block below.

packages/hoppscotch-common/src/components/app/Header.vue
```diff
@@ -281,8 +282,21 @@ const showTeamsModal = ref(false)
...
+watch(
+  network,
+  () =>
+    (banner.value = network.isOnline
+      ? undefined
+      : {
+          type: "info",
+          text: t("helpers.offline"),
+          alternateText: t("helpers.offline_short"),
+        })
```
AndrewBastin, line 297: Readability.

```suggestion
  () => {
    if (network.isOnline) {
      banner.value = {
          type: "info",
          text: t("helpers.offline"),
          alternateText: t("helpers.offline_short"),
      }
    }
  }
```

AndrewBastin (changes requested):

packages/hoppscotch-common/src/components/app/Banner.vue
```diff
@@ -0,0 +1,50 @@
+<template>
+  <div
+    class="flex items-center px-4 py-2 text-tiny"
+    :class="bannerColor"
+    :role="bannerRole"
```
jamesgeorge007, line 5:
```suggestion
    :role="bannerRole"
    class="flex items-center px-4 py-2 text-tiny"
    :class="bannerColor"
```

We can keep `class` towards the end of the list; that way there's a clear distinction.

packages/hoppscotch-common/src/components/app/Banner.vue
```diff
@@ -0,0 +1,50 @@
...
+  >
+    <icon-lucide-info
+      v-if="banner.type === 'info'"
+      class="mr-2 text-secondaryDark"
+    />
+    <icon-lucide-alert-circle
+      v-else-if="banner.type === 'warning'"
+      class="mr-2 text-secondaryDark"
+    />
+    <icon-lucide-alert-triangle v-else class="mr-2 text-secondaryDark" />
```
jamesgeorge007, line 15: We could simplify this by having an object with a map of banner types and the component references, and we can dynamically render the icon, [Ref](https://vuejs.org/guide/essentials/component-basics.html#dynamic-components).

jamesgeorge007 (commented):
Aren't we using `text-white` as the color for the icon and the text?

packages/hoppscotch-common/src/components/app/Banner.vue
```diff
@@ -0,0 +1,50 @@
...
+  banner: BannerContent
+}>()
+
+const bgColors: Record<BannerType, string> = {
+  error: "bg-red-700",
+  warning: "bg-yellow-700",
+  info: "bg-stone-800",
+}
+
+const role: Record<BannerType, string> = {
```
jamesgeorge007, line 42: How about calling this `ariaRoles` that'll give more context about the purpose?

packages/hoppscotch-common/src/components/app/Banner.vue
jamesgeorge007: Isn't white the one fitting well in this context?

packages/hoppscotch-common/src/services/banner.service.ts
```diff
@@ -0,0 +1,20 @@
...
+export type BannerType = "info" | "warning" | "error"
+
+export type BannerContent = {
+  type: BannerType
+  text: string
+  // Can be used to display an alternate text when display size is small
+  alternateText?: string
+}
+
+export class BannerService extends Service {
```
AndrewBastin, line 13: Provide a small sentence that defines what this Service does as a JSDoc comment.
See: https://github.com/hoppscotch/hoppscotch/blob/73b3ff8e411705dd842abb3088b1a20bc9aa7df5/packages/hoppscotch-common/src/services/interceptor.service.ts#L123-L127

packages/hoppscotch-common/src/services/banner.service.ts
```diff
@@ -0,0 +1,20 @@
...
+  public static readonly ID = "BANNER_SERVICE"
+  constructor() {
+    super()
+  }
```
AndrewBastin, line 17: You can remove unused constructor.

packages/hoppscotch-common/src/services/banner.service.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  public content = ref<BannerContent | undefined>(undefined)
```
AndrewBastin, line 19: `public` properties of services should have JSDoc comments describing what it does.

AndrewBastin (changes requested):

packages/hoppscotch-common/src/components/app/Header.vue
```diff
@@ -281,8 +282,19 @@ const showTeamsModal = ref(false)
...
 const network = reactive(useNetwork())
 
+watch(network, () => {
+  if (!network.isOnline) {
+    banner.value = {
+      type: "info",
+      text: t("helpers.offline"),
+      alternateText: t("helpers.offline_short"),
+    }
+  } else banner.value = undefined
```
jamesgeorge007, line 295:
```suggestion
  if (network.isOnline) {
    banner.value = undefined;
    return;
  }

  banner.value = {
    type: "info",
    text: t("helpers.offline"),
    alternateText: t("helpers.offline_short"),
  };
```

How about the early return approach?

packages/hoppscotch-common/src/services/banner.service.ts
```diff
@@ -0,0 +1,28 @@
...
+ * It can used to display information, warnings or errors.
+ */
+export class BannerService extends Service {
+  public static readonly ID = "BANNER_SERVICE"
+
+  /**
+   * This is a reactive variable that can be used to set the contents of the banner
+   * and use it to render the banner on components.
+   */
+  public content = ref<BannerContent | undefined>(undefined)
```
jamesgeorge007, line 27: Doesn't `undefined` indicate the absence of value. Is `null` more appropriate here, signifying emptiness 🤔

jamesgeorge007:
@JoelJacobStephen, can you add relevant screenshots highlighting the different banner types to the PR description?

JoelJacobStephen:
@jamesgeorge007 Updated the PR description with screenshots.

jamesgeorge007 (approved):

amk-dev (approved):
LGTM 👍🏼

PR #3476, opened 2023-10-26
feat: introduction of shared-requests

balub:
Closes HBE-289

### Description

In this PR we introduce a new module into the backend along with a new `SharedRequest` type in the gql layer. This will now replace the previous `Shortcode` module/type. The previous `Shortcode` type/module has been marked `deprecated` and will be removed shortly.

```
type SharedRequest {
	id: ID!
	request: String!
	properties: String? // null if shortcode
	createdOn: DateTime!
}
```

### Checks

### Additional Information

The `Shortcode` module and type is now `deprecated`.


packages/hoppscotch-backend/src/shared-request/shared-request.resolver.ts
```diff
@@ -0,0 +1,130 @@
...
+
+    if (E.isLeft(result)) throwErr(result.left);
+    return result.right;
+  }
+
+  @Query(() => [SharedRequest], {
+    description: 'List all shared-request the current user has generated',
+  })
+  @UseGuards(GqlAuthGuard)
+  async mySharedRequest(
```
mirarifhasan, line 52:
```suggestion
  async mySharedRequests(
```

packages/hoppscotch-backend/src/shared-request/shared-request.service.ts
```diff
@@ -0,0 +1,217 @@
...
+  /**
+   * Create a new SharedRequest
+   *
+   * @param request JSON string of request details
+   * @param properties JSON string of embed properties, if present
+   * @returns Either of SharedRequest or error
+   */
+  async createSharedRequest(
+    request: string,
+    properties: string | null,
```
mirarifhasan, line 128: To avoid undefined
```suggestion
    properties: string | null = null,
```

packages/hoppscotch-backend/src/shared-request/shared-requests.model.ts
```diff
@@ -0,0 +1,25 @@
...
+  request: string;
+
+  @Field({
+    description: 'JSON string representing the properties for an embed',
+    nullable: true,
+  })
+  properties: string;
+
+  @Field({
+    description: 'Timestamp of when the Shortcode was created',
```
mirarifhasan, line 22:
```suggestion
    description: 'Timestamp of when the SharedRequest created',
```

packages/hoppscotch-backend/src/shared-request/shared-request.service.ts
```diff
@@ -0,0 +1,217 @@
...
+    const requestData = stringToJson(request);
+    if (E.isLeft(requestData))
+      return E.left(SHARED_REQUEST_INVALID_REQUEST_JSON);
+
+    let propertiesData;
+    if (!properties) propertiesData = undefined;
+    const parsedProperties = stringToJson(properties);
+    if (E.isLeft(parsedProperties))
+      return E.left(SHARED_REQUEST_INVALID_PROPERTIES_JSON);
+    propertiesData = parsedProperties.right;
```
mirarifhasan, line 140: I think this will be more readable. Am I missing any corner case?

see my next review as well to better understand in `this.prisma.shortcode.create`

```suggestion
    const parsedProperties = stringToJson(properties);
    if (E.isLeft(parsedProperties))
      return E.left(SHARED_REQUEST_INVALID_PROPERTIES_JSON);
```

mirarifhasan (changes requested):

mirarifhasan (commented):
@balub Are we providing public query `sharedRequest` for both shortcode/embed?

packages/hoppscotch-backend/prisma/schema.prisma
```diff
@@ -67,10 +67,21 @@ model TeamRequest {
...
+//   createdOn  DateTime @default(now())
+
+//   @@unique(fields: [id, creatorUid], name: "creator_uid_shortcode_unique")
+// }
+
 model Shortcode {
-  id         String   @id
+  id         String   @id @unique
   request    Json
+  properties Json? // if properties is null, SharedRequest is an embed else shortcode
```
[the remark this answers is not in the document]
balub: So, we are going to refactor shortcodes into share-requests which now contain embeds and shortcodes this is a big change on the FE. From the BE perspective nothing really changes the new shared-requests are built on top of the existing shortcodes table with the addition of the `properties` property, which if null means the shared-request is a shortcode and if not null it is an embed.

balub:
Yes, we need a public query to maintain backwards compatibility with previously created shortcodes.

packages/hoppscotch-backend/prisma/schema.prisma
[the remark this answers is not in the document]
mirarifhasan: Then, I think, the comment should be like this?
```suggestion
  properties Json? // if properties is null, SharedRequest is an shortcode else embed
```

packages/hoppscotch-backend/src/shared-request/shared-request.service.ts
```diff
@@ -0,0 +1,217 @@
...
+   * Converts a Prisma SharedRequest type into the SharedRequest model
+   *
+   * @param sharedRequestInfo Prisma SharedRequest type
+   * @returns GQL SharedRequest
+   */
+  private cast(sharedRequestInfo: DBSharedRequest): SharedRequest {
+    return <SharedRequest>{
+      id: sharedRequestInfo.id,
+      request: JSON.stringify(sharedRequestInfo.request),
+      properties: JSON.stringify(sharedRequestInfo.properties),
```
mirarifhasan, line 70: If `properties` is `null`, then we should avoid to `JSON.stringify` it.

Screenshot 2023-10-27 at 6 10 18 AM

packages/hoppscotch-backend/src/shared-request/shared-request.service.ts
```diff
@@ -0,0 +1,259 @@
...
+   */
+  async updateSharedRequest(
+    sharedRequestID: string,
+    uid: string,
+    updatedProps: string,
+  ) {
+    if (!updatedProps) return E.left(SHARED_REQUEST_PROPERTIES_NOT_FOUND);
+
+    const parsedProperties = stringToJson(updatedProps);
+    if (E.isLeft(parsedProperties))
```
mirarifhasan, line 233: There is a corner case for specific input for `properties` (null as string - see screenshot)

Screenshot 2023-11-02 at 11 58 50 AM

```suggestion
    if (E.isLeft(parsedProperties) || !parsedProperties.right)
```

mirarifhasan (changes requested):

packages/hoppscotch-backend/src/shared-request/shared-request.service.ts
```diff
@@ -0,0 +1,259 @@
...
+    try {
+      const updatedSharedRequest = await this.prisma.shortcode.update({
+        where: {
+          creator_uid_shortcode_unique: {
+            creatorUid: uid,
+            id: sharedRequestID,
+          },
+        },
+        data: {
+          properties: updatedProps,
```
mirarifhasan, line 245:
```suggestion
          properties: parsedProperties.right,
```

mirarifhasan (changes requested):

packages/hoppscotch-backend/prisma/schema.prisma
```diff
@@ -68,9 +68,11 @@ model TeamRequest {
 }
 
 model Shortcode {
-  id         String   @id
+  id         String   @id @unique
   request    Json
+  properties Json? // if properties is null, SharedRequest is an shortcode else embed
   creatorUid String?
+  User       User?    @relation(fields: [creatorUid], references: [uid])
   createdOn  DateTime @default(now())
```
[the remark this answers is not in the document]
ankitsridhar16: We should add an updatedOn column. Usage wise very easy to sort based on last accessed or something.

packages/hoppscotch-backend/prisma/schema.prisma
```diff
@@ -68,9 +68,11 @@ model TeamRequest {
 }
 
 model Shortcode {
-  id         String   @id
+  id         String   @id @unique
   request    Json
+  properties Json? // if properties is null, SharedRequest is an shortcode else embed
```
ankitsridhar16, line 73: if this will be always embeds then why name it properties instead of embedsInfo or something like that

balub: I gave it a generic name because while the property may belong to that of an embed in the DB layer they are all stored in the same table, hence I gave it a generic name that conveys the meaning.

ankitsridhar16: so embeds are a property of shortcodes and not the other way around ?

packages/hoppscotch-backend/prisma/schema.prisma
```diff
@@ -102,6 +104,7 @@ model User {
   currentGQLSession  Json?
   createdOn          DateTime            @default(now()) @db.Timestamp(3)
   invitedUsers       InvitedUsers[]
+  shortcodes         Shortcode[]
 }
```
ankitsridhar16, line 108: cant users query shared requests from `user` type like shortcodes

mirarifhasan (commented):
@balub 
1. Undo the docker-compose.yml file changes
2. Rebase & solve conflicts

mirarifhasan (approved):

packages/hoppscotch-backend/src/shortcode/shortcode.model.ts
```diff
@@ -13,7 +13,13 @@ export class Shortcode {
...
 
   @Field({
-    description: 'Timestamp of when the Shortcode was created',
+    description: 'JSON string representing the properties for an embed',
+    nullable: true,
+  })
+  properties: string;
+
+  @Field({
+    description: 'Timestamp of when the SharedRequest was created',
```
ankitsridhar16, line 22: Shortcode or SharedRequest ?

mirarifhasan (approved):

packages/hoppscotch-backend/src/shortcode/shortcode.service.ts
```diff
@@ -104,35 +112,44 @@ export class ShortcodeService implements UserDataHandler, OnModuleInit {
...
    * @param request JSON string of request details
-   * @param userUID user UID, if present
+   * @param userInfo user UI
+   * @param properties JSON string of embed properties, if present
    * @returns Either of ShortCode or error
    */
-  async createShortcode(request: string, userUID: string | null) {
-    const shortcodeData = stringToJson(request);
-    if (E.isLeft(shortcodeData)) return E.left(SHORTCODE_INVALID_JSON);
+  async createShortcode(
```
mirarifhasan, line 119: If we pass `"null"` stringify null value in the properties, should we treat as invalid input?

Screenshot 2023-11-07 at 11 17 37 AM

packages/hoppscotch-backend/src/shortcode/shortcode.resolver.ts
```diff
@@ -64,20 +59,53 @@ export class ShortcodeResolver {
...
+
+    if (E.isLeft(result)) throwErr(result.left);
+    return result.right;
+  }
+
+  @Mutation(() => Shortcode, {
+    description: 'Update a user generated Shortcode',
+  })
+  @UseGuards(GqlAuthGuard)
+  async updateShortcode(
```
ankitsridhar16, line 91: do you think `updateShortcodeEmbed` will be more clear ?

PR #3478, opened 2023-10-26
feat: group requests by tag when importing OpenAPI

### LINKED ISSUE #3411: [feature]: Group request by tag when importing OpenAPI

### Summary

Requests in OpenAPI can be grouped  using tags, for example:
image

But Hoppscotch imports them as a one flat structure:
image

It will be a good idea to make a folder for each tag in OpenAPI document

It can be hard to work with a big ungrouped collection which is generated form OpenAPI

pierre-borckmans:
Closes #3411

### Description
When importing openapi, this PR makes sure to create a Folder for each tag present in the openapi file.
The endpoints are then added in the corresponding folder.
Endpoints without a tag are added to the root of the collection (same behaviour as before).

### Checks

### Additional Information
One caveat is that if multiple tags are assigned to a given endpoint, the same request will be populated twice in the corresponding 2 hoppscotch folders.

### DIFF
--- a/packages/hoppscotch-common/src/helpers/import-export/import/openapi.ts
+++ b/packages/hoppscotch-common/src/helpers/import-export/import/openapi.ts
@@ -543,7 +543,8 @@ const parseOpenAPIUrl = (
 const convertPathToHoppReqs = (
   doc: OpenAPI.Document,
   pathName: string,
-  pathObj: OpenAPIPathInfoType
+  pathObj: OpenAPIPathInfoType,
+  filterTag?: string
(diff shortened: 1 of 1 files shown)


nivedin:
Thank you so much for your contribution and for opening this pull request to fix the issue. We really appreciate your effort and the time you've invested in this.
I wanted to let you know that this [PR](https://github.com/hoppscotch/hoppscotch/pull/3846) fixed the issue and has been merged. Your contribution is valuable to us, and we encourage you to keep contributing in the future.

PR #3486, opened 2023-11-01
feat: shared request

nivedin:
Closes HFE-256

### Description
This PR introduces a new way of sharing request. Previously the user created  a shortcode that contains a link of the request now a new flow is introduced.

Users can now share requests from:
- The dropdown menu on the save button
- The more options menu for each request in the collection panel

Upon clicking the "Share Request" option, a modal window opens, offering different ways to share the request:
- Embed (WIP - HFE-250)
- Button
- Link

image

After clicking the "Create" button, a shared request is generated, and the user can modify the selected option or change the widget style in the next section.

image

Subsequently, the user can copy the widget style in HTML, Markdown, or simply copy the link.

Users can now access all the shared requests in the newly introduced "Shared Request" section in the sidebar.

image

The user has the option to open the request in a new tab, delete, or customize the created shared request from this section.

### Checks
- [x] Updates required to documentation are tracked in [HP-184](https://linear.app/hopp/issue/HP-184/documentation).

### Additional Information
- This PR removes the shortcode and it's related components.
- The shared request is defined as shortcode in the BE so the related adapter and BE queries is named shortcode and the FE releated components as SharedRequest.


jamesgeorge007 (changes requested):
- The `Share Request` modal highlights the widget corresponding to the previously created shared request entry. Is this the expected behavior? I noticed creating shared requests with the same widget multiple times for a request is permitted. But do we want to prevent creating further entries if shared request entries have been created under all the widgets for a request? In that case, we can highlight all the selected widgets for a request with associated shared request entries. For others, the selection falls back to `Embed`.

  https://github.com/hoppscotch/hoppscotch/assets/25279263/3ea6e68c-2d2c-4e36-9f63-970769892641

- Creating a shared request overrides the widgets selection in the previously created entries. This happens with the customize flow as well. Changing a widget type in one shared request entry affects others.

  https://github.com/hoppscotch/hoppscotch/assets/25279263/e70bb2da-64d7-4f8d-ad22-34cc88abfb0b

- Performing a full reload resets the entries to the default widget of `Embed`.

  https://github.com/hoppscotch/hoppscotch/assets/25279263/e1fa82b8-4fa1-43df-be12-7e1b761f0cb9

- The create shared request option in the request dropdown as part of the collection entries should be disabled/ the log-in modal should be displayed if the user is not logged in. This also applies to the delete option for shared request entries.

  https://github.com/hoppscotch/hoppscotch/assets/25279263/77c1949f-fa88-4833-b8fa-611fea1885cf

- If possible can we unify the share request modal flows into a single modal component? Instead of dismissing the first modal and showing the subsequent one, we can offer a smooth experience with a step-based flow in a single modal component. The `customize` option for the shared request entries can also convey it's the second step that needs to be rendered via a prop.

packages/hoppscotch-common/src/components/share/CustomizeModal.vue
```diff
@@ -0,0 +1,454 @@
...
+
+const baseURL = import.meta.env.VITE_SHORTCODE_BASE_URL ?? "https://hopp.sh"
+
+const copyEmbed = () => {
+  const options = embedOptions.value
+  const enabledEmbedOptions = options.tabs
+    .filter((tab) => tab.enabled)
+    .map((tab) => tab.value)
+    .toString()
+  return `<iframe src="${baseURL}/e/${props.request?.id}/${enabledEmbedOptions}' style='width: 100%; height: 500px; border: 0; border-radius: 4px; overflow: hidden;'></iframe>`
```
jamesgeorge007, line 395:
```suggestion
  return `<iframe src="${baseURL}/r/${props.request?.id}/${enabledEmbedOptions}" style='width: 100%; height: 500px; border: 0; border-radius: 4px; overflow: hidden;"></iframe>`
```

Currently, the embed gives a `404`. I think the route here should be `/r` with the embed options supplied via query params. Also, shouldn't tabs in the embed displayed be based on the above options?

liyasthomas: `/r/*` is reserved for shortcodes and it's redirected to the home page with it's resolved request setup.

We're hoping to reserve `/e/*` for embeds to render the embedded page.

jamesgeorge007: Gotcha, shouldn't there be a new page added under the `e` route?

liyasthomas: Yep, but that's part of another ticket. Out of scope of this feature PR.

liyasthomas: But can have a placeholder blank page in `/e` I guess. @nivedin can you add a blank page in `/e` path along with this PR?

jamesgeorge007: Or maybe mention it in the PR description such that the embed view implementation will be handled separately with a link to the corresponding Linear ticket.

packages/hoppscotch-common/src/components/share/index.vue
```diff
@@ -0,0 +1,294 @@
...
+        :section="t('tab.shared_requests')"
+        :is-only-personal="true"
+      />
+    </div>
+    <div
+      class="sticky top-sidebarPrimaryStickyFold z-10 flex flex-1 flex-shrink-0 justify-end overflow-x-auto border-b border-dividerLight bg-primary"
+    >
+      <HoppButtonSecondary
+        v-tippy="{ theme: 'tooltip' }"
+        to="https://docs.hoppscotch.io/documentation/features/shared-request"
```
jamesgeorge007, line 16: Are the documentation updates tracked on Linear? If not, can you create an issue and mention it next to the point under `Checks`?

packages/hoppscotch-common/src/pages/e/_id.vue
jamesgeorge007, line 1: The generated URL refers to a nested page under `/e/_id` with the enabled embed options `/e/_id/_embedOptions`, right? If no options are enabled, it'd be just `/e/_id`. Are we keeping it like this or always pointing to the nested page (keeping a placeholder if no options are enabled for the `embedOptions` route)?

packages/hoppscotch-common/src/components/share/index.vue
```diff
@@ -0,0 +1,294 @@
+<template>
+  <div>
+    <div
+      class="sticky top-0 z-10 flex flex-shrink-0 flex-col overflow-x-auto bg-primary"
+    >
+      <WorkspaceCurrent
+        :section="t('tab.shared_requests')"
+        :is-only-personal="true"
```
jamesgeorge007, line 8: The current workspace title shows up as `My workspace > Shared Requests` all the time. This is intentional, right? I noticed that shared requests are always associated with a user.

jamesgeorge007 (commented):
- For the shared request entries, is it intentional that the customize widgets modal shows `Embed` regardless of the type opted by the user initially? Since we say, `Select a widget, you can change and customize this later`, doesn't that convey that the selection will be persisted? I can see that the API doesn't send/take in the shared request type. In that case, we might need to convey the same in some form.

  https://github.com/hoppscotch/hoppscotch/assets/25279263/3ca10171-0c60-4ad6-af0b-1ed35094a324

- For the empty state when the user is not logged in, how about asking them to log in to view shared requests? For instance, we follow a similar approach in the `profile` page.
  
  image
  
  image

packages/hoppscotch-common/src/components/share/index.vue
nivedin: yup this is intentional since shared requests are associated with a user only.

packages/hoppscotch-common/src/pages/e/_id.vue
nivedin: the `/e/_id `is just a placeholder page at the moment i think we can tackle this part when implementing embeds.

nivedin:
@jamesgeorge007  so shared request is only created once and the customize flow doesn't update the shared request in the BE we only change the format of the sharable link in FE either a embed, link or a button. Also we don't persist the type of shared request atm user can copy the type of shared request any time without hitting the BE.

jamesgeorge007:
@nivedin, I've made a few changes in https://github.com/hoppscotch/hoppscotch/pull/3486/commits/ede2f6d4c35925aea894a0cf67a2ba54f3b76093. Can you take a look as you get the chance? Also, I've resolved the conflicts and rebased the branch.

jamesgeorge007 (approved):

jamesgeorge007:
> > @jamesgeorge007 so shared request is only created once and the customize flow doesn't update the shared request in the BE we only change the format of the sharable link in FE either a embed, link or a button. Also we don't persist the type of shared request atm user can copy the type of shared request any time without hitting the BE.
> 
> I'll get a confirmation on this. Cause, in the case of embeds, the config part is customizable hence this needs to be saved to the BE.

Discussed separately, persistence will be handled in subsequent iterations.

PR #3488, opened 2023-11-02
fix: graphql page crashing and broken syntax highlighting 

amk-dev:
**Before**

Our graphql page was crashing and syntax highlighting was not working properly. The error was due to having multiple  versions of `@lezer/highlight` package being installed. `hoppscotch-ui`  uses `^1.0.0`  and `hoppscotch-common`  uses `^1.1.6`

**After**

We can fix the problem by bumping `@lezer/highlight` to `1.1.6`, but since `@lezer/highlight` is an unused dependency in `hoppscotch-ui`, we decided to remove it.

### DIFF
--- a/packages/hoppscotch-ui/package.json
+++ b/packages/hoppscotch-ui/package.json
@@ -21,7 +21,6 @@
     "@fontsource-variable/material-symbols-rounded": "^5.0.5",
     "@fontsource-variable/roboto-mono": "^5.0.6",
     "@hoppscotch/vue-toasted": "^0.1.0",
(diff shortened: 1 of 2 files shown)


anwarulislam:
@amk-dev I think we can remove `@lazer/highlight` from `hoppscotch-ui` package as it is not being used there.

amk-dev:
> @amk-dev I think we can remove `@lazer/highlight` from `hoppscotch-ui` package as it is not being used there.

Ohh, we can do that then. i think there are also other packages that are not used in there. we can remove them also.

AndrewBastin (approved):

PR #3493, opened 2023-11-03
refactor: move persistence logic into a dedicated service

jamesgeorge007:
### Description

This PR moves the `localpersistence` logic to a dedicated service and adds `Zod` based schema validations for the data read from `localStorage` within various methods.

Closes HFE-248.

### Changes

- Introduces a new `persistence service` compiling the logic from the `localpersistence` module.
- Replace the instances of helper function invocations with the service methods.
- Adds schema based validations for the data read from `localStorage`. An error toast is shown on the course of schema parse failure along with keeping a backup in `localStorage`.
- Adds associated test suite for the newly added service.
- Adds `verzod` as a dependency under the `@hoppscotch/common` package.

### Checks


packages/hoppscotch-common/src/services/persistence/validation-schemas/index.ts
```diff
@@ -0,0 +1,444 @@
...
+  postwoman: z.optional(
+    z.object({
+      settings: z.optional(SettingsDefSchema),
+      //! Versioned entities
+      collections: z.optional(z.array(HoppRESTCollectionSchema)),
+      collectionsGraphql: z.optional(z.array(HoppGQLCollectionSchema)),
+      environments: z.optional(z.array(entityReference(Environment))),
+    })
+  ),
+})
```
jamesgeorge007, line 80: There wasn't a well-defined type for the data associated with the key `vuex` read from `localStorage`. I've made the schema not strict and accept unknown keys, appreciate suggestions.

packages/hoppscotch-common/src/services/persistence/validation-schemas/index.ts
```diff
@@ -0,0 +1,467 @@
...
+    .strict(),
+])
+
+export const SETTINGS_SCHEMA = z.union([
+  z.object({}).strict(),
+  SettingsDefSchema.extend({
+    EXTENSIONS_ENABLED: z.optional(z.boolean()),
+    PROXY_ENABLED: z.optional(z.boolean()),
+  }),
+])
```
jamesgeorge007, line 101: Similar to `vuex`, the data associated with the key `settings` read from `localStorage` didn't have a well-defined type to be inferred from. From the function [here](https://github.com/jamesgeorge007/hoppscotch/blob/ae079af0cd5ab7b8f50a675a21ed730dd09b1859/packages/hoppscotch-common/src/newstore/settings.ts#L162-L182) as invoked within [setupSettingsPersistence method](https://github.com/jamesgeorge007/hoppscotch/blob/refactor/persistence-service/packages/hoppscotch-common/src/services/persistence/index.ts#L244-L273), there were two additional properties expected in the incoming object.

packages/hoppscotch-common/src/services/persistence/__tests__/index.spec.ts
```diff
@@ -0,0 +1,2027 @@
...
+    const toast = useToast()
+    const service = getServiceInstance()
+
+    // @ts-expect-error Testing private method
+    service.showErrorToast(testKey)
+
+    expect(toast.error).toHaveBeenCalledWith(
+      `There's a mismatch with the expected schema for the value corresponding to ${testKey} read from localStorage, keeping a backup in ${testKey}-backup`
+    )
+  })
```
jamesgeorge007, line 138: This test case is failing atm. Am I missing something wrt mocking? `showErrorToast` method -> [Ref](https://github.com/jamesgeorge007/hoppscotch/blob/refactor/persistence-service/packages/hoppscotch-common/src/services/persistence/index.ts#L90-L95).

packages/hoppscotch-common/src/services/persistence/validation-schemas/index.ts
AndrewBastin: Yep, the old schema was very iffy, most of our users should have migrated over to the new one, so we just need to worry about the existing legacy conversions tbh.

PR #3504, opened 2023-11-07
feat: introducing shortcode into admin module

balub:
Closes HBE-289

### Description

In this PR we introduce queries and mutations into the `Infra` and `Admin` types respectively to be able to view and delete shortcodes present in the database. The `allShortcodes` query in the `Infra` type returns a new `Shortcode` type with `creatorUid` and `creatorEmail` properties added in addition to the usual `Shortcode` properties to satisfy requirements stated in ticket.

```
type ShortcodeWithUserEmail {
	id: ID!
	request: String!
	properties: String?
	createdOn: DateTime!
        creatorUid: String!
       creatorEmail: String!
}
```

### Checks


.vscode/extensions.json
mirarifhasan, line 1: Are we sure, what to delete this?

packages/hoppscotch-backend/src/admin/admin.service.ts
```diff
@@ -432,4 +434,30 @@ export class AdminService {
...
     return E.right(teamInvite.right);
   }
+
+  /**
+   * Fetch all created ShortCodes
+   *
+   * @param args Pagination arguments
+   * @returns ShortcodeWithUserEmail
+   */
+  async fetchAllShortcodes(cursorID: string, take: number, userEmail = null) {
```
mirarifhasan, line 444:
```suggestion
  async fetchAllShortcodes(cursorID: string, take: number, userEmail: string = null) {
```

packages/hoppscotch-backend/src/shortcode/shortcode.service.ts
```diff
@@ -263,4 +283,55 @@ export class ShortcodeService implements UserDataHandler, OnModuleInit {
...
+          request: JSON.stringify(code.request),
+          properties:
+            code.embedProperties != null
+              ? JSON.stringify(code.embedProperties)
+              : null,
+          createdOn: code.createdOn,
+          creator: {
+            uid: code.User.uid,
+            email: code.User.email,
+          },
```
mirarifhasan, line 330: Can we create a `@ResolveField` for this instead of joining? What do you think?

mirarifhasan, line 330: And just to backward compatibility, can we allow that `creator` can be nullable?

mirarifhasan (changes requested):

.vscode/extensions.json
balub: Yup, its just some settings from my vs code instance.

packages/hoppscotch-backend/src/shortcode/shortcode.service.ts
balub: Nope, I don't think it's necessary. By creating this as a field resolver it will be available to all the created shortcodes, we only want this property to be available to that one specific query in the admin module.

mirarifhasan (approved):

docker-compose.yml
```diff
@@ -107,7 +107,7 @@ services:
     build:
       dockerfile: packages/hoppscotch-backend/Dockerfile
       context: .
-      target: prod
+      target: dev
```
ankitsridhar16, line 110: I think changes are required here in docker-compose

PR #3531, opened 2023-11-10
fix: Oauth 2.0 authentication type is breaking

amk-dev:
**Before**

Currently authentication type Oauth is breaking when trying to generate token.  The breakage is caused by an incorrect destructuring we do when extracting info from OIDC discovery URL.

**After**

Added necessery validations when parsing OIDC discovery URL content.

Since we haven't touched the oauth code in a long time. I've done some quick improvements to the code. 

* basic migration of `oauth.js` to typescript
* move the oauth redirect uri handling logic from `index.vue` to a seperate page `oauth.vue`
* use zod validations & improve error handling


packages/hoppscotch-common/src/pages/oauth.vue
liyasthomas, line 1: Let's do this inside the Authorization Tab itself. Don't need a separate page.

liyasthomas (changes requested):
~I don't encourage the UX to navigate to another page while generating a Token.~

AndrewBastin (approved):
lgtm, although, the error messages should have a mechanism to show proper i18n friendly error messages, I will add it in from the branch.

Issue #3381, opened 2023-09-21
[bug]: Workspaces/Team Environments get's deleted by changing Team


RaHehl:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

I c
![image](https://github.com/hoppscotch/hoppscotch/assets/7577984/21d32d13-e436-45b2-ae8a-59e5a4fa494b)

### Steps to reproduce

- I create a environment with var's for a Team, it get's send to the server via graphql

![image](https://github.com/hoppscotch/hoppscotch/assets/7577984/21d32d13-e436-45b2-ae8a-59e5a4fa494b)

- I select another team
- i select the prior team
- my environments are gone

### Environment

Production

### Version

Self-hosted

RaHehl:
it's a little strange. I created the bug based on repeated feedback from an internal user. After I was able to reproduce the behavior even in incognito mode. env created team changed, everything empty. Now suddenly all the previously disappeared env are back and I can no longer reproduce it.

RaHehl:
@balub Right, that was the behavior I had, but now I can't reproduce it.

But I just noticed something: maybe it has something to do with bug #3393. I originally attached a screenshot of the request here, but when I saw that a CreateUserEnvironment was visible in the image instead of a CreateTeamEnvironment, I took the screenshot out again because I thought I had included it incorrectly.

RaHehl:
A user saw something like this in the dev tools when it occurred:
operationName: GetTeamEnvironments
Throttlinger 419

balub:
@RaHehl I apologise for the time it took to get to this issue, but after looking into this it seems certain field-resolvers for certain modules do not have any rate-limiting on them. In addition to that maybe the lack of extracting the source IP of the requests in the `GqlThrottlerGuard` could cause issues when working with proxies. Will run some tests to confirm the same.

balub:
@RaHehl are you still encountering this issue, if not I believe its safe to close this issue ?

RaHehl:
@balub At least I haven't had any reports from users about this for a while, so I agree with you closing this

PR #3535, opened 2023-11-10
feat(backend): add `getTracker` method for IP extraction

RaHehl:
(maybe) Closes # 3381

### Description
For bug #3381, the cause appears to be @nestjs/throttler, at least according to the report of an internal user who saw something with "throttling" in the log when the team environments disappeared. I suspect this is because the Hoppscotch API runs in Kubernetes and therefore there is an ingress proxy running between the API and the client. I don't know anything about NestJS, I think the solution to this lies in the direction of this PR.


balub:
@RaHehl I believe adding the `getTracker` function only in the `gql-throttler.guard.ts` file is enough, so can you remove it from the auth-guard.

balub (changes requested):
@RaHehl I believe adding the getTracker function only in the gql-throttler.guard.ts file is enough, so can you remove it from the auth-guard.

RaHehl:
> @RaHehl I believe adding the getTracker function only in the gql-throttler.guard.ts file is enough, so can you remove it from the auth-guard.

done

balub (approved):

mirarifhasan (approved):
LGTM

PR #3537, opened 2023-11-10
feat: introducing shared requests to admin dashboard

JoelJacobStephen:
### Ticket

- Closes HFE-280
- Depends on #3504 

### Description
This PR introduces Shared Requests to the Admin Dashboard. The Individual User's page is now modified to include the shared requests created by that particular user. The new Shared Requests module allows admins to:

1. View shared requests created by any user
2. Open the shared requests in a new tab
3. Copy the link of the selected shared request 
4. Delete any shared request

### Screenshot
image

### Objectives
- [x] Allow admins to view shared requests created by any user
- [x] Add the ability to open a shared requests in a new tab
- [x] Add the abiility to copy the link of the selected shared request 
- [x] Add the ability to delete any shared request

### Checks


packages/hoppscotch-sh-admin/src/composables/usePagedQuery.ts
```diff
@@ -13,26 +13,29 @@ export function usePagedQuery<
...
-  //Fetch All Users
   const { client } = useClientHandle();
   const fetching = ref(true);
   const error = ref(false);
   const list = ref<ListItem[]>([]);
   const currentPage = ref(0);
   const hasNextPage = ref(true);
 
-  const fetchNextPage = async () => {
+  const fetchNextPage = async (email?: string) => {
```
AndrewBastin, line 23: Why is email hardcoded here ? Isn't this composable supposed to be more general ?

packages/hoppscotch-sh-admin/src/composables/usePagedQuery.ts
```diff
@@ -57,11 +61,11 @@ export function usePagedQuery<
     }
   };
 
-  const refetch = async () => {
+  const refetch = async (email?: string) => {
```
AndrewBastin, line 64: Same as above. Why is email hardcoded on this generic composable ?

packages/hoppscotch-sh-admin/locales/en.json
```diff
@@ -19,6 +19,25 @@
     "owner": "OWNER",
     "viewer": "VIEWER"
   },
+  "sharedRequests": {
```
AndrewBastin, line 22: Follow snake case as the norm for i18n strings.

AndrewBastin (changes requested):

packages/hoppscotch-sh-admin/src/components/users/Details.vue
```diff
@@ -0,0 +1,137 @@
...
+
+      <div v-else class="bg-primaryDark w-16 p-3 rounded-2xl mb-3 relative">
+        <icon-lucide-user class="text-4xl" />
+        <span
+          v-if="user.isAdmin"
+          class="absolute left-16 bottom-0 text-xs font-medium px-3 py-0.5 rounded-full bg-green-900 text-green-300"
+        >
+          {{ t('users.admin') }}
+        </span>
+      </div>
```
jamesgeorge007, line 22: This can be unified. The dynamic things here are deciding whether to render an image or the icon based on the avatar URL availability and conditionally binding the classes to the parent `div`.

packages/hoppscotch-sh-admin/src/components/users/Details.vue
```diff
@@ -0,0 +1,137 @@
...
+        <label class="text-secondaryDark" for="username">{{
+          t('users.created_on')
+        }}</label>
+        <div
+          class="w-full p-3 mt-2 bg-divider border-gray-600 rounded-md focus:border-emerald-600 focus:ring focus:ring-opacity-40 focus:ring-emerald-500"
+        >
+          {{ getCreatedDateAndTime(user.createdOn) }}
+        </div>
+      </div>
+    </div>
```
jamesgeorge007, line 67: This is an ideal candidate where the markup can be constructed based on state maintained locally. For instance, have an object like

```js
const userInfo = {
    uid: { 
       condition: user.uid,
       labelText: t('users.uid'),
       value: user.uid
     }
     ...
  }
```

Iterate over the objects and construct the markup. The `condition` field can be applied to the `v-if` directive.

packages/hoppscotch-sh-admin/src/components/users/Details.vue
```diff
@@ -0,0 +1,137 @@
...
+
+      <HoppButtonSecondary
+        v-if="user.isAdmin"
+        class="mr-4 bg-red-600 text-white hover:text-gray-100"
+        filled
+        outline
+        :icon="IconTrash"
+        :label="t('users.delete')"
+        @click="toast.error(t('state.remove_admin_to_delete_user'))"
+      />
```
jamesgeorge007, line 107: These can be unified as well. The props can be dynamically bound based on the `user.isAdmin` check.

jamesgeorge007 (changes requested):

jamesgeorge007:
Are we optimizing for mobile? In that case, can you look into the following?

- The `Delete` button in the user details page appears truncated when the user is an admin.
  
  image

- The title appears far apart and might need to be truncated. Also, the shortcode URL appears vertically centered while others align to the top in the table view for shared requests.
  
  image

> iPhone SE (375 x 667)

packages/hoppscotch-sh-admin/src/components/users/Details.vue
JoelJacobStephen: Thanks for the suggestion. Helped in reducing the markup significantly

JoelJacobStephen:
@jamesgeorge007 The Admin Dashboard as a whole is not optimised for mobile as I don't believe there is a significant use case for this. If we are aiming for optimising the UI for mobile devices, I would like to tackle this properly in a separate PR where I can make the entire dashboard optimised for mobile. @liyasthomas Can I have your feedback regarding this?

liyasthomas:
Yeah that's correct. Dashboard project is not responsively designed. We can bring mobile-friendly design later, hence out of scope for this PR.

Cc: @JoelJacobStephen @jamesgeorge007

packages/hoppscotch-sh-admin/src/components/users/SharedRequests.vue
```diff
@@ -0,0 +1,191 @@
...
+                :blank="true"
+                :icon="IconExternalLink"
+                class="px-3 text-emerald-500 hover:text-accent"
+              />
+
+              <UiAutoResetIcon
+                :title="t('shared_requests.copy')"
+                :icon="{ default: IconCopy, temporary: IconCheck }"
+                @icon-clicked="copySharedRequest(request.id)"
+              />
```
jamesgeorge007, line 63: This component is similar to the implementation in the app. Am I right? Can we use the `HoppSmartItem` component here instead?

https://github.com/hoppscotch/hoppscotch/blob/23afc201a12e221c01f79d731513fcfeff928fdd/packages/hoppscotch-common/src/components/http/Request.vue#L180-L190

JoelJacobStephen: @jamesgeorge007 I tried this implementation, but I constantly face the issue where clicking on the copy icon triggers all the icons in each row, however the one selected gets copied. Separating it into a new component fixed the problem where the icon from only one row gets triggered. I have also added better functionality to this component where it can be customized and reused. I am open to ways where using SmartItem actually accomplishes the intended functionality here.

jamesgeorge007: Gotcha, there will be a shared state problem. Let's go ahead with the component-based approach 👍
It'll be handy when the pattern is used in another part of the app down the lane.

jamesgeorge007:
@JoelJacobStephen, I've made a few changes [here](https://github.com/hoppscotch/hoppscotch/pull/3537/files/6cb7689ec6dee2c8fed2f120d4f3bfafb362ecb9..5b3329d030bbbcf52cfff0139abd8c04888bf14d). Please take a look as you get the chance. I've made the icon `user-minus` for the primary button in the user details view to be displayed only when the user is an admin (`Remove Admin Status` state) as was the case earlier, [Ref](https://github.com/hoppscotch/hoppscotch/pull/3537#discussion_r1409276258). Is it intentional that there isn't an icon associated with the `Make admin` state?

jamesgeorge007 (approved):

packages/hoppscotch-sh-admin/src/components/ui/AutoResetIcon.vue
```diff
@@ -0,0 +1,47 @@
...
+    resetColor?: string;
+    icon: {
+      default: FunctionalComponent<SVGAttributes, {}>;
+      // icon to be displayed temporarily until reset
+      temporary: FunctionalComponent<SVGAttributes, {}>;
+    };
+  }>(),
+  {
+    color: 'white',
+    resetColor: 'green',
```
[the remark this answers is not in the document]
JoelJacobStephen: Yes. I noticed that text color is not being applied when Secondary button is using an icon. Making the change now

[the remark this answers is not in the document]
jamesgeorge007: I don't think changes to the `HoppButtonSecondary` component source are required here; have updated the default value to be `emerald` for the `AutoResetIcon` component `resetColor` prop. Lmk your thoughts.

jamesgeorge007 (approved):

PR #3567, opened 2023-11-19
fix: XML body disappearing with invalid XML

### LINKED ISSUE #3236: [bug]: XML Body disappearing

### Current behavior

When I enter a more complex XML body for a request with content type application/xml and I send some requests, after some time the XML body disappears (Body Tab, Raw Request Body) and is no longer displayed.
The body is still available and sent when I send a request but it is no longer displayed.
I can only change it if I clear the body (with the trashcan icon) and then enter it agan

### Steps to reproduce

1. Create a POST request
2. Go to Body, enter content type application/xml
3. enter a bigger XML as Raw Request Body
4. send the request
5. Save the request
6. Close the request tab, click around in the UI, open the request again
7. Go to the Body tab
8. Raw Request Body is not displayed
9. Send the request and detect that the body is still available but not displayed

### Environment

Production

### Version

Self-hosted

amk-dev:
Fixes HFE-213
Closes #3236 

**Before**

When formatting/parsing XML files, if the xml is invalid, request body disappeares. We were not catching the error thrown by the XMLFormatter function.

**After**

We capture and handle the thrown error with a try-catch. Also wrapped the try catch in a utility function that returns an Either to make it a little bit tidy.

### DIFF
--- a/packages/hoppscotch-common/src/composables/codemirror.ts
+++ b/packages/hoppscotch-common/src/composables/codemirror.ts
@@ -44,6 +44,8 @@ import { invokeAction } from "~/helpers/actions"
 import { useDebounceFn } from "@vueuse/core"
 // TODO: Migrate from legacy mode
 
+import * as E from "fp-ts/Either"
+
 type ExtendedEditorConfig = {
   mode: string
   placeholder: string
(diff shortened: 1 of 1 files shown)


AndrewBastin (approved):

nivedin (approved):
LGTM 💯

PR #3573, opened 2023-11-21
bug: remove mandatory client secret variable

### LINKED ISSUE #3572: [bug]: Allow Client Secret for OAuth2 Authentication

### Current behavior

Previously Hoppscotch would allow setting no client secret for OAuth2 authentication, which would then trigger the client credentials flow for providers such as Azure AD.

The recent 2023.8.4 release adds [validation for the `clientSecret`](https://github.com/hoppscotch/hoppscotch/pull/3531/files#diff-38391f555a8116c520a32b72d05b184e3351b1c25ba61436e568b6a97d6812ceR264) that forces users to set it, which will trigger the authorization code flow.

However, authorization code cannot be used, because Hoppscotch's final token exchange happens on the client, where Azure AD has CORS headers that block the request as well as a check on the `Origin` header, throwing the following error with a 400 Bad Request:

> AADSTS9002326: Cross-origin token redemption is permitted only for the 'Single-Page Application' client-type. Request origin: 'https://hoppscotch.io'.

### Steps to reproduce

1. Create a request to an authorized API with OAuth2.
2. Set authorization type to "OAuth 2.0".
3. Configure the authorization code flow by setting client secret / client credentials flow by removing it.
4. Click "Generate Token".

### Environment

Production

### Version

Cloud

Dan6erbond:
Closes #3572 

### Description

Removes client secret validation to trigger client credentials flow when it isn't set.

### Checks

### Additional Information

None.

### DIFF
--- a/packages/hoppscotch-common/src/helpers/oauth.ts
+++ b/packages/hoppscotch-common/src/helpers/oauth.ts
@@ -261,10 +261,6 @@ const handleOAuthRedirect = async () => {
     return E.left("NO_CLIENT_ID" as const)
   }
 
-  if (!clientSecret) {
-    return E.left("NO_CLIENT_SECRET" as const)
-  }
-
   if (!codeVerifier) {
     return E.left("NO_CODE_VERIFIER" as const)
   }


amk-dev:
> which would then trigger the client credentials flow for providers such as Azure AD.

Can you give me steps to reproduce your configuration in Azure AD or any other service that supports a client credential flow without a client secret.

StormicK:
Is it possible to help out here? The PR seems quite dead and I need the oauth2 pkce flow without a client secret urgently.

Or do you have an idea how to get around this issue

PR #3574, opened 2023-11-21
Bosnian translation

Kolesar:
### Description
Added translation for Bosnian language.


liyasthomas:
Thanks for making efforts to contribute to Hoppscotch i18n, but since Hoppscotch doesn't have plans to add a Bosnian locale, I'm closing this PR. We'll consider this feature if we get enough requests.

Kolesar:
Does not have a plan to add the Bosnian language? What it means?

I just added one language more, nothing more, nothing less. Where is the issue actually?

JaouherK:
@liyasthomas I do not understand why not. Would be great to understand the reason to limit the locales when you already have a system that can handle multiple. 
I am just curious.

PR #3575, opened 2023-11-21
chore: bump deps for `hoppscotch-common` and `hoppscotch-selfhost-web`

amk-dev:
Fixes HFE-262, HFE-264

This PR updates the dependencies `hoppscotch-common` and `hoppscotch-selfhost-web`.

- we didn't upgrade to Vite version 5 because some of the plugins we rely on haven't been updated to work with Vite 5 yet.

- Additionally, there are two packages, `@vitejs/plugin-legacy` and `vite-plugin-static-copy`, which dropped support for Vite 4. We didn't update them to their major versions either.

- this PR removes some unwanted deps from `@hoppscotch/selfhost-web`. we could also do the same for `@hoppscotch/common`. will tackle that in another PR.

- The package `unhead/vue` is now used instead of `vue/head` because `vue/head` is no longer supported and has been deprecated.


packages/hoppscotch-selfhost-web/package.json
```diff
@@ -23,62 +23,58 @@
     "postinstall": "pnpm run gql-codegen"
   },
   "dependencies": {
-    "@fontsource-variable/inter": "^5.0.8",
-    "@fontsource-variable/material-symbols-rounded": "^5.0.7",
-    "@fontsource-variable/roboto-mono": "^5.0.9",
```
liyasthomas, line 28: Aren't these font packages required by `unplugin-fonts` anymore?

refer: https://github.com/cssninjaStudio/unplugin-fonts#fontsource

liyasthomas (changes requested):

packages/hoppscotch-selfhost-web/package.json
amk-dev: i tested it and the fonts worked fine. 

but digging a little in it seems our app was not using Unfont generated font css. we were directly importing the font files from the fontsource packages.

in `hoppscotch-common/src/index.ts`, we have direct imports from the fontsource packages. in order to use the unfont generated one, we might have to replace this with 'unfonts.css'

```diff
- import "@fontsource-variable/inter"
- import "@fontsource-variable/material-symbols-rounded"
- import "@fontsource-variable/roboto-mono"
+ import "unfonts.css"
```

this way we can remove these packages from `hoppscotch-common` and keep them in font-source packages only in `hoppscotch-selfhost-web` because that's where the vite config picks up packages.

let me know what you think.

liyasthomas: @amk-dev can you verify once again al packages works if you only add font deps in common?

amk-dev: the unfonts usage is fixed. the @fontsource deps are removed from `hoppscotch-common`. but added back to `hoppscotch-selfhost-web`

liyasthomas:
Alright, does that also means central needs any updates?

amk-dev:
> Alright, does that also mean central needs any updates? @amk-dev

it would work in the current state. but the 'import unfonts.css' should also go there. will add that in a PR to central. the 'sh-admin' parts are also not touched. it will work with out any problems without the change.  but better we bring these there also. but right now, the dependency situation there is a little behind. eg: vite is v3 there. it needs a little work and testing. defering it currently.

liyasthomas:
Cool. Approving this rn, looking to the central implementation.

liyasthomas (approved):
LGTM

AndrewBastin (approved):

PR #3603, opened 2023-11-28
fix: minor ui improvements

liyasthomas:
- [x] The icon inside the FileChip component was invisible for some reason, this PR fixes this behavior.
- [x] The drag-and-drop handle color is now even with UI design.
- [x] Fixed wrong tooltip content in the computed headers action buttons.


nivedin (approved):
LGTM 💯

AndrewBastin (approved):

PR #3615, opened 2023-12-01
feat: init boring avatars

liyasthomas:
Introduced [boring avatars](https://cmgriffing.github.io/boringer-avatars) to replace the current `Smart/Picture.vue` component.


AndrewBastin (approved):
lgtm. @anwarulislam we will need to sync the changes up to the detached Hoppscotch UI version.

JoelJacobStephen:
@liyasthomas If a user has a profile picture associated with his account, this component no longer shows that but shows these avatars instead. Is this an intended change? For example, in the Admin Dashboard, shouldn't we show the profile picture of the logged in user if it is present and default back to an avatar if there is no picture available? Or are we going to show these avatars only from now on?

JoelJacobStephen:
@liyasthomas If a user has a profile picture associated with his account, this component no longer shows that but shows these avatars instead. Is this an intended change? For example, in the Admin Dashboard, shouldn't we show the profile picture of the logged in user if it is present and default back to an avatar if there is no picture available? Or are we going to show these avatars only from now on?

JoelJacobStephen:
> @JoelJacobStephen actually yes, we won't be showing photo URLs from auth providers anymore, this change is partly due to a few things:
> 
> 1. bring brand identity to the profile
> 2. bring unified profile pictures to users
> 3. use a defined and limited color palette

Ok. So in context of the Admin Dashboard, we do show the photo URLs from auth providers in the individual user's details page as well. Are we replacing that with these avatars too? If so, can you make that change in this PR?

liyasthomas:
We're replacing profile pictures with avatars all across the platforms. So changes on the dashboard were also a intentional change.

PR #3619, opened 2023-12-03
refactor(scripting-revamp): migrate `js-sandbox` to `web worker/Node vm` based implementation

jamesgeorge007:
### Description

This PR marks the initial step for the efforts that will be made towards the scripting revamp front. We're moving from [quickjs-emscripten](https://github.com/justjake/quickjs-emscripten) to a Web worker/Node VM-based implementation for `js-sandbox`.

Closes HFE-318.

### Changes

- Separate implementation based on environments for `js-sandbox` with subpath-based exports. The functions to execute pre-request and test scripts are exposed under subpaths after the respective environment, while types are accessible from the package root.
  ```ts
    import { runPreRequestScript, runTestScript } from "@hoppscotch/js-sandbox/web"
  ```

  ```ts
    import { runPreRequestScript, runTestScript } from "@hoppscotch/js-sandbox/node"
  ```

  ```ts
    import { TestDescriptor } from "@hoppscotch/js-sandbox"
  ```
  - The web worker-based implementation `{pre-request,test-runner}/web-worker/*.ts` includes a script compiling the main thread logic for communicating with the worker thread via the exposed function. The received arguments (script, envs, etc) from the place of invocation are supplied to the worker script residing relative to the main thread source, where the script gets executed via the [Function constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/Function), and the results are sent back to the main thread which in turn is returned to the system.

  - The Node VM-based implementation `{pre-request,test-runner}/node-vm/*.ts` creates a new VM context via the [createContext](https://nodejs.org/api/vm.html#vmcreatecontextcontextobject-options) method, executes the script in a new context via the [runInContext](https://nodejs.org/api/vm.html#scriptrunincontextcontextifiedobject-options) method and returns the relevant results.

  - `getPreRequestScriptMethods` & `getTestRunnerScriptMethods` functions under `utils` compiles the methods to be exposed under the `pw` namespace.
- Move the `js-sandbox` build system from `tsup` to `vite`.
  - There are two entry points corresponding to the environment.
    - `src/web.ts`
    - `src/node.ts`
  - [vite-plugin-dts](https://github.com/qmhc/vite-plugin-dts) ensures the declaration files are emitted.
- The `marshlObjecToVm` function under `utils` has been made to prevent cyclic objects under the name `preventCyclicObjects` since there isn't a requirement for marshaling.
- Enable CLI e2e tests on CI.

### Checks


packages/hoppscotch-cli/src/__tests__/samples/env-flag-tests.json
```diff
@@ -12,7 +12,7 @@
...
+      "testScript": "const HOST = pw.env.get(\"HOST\");\nconst UNSET_ENV = pw.env.get(\"UNSET_ENV\");\nconst EXPECTED_URL = \"https://echo.hoppscotch.io\";\nconst URL = pw.env.get(\"URL\");\nconst X_COUNTRY = [\"IN\", \"US\"];\nconst BODY_VALUE = pw.env.get(\"BODY_VALUE\");\n\n// Check JSON response property\npw.test(\"Check headers properties.\", ()=> {\n    pw.expect(pw.response.body.headers.host).toBe(HOST);\n\t   pw.expect(X_COUNTRY).toInclude(pw.response.body.headers[\"x-country\"]);  \n});\n\npw.test(\"Check data properties.\", () => {\n\tconst DATA = pw.response.body.data;\n  \n    pw.expect(DATA).toBeType(\"string\");\n    pw.expect(JSON.parse(DATA).body_key).toBe(BODY_VALUE);\n});\n\npw.test(\"Check request URL.\", () => {\n  pw.expect(URL).toBe(EXPECTED_URL);\n})\n\npw.test(\"Check unset ENV.\", () => {\n  pw.expect(UNSET_ENV).toBeType(\"undefined\");\n})",
```
jamesgeorge007, line 15: While running tests in the CI, the above header came up as `US`; thought to maintain a static list since CLI doesn't support arrays as values for environment entries. It can be an issue if one residing in a different nation tries to run the tests locally, and the test starts failing. Should we remove the above test?

packages/hoppscotch-js-sandbox/src/utils.ts
```diff
@@ -1,89 +1,439 @@
...
+  }
+
+  const toBeLevelXxx = (
+    level: string,
+    rangeStart: number,
+    rangeEnd: number
+  ) => {
+    const parsedExpectVal = parseInt(expectVal)
+
+    if (!Number.isNaN(parsedExpectVal)) {
```
jamesgeorge007, line 216:
```js
pw.expect("404").toBeLevel4xx();
```

Added support for the expect value being a `string` with `toBeLevel` matchers. It was mentioned in the [documentation](https://docs.hoppscotch.io/documentation/features/scripts#pwtobelevelxxx) such that it was supported already.

packages/hoppscotch-js-sandbox/src/utils.ts
```diff
@@ -1,89 +1,439 @@
...
-    return E.left("Marshaling stringification failed")
+const getEnv = (envName: string, envs: TestResult["envs"]) => {
+  return O.fromNullable(
+    envs.selected.find((x: SelectedEnvItem) => x.key === envName) ??
+      envs.global.find((x: GlobalEnvItem) => x.key === envName)
+  )
+}
+
+// Compiles shared scripting API methods for use in both pre and post request scripts
+const getSharedMethods = (envs: TestResult["envs"]) => {
```
jamesgeorge007, line 22: The `envs` argument received here is a deep copy. From this point, we rely on mutation since most of them happen in the API methods, which, while invoked, can't return the updated value to the surrounding context.
For instance, the environment gets updated in the `setEnvFn` method.

This is the case with `createExpectation` function as well (`currTestStack`).

packages/hoppscotch-js-sandbox/src/utils.ts
```diff
@@ -1,89 +1,439 @@
...
-}
+    updatedEnvs = setEnv(key, value, updatedEnvs)
 
-export function getEnv(envName: string, envs: TestResult["envs"]) {
-  return O.fromNullable(
-    envs.selected.find((x) => x.key === envName) ??
-      envs.global.find((x) => x.key === envName)
-  )
+    return undefined
+  }
```
jamesgeorge007, line 80: As mentioned earlier, when this method is invoked, there isn't a way to return the updated environment entries to the surrounding context. The return value below from the outside function gets processed beforehand. Hence, we rely on mutation. This only applies to `utils`; we obtain a deep copy of the arg supplied from the outside world.

AndrewBastin (approved):

PR #3625, opened 2023-12-05
feat: import environments from insomnia

amk-dev:
Fixes HFE-235

wip


jamesgeorge007 (changes requested):
- Are we not considering environments maintained at the folder level? For instance, in the below example, the `Random Test User` folder has an environment set overriding the `baseURL` and a new `randomValue` variable.

  > Discussed separately, this will require adding support for environments at the folder level.

- Insomnia adds a `_.` prefix for environment variables. This is stripped off while displaying the values under `Environments`. Similarly, this is to be implemented for requests.

https://github.com/hoppscotch/hoppscotch/assets/25279263/2157042a-81f4-40fc-aa77-1ca81631726f

amk-dev:
> * Are we not considering environments maintained at the folder level? For instance, in the below example, the `Random Test User` folder has an environment set overriding the `baseURL` and a new `randomValue` variable.

Currently we don't have a concept of folder level environments. and not a logical way to add them to our envs. So this is a gotcha we've to keep in mind.

> * Insomnia adds a `_.` prefix for environment variables. This is stripped off while displaying the values under `Environments`. Similarly, this is to be implemented for requests.

Now this is implemented to the Insomnia Collection imports also.

jamesgeorge007 (changes requested):
The Insomnia environment manager supports specifying a non-string value since it offers a JSON-based editor. Currently, environments with non-string values are removed while importing since it's a mismatch with the expected schema, and there is no feedback to the user regarding the removal. We can convert all the values to a string since we expect that format.

Please use [this export](https://github.com/hoppscotch/hoppscotch/files/13651032/insomnia-test.json) from Insomnia to verify this behavior.

jamesgeorge007:
@amk-dev, as discussed, I've pushed the changes converting the environment data values to string while importing in https://github.com/hoppscotch/hoppscotch/pull/3625/commits/e23ce50c278c0eb1e9cddd362f1644b62b9d26e9. Please take a look.

jamesgeorge007 (approved):

PR #3626, opened 2023-12-05
Update mailer.module.ts -> better configuration and accept custom certs

RaHehl:
This update provides an option to disable certificate checking for the SMTP server. While we recognize that this may not be the ultimate solution, we believe it provides a more flexible approach for users who require additional control over certificate validation.
In addition, it allows better configuration of username and password.


packages/hoppscotch-backend/src/mailer/mailer.module.ts
```diff
@@ -1,27 +1,36 @@
...
-      },
-      template: {
-        dir: __dirname + '/templates',
-        adapter: new HandlebarsAdapter(),
-      },
+    NestMailerModule.forRootAsync({
+      useFactory: () => ({
+        transport: {
+          host: process.env.SMTP_DOMAIN,
+          port: 587,
```
r-brown, line 14: hardcoded port?

RaHehl: It was actually intended as a quick example, since pull requests are usually implemented by the Hoppscotch team themselves. But I refreshed it again with the current main because there were breaking changes.

mirarifhasan:
Hi @RaHehl,

Thank you for your contribution! 🙌

We've implemented the custom email configuration in a different way. Your initiative in raising this concern encouraged us to make the mailer module more configurable. We truly appreciate your input.

As the changes have been addressed through another approach, we'll be closing this PR. Looking forward to your future contributions!

PR #3635, opened 2023-12-08
fix: use base url instead of hardcoded url

liyasthomas:
- Use `env.VITE_SHORTCODE_BASE_URL` instead of hardcoded value for shared requests.
- Minor UI improvements.

### DIFF
--- a/packages/hoppscotch-common/src/components/history/index.vue
+++ b/packages/hoppscotch-common/src/components/history/index.vue
@@ -3,7 +3,7 @@
     <div
       class="sticky top-0 z-10 flex flex-shrink-0 flex-col overflow-x-auto border-b border-dividerLight bg-primary"
     >
-      <WorkspaceCurrent :section="t('tab.history')" />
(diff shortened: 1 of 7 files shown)


packages/hoppscotch-common/src/components/history/index.vue
```diff
@@ -3,7 +3,7 @@
     <div
       class="sticky top-0 z-10 flex flex-shrink-0 flex-col overflow-x-auto border-b border-dividerLight bg-primary"
     >
-      <WorkspaceCurrent :section="t('tab.history')" />
+      <WorkspaceCurrent :section="t('tab.history')" :is-only-personal="true" />
```
jamesgeorge007, line 6: Is this intentional? Is history not synced to the team workspace currently?

liyasthomas: History entries are personal hence won't be synced across the team. Somehow missing this notation went unnoticed all this time.

jamesgeorge007 (approved):

PR #3647, opened 2023-12-13
feat: persist line wrap setting

### LINKED ISSUE #3641: [feature]: Persist wrap lines settings
I need to report about wrap line is always reset to default on every request

ajmalmohad:
Persisted line wrap setting across various fields

Closes #3641 

### Description
Added settings for wrap line toggle for editor fields

**HTTP:** Request Body, Response Body, Headers, Parameters, URL Encoded, Pre-request, Test
**GraphQL:** Query, Response Body, Headers, Variables, Schema
**Realtime:** Communication Body, Realtime Log
**Others:** Import cURL, Code Generation, Cookie

### Checks

### Additional Information
https://github.com/hoppscotch/hoppscotch/assets/55492635/7e9e152e-226e-4af1-ad04-d515fbaa3412


packages/hoppscotch-common/src/newstore/settings.ts
```diff
@@ -30,6 +30,31 @@ export type SettingsDef = {
...
+  WRAP_LINES_GRAPHQL_SCHEMA: boolean
+
+  // Realtime Wrap Lines
+  WRAP_LINES_REALTIME_BODY: boolean
+  WRAP_LINES_REALTIME_LOG: boolean
+
+  // Other Wrap Lines
+  WRAP_LINES_IMPORT_CURL: boolean
+  WRAP_LINES_CODE_GEN: boolean
+  WRAP_LINES_COOKIE: boolean
```
jamesgeorge007, line 56: Can we group these under a property, say `WRAP_LINES` similar to `URL_EXCLUDES? It might require changes to the [toggleSetting function](https://github.com/hoppscotch/hoppscotch/blob/main/packages/hoppscotch-common/src/newstore/settings.ts#L138-L145) and the corresponding [settings store dispatcher](https://github.com/hoppscotch/hoppscotch/blob/main/packages/hoppscotch-common/src/newstore/settings.ts#L87-L102) since it only accepts a top-level setting key currently.

ajmalmohad: Okay, I'll change that

ajmalmohad: > Can we group these under a property, say `WRAP_LINES` similar to `URL_EXCLUDES`? It might require changes to the [toggleSetting function](https://github.com/hoppscotch/hoppscotch/blob/main/packages/hoppscotch-common/src/newstore/settings.ts#L138-L145) and the corresponding [settings store dispatcher](https://github.com/hoppscotch/hoppscotch/blob/main/packages/hoppscotch-common/src/newstore/settings.ts#L87-L102) since it only accepts a top-level setting key currently.

I've created seperate functions instead of extending the current ones, parsing by `.` and changing state from the current function doesn't provide flexibility for proper type checking

packages/hoppscotch-common/src/newstore/settings.ts
```diff
@@ -146,6 +140,26 @@ const dispatchers = defineDispatchers({
...
+  toggleNestedSetting(
+    currentState: SettingsDef,
+    {
+      settingKey,
+      property,
+    }: { settingKey: KeysMatching<SettingsDef, object>; property: string }
+  ) {
+    const result: Partial<SettingsDef> = {}
+
+    if (!has(currentState, settingKey + "." + property)) {
```
jamesgeorge007, line 152:
```suggestion
    if (!has(currentState, [settingKey, property])) {
```

Let's specify the path as an array.

packages/hoppscotch-common/src/newstore/settings.ts
```diff
@@ -146,6 +140,26 @@ const dispatchers = defineDispatchers({
 
     return result
   },
+  toggleNestedSetting(
+    currentState: SettingsDef,
+    {
+      settingKey,
+      property,
+    }: { settingKey: KeysMatching<SettingsDef, object>; property: string }
```
jamesgeorge007, line 148:
```suggestion
    }: { settingKey: KeysMatching<SettingsDef, Record<string, boolean>>; property: string }
```

Also, can we have a well-defined type for the `property`?

packages/hoppscotch-common/src/newstore/settings.ts
```diff
@@ -146,6 +140,26 @@ const dispatchers = defineDispatchers({
...
+    const result: Partial<SettingsDef> = {}
+
+    if (!has(currentState, settingKey + "." + property)) {
+      return {}
+    }
+
+    result[settingKey] = {
+      ...currentState[settingKey],
+      [property]: !currentState[settingKey][property],
+    }
```
jamesgeorge007, line 159: Can you look into the type errors arising here?

packages/hoppscotch-common/src/newstore/settings.ts
```diff
@@ -190,6 +204,19 @@ export function toggleSetting(settingKey: KeysMatching<SettingsDef, boolean>) {
...
 
+export function toggleNestedSetting<
+  K extends KeysMatching<SettingsDef, object>,
+  P extends keyof SettingsDef[K],
+>(settingKey: K, property: P) {
+  settingsStore.dispatch({
+    dispatcher: "toggleNestedSetting",
+    payload: {
+      settingKey,
+      property,
```
jamesgeorge007, line 215: There's a type error arising here.

ajmalmohad: > There's a type error arising here.

Added // @ts-error here as TypeScript is not able to figure out the passed property

jamesgeorge007 (commented):
I was wondering if we would want to persist line wrap preference (communication message editor instance/ response logs) across real-time protocols? It was appropriate for REST/GQL requests, being the same entity, and since they persisted across tabs. But shouldn't we keep that distinction and scope the settings in the case of WebSocket, SSE, etc, being separate protocols?

Also, for the individual log entries, updating the line wrap preference next to one updates it across, which might not be ideal since having a line wrap toggle next to each log item conveys that it is specific to the respective entry. It should be moved to the Log section title if keeping this behavior.

https://github.com/hoppscotch/hoppscotch/assets/25279263/26b47ba1-540e-4187-9780-e0bb6e55384d

cc @liyasthomas

jamesgeorge007:
@ajmalmohad, I discussed this with @liyasthomas. Let's not persist the line wrap preference for the realtime view and keep the behavior as earlier for the communication message editor instance and log entries where the preference is kept in the local component state.

ajmalmohad:
@jamesgeorge007 I've reverted the changes made to the realtime view components and removed the associated settings.

jamesgeorge007 (approved):

jamesgeorge007 (approved):

AndrewBastin (approved):

PR #3653, opened 2023-12-14
fix(common): GraphQL query syntax highlighting

jamesgeorge007:
### Description

This PR aims at bringing back the syntax highlighting for GraphQL queries. Bumps all the `codemirror` related packages to the latest version in both `hoppscotch/common` and `codemirror-lang-graphql` packages.

> Context

We had [@lezer/highlight](https://github.com/lezer-parser/highlight) pinned to `v1.1.4` since any version above was found to make the GQL page crash/affect syntax highlighting during that time. It was found to be fixed in the recent versions; hence, we're moving to the latest version.

### Before

  image

### After

  image

### Checks


AndrewBastin (changes requested):
the highlight is still broken when we do a prod build.

AndrewBastin (approved):

PR #3664, opened 2023-12-18
fix(common): ensure the add-environment modal value field is empty when opened via the inspector

jamesgeorge007:
### Description

When opened via the inspector, the add environment modal showed the value associated with the previous attempt to set an environment via the context menu. This PR aims to fix this behavior by removing the truthy check before assigning environment values at the parent environments component, thereby permitting empty strings.

https://github.com/hoppscotch/hoppscotch/assets/25279263/9dd2e560-bfb8-4152-819c-90a0e07f84ce

Closes HFE-347.

### Checks

### DIFF
--- a/packages/hoppscotch-common/src/components/environments/index.vue
+++ b/packages/hoppscotch-common/src/components/environments/index.vue
@@ -296,7 +296,7 @@ watch(
 
 defineActionHandler("modals.environment.add", ({ envName, variableName }) => {
   editingVariableName.value = envName
-  if (variableName) editingVariableValue.value = variableName
+  editingVariableValue.value = variableName
   displayModalNew(true)
 })
 </script>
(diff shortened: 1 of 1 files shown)


liyasthomas (approved):

packages/hoppscotch-common/src/components/environments/index.vue
```diff
@@ -296,7 +296,7 @@ watch(
 
 defineActionHandler("modals.environment.add", ({ envName, variableName }) => {
   editingVariableName.value = envName
-  if (variableName) editingVariableValue.value = variableName
```
jamesgeorge007, line 299: @nivedin, was this truthy check added for eliminating `null/undefined` values? I don't think there can be a case where we would end up with such values since we've control over the value supplied. Regardless, we could have a fallback to an empty string.

nivedin: it was added as a undefined/null blocker ig, I think empty string will work in this case

nivedin (approved):

PR #3686, opened 2023-12-21
refactor: added input validation to server configurations in admin dashboard

JoelJacobStephen:
### Ticket
Closes HFE-363

### Description
This PR introduces input validations to all server configuration input fields in the admin dashboard

### Checks


packages/hoppscotch-sh-admin/src/composables/useClientHandler.ts
jamesgeorge007, line 1: This composable doesn't seem to be consumed anywhere with the addition of `useClientListHandler`. We can unify both implementations by moving the `getList` parameter in the `useClientListHandler` composable to the end of the list, making it optional and determining how to parse the result based on the availability of the same. Since the current use cases are about supplying the transformer functions, we can get rid of the `useClientHandler` composable and revisit this separately. Feel free to take the call here.

```ts
  if (getList) {
    const resultList = getList(result.data!);
    list.value.push(...resultList);
  } else {
    list.value = result.data
  }
```

jamesgeorge007 (commented):
Would it be helpful to restore the fields emptied after an attempt to save the changes?

JoelJacobStephen:
> Would it be helpful to restore the fields emptied after an attempt to save the changes?

@jamesgeorge007 It would bring quite a overhead to check each input and see if they are empty and then restore the old value. This can be done by calling the backend query to refetch new values or revert the workingConfigs value back to currentConfig value which will also revert any other input changed by the user when trying to save the changes.

packages/hoppscotch-sh-admin/src/components/settings/ServerRestart.vue
```diff
@@ -72,13 +72,14 @@ onMounted(async () => {
...
     const infraResult = await updateInfraConfigs(updateInfraConfigsMutation);
 
-    success = authResult && infraResult;
-    if (!success) return;
+    if (infraResult) {
+      const authResult = await updateAuthProvider(
+        updateAllowedAuthProviderMutation
+      );
+      if (!authResult) return;
+    } else return;
```
jamesgeorge007, line 82:
```suggestion
    if (!infraResult) {
      return;
    }

    const authResult = await updateAuthProvider(
      updateAllowedAuthProviderMutation
    );
    if (!authResult) {
      return
    }
```

We could perform an early return instead of the trailing `else` block.

packages/hoppscotch-sh-admin/src/composables/useClientHandler.ts
```diff
@@ -31,9 +32,12 @@ export function useClientHandler<
         })
         .toPromise();
 
-      const resultList = getList(result.data!);
-
-      list.value.push(...resultList);
+      if (getList) {
+        const resultList = getList(result.data!);
+        dataAsList.value.push(...resultList);
```
jamesgeorge007, line 37: There's a type error arising here; might need an assertion.

packages/hoppscotch-sh-admin/src/pages/settings.vue
```diff
@@ -16,15 +16,15 @@
...
-    <div v-else class="flex flex-col py-8">
-      <HoppSmartTabs v-model="selectedOptionTab" render-inactive-tabs>
-        <HoppSmartTab :id="'config'" :label="t('configs.title')">
-          <SettingsConfigurations
-            v-model:config="workingConfigs"
-            class="py-8 px-4"
-          />
-        </HoppSmartTab>
-      </HoppSmartTabs>
+  <div v-if="workingConfigs" class="flex flex-col py-8">
```
jamesgeorge007, line 19:
```suggestion
  <div v-else-if="workingConfigs" class="flex flex-col py-8">
```

packages/hoppscotch-sh-admin/src/composables/useConfigHandler.ts
```diff
@@ -255,7 +259,21 @@ export function useConfigHandler(updatedConfigs?: Config) {
...
 
-  // Trasforming the working configs back into the format required by the mutations
+  // Checking if any of the config fields are empty
+  const isFieldEmpty = (field: string) => field === '';
+
+  const AreAnyConfigFieldsEmpty = (config: Config): boolean => {
+    const providerFieldsEmpty = Object.values(config.providers).some(
+      (provider) => [...Object.values(provider.fields)].some(isFieldEmpty)
+    );
+    const mailFieldsEmpty = [...Object.values(config.mailConfigs.fields)].some(
```
jamesgeorge007, line 269: Isn't it redundant to create another array since `Object.values` already return a string array?

```suggestion
      (provider) => Object.values(provider.fields).some(isFieldEmpty)
    );
    const mailFieldsEmpty = Object.values(config.mailConfigs.fields).some(
```

packages/hoppscotch-sh-admin/src/pages/settings.vue
```diff
@@ -78,4 +81,17 @@ const isConfigUpdated = computed(() =>
     ? !isEqual(currentConfigs.value, workingConfigs.value)
     : false
 );
+
+// Check if any of the fields in workingConfigs are empty
+const areAnyFieldsEmpty = computed(() =>
+  workingConfigs.value ? AreAnyConfigFieldsEmpty(workingConfigs.value) : true
```
jamesgeorge007, line 87: The fallback to `true` here is intentional, right?

packages/hoppscotch-sh-admin/src/composables/useClientHandler.ts
JoelJacobStephen: @jamesgeorge007 When pushing data into dataAsList, the value is of type `UnWrapRefSimple<ListItem>`.  However I am not able to import this type. I am able to import `UnwrapRef` from vue. I am not able to resolve this using assertion. Can you suggest a way to fix the type inconsistency?

jamesgeorge007: We'd require an assertion as `Ref<ListItem[]>` while initializing `dataAsList` in this case - Ref: https://github.com/vuejs/core/issues/2136.

Thought to specify the type directly in https://github.com/hoppscotch/hoppscotch/pull/3686/commits/c2c232add299785bdbe0c2842af3b175dfcd9f00 since the assertion was felt redundant.

jamesgeorge007 (approved):

jamesgeorge007:
Please ensure to update the target branch once the release branch is up.

PR #3690, opened 2023-12-22
fix: minor updation for shared request and auth-headers features

nivedin:
Closes HFE-367

### Description
This PR fixes some of the minor bugs in the shared request and auth-headers in collection features.
 - Added loading state in embeds
 - fixed i18n string when deleting a already deleted shared request
 - minor code update on team auth-header cascading folder path
 - Saving team request failed to inherit property.
 
### Checks

### DIFF
--- a/packages/hoppscotch-common/src/helpers/teams/TeamCollectionAdapter.ts
+++ b/packages/hoppscotch-common/src/helpers/teams/TeamCollectionAdapter.ts
@@ -1034,6 +1034,11 @@ export default class NewTeamCollectionAdapter {
     }
   }
 
+  /**
+   * Used to obtain the inherited auth and headers for a given folder path, used for both REST and GraphQL team collections
+   * @param folderPath the path of the folder to cascade the auth from
+   * @returns the inherited auth and headers for the given folder path
+   */
   public cascadeParentCollectionForHeaderAuth(folderPath: string) {
     let auth: HoppInheritedProperty["auth"] = {
       parentID: folderPath ?? "",
@@ -1089,7 +1094,10 @@ export default class NewTeamCollectionAdapter {
       const parentFolderAuth = data.auth
       const parentFolderHeaders = data.headers
 
-      if (parentFolderAuth?.authType === "inherit" && path.length === 1) {
+      if (
+        parentFolderAuth?.authType === "inherit" &&
+        [...path.slice(0, i + 1)].length === 1
+      ) {
         auth = {
           parentID: [...path.slice(0, i + 1)].join("/"),
           parentName: parentFolder.title,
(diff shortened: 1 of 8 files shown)


packages/hoppscotch-common/src/helpers/teams/TeamCollectionAdapter.ts
jamesgeorge007, line 1: I noticed the following issues while setting authorization/headers at the collection level in a team workspace.

- Setting the auth type as `inherit` for an immediate child request to the root collection results in the message, `Please save this request in any collection to inherit the authorization`.
- Updates at the parent level were not reflected immediately in the currently active request tab.

https://github.com/hoppscotch/hoppscotch/assets/25279263/145cbb81-cfb4-43fb-ab0c-bf316cb9deba

Could you look into the issues mentioned above?

jamesgeorge007 (changes requested):
The following issues persist for the team workspace.

- Updates made at the parent level were not reflected immediately in the currently active request tab. This happens only when the request is created for the first time, as observable from the below recordings; it works as expected while reopening.

    https://github.com/hoppscotch/hoppscotch/assets/25279263/a289d888-d4fd-438f-958c-dc53b88764be 

  https://github.com/hoppscotch/hoppscotch/assets/25279263/992affcb-7814-431f-a1f4-81fc0e768a65

> Mentioned below are a few issues encountered while verifying the flows not necessarily specific to the team workspace. Feel free to address them separately if you feel they are adding to the scope of this PR.

- If there's a header value set at the parent level, a folder below overrides it, and an update is performed for the value set at the ancestor level. The deeply nested request inherits the above value instead of consuming the overridden value from the immediate parent.

  https://github.com/hoppscotch/hoppscotch/assets/25279263/01b676cc-fef3-4f96-855c-16c3c3c4b79c

- If an authorization type is specified for a request and an Authorization header is set at any level above it in the hierarchy, the request will end up having duplicate Authorization headers, and the one from the parent takes precedence. Ideally, the authorization set at the request level should override any inherited values.

  https://github.com/hoppscotch/hoppscotch/assets/25279263/80ff1958-8a7a-4453-94b4-1b38c92f91f0

- Inheriting headers is dependent on the authorization type set at the folder level. For instance, if the authorization type is set to something other than inherit at a folder level, then the request within it won't be able to inherit the headers set at the ancestor level.

  https://github.com/hoppscotch/hoppscotch/assets/25279263/f01e3476-682c-4f61-b32a-f8c87ff15b00

jamesgeorge007:
Tracking the above issues in [HFE-375](https://linear.app/hopp/issue/HFE-375/issues-found-with-setting-authorizationheaders-at-the-collection-level).

PR #3705, opened 2023-12-29
fix(selfhost-web): perform logout if the silent refresh attempt fails

### LINKED ISSUE #3701: [bug]: Auth cookies are not removal from selfhost-web after expired it

### Current behavior

When I log into my account, I wait for the expiration of the refresh_token JWT (specified in the REFRESH_TOKEN_VALIDITY parameter). After that, I log into the selfhost-web component and see that my account is still active, but any GraphQL method  related with account is not working.

I expect that after expiration refresh token I will be logout and my cookie will be removed.

### Steps to reproduce

1. Log in into my account through selfhost-web component
2. Waiting for the expiration of the cookie parameter JWT refresh_token 
3. Open page of the selfhost-web component
4. See that account is still be active.

### Environment

Production

### Version

Self-hosted

shipko:
Closes #3701

### Description
This PR fixes the bug found in refresh token feature

- Fix problem with refreshToken method when backend returns 401 code
- When refreshToken returns false then completing the user session

- [ ] Not Completed
- [x] Completed

### Checks

### DIFF
--- a/packages/hoppscotch-selfhost-web/src/platform/auth/auth.platform.ts
+++ b/packages/hoppscotch-selfhost-web/src/platform/auth/auth.platform.ts
@@ -153,7 +153,7 @@ async function refreshToken() {
       {
         withCredentials: true,
       }
-      )
+    )
 
     const isSuccessful = res.status === 200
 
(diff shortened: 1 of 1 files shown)


shipko:
@balub please review it

jamesgeorge007 (approved):

jamesgeorge007:
Based on the existing implementation, the silent refresh flow would kick in only when the user performs a page reload and a log-out during the silent refresh flow when the refresh token has expired. We'd want to explore having this during page visits so that any actions performed while the tokens are expired will log them out, preventing exceptions due to unauthorized access.

jamesgeorge007 (approved):

balub (approved):
Tested locally LGTM 🚀 .

Ashraff-547:
tested locally LGTM.

PR #3718, opened 2024-01-04
HBE-345 feature: sso callback url and scope added in infra-config

mirarifhasan:
Closes HBE-345

### Description

This PR enables the Admin Dashboard to configure more attributes as a part of Server Config feature. Those attributes are:
```
GOOGLE_CALLBACK_URL
GOOGLE_SCOPE

MICROSOFT_CALLBACK_URL
MICROSOFT_SCOPE
MICROSOFT_TENANT

GITHUB_CALLBACK_URL
GITHUB_SCOPE
```

### Checks

### Additional Information

Nil


packages/hoppscotch-backend/src/utils.ts
```diff
@@ -173,6 +173,32 @@ export const validateSMTPUrl = (url: string) => {
...
+  if (!scope || scope.length === 0) return false;
+
+  const scopes = scope.split(',');
+  scopes.forEach((aScope) => {
+    if (aScope.length === 0) return false;
+  });
+
+  return true;
+};
+
```
balub, line 201: I think we need to remove this function, every SSO provider defines scopes in their own way, so even if the ones we use right now can use this function there is no guarantee any new SSO provider we add in the future can use this too.

balub (changes requested):

packages/hoppscotch-backend/src/utils.ts
mirarifhasan: Okay @balub

packages/hoppscotch-backend/src/infra-config/helper.ts
```diff
@@ -54,6 +64,125 @@ export async function loadInfraConfiguration() {
   }
 }
 
+/**
+ * Read the default values from .env file and return them as an array
+ * @returns Array of default infra configs
+ */
+export async function getDefaultInfraConfigs(): Promise<
```
balub, line 71: I think this function can be inside `infra-config.service.ts` file itself, was there particular reason we are moving it here?

packages/hoppscotch-backend/src/infra-config/helper.ts
```diff
@@ -54,6 +64,125 @@ export async function loadInfraConfiguration() {
...
+  { name: InfraConfigEnum; value: string }[]
+> {
+  const prisma = new PrismaService();
```
balub, line 74: I have a suggestion, rather than initialize a new instance of `PrismaService` here can't we just pass in the already initiated `PrismaService` from the file this function is being used in?

packages/hoppscotch-backend/src/types/InfraConfig.ts
```diff
@@ -24,12 +31,19 @@ export enum InfraConfigEnumForClient {
```
balub, line 31: Can we introduce the exclude const mechanism we use in SHE here as well @mirarifhasan and remove the `InfraConfigEnumForClient` enum from SHC as well, this is greatly help in merges in SHE in the future.

packages/hoppscotch-backend/src/auth/auth.module.ts
```diff
@@ -34,6 +37,11 @@ import { InfraConfigModule } from 'src/infra-config/infra-config.module';
 })
 export class AuthModule {
   static async register() {
+    const isInfraConfigPopulated = await isInfraConfigTablePopulated();
+    if (!isInfraConfigPopulated) {
+      return { module: AuthModule };
+    }
+
```
balub, line 44: Can we move this code outside of the auth module?

balub (changes requested):

packages/hoppscotch-backend/src/infra-config/helper.ts
mirarifhasan: There is a function `isInfraConfigTablePopulated()` in helper.ts that needs `getDefaultInfraConfigs()`. Since, `getDefaultInfraConfigs()` was in infra-config.service.ts, could not able call from `isInfraConfigTablePopulated()` (from helper file without creating object of service class)

balub: Alright, understood.

packages/hoppscotch-backend/src/infra-config/helper.ts
mirarifhasan: One caller function (form helper.ts) `isInfraConfigTablePopulated()` can not provide prismaService object.

packages/hoppscotch-backend/src/types/InfraConfig.ts
mirarifhasan: Sure

packages/hoppscotch-backend/src/infra-config/infra-config.service.ts
```diff
@@ -35,84 +32,32 @@ export class InfraConfigService implements OnModuleInit {
     private readonly configService: ConfigService,
   ) {}
 
+  // Following fields are not updatable by `infraConfigs` Mutation. Use dedicated mutations for these fields instead.
+  EXCLUDE_FROM_UPDATE_CONFIGS = [
+    InfraConfigEnum.VITE_ALLOWED_AUTH_PROVIDERS,
+    InfraConfigEnum.ALLOW_ANALYTICS_COLLECTION,
+    InfraConfigEnum.ANALYTICS_USER_ID,
+    InfraConfigEnum.IS_FIRST_TIME_INFRA_SETUP,
```
balub, line 40: The `IS_FIRST_TIME_INFRA_SETUP` config is updated via the `update` method in the `infra-config.controller.ts` file.

balub (changes requested):

balub (approved):
We should keep an eye on the `isInfraConfigTablePopulated` function in the auth module for any issues it may throw and figure out alternatives to the way we currently have implemented it other than that LGTM.

PR #3723, opened 2024-01-06
feat(graphql): Environments

Timeraa:
Closes #2172 

### Description

This PR adds the environments that are present in the REST tab to the GraphQL tab.

### TODO:
[ ] Parse Graphql variables
[ ] Add tests to check if functionality works

### Checks

- [ ] My pull request adheres to the code style of this project
- [ ] My code requires changes to the documentation
- [ ] I have updated the documentation as required
- [ ] All the tests have passed

PR #3730, opened 2024-01-07
fix: documentation is not being generated on GQL

anwarulislam:
Closes HFE-380

### Description
Updated the `resolveRootType` function to correctly handle GraphQL types that do not have an `ofType` property. The function now uses a while loop to traverse the `ofType` chain until it reaches a type that does not contain any other types (i.e., its ofType property is null). This ensures that the function always returns the correct root type for any given GraphQL type.

- [ ] Not Completed
- [x] Completed

### DIFF
--- a/packages/hoppscotch-common/src/components/graphql/Sidebar.vue
+++ b/packages/hoppscotch-common/src/components/graphql/Sidebar.vue
@@ -187,7 +187,7 @@ import IconClock from "~icons/lucide/clock"
 import IconCopy from "~icons/lucide/copy"
 import IconBox from "~icons/lucide/box"
 import { computed, nextTick, reactive, ref } from "vue"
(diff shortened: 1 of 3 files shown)


Timeraa:
Encountered this issue as well when I was working on #3723, glad I wasn't the only one

amk-dev (changes requested):
there is a method from the `graphql` package `getNamedType` that does the same thing as `resolveRootType`. can we use that instead of `resolveRootType`

https://graphql.org/graphql-js/type/#getnamedtype

liyasthomas:
Updated to `getNamedType` instead of `resolveRootType`.

amk-dev (approved):

jamesgeorge007 (approved):

PR #3742, opened 2024-01-11
refactor: persist running requests while switching tabs

nivedin:
Closes HFE-384

### Description
This PR introduces a feature that empowers tabs to continue executing their respective requests seamlessly even during tab-switching scenarios, preventing premature cancellations that were occurring previously.

### Checks

### DIFF
--- a/packages/hoppscotch-common/src/helpers/RequestRunner.ts
+++ b/packages/hoppscotch-common/src/helpers/RequestRunner.ts
@@ -154,6 +154,9 @@ export function runRESTRequest$(
           )
 
           if (E.isRight(runResult)) {
+            // set the response in the tab so that multiple tabs can run request simultaneously
+            tab.value.document.response = res
+
             tab.value.document.testResults = translateToSandboxTestResults(
               runResult.right
             )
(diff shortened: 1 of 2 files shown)


packages/hoppscotch-common/src/helpers/RequestRunner.ts
```diff
@@ -154,6 +154,9 @@ export function runRESTRequest$(
           )
 
           if (E.isRight(runResult)) {
+            // set the response in the tab so that multiple tabs can run request simultaneously
+            tab.value.document.response = res
```
anwarulislam, line 158: We have `updateRESTResponse` method in `Request.vue` component which updates the tab document with the response it received from the network call.

So, when we change the tab updating response into tab document from `RequestRunner` works fine. But, if we don't change the tab the it gets updated twice because the component didn't unmount hence the existing code in that component also updates the response

anwarulislam (approved):
LGTM 🚀

amk-dev (approved):

jamesgeorge007 (commented):
For an ongoing request, should we cancel it if the tab gets closed?

liyasthomas:
> For an ongoing request, should we cancel it if the tab gets closed?

If the tab gets closed, while waiting for a response, we should cancel it. This should also help to regain performance.

jamesgeorge007 (approved):

PR #3752, opened 2024-01-16
simple changes for assigning null values

mir-ishfaq-hussain:


packages/hoppscotch-backend/src/user/user.service.ts
```diff
@@ -41,12 +41,8 @@ export class UserService {
...
     return {
       ...dbUser,
-      currentRESTSession: dbCurrentRESTSession
-        ? JSON.stringify(dbCurrentRESTSession)
-        : null,
-      currentGQLSession: dbCurrentGQLSession
-        ? JSON.stringify(dbCurrentGQLSession)
-        : null,
+      currentRESTSession: JSON.stringify(dbCurrentRESTSession) ?? null,
+      currentGQLSession:  JSON.stringify(dbCurrentGQLSession) ?? null,
```
anwarulislam, line 45: `JSON.stringify(null)` output for this will be `"null"`. You have to check if the value is null or not in the first place. And then stringify later

anwarulislam (changes requested):
Look at the comment

packages/hoppscotch-backend/src/user/user.service.ts
balub: @mir-ishfaq-hussain Please revert these changes to the old code.

balub (changes requested):

packages/hoppscotch-backend/src/user/user.service.ts
mir-ishfaq-hussain: @balub i have done the required changes

PR #3754, opened 2024-01-17
feat: press tab on current input field to focus the next input field

anwarulislam:
Closes HFE-385

### Description
This PR modifies the keyboard event handling in the `EnvInput.vue` component. 
- Pressing the `Tab` will focus the next input field for better accessibility
- Previously, an `enter` event was emitted to hide suggestions popover and run request at the same time while `Tab` was used to select and hide suggestions popover. Now, `Tab` will do its default behavior. `Enter` will select the suggestion from the suggestions list and emit the event to run the request when the suggestions popover is not opened.

- [ ] Not Completed
- [x] Completed


liyasthomas (approved):
LGTM 💯

packages/hoppscotch-common/src/components/smart/EnvInput.vue
```diff
@@ -176,10 +178,15 @@ const handleKeystroke = (ev: KeyboardEvent) => {
     return
   }
 
-  showSuggestionPopover.value = true
+  // show suggestions only when the user types a character other than enter or tab
+  if (ev.key === "Tab") {
```
JoelJacobStephen, line 182: Can you make this into one line where if it is Enter, make the suggestions false?

JoelJacobStephen (approved):
Have a look at the comment below. Just a minor nitpick. Other than that, LGTM

JoelJacobStephen (approved):

packages/hoppscotch-common/src/components/smart/EnvInput.vue
```diff
@@ -176,10 +178,11 @@ const handleKeystroke = (ev: KeyboardEvent) => {
     return
   }
 
-  showSuggestionPopover.value = true
+  // show suggestions only when the user types a character other than enter or tab
+  showSuggestionPopover.value = !["Enter", "Tab"].includes(ev.key)
 
   if (
-    ["Enter", "Tab"].includes(ev.key) &&
+    ["Enter"].includes(ev.key) &&
```
jamesgeorge007, line 185: Do we need the above condition since we intend not to show suggestions while pressing `Enter`?

San-77x (commented):
--

PR #3777, opened 2024-01-25
chore(cli): emit bundle in ESM format

jamesgeorge007:
### Description

This PR brings the necessary updates to the build system for the CLI to emit an ESM-only bundle. Along with that, the dependencies are bumped, and the following are removed:
  - [esm](https://www.npmjs.com/package/esm)
  - [io-ts](https://www.npmjs.com/package/io-ts)

Follow up of #3441.

Closes HFE-278.

### Checks

### Note to reviewers

The dependencies, while inlined in the bundle, caused issues relating to the usage of `require` in the ESM context; hence they're no longer inlined ([officially recommended](https://tsup.egoist.dev/#excluding-all-packages)) and moved under `dependencies`. For `fp-ts`, it's inlined as part of the bundle and kept under `devDependencies` since the import paths were expected to specify the extensions `import * as TE from fp-ts/lib/TaskEither.js`.

### DIFF
--- a/packages/hoppscotch-cli/bin/hopp
+++ b/packages/hoppscotch-cli/bin/hopp
@@ -1,3 +1,0 @@
-#!/usr/bin/env node
-// * The entry point of the CLI
-require("../dist").cli(process.argv);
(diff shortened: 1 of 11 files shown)


AndrewBastin (approved):
lgtm 💯

PR #3779, opened 2024-01-25
feat: secret variables in environments

nivedin:
Closes HFE-220 HFE-393 HFE-358

### Description
**A user** will now be able to create a secret variable in the **personal environment**;
**Any owner or editor** will now be able to create a “**secret**” variable. in a **team environment**;

1. A normal environment variable: This is the current implementation of environment variables. Normal variables in team environments are shared across the team workspace.
2. A secret environment variable: The value of the secret environment variable will be masked and never synced to servers/DB.

This PR also add a additional check in the environment inspector which alerts the user if a environment variable has empty value.

### Checks

### DIFF
--- a/packages/hoppscotch-common/src/components/environments/ImportExport.vue
+++ b/packages/hoppscotch-common/src/components/environments/ImportExport.vue
@@ -9,7 +9,7 @@
 </template>
 
 <script setup lang="ts">
-import { Environment } from "@hoppscotch/data"
+import { Environment, NonSecretEnvironment } from "@hoppscotch/data"
 import * as E from "fp-ts/Either"
 import { ref } from "vue"
 
@@ -340,16 +340,12 @@ const showImportFailedError = () => {
 
 const handleImportToStore = async (
   environments: Environment[],
-  globalEnv?: Environment
+  globalEnv?: NonSecretEnvironment
 ) => {
   // if there's a global env, add them to the store
   if (globalEnv) {
-    globalEnv.variables.map((env) => {
-      if (env.secret) {
-        addGlobalEnvVariable({ key: env.key, secret: true })
-      } else {
-        addGlobalEnvVariable({ key: env.key, value: env.value, secret: false })
-      }
+    globalEnv.variables.map(({ key, value, secret }) => {
+      addGlobalEnvVariable({ key, value, secret })
     })
   }
 
--- a/packages/hoppscotch-common/src/components/environments/ImportExport.vue
+++ b/packages/hoppscotch-common/src/components/environments/ImportExport.vue
@@ -344,9 +344,9 @@ const handleImportToStore = async (
 ) => {
   // if there's a global env, add them to the store
   if (globalEnv) {
-    globalEnv.variables.map(({ key, value, secret }) => {
+    globalEnv.variables.forEach(({ key, value, secret }) =>
       addGlobalEnvVariable({ key, value, secret })
-    })
+    )
   }
 
   if (props.environmentType === "MY_ENV") {
--- a/packages/hoppscotch-common/src/components/environments/Selector.vue
+++ b/packages/hoppscotch-common/src/components/environments/Selector.vue
@@ -497,15 +497,20 @@ const selectedEnv = computed(() => {
         type: "MY_ENV",
         index: props.modelValue.index,
         name: props.modelValue.environment?.name,
+        variables: props.modelValue.environment?.variables,
       }
     } else if (props.modelValue?.type === "team-environment") {
       return {
         type: "TEAM_ENV",
         name: props.modelValue.environment.environment.name,
         teamEnvID: props.modelValue.environment.id,
+        variables: props.modelValue.environment.environment.variables,
       }
     }
-    return { type: "global", name: "Global" }
+    return {
+      type: "global",
+      name: "Global",
+    }
   }
   if (selectedEnvironmentIndex.value.type === "MY_ENV") {
     const environment =
--- a/packages/hoppscotch-common/src/components/environments/Selector.vue
+++ b/packages/hoppscotch-common/src/components/environments/Selector.vue
@@ -593,9 +593,7 @@ const environmentVariables = computed(() => {
 })
 
 const editGlobalEnv = () => {
-  invokeAction("modals.my.environment.edit", {
-    envName: "Global",
-  })
+  invokeAction("modals.global.environment.update", {})
 }
 
 const editEnv = () => {
--- a/packages/hoppscotch-common/src/components/environments/Selector.vue
+++ b/packages/hoppscotch-common/src/components/environments/Selector.vue
@@ -209,17 +209,11 @@
                 <span class="min-w-[9rem] w-1/4 truncate text-secondaryLight">
                   {{ variable.key }}
                 </span>
-                <span
-                  v-if="!variable.secret"
-                  class="min-w-[9rem] w-full truncate text-secondaryLight"
-                >
-                  {{ variable.value }}
-                </span>
-                <span
-                  v-else
-                  class="min-w-[9rem] w-full truncate text-secondaryLight"
-                >
-                  ********
+                <span class="min-w-[9rem] w-full truncate text-secondaryLight">
+                  <template v-if="variable.secret"> ******** </template>
+                  <template v-else>
+                    {{ variable.value }}
+                  </template>
                 </span>
               </div>
               <div v-if="globalEnvs.length === 0" class="text-secondaryLight">
@@ -273,17 +267,11 @@
                 <span class="min-w-[9rem] w-1/4 truncate text-secondaryLight">
                   {{ variable.key }}
                 </span>
-                <span
-                  v-if="!variable.secret"
-                  class="min-w-[9rem] w-full truncate text-secondaryLight"
-                >
-                  {{ variable.value }}
-                </span>
-                <span
-                  v-else
-                  class="min-w-[9rem] w-full truncate text-secondaryLight"
-                >
-                  ********
+                <span class="min-w-[9rem] w-full truncate text-secondaryLight">
+                  <template v-if="variable.secret"> ******** </template>
+                  <template v-else>
+                    {{ variable.value }}
+                  </template>
                 </span>
               </div>
               <div
--- a/packages/hoppscotch-common/src/helpers/import-export/export/environment.ts
+++ b/packages/hoppscotch-common/src/helpers/import-export/export/environment.ts
@@ -12,8 +12,6 @@ const getEnvironmentJson = (
       ? cloneDeep(environmentObj.environment)
       : cloneDeep(environmentObj)
 
-  delete newEnvironment.id
-
   const environmentId =
     environmentIndex || environmentIndex === 0
       ? environmentIndex
--- a/packages/hoppscotch-common/src/helpers/preRequest.ts
+++ b/packages/hoppscotch-common/src/helpers/preRequest.ts
@@ -25,7 +25,6 @@ const unsecretEnvironments = (
     if (secretVar) {
       return {
         ...globalVar,
-        secret: false,
         value: secretVar.value,
       }
     }
@@ -41,7 +40,6 @@ const unsecretEnvironments = (
       if (secretVar) {
         return {
           ...selectedVar,
-          secret: false,
           value: secretVar.value,
         }
       }
--- a/packages/hoppscotch-common/src/helpers/preRequest.ts
+++ b/packages/hoppscotch-common/src/helpers/preRequest.ts
@@ -27,6 +27,11 @@ const unsecretEnvironments = (
         ...globalVar,
         value: secretVar.value,
       }
+    } else if (!("value" in globalVar) || !globalVar.value) {
+      return {
+        ...globalVar,
+        value: "",
+      }
     }
     return globalVar
   })
--- a/packages/hoppscotch-common/src/helpers/preRequest.ts
+++ b/packages/hoppscotch-common/src/helpers/preRequest.ts
@@ -7,7 +7,15 @@ import {
   getCurrentEnvironment,
   getGlobalVariables,
 } from "~/newstore/environments"
-import { TestResult } from "@hoppscotch/js-sandbox"
+import { TestDescriptor } from "@hoppscotch/js-sandbox"
+
+type TestResultWithSelectedEnv = {
+  tests: TestDescriptor[]
+  envs: {
+    global: Environment["variables"]
+    selected: Environment
+  }
+}
 
 export const getCombinedEnvVariables = () => ({
   global: cloneDeep(getGlobalVariables()),
@@ -20,5 +28,5 @@ export const getFinalEnvsFromPreRequest = (
     global: Environment["variables"]
     selected: Environment
   }
-): Promise<E.Either<string, TestResult["envs"]>> =>
+): Promise<E.Either<string, TestResultWithSelectedEnv["envs"]>> =>
   runPreRequestScript(script, envs)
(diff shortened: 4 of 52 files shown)


jamesgeorge007 (changes requested):
- Updates to the secret environment variable values don't persist on saving while at the variables tab.

  https://github.com/hoppscotch/hoppscotch/assets/25279263/c2b31d53-ef8b-4044-afa3-a12e2200addd

- In the personal and team workspace, values specified for secret environment variables are not saved in the initial attempt. It requires re-opening the modal and making the changes.

  https://github.com/hoppscotch/hoppscotch/assets/25279263/38c22443-41a4-409c-a4a2-cbaea83d29f2

- Secret environment variables under the team workspace do not persist in the request and scripting context.

  https://github.com/hoppscotch/hoppscotch/assets/25279263/f4fda971-437b-4f37-8576-7a11cae64cf2

- When a secret environment variable is specified as the Bearer token, there isn't an indication of the value under the Authorization header. Shouldn't a masked representation of the variable be present next to `Bearer`? Similarly, for the `API key` Auth type, referring to secret environment variables as part of the key-value pairs (supplied via headers), the value appears empty instead of a masked entry.

  https://github.com/hoppscotch/hoppscotch/assets/25279263/2d79537f-633c-4522-ae4d-87a8235f8329

- Opting for importing from Hoppscotch environments and importing files based on Postman and Insomnia formats should be restricted with an error toast conveying the same.

- Is it intentional that global secret environment variables can be updated by a team member with a role viewer?

packages/hoppscotch-js-sandbox/src/utils.ts
```diff
@@ -104,14 +111,20 @@ const getSharedMethods = (envs: TestResult["envs"]) => {
...
+          E.map((e) =>
+            pipe(
+              parseTemplateStringE(!e.secret ? e.value?.toString() : "", [
+                ...updatedEnvs.selected,
+                ...updatedEnvs.global,
+              ]),
+              // If the recursive resolution failed, return the unresolved value
+              E.getOrElse(() => (!e.secret ? e.value?.toString() : ""))
+            )
+          )
```
jamesgeorge007, line 127:
```ts
      E.map(({ value }) =>
        pipe(
          parseTemplateStringE(value ?? "", [
            ...updatedEnvs.selected,
            ...updatedEnvs.global,
          ]),
          // If the recursive resolution failed, return the unresolved value
          E.getOrElse(() => value ?? "")
        )
      ),
```

Won't the `E.fromOption` above handle the case with invalid keys? Since secret environment variables are unwrapped before reaching the scripting context, we can specify value directly with a fallback to an empty string.

packages/hoppscotch-data/src/environment/v/1.ts
```diff
@@ -0,0 +1,44 @@
+import { z } from "zod"
+import { defineVersion } from "verzod"
+import { V0_SCHEMA } from "./0"
+
+export const V1_SCHEMA = z.object({
+  v: z.literal(1),
+  id: z.string(),
```
jamesgeorge007, line 7: Is `id` a required field? It's marked optional in the `v0` schema.

jamesgeorge007:
- Updates made to environments from the scripting context aren't persisted in the UI. This can be reproduced with the following.
  - Open a request and add the following pre-request script.
    
    ```js
      pw.env.set("preReqVar", "value")
    ```
  - Now add the following post-request script
  
    ```js
      pw.expect(pw.env.get("preReqVar")).toBe("value")

      pw.env.set("postReqVarOne", "value")

      pw.expect(pw.env.get("postReqVarOne")).toBe("value")

      pw.env.set("postReqVarTwo", "new-value")
      pw.expect(pw.env.get("postReqVarTwo")).toBe("new-value")
    ```

  - Perform the request; observe all the tests pass, but the newly added environment variables aren't added.

    https://github.com/hoppscotch/hoppscotch/assets/25279263/3d92d598-46d9-4e67-9ff9-cfff5339224f

- For the case when there isn't an environment selected, and the user decides to add environment variables created from the scripting context to the Global environment, we might need to open the corresponding modal so that they can choose between regular/secret variables.
  
  image

- In the create/edit environments modal, values for variables appear empty while hitting enter and get restored while typing any other character.

  https://github.com/hoppscotch/hoppscotch/assets/25279263/25a2e8ac-cc08-4f1b-a59f-b5520f560972

packages/hoppscotch-common/src/helpers/preRequest.ts
```diff
@@ -8,11 +8,63 @@ import {
...
+      index
+    )
+    if (secretVar) {
+      return {
+        ...globalVar,
+        secret: false,
+        value: secretVar.value,
+      }
+    }
+    return globalVar
```
jamesgeorge007, line 32:
```suggestion
    return {
      ...globalVar,
      value:
        (
          globalVar as {
            value: string
            key: string
            secret: false
          }
        ).value ?? "",
    }
```

I noticed once while importing an environment with secret values being empty, and attempting to update it from the scripting context didn't work as expected since the `value` field wasn't present in the incoming data. Would it be helpful to ensure the `value` field exists for the fallback?

packages/hoppscotch-common/src/helpers/preRequest.ts
```diff
@@ -8,11 +8,63 @@ import {
...
+        index
+      )
+      if (secretVar) {
+        return {
+          ...selectedVar,
+          secret: false,
+          value: secretVar.value,
+        }
+      }
+      return selectedVar
```
jamesgeorge007, line 48:
```suggestion
      return {
        ...selectedVar,
        value:
          (
            selectedVar as {
              value: string
              key: string
              secret: false
            }
          ).value ?? "",
      }
```

Similar to the above.

packages/hoppscotch-common/src/services/persistence/index.ts
```diff
@@ -438,6 +445,56 @@ export class PersistenceService extends Service {
     })
   }
 
+  private setupSecretEnvironmentsPersistence() {
+    const secretEnvironmentsKey = "secretEnvironments"
+    const secretEnvironmentsData = window.localStorage.getItem(
+      secretEnvironmentsKey
+    )
```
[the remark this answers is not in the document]
nivedin: yes we are not storing the values for the secret environment variables in the environment store hence a seperation.

packages/hoppscotch-common/src/services/secret-environment.service.ts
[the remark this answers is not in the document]
nivedin: added

packages/hoppscotch-data/src/environment/v/1.ts
nivedin: its optional because in older version there is case where id will not be present, and in newer version its mandatory

jamesgeorge007 (changes requested):
- Environments created from the scripting context are added straightaway to the Global environment without confirmation from the modal.

  https://github.com/hoppscotch/hoppscotch/assets/25279263/ba208443-10eb-4c83-92cd-a85b87547b45

- When duplicating an environment, the secret variable values set in the original environment are lost.

  https://github.com/hoppscotch/hoppscotch/assets/25279263/e70c1fc8-63d3-423a-a23c-113c50982a00

- Opting for importing from Hoppscotch environments and importing other file formats should be restricted.

- Secret environment variables referred to in the request body aren't resolved properly, [Ref](https://github.com/hoppscotch/hoppscotch/pull/3779#discussion_r1468832630).

- Let's remove the `strict` requirement from the schemas.

packages/hoppscotch-common/src/services/persistence/index.ts
```diff
@@ -438,6 +444,56 @@ export class PersistenceService extends Service {
...
+        window.localStorage.setItem(
+          secretEnvironmentsKey,
+          JSON.stringify(newSecretEnvironment)
+        )
+      },
+      {
+        debounce: 200,
+      }
+    )
+  }
```
jamesgeorge007, line 495: Do we want to track changes made to the nested properties? In that case, supplying `deep: true` will be required. Other instances use a debounce delay of `500ms`. Should we keep it the same here?

packages/hoppscotch-common/src/services/persistence/validation-schemas/index.ts
```diff
@@ -212,13 +211,23 @@ export const MQTT_REQUEST_SCHEMA = z.nullable(
...
+          secret: z.literal(true),
+        })
+        .strict(),
+      z
+        .object({
+          key: z.string(),
+          value: z.string(),
+          secret: z.literal(false),
+        })
+        .strict(),
```
jamesgeorge007, line 229: Let's remove the strict requirement since the internal representation doesn't respect this every time.

packages/hoppscotch-common/src/services/persistence/validation-schemas/index.ts
```diff
@@ -358,7 +393,11 @@ const HoppTestResultSchema = z
           .object({
             additions: z.array(EnvironmentVariablesSchema),
             updations: z.array(
-              EnvironmentVariablesSchema.extend({ previousValue: z.string() })
+              EnvironmentVariablesSchema.and(
+                z.object({
+                  previousValue: z.string(),
+                })
+              )
             ),
```
jamesgeorge007, line 401: Is it not necessary to narrow this down to the case where `secret` is set to `false` that has a `value` associated?

packages/hoppscotch-common/src/services/persistence/validation-schemas/index.ts
```diff
@@ -367,7 +406,11 @@ const HoppTestResultSchema = z
           .object({
             additions: z.array(EnvironmentVariablesSchema),
             updations: z.array(
-              EnvironmentVariablesSchema.extend({ previousValue: z.string() })
+              EnvironmentVariablesSchema.and(
+                z.object({
+                  previousValue: z.string(),
+                })
+              )
```
jamesgeorge007, line 413: Is it not necessary to narrow this down to the case where secret is set to false that has a value associated?

packages/hoppscotch-common/src/services/persistence/validation-schemas/index.ts
```diff
@@ -339,12 +348,38 @@ const HoppTestDataSchema = z.lazy(() =>
...
+      z
+        .object({
+          key: z.string(),
+          value: z.string(),
+          varIndex: z.number(),
+        })
+        .strict()
+    )
+  ),
+])
```
jamesgeorge007, line 382: Let's remove the strict requirements.

packages/hoppscotch-common/src/services/persistence/validation-schemas/index.ts
```diff
@@ -339,12 +348,38 @@ const HoppTestDataSchema = z.lazy(() =>
...
+    .strict(),
+  z
+    .object({
+      key: z.string(),
+      secret: z.literal(true),
+    })
+    .strict(),
+])
+
+export const SecretEnvironmentVariable = z.union([
```
jamesgeorge007, line 367:
```suggestion
export const SECRET_ENVIRONMENT_VARIABLE_SCHEMA = z.union([
```

Let's follow the uppercase convention used for exposed schemas.

packages/hoppscotch-js-sandbox/src/utils.ts
```diff
@@ -86,9 +93,18 @@ const getSharedMethods = (envs: TestResult["envs"]) => {
...
         () => undefined,
-        ({ value }) => String(value)
+        (env) =>
+          String(
+            (
+              env as {
+                key: string
+                value: string | undefined
+                secret: boolean
+              }
```
jamesgeorge007, line 105: Since we know from the sandbox, that environment variables will follow the above shape, can we maintain a custom type under `src/types` or similar that can be reused across?

packages/hoppscotch-common/src/components/environments/ImportExport.vue
```diff
@@ -344,8 +344,12 @@ const handleImportToStore = async (
...
-    globalEnv.variables.forEach(({ key, value }) => {
-      addGlobalEnvVariable({ key, value })
+    globalEnv.variables.map((env) => {
+      if (env.secret) {
+        addGlobalEnvVariable({ key: env.key, secret: true })
+      } else {
+        addGlobalEnvVariable({ key: env.key, value: env.value, secret: false })
+      }
     })
   }
```
jamesgeorge007, line 354:
```ts
  if (globalEnv) {
    globalEnv.variables.map(({ key, value, secret }) => {
      addGlobalEnvVariable({ key, value, secret })
    })
  }
```

Since we're setting `secret: false` while parsing environment variables based on Insomnia export format, we can set the incoming values straightaway specifying the type as `globalEnv?: NonSecretEnvironment`.

packages/hoppscotch-common/src/helpers/import-export/export/environment.ts
```diff
@@ -12,8 +12,6 @@ const getEnvironmentJson = (
       ? cloneDeep(environmentObj.environment)
       : cloneDeep(environmentObj)
 
-  delete newEnvironment.id
```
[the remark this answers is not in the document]
nivedin: earlier the id was optional while exporting, now its mandatory

packages/hoppscotch-common/src/services/persistence/index.ts
nivedin: i think there is not much nesting happening in this case `persistableSecretEnvironments` is of the type `Record<string, {
  key: string
  value: string
  varIndex: number
}[]>` 

should we add deep, wdyt ?

packages/hoppscotch-common/src/components/environments/Selector.vue
```diff
@@ -209,9 +209,18 @@
...
+                  class="min-w-[9rem] w-full truncate text-secondaryLight"
+                >
                   {{ variable.value }}
                 </span>
+                <span
+                  v-else
+                  class="min-w-[9rem] w-full truncate text-secondaryLight"
+                >
+                  ********
+                </span>
```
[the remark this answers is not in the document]
nivedin: yes

packages/hoppscotch-common/src/services/persistence/index.ts
jamesgeorge007: Gotcha; a shallow watcher would suffice since the operations are only at the top level.

jamesgeorge007 (approved):

anwarulislam:
There is a simple accessibility issue here mentioned in the video.
If I press to add a new variable give the name of the variable and press `tab` then it doesn't focus on the next input field (value for the variable). Instead, it focuses on the lock icon.

https://github.com/hoppscotch/hoppscotch/assets/26295990/d6ac96cc-7ee6-4b2a-b565-7342ca3432c6

anwarulislam (changes requested):
There is a simple accessibility issue here mentioned in the video.
If I press to add a new variable give the name of the variable and press `tab` then it doesn't focus on the next input field (value for the variable). Instead, it focuses on the lock icon.

It is because the input field is disabled as initially, the secret visibility is disabled. I think we should keep it open and after writing the secret we can make the state disabled again

https://github.com/hoppscotch/hoppscotch/assets/26295990/d6ac96cc-7ee6-4b2a-b565-7342ca3432c6

jamesgeorge007:
I noticed the focus goes to the value field while adding new non-secret environment variable entries. This seems like a pre-existing issue. Can we investigate this behavior as well?

https://github.com/hoppscotch/hoppscotch/assets/25279263/5744fca8-94d0-49d8-838b-e82892eff893

jamesgeorge007 (changes requested):
Attempting to set a value for an empty secret environment variable that is currently active and opening the `Global Edit Environment` modal picks Secret as the active tab. It goes back to Variables for further attempts. This happens in the personal workspace and while consuming personal environments in the team workspace.

https://github.com/hoppscotch/hoppscotch/assets/25279263/e5af97cd-a866-4cb0-ba65-9cc6e4101235

jamesgeorge007 (approved):

anwarulislam (approved):
LGTM 🚀

PR #3786, opened 2024-01-30
HSB-358 feat: admin user management (backend)

mirarifhasan:
Closes HSB-358

### Description

This PR introduces the following features (for Admin Dashboard) in the `hoppscotch-backend` package:

1. Admin can update user's displayName
2. Show only the pending users list (in the invited user tab)
3. Pending user invite revocation
4. The user's list will be sorted by displayName ('asc') and the Admin users will be at the top
5. User search by their name & email
6. Bulk: user to admin
7. Bulk: Admin to user
8. Bulk: User delete

### Checks

### How to Test
The following mutation/queries are created/updated. Testing these operations will be good.

New Mutations::

1. updateUserDisplayNameByAdmin
2. revokeUserInvitationsByAdmin
3. removeUsersByAdmin
4. removeUsersAsAdmin
5. makeUsersAdmin

Updated Queries::

1. infra.allUsersV2 -> Input parameter changed. Response format unchanged
2. infra.invitedUsers -> Input parameter changed. Response format unchanged

### Additional Information

Nil


packages/hoppscotch-backend/src/admin/admin.resolver.ts
```diff
@@ -269,6 +70,27 @@ export class AdminResolver {
...
+  async revokeUserInvitationByAdmin(
+    @GqlAdmin() adminUser: Admin,
+    @Args({
+      name: 'inviteeEmail',
+      description: 'Invite Email',
+      type: () => ID,
+    })
+    inviteeEmail: string,
+  ): Promise<boolean> {
+    const invite = await this.adminService.revokeUserInvite(
```
balub, line 86: I can see this also becoming a bulk actionable feature in the future, shall we make it accept an array or invitee emails. I think we should do it this way since its a very small change.

packages/hoppscotch-backend/src/user/user.model.ts
```diff
@@ -56,3 +56,22 @@ export enum SessionType {
 registerEnumType(SessionType, {
   name: 'SessionType',
 });
+
+@ObjectType()
+export class UserDeleteData {
```
balub, line 61:
```suggestion
export class UserDeletionResult {
```

packages/hoppscotch-backend/src/user/user.model.ts
```diff
@@ -56,3 +56,22 @@ export enum SessionType {
...
+  @Field(() => ID, {
+    description: 'UID of the user',
+  })
+  userUID: string;
+
+  @Field(() => Boolean, {
+    description: 'Flag to determine if user deletion was successful or not',
+  })
+  success: Boolean;
```
balub, line 70:
```suggestion
  isDeleted: Boolean;
```

packages/hoppscotch-backend/src/admin/admin.resolver.ts
```diff
@@ -281,10 +103,30 @@ export class AdminResolver {
...
     userUID: string,
   ): Promise<boolean> {
-    const invitedUser = await this.adminService.removeUserAccount(userUID);
-    if (E.isLeft(invitedUser)) throwErr(invitedUser.left);
-    return invitedUser.right;
+    const removedUser = await this.adminService.removeUserAccount(userUID);
+    if (E.isLeft(removedUser)) throwErr(removedUser.left);
+    return removedUser.right;
   }
+
```
balub, line 110: I dont like the fact we have 2 seperate mutations to do the same operation, I understand maybe its done this way due to backwards compatability reasons but this mutation is only used on one location in the admin app so I think changing it is there is fine.

packages/hoppscotch-backend/src/user/user.service.ts
```diff
@@ -88,6 +89,20 @@ export class UserService {
     }
   }
 
+  /**
+   * Find Non-Admin Users with given IDs
```
balub, line 93: The function description is incorrect, the fuction does not do anything explicit to return only non-admin users.

packages/hoppscotch-backend/src/admin/admin.service.ts
```diff
@@ -349,6 +432,59 @@ export class AdminService {
...
   }
 
+  /**
+   * Remove user (not Admin) accounts by UIDs
+   * @param userUid User UIDs
+   * @returns an Either of boolean or error
+   */
+  async removeUserAccounts(userUIDs: string[]) {
+    const allUsers = await this.userService.findUsersByIds(userUIDs);
+    if (allUsers.length === 0) return E.left(USER_NOT_FOUND);
```
balub, line 442: I think we should throw a different error for multiple users since the current one is used in many locations when a user account cannot be found.

packages/hoppscotch-backend/src/admin/admin.service.ts
```diff
@@ -349,6 +432,59 @@ export class AdminService {
...
+
+    const userDeleteResult: UserDeleteData[] = [];
```
balub, line 444:
```suggestion
    const userDeletionResults: UserDeleteData[] = [];
```

packages/hoppscotch-backend/src/admin/admin.service.ts
```diff
@@ -349,6 +432,59 @@ export class AdminService {
...
+      if (user.isAdmin) {
+        userDeleteResult.push({
+          userUID: user.uid,
+          success: false,
+          errorMessage: ADMIN_CAN_NOT_BE_DELETED,
+        });
+      }
+    });
+
+    const normalUsers = allUsers.filter((user) => !user.isAdmin);
```
balub, line 457:
```suggestion
    const nonAdminUsers = allUsers.filter((user) => !user.isAdmin);
```

packages/hoppscotch-backend/src/admin/admin.service.ts
```diff
@@ -349,6 +432,59 @@ export class AdminService {
     return E.right(delUser.right);
   }
 
+  /**
+   * Remove user (not Admin) accounts by UIDs
+   * @param userUid User UIDs
+   * @returns an Either of boolean or error
+   */
+  async removeUserAccounts(userUIDs: string[]) {
+    const allUsers = await this.userService.findUsersByIds(userUIDs);
```
balub, line 441:
```suggestion
    const allUsersList = await this.userService.findUsersByIds(userUIDs);
```

balub (changes requested):

packages/hoppscotch-backend/src/admin/admin.service.spec.ts
```diff
@@ -134,6 +145,75 @@ describe('AdminService', () => {
     });
   });
 
+  describe('revokeUserInvite', () => {
+    test('should resolve left and return error if email not invited', async () => {
+      mockPrisma.invitedUsers.delete.mockRejectedValueOnce(new Error());
```
balub, line 150:
```suggestion
      mockPrisma.invitedUsers.delete.mockRejectedValueOnce('RecordNotFound');
```

packages/hoppscotch-backend/src/admin/admin.service.spec.ts
```diff
@@ -134,6 +145,75 @@ describe('AdminService', () => {
...
+        invitedUsers[0],
+      );
+    });
+  });
+
+  describe('removeUsersAsAdmin', () => {
+    test('should resolve right and make admins to users', async () => {
+      mockUserService.fetchAdminUsers.mockResolvedValueOnce([
+        { uid: '123' } as any,
+      ]);
```
balub, line 197: This should return valid `User` objects instead of what is it currently.

packages/hoppscotch-backend/src/admin/admin.service.spec.ts
```diff
@@ -134,6 +145,75 @@ describe('AdminService', () => {
...
+
+      expect(mockPubSub.publish).toHaveBeenCalledWith(
+        `admin/${adminUid}/invitation_revoked`,
+        invitedUsers[0],
+      );
+    });
+  });
+
+  describe('removeUsersAsAdmin', () => {
+    test('should resolve right and make admins to users', async () => {
```
balub, line 194: Please rewrite this test it is incorrect.

packages/hoppscotch-backend/src/admin/admin.service.spec.ts
```diff
@@ -134,6 +145,75 @@ describe('AdminService', () => {
...
+
+      return expect(
+        await adminService.removeUsersAsAdmin(['123456']),
+      ).toEqualRight(true);
+    });
+
+    test('should resolve left and return error if only one admin in the infra', async () => {
+      mockUserService.fetchAdminUsers.mockResolvedValueOnce([
+        { uid: '123' } as any,
+      ]);
```
balub, line 208: Here too return the correct object.

packages/hoppscotch-backend/src/user/user.service.spec.ts
```diff
@@ -176,6 +176,26 @@ describe('UserService', () => {
...
+      expect(result).toEqual(users);
+    });
+
+    test('should return empty array of users given a invalid user UIDs', async () => {
+      mockPrisma.user.findMany.mockResolvedValueOnce([]);
+
+      const result = await userService.findNonAdminUsersByIds(['sdcvbdbr']);
+      expect(result).toEqual([]);
+    });
+  });
```
balub, line 197: This method does not exist, please take a look at this test.

packages/hoppscotch-backend/src/user/user.service.spec.ts
```diff
@@ -414,6 +434,41 @@ describe('UserService', () => {
...
+    test('should resolve right and update user', async () => {
+      mockPrisma.user.update.mockResolvedValueOnce(user);
+
+      const result = await userService.updateUser(user.uid, user.displayName);
+      expect(result).toEqualRight({
+        ...user,
+        currentGQLSession: JSON.stringify(user.currentGQLSession),
+        currentRESTSession: JSON.stringify(user.currentRESTSession),
+      });
+    });
```
balub, line 447: This test is technically not invalid, but I think we should actually write the test to update to a different name and check.

packages/hoppscotch-backend/src/user/user.service.spec.ts
```diff
@@ -424,7 +479,7 @@ describe('UserService', () => {
     test('should resolve right and return next 20 users when cursor is provided', async () => {
       mockPrisma.user.findMany.mockResolvedValueOnce(users);
 
-      const result = await userService.fetchAllUsers('123344', 20);
+      const result = await userService.fetchAllUsers('123456', 20);
```
balub, line 482:
```suggestion
      const result = await userService.fetchAllUsers('123344', 20);
```

packages/hoppscotch-backend/src/user/user.service.spec.ts
```diff
@@ -435,6 +490,36 @@ describe('UserService', () => {
...
+    test('should resolve right and return first 20 users when searchString is null', async () => {
+      mockPrisma.user.findMany.mockResolvedValueOnce(users);
+
+      const result = await userService.fetchAllUsersV2(null, {
+        take: 20,
+        skip: 0,
+      });
+      expect(result).toEqual(users);
+    });
+    test('should resolve right and return next 20 users when cursor is provided', async () => {
```
balub, line 503:
```suggestion
    test('should resolve right and return next 20 users when searchString is provided', async () => {
```

packages/hoppscotch-backend/src/user/user.service.spec.ts
```diff
@@ -556,4 +641,12 @@ describe('UserService', () => {
       expect(result).toEqual(10);
     });
   });
+
+  describe('removeUsersAsAdmin', () => {
```
balub, line 645: Please add more test cases for different situations like if userUids provided was not present in the db etc.

balub (changes requested):

packages/hoppscotch-backend/src/admin/admin.resolver.ts
mirarifhasan: Done

packages/hoppscotch-backend/src/admin/admin.service.ts
```diff
@@ -110,14 +133,68 @@ export class AdminService {
...
+    return E.right(true);
+  }
+
+  /**
+   * Revoke infra level user invitations
+   * @param inviteeEmails Invitee's emails
+   * @param adminUid Admin Uid
+   * @returns an Either of boolean or error string
+   */
+  async revokeUserInvites(inviteeEmails: string[], adminUid: string) {
```
balub, line 158: Why are we passing in the `adminUid` if we are not using it, please remove it.

balub:
```suggestion
  async revokeUserInvitations(inviteeEmails: string[], adminUid: string) {
```

packages/hoppscotch-backend/src/admin/admin.service.spec.ts
```diff
@@ -58,20 +63,87 @@ const invitedUsers: InvitedUsers[] = [
...
+  email: 'email@email.com',
+  photoURL: 'photoURL',
+  isAdmin: false,
+  refreshToken: 'refreshToken',
+  currentRESTSession: '',
+  currentGQLSession: '',
+  createdOn: new Date(),
+};
+
 describe('AdminService', () => {
```
balub, line 103: Please rewrite the testcases for this method.

balub (commented):
Please add the deprecated tab in jsdoc to all the service methods we will be removing.

balub (commented):
Please add the deprecated tab in jsdoc to all the service methods we will be removing.

balub (approved):
LGTM 🚀

PR #3702, opened 2023-12-28
refactor: improvements to users module in admin dashboard

JoelJacobStephen:
### Ticket
Closes HFE-377

### Description
This PR focuses on improving the users module in Admin Dashboard. Each component related to the users module is refactored and improved upon and minor bugs found are fixed.

### Objectives
- [x] Refactor users module in admin dashboard

### Checks

- [x] My pull request adheres to the code style of this project
- [ ] My code requires changes to the documentation
- [ ] I have updated the documentation as required
- [x] All the tests have passed

PR #3790, opened 2024-01-30
refactor: consolidated admin dashboard improvements

JoelJacobStephen:
### Ticket

* Closes 
	* HFE-363
	* HFE-373
	* HFE 377

* Suppresses 
	* #3686 
	* #3694 
	* #3702

### Description
This PR combines the improvements and changes made to different modules in the Admin Dashboard into one branch with all merge conflicts resolved. This allows seamless testing of all the new refactors introduced in the Admin Dashboard without any breaking changes 

### Checks


packages/hoppscotch-sh-admin/src/composables/useConfigHandler.ts
```diff
@@ -255,7 +259,21 @@ export function useConfigHandler(updatedConfigs?: Config) {
     return config;
   });
 
-  // Trasforming the working configs back into the format required by the mutations
+  // Checking if any of the config fields are empty
+  const isFieldEmpty = (field: string) => field === '';
```
anwarulislam, line 263: Can a field be `null` or `undefined`?

anwarulislam: If it can be `null` or `undefined` then we can just check `falsy` value like this:

```
const isFieldEmpty = (field: string) => !field;
```

anwarulislam (commented):
LGTM 🚀
Just check the comment

packages/hoppscotch-sh-admin/src/composables/useConfigHandler.ts
AndrewBastin: @anwarulislam since the defined type here forbids null/undefined, I don't think we need to check for that.

Instead the trim check is enough here.

AndrewBastin (approved):

PR #3792, opened 2024-01-30
chore: Oauth temporary ux improvements

amk-dev:
Fixes HFE-382

**Changes**

1. Disable the OpenID Connect Discovery URL when Authorization/Access Token URLs are present.
2. Disable Authorization/Access Token URLs when the OpenID Connect Discovery URL is present.
3. Implement OAuth fetch calls through interceptors.
4. Fix an if statement that made OpenID Connect discovery URLs mandatory

**Note**
We will be removing OpenID Connect Discovery URL very soon. this is temporary solution till then.

### DIFF
--- a/packages/hoppscotch-common/src/helpers/oauth.ts
+++ b/packages/hoppscotch-common/src/helpers/oauth.ts
@@ -297,19 +297,7 @@ const clearPKCEState = () => {
 }
 
 async function runRequestThroughInterceptor(config: AxiosRequestConfig) {
-  if (interceptorService.currentInterceptorID.value === "extension") {
-    // wait till the extension resolves
-    await until(extensionService.extensionStatus).toMatch(
-      (status) => status !== "waiting",
-      {
-        timeout: 1000,
-      }
-    )
-  }
-
   const res = await interceptorService.runRequest(config).response
-
-  // wait for the extension to resolve if the interceptor is extension
 
   if (E.isLeft(res)) {
     return E.left("REQUEST_FAILED")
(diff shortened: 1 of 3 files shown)


anwarulislam:
@amk-dev there is an issue you can see in the video. I think we need to handle empty value here.


https://github.com/hoppscotch/hoppscotch/assets/26295990/5ab58c80-7ad5-4cac-b0fc-8f83e1950d79

anwarulislam (changes requested):
@amk-dev there is an issue you can see in the video. I think we need to handle empty value here.


https://github.com/hoppscotch/hoppscotch/assets/26295990/5ab58c80-7ad5-4cac-b0fc-8f83e1950d79

packages/hoppscotch-common/src/helpers/oauth.ts
```diff
@@ -309,4 +296,32 @@ const clearPKCEState = () => {
...
+async function runRequestThroughInterceptor(config: AxiosRequestConfig) {
+  if (interceptorService.currentInterceptorID.value === "extension") {
+    // wait till the extension resolves
+    await until(extensionService.extensionStatus).toMatch(
+      (status) => status !== "waiting",
+      {
+        timeout: 1000,
+      }
+    )
+  }
```
AndrewBastin, line 308: I would recommend to keep specific Interceptor logic out of the general app code.

What if we do this wait within `ExtensionInterceptorService`'s `runRequest` call ?

AndrewBastin (changes requested):

packages/hoppscotch-common/src/helpers/oauth.ts
amk-dev: yeah. that would work.

anwarulislam (approved):
LGTM 🚀

AndrewBastin (approved):
lgtm

PR #3825, opened 2024-02-14
feat: request variables

nivedin:
Closes HFE-404 HFE-432

### Description
This PR add a feature which will provide users to specify a variable that is scoped to a request.

Request variables are variables whose scope is limited to the request. Users can now specify request variables using the notation `<<variable>`.

- Request variables vs query params - Request variables can be part of a URL, for example: 
 `api.hoppscotch.io/user/<<variable>>`
- Request variables vs environment variables: The scope of request variables is limited to the request only, and they can be used in embeds.

### Checks


jamesgeorge007 (changes requested):
- If entries from the request & environment variables share the same name, the empty entries from the request variable should fall back to the environment variables.

  https://github.com/hoppscotch/hoppscotch/assets/25279263/d707bcb4-9926-46fc-841c-7aef246f434e  

- The autocomplete for any request values will have an entry saying undefined if there exists, at least an entry under request variables that is turned off.

  https://github.com/hoppscotch/hoppscotch/assets/25279263/78db956f-afcf-4c10-8265-708738a317d8

- If the value for a request variable is set to an environment variable under the same name (key), hovering over the same results in the tooltip text says error.

  https://github.com/hoppscotch/hoppscotch/assets/25279263/2ee692ce-7e92-4911-b2a1-f12cf02a37f6

- Inspector shouldn't warn if the selected environment has a value set while the same is empty under `Global` - [HFE-432](https://linear.app/hoppscotch/issue/HFE-432/inspector-fails-to-recognize-value-in-environment).

- If a request variable refers to a secret environment variable, should the values appear masked in the tooltip text throughout?

- Encountered the error toast related to schema mismatch for `restTabState` once. We might need to make the `previousValue` field optional.
 https://github.com/hoppscotch/hoppscotch/blob/faab1d20fde9a6be660db40fc73dcf28f9038008/packages/hoppscotch-common/src/services/persistence/validation-schemas/index.ts#L400-L415

jamesgeorge007:
@nivedin, can you rebase against `release/2023.4.0` (FYI, there're merge conflicts) and update the base branch accordingly?

packages/hoppscotch-common/src/components/smart/EnvInput.vue
```diff
@@ -357,25 +360,63 @@ const aggregateEnvs = useReadonlyStream(aggregateEnvsWithSecrets$, []) as Ref<
...
+                  key: x.key,
+                  value: x.value,
+                  sourceEnv: "RequestVariable",
+                  secret: false,
+                }
+              : ([] as unknown as AggregateEnvironment)
+        ),
+        ...aggregateEnvs.value,
+      ]
 })
```
jamesgeorge007, line 397:
```suggestion
const envVars = computed(() => {
  if (props.envs) {
    return props.envs.map((x) => {
      const { key, secret } = x

      const value = secret ? "********" : x.value
      const sourceEnv = "sourceEnv" in x ? x.sourceEnv : null

      return {
        key,
        value,
        sourceEnv,
        secret,
      }
    })
  }

  return [
    ...tabs.currentActiveTab.value.document.request.requestVariables.map(
      ({ active, key, value }) =>
        active
          ? {
              key,
              value,
              sourceEnv: "RequestVariable",
              secret: false,
            }
          : ({} as AggregateEnvironment)
    ),
    ...aggregateEnvs.value,
  ]
})
```

Can you check if the above works? I've converted the ternary-based approach to follow an early return-based one as I felt it enhances readability. Also, I've grouped the checks in the first block (`props.envs` is available).

packages/hoppscotch-common/src/components/smart/EnvInput.vue
```diff
@@ -357,25 +360,63 @@ const aggregateEnvs = useReadonlyStream(aggregateEnvsWithSecrets$, []) as Ref<
...
+function envAutoCompletion(context: CompletionContext) {
+  const options = envVars.value
+    ? envVars.value.map((env) => {
+        return {
+          label: `<<${env.key}>>`,
+          info: env.value,
+          apply: `<<${env.key}>>`,
+        }
+      })
+    : []
```
jamesgeorge007, line 408:
```suggestion
  const options =
    envVars.value?.map((env) => {
      return {
        label: `<<${env.key}>>`,
        info: env.value,
        apply: `<<${env.key}>>`,
      }
    }) ?? []

```

packages/hoppscotch-common/src/composables/codemirror.ts
```diff
@@ -243,7 +243,7 @@ export function useCodemirror(
       if (from === to) return
       const text = view.value?.state.doc.sliceString(from, to)
       const { top, left } = view.value?.coordsAtPos(from)
-      if (text) {
+      if (text && text.trim()) {
```
jamesgeorge007, line 246:
```suggestion
      if (text?.trim()) {
```

packages/hoppscotch-common/src/helpers/RequestRunner.ts
```diff
@@ -175,15 +186,36 @@ export function runRESTRequest$(
...
+      tab.value.document.request.requestVariables.map(
+        (v: HoppRESTRequestVariable) => {
+          if (v.active) {
+            return {
+              key: v.key,
+              value: v.value,
+              secret: false,
+            }
+          }
+          return []
```
jamesgeorge007, line 199:
```suggestion
          return {}
```
Is it possible to filter out the inactive entries instead of preserving empty entries? Is it intentional to keep the list length the same as the original one?

packages/hoppscotch-common/src/helpers/editor/extensions/HoppEnvironment.ts
```diff
@@ -123,11 +128,16 @@ const cursorTooltipField = (aggregateEnvs: AggregateEnvironment[]) =>
...
-            envName: tooltipEnv?.sourceEnv !== "Global" ? envName : "Global",
-            variableName: parsedEnvKey,
-            isSecret: tooltipEnv?.secret,
-          })
+          if (tooltipEnv?.sourceEnv === "RequestVariable") {
+            restTabs.currentActiveTab.value.document.optionTabPreference =
+              "requestVariables"
+          } else {
+            invokeAction(invokeActionType, {
+              envName: tooltipEnv?.sourceEnv !== "Global" ? envName : "Global",
```
jamesgeorge007, line 136:
```suggestion
              envName: tooltipEnv?.sourceEnv === "Global" ? "Global" : envName ,
```

Let's check for the positive condition.

packages/hoppscotch-common/src/helpers/editor/extensions/HoppEnvironment.ts
```diff
@@ -209,10 +224,43 @@ export class HoppEnvironmentPlugin {
...
+    const aggregateEnvs = getAggregateEnvsWithSecrets()
+    const currentTab = restTabs.currentActiveTab.value
+
+    watch(
+      currentTab.document.request,
+      (reqVariables) => {
+        this.envs = [
+          ...reqVariables.requestVariables.map((variable) => ({
+            key: variable.key,
+            value: variable.value,
```
jamesgeorge007, line 236:
```suggestion
          ...reqVariables.requestVariables.map(({ key, value }) => ({
            key,
            value,
```

packages/hoppscotch-common/src/helpers/editor/extensions/HoppEnvironment.ts
```diff
@@ -209,10 +224,43 @@ export class HoppEnvironmentPlugin {
...
+      },
+      { immediate: true, deep: true }
+    )
 
     subscribeToStream(aggregateEnvsWithSecrets$, (envs) => {
-      this.envs = envs
+      this.envs = [
+        ...currentTab.document.request.requestVariables.map((variable) => ({
+          key: variable.key,
+          value: variable.value,
```
jamesgeorge007, line 257:
```suggestion
        ...currentTab.document.request.requestVariables.map(({ key, value }) => ({
          key,
          value,
```

packages/hoppscotch-common/src/services/inspection/inspectors/environment.inspector.ts
```diff
@@ -199,14 +219,19 @@ export class EnvironmentInspectorService extends Service implements Inspector {
...
-                        })
+                        if (env.sourceEnv === "RequestVariable") {
+                          currentTab.document.optionTabPreference =
+                            "requestVariables"
+                        } else {
+                          invokeAction(invokeActionType, {
+                            envName:
+                              env.sourceEnv !== "Global"
+                                ? currentSelectedEnvironment.name
+                                : "Global",
```
jamesgeorge007, line 230:
```suggestion
                              env.sourceEnv === "Global"
                                ? "Global"
                                : currentSelectedEnvironment.name,
```

packages/hoppscotch-common/src/helpers/import-export/import/openapi.ts
```diff
@@ -82,6 +83,27 @@ const parseOpenAPIParams = (params: OpenAPIParamsType[]): HoppRESTParam[] =>
...
+    variables,
+
+    A.filterMap(
+      flow(
+        O.fromPredicate((param) => param.in === "path"),
+        O.map(
+          (param) =>
+            <HoppRESTRequestVariable>{
+              key: param.name,
+              value: "", // TODO: Can we do anything more ? (parse default values maybe)
```
jamesgeorge007, line 99: `value` won't be present in the incoming data, right? Are we keeping it empty so that filling up the value is up to the user?

packages/hoppscotch-common/src/components/http/RequestVariables.vue
```diff
@@ -0,0 +1,411 @@
...
+          :icon="IconPlus"
+          :disabled="bulkMode"
+          @click="addVariable"
+        />
+      </div>
+    </div>
+    <div v-if="bulkMode" class="h-full relative">
+      <div ref="bulkEditor" class="absolute inset-0"></div>
+    </div>
+    <div v-else>
```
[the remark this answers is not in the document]
nivedin: the empty placeholder is inside the div, so I think we can keep it there

packages/hoppscotch-common/src/helpers/import-export/import/openapi.ts
nivedin: yup the value will be empty while importing

packages/hoppscotch-common/src/components/http/RequestVariables.vue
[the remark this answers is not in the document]
jamesgeorge007: Good catch. I don't think it's necessary though (we could replace it with a `template` wrapper). But, let's keep it as is for now.

jamesgeorge007 (commented):
The following issues persist.

- If entries from the request & environment variables share the same name, the empty entries from the request variable should fall back to the environment variables.

- Inspector shouldn't warn if the selected environment has a value set while the same is empty under Global - [HFE-432](https://linear.app/hoppscotch/issue/HFE-432/inspector-fails-to-recognize-value-in-environment).

jamesgeorge007 (approved):

mabi:
How can I get/set these variables from a pre-request script?
I tried using `pw.env.get("my_reqest_variable")` but it returns `undefined`
Thank you.

PR #3846, opened 2024-02-20
feat: use tags as folders when importing from openapi

### LINKED ISSUE #3411: [feature]: Group request by tag when importing OpenAPI

amk-dev:
fixes #3411
fixes HFE-388

**Changes**

* When importing from a openapi spec, we'll group the requests to folders based on the tags provided

     * If no tags are provided, they'll be added to the root of the imported collection
    * For each tag, a folder will be created and the requests will be added to the folder
    * If a request has multiple tags, it'll be added to all the folders corresponding to the tags
    
 for testing this PR, you can use openapi spec in this URL https://gist.githubusercontent.com/amk-dev/cc17964b48da497257502b8bf4db7e93/raw/50752cc7377602b80261370c4faf93f87c5d7626/openapi-with-tags-example
 
 it has requests with multiple tags, single tag and no tag

### DIFF
--- a/packages/hoppscotch-common/src/helpers/import-export/import/openapi.ts
+++ b/packages/hoppscotch-common/src/helpers/import-export/import/openapi.ts
@@ -17,6 +17,7 @@ import {
   HoppCollection,
   makeCollection,
   HoppRESTRequestVariable,
+  HoppRESTRequest,
 } from "@hoppscotch/data"
 import { pipe, flow } from "fp-ts/function"
 import * as A from "fp-ts/Array"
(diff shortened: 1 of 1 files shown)


nivedin (approved):
LGTM 💯

PR #3851, opened 2024-02-21
refactor: possible links in email templates do not highlight

AndrewBastin:
NOTE: This PR is made in response to a moderate security advisory

### Goal
If something representable as a site name or something is present in a Team Invitation (which is possible if the team name or the user name contains things which can look like phone numbers, sites or emails), those text should not be converted into links that the user can click through.

### Implementation
We wrap around the variable bits in the template in an anchor tag to prevent the Email clients doing it on their end and style it so no UI weirdness happens.

### Checks

### DIFF
--- a/packages/hoppscotch-backend/src/mailer/mailer.service.ts
+++ b/packages/hoppscotch-backend/src/mailer/mailer.service.ts
@@ -25,7 +25,7 @@ export class MailerService {
   ): string {
     switch (mailDesc.template) {
       case 'team-invitation':
-        return `${mailDesc.variables.invitee} invited you to join ${mailDesc.variables.invite_team_name} in Hoppscotch`;
+        return `A user has invited you to join a team workspace in Hoppscotch`;
 
       case 'user-invitation':
         return 'Sign in to Hoppscotch';


balub (commented):
Screenshot 2024-02-22 at 11 55 47 PM

Is there any way we can change the subject from showing the team name as a link too? 

PS: This was apple mail

balub:
Do we want to do that though ?

AndrewBastin:
Sure, why not, its doable.

balub:
Also, just a note, the subject appearing as a link does not seem to happen in Gmail, but the mouse `cursor` property doesn't seem to work in Gmail.

Screenshot 2024-02-23 at 12 00 46 AM

balub:
Yup, only the CSS thing fails. The behaviour is still as intended.

balub:
Okay, everything looks good to me, tested on the following email clients:

- Gmail
- Gmail mobile
- Apple Mail
- Thunderbird
- Outlook
- Yahoo Mail

The cursor property of the mouse fails in a few clients ( Gmail and Outlook) but the rest is working as intended. Will approve after the subject fix is pushed.

AndrewBastin:
@balub the subject fix is pushed.

balub (approved):

PR #3852, opened 2024-02-22
fix: collection level authorization inheritance issue

### LINKED ISSUE #3713: [bug]: Authorization on Collection level does not work

### Current behavior

When I move the Authorization settings from Request level to the Collection level, I see the following error message when I click `Generate Token`:
auth server error

What I expect is that the token is successfully generated and populated in the corresponding field in the dialog and subsequent Requests in this Collection.

### Steps to reproduce

1. Start with a successful Request that uses OAuth 2.0 Authorization
2. Move the Authorization settings from Request level to Collection level
    - In my case this means entering the `OpenID Connect Discovery URL`, `Authorization URL`, ' Access Token URL` and `Client ID`
    - I do not need to specify `Client Secret` and `Scope`
3. Click on `Generate Token`
4. See error: `Auth Server returned an error state`

Please note that this error does not occur, when you enter any value in the `Scope` field and then remove it before clicking on `Generate Token`.
Therefore I suspect that this field or the corresponding variable is not properly initialized or passed to the URL.
Without this workaround, I can see `scope=undefined` in the resulting request. Using the workaround, this turns into `scope=`, which is what I expected because this is also used in the Request level Authorization.

Unfortunately the workaround does not completely fix Collection level Authorization.
When I click on `Generate Token` after using the workaround, I see a different error:
client secret error

I do not need to specify a `Client Secret`  on the Request level Authorization and the workaround for the `Scope` does not work for the `Client Secret`.

### Environment

Production

### Version

Self-hosted

anwarulislam:
Closes HFE-422 #3713

### Description
This pull request resolves the issue of collection-level authorization inheritance in its requests. Currently, when we create a request in a collection that is supposed to inherit the auth from collection properties, we have to manually choose an auth-type called "inherit" to inherit authorization from the collection. 

Although it works when we select the option manually, it should by default inherit the auth options from the collection and prioritize request auth over collection auth without creating or selecting any auth-type called "inherit". This PR addresses the issue and changes the flow to achieve this.

### Checks

### DIFF
--- a/packages/hoppscotch-common/src/components/graphql/Authorization.vue
+++ b/packages/hoppscotch-common/src/components/graphql/Authorization.vue
@@ -32,17 +32,6 @@
               @keyup.escape="hide()"
             >
               <HoppSmartItem
-                label="None"
-                :icon="authName === 'None' ? IconCircleDot : IconCircle"
-                :active="authName === 'None'"
-                @click="
-                  () => {
-                    auth.authType = 'none'
-                    hide()
-                  }
-                "
-              />
-              <HoppSmartItem
                 v-if="!isRootCollection"
                 label="Inherit"
                 :icon="authName === 'Inherit' ? IconCircleDot : IconCircle"
@@ -50,6 +39,17 @@
                 @click="
                   () => {
                     auth.authType = 'inherit'
+                    hide()
+                  }
+                "
+              />
+              <HoppSmartItem
+                label="None"
+                :icon="authName === 'None' ? IconCircleDot : IconCircle"
+                :active="authName === 'None'"
+                @click="
+                  () => {
+                    auth.authType = 'none'
                     hide()
                   }
                 "
@@ -284,7 +284,7 @@ const authActive = pluckRef(auth, "authActive")
 
 const clearContent = () => {
   auth.value = {
-    authType: "none",
+    authType: "inherit",
     authActive: true,
   }
 }
--- a/packages/hoppscotch-common/src/components/graphql/Authorization.vue
+++ b/packages/hoppscotch-common/src/components/graphql/Authorization.vue
@@ -232,6 +232,7 @@ import { useI18n } from "@composables/i18n"
 import { useColorMode } from "@composables/theming"
 import { useVModel } from "@vueuse/core"
 import { HoppInheritedProperty } from "~/helpers/types/HoppInheritedProperties"
+import { onMounted } from "vue"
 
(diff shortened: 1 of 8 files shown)


packages/hoppscotch-common/src/components/http/Authorization.vue
```diff
@@ -233,15 +232,6 @@ const emit = defineEmits<{
...
 
-onMounted(() => {
-  if (props.isRootCollection && auth.value.authType === "inherit") {
-    auth.value = {
-      authType: "none",
-      authActive: true,
-    }
-  }
-})
-
```
nivedin, line 244: I think we can keep this since a root collection can't inherit auth so we should set the auth to "None" in this case.

nivedin: also for a newly created collection the auth active is disabled by default can you change it to active

packages/hoppscotch-common/src/helpers/rest/default.ts
```diff
@@ -8,7 +8,7 @@ export const getDefaultRESTRequest = (): HoppRESTRequest => ({
   headers: [],
   method: "GET",
   auth: {
-    authType: "none",
```
nivedin, line 11: can you set the default graphql auth to inherit as well

packages/hoppscotch-common/src/components/graphql/Authorization.vue
```diff
@@ -32,24 +32,24 @@
               @keyup.escape="hide()"
             >
               <HoppSmartItem
-                label="None"
-                :icon="authName === 'None' ? IconCircleDot : IconCircle"
-                :active="authName === 'None'"
```
nivedin, line 37: can you add the same root check in graphql as well

```
onMounted(() => {
  if (props.isRootCollection && auth.value.authType === "inherit") {
    auth.value = {
      authType: "none",
      authActive: true,
    }
  }
})
```

nivedin (changes requested):
please look into the given comments

packages/hoppscotch-common/src/components/http/Authorization.vue
anwarulislam: We always keep the authType to `inherit` even if the request isn't in any collection (it will work like none). That we have discussed already with @thetronjohnson

nivedin (approved):
LGTM 🚀

PR #3874, opened 2024-03-01
fix: context menu bug and incorrect position

nivedin:
Closes HFE-403

### Description
This PR fixes the issue where sometimes when scrolling the response section while the context menu is opened will break the syntax highlighting in codemirror.
This PR also fixes a slight change in context menu position due the the z position of the splitpanes changed.

### Checks

### DIFF
--- a/packages/hoppscotch-common/src/composables/codemirror.ts
+++ b/packages/hoppscotch-common/src/composables/codemirror.ts
@@ -242,7 +242,9 @@ export function useCodemirror(
       const { from, to } = selection
       if (from === to) return
       const text = view.value?.state.doc.sliceString(from, to)
-      const { top, left } = view.value?.coordsAtPos(from)
+      const coords = view.value?.coordsAtPos(from)
+      const top = coords?.top ?? 0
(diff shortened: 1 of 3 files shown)


packages/hoppscotch-common/src/composables/codemirror.ts
```diff
@@ -324,7 +327,8 @@ export function useCodemirror(
       EditorView.domEventHandlers({
         scroll(event) {
           if (event.target && options.contextMenuEnabled) {
-            handleTextSelection()
+            // Debounce to make the performance better
+            debouncedTextSelection(30)()
```
jamesgeorge007, line 331: Do we need a similar optimization here?

https://github.com/hoppscotch/hoppscotch/blob/faab1d20fde9a6be660db40fc73dcf28f9038008/packages/hoppscotch-common/src/components/smart/EnvInput.vue#L467-L469

nivedin: yup missed this

jamesgeorge007 (approved):

PR #3881, opened 2024-03-07
fix: secret env bug in firebase due to undefined value

nivedin:
Closes HFE-440

### Description
This PR fixes the issue in secret environment feature where when pushing the environment with `undefined` value failed, resulting in empty secret env's.
Also add id while upgrading env version.

### Checks

### DIFF
--- a/packages/hoppscotch-data/src/environment/index.ts
+++ b/packages/hoppscotch-data/src/environment/index.ts
@@ -5,7 +5,7 @@ import { InferredEntity, createVersionedEntity } from "verzod"
 import { z } from "zod"
 
 import V0_VERSION from "./v/0"
-import V1_VERSION from "./v/1"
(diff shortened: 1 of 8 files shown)


jamesgeorge007 (approved):

packages/hoppscotch-data/src/environment/index.ts
```diff
@@ -165,7 +165,7 @@ export const translateToNewEnvironment = (x: any): Environment => {
   if (x.v && x.v === EnvironmentSchemaVersion) return x
 
   // Legacy
-  const id = x.id ?? ""
+  const id = x.id ?? uniqueID()
```
jamesgeorge007, line 168: We could maybe replace this with `||` to ensure IDs as empty strings fall back to the unique ID variant.

PR #3889, opened 2024-03-11
hotfix: request variable version syncing bug

nivedin:
Closes HFE-443

### Description
This PR fixed the issue with request variable where the app crashes since the request schema failed to upgrade version while syncing returning a undefined value.

### Checks

### DIFF
--- a/packages/hoppscotch-data/src/collection/index.ts
+++ b/packages/hoppscotch-data/src/collection/index.ts
@@ -52,8 +52,6 @@ export function makeCollection(x: Omit<HoppCollection, "v">): HoppCollection {
  * @returns The proper new collection format
  */
 export function translateToNewRESTCollection(x: any): HoppCollection {
-  if (x.v && x.v === CollectionSchemaVersion) return x
-
   // Legacy
   const name = x.name ?? "Untitled"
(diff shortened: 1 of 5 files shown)


jamesgeorge007 (changes requested):
We have to account for the case when a collection conforming to the old version format is imported and a request is chosen from the collection tree straightaway.

packages/hoppscotch-data/src/collection/index.ts
nivedin, line 1: there is case where a collection version is upto date but the request version will be behind in this case the translate will not return the latest schema. So the version check is removed here.
cc @AndrewBastin

jamesgeorge007 (approved):

packages/hoppscotch-data/src/collection/index.ts
AndrewBastin: Actaully these can be updated to use Verzod functions now. Since it can perform validations as well ?

jamesgeorge007: I think we'll have to separate them into REST & GQL implementations for better control and update the transformation functions accordingly.

Issue #3685, opened 2023-12-21
[bug]: Setting body type to none sets the content type in the request to application/x-www-form-urlencoded


flleeppyy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When setting the body type to none, the request headers become "Content-Type: application/x-www-form-urlencoded", when it should be removing "Content-Type" from the request headers.

This behavior **only happens** without an interceptor selected.

### Steps to reproduce

![image](https://github.com/hoppscotch/hoppscotch/assets/18307183/57b2b629-3a63-46bf-ae5a-39495a394aef)
![image](https://github.com/hoppscotch/hoppscotch/assets/18307183/5cf2d6dd-5990-4f02-a1c7-03ce18ac49c5)

### Environment

Production

### Version

Cloud

liyasthomas:
@AndrewBastin can you look into this?

anwarulislam:
The issue regarding the bug where setting the body type to none inadvertently sets the content type in the request to application/x-www-form-urlencoded has been addressed and resolved. This issue is now considered closed and no longer exists.

PR #3906, opened 2024-03-18
[bug]: Setting body type to none sets the content type in the request to application/x-www-form-urlencoded #3685

ialloyd:
This PR addresses an issue where the 'Content-Type' header was incorrectly set to 'application/x-www-form-urlencoded' when the body type was set to 'none'. The changes ensure that the 'Content-Type' header is removed as expected in this scenario.

I would appreciate if you could review the changes and provide your valuable feedback. Thank you!

### DIFF
--- a/packages/hoppscotch-common/src/components/http/Body.vue
+++ b/packages/hoppscotch-common/src/components/http/Body.vue
@@ -185,11 +185,7 @@ const isContentTypeAlreadyExist = () => {
 
 const updateContentType = (contentTypeItem, hide) => {
   if (contentTypeItem === 'none') {
-      body.contentType = null
-      body.contentType = null
-      // And remove the 'Content-Type' header
     body.contentType = null
-      // And remove the 'Content-Type' header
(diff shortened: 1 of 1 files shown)


packages/hoppscotch-common/src/components/http/Body.vue
```diff
@@ -190,4 +190,21 @@ const isContentTypeAlreadyExist = () => {
...
+      // Set the body.contentType to null
+      body.contentType = null
+      // And remove the 'Content-Type' header
+      headers.value = headers.value.filter(header => header.key.toLowerCase() !== 'content-type');
+    } else {
+      // Otherwise, set the body.contentType as usual
+      body.contentType = contentTypeItem
+    }
+    hide()
+  }
```
anwarulislam, line 208: You have included Vue template code inside the script tag. Please revise your pull request.

anwarulislam (changes requested):
You've

anwarulislam:
@ialloyd, I could not reproduce the issue. I see it's working completely fine. Can you please tell me the process to reproduce this issue?

PR #3912, opened 2024-03-19
feat: add support for versioned entities in the CLI

jamesgeorge007:
### Description

The public data structures like `collections`, `requests` & `environments` follow versioning where the new fields being added are introduced as a new schema version, Ref #3457. When one of the above files is imported into the app, it is translated to the latest version so that the same variant of the incoming data can be consumed without ambiguities. This wasn't the case with the CLI, especially with REST requests, previously it would expect the incoming request entry to conform to the latest schema and rejects if not. This PR adds support for versioned entities to the CLI such that the incoming data is translated to the latest version before consumption.

### Changes

- Leverage [verzod](https://github.com/AndrewBastin/verzod) to parse the incoming data and translate them to the latest version, if valid.
- Fix the `v` field parsing for the `HoppCollection` versioned entity. The intended type is `number`.
- Test suite updates ensuring different version ranges of the data structures are supported.
- Remove stale `isHoppRESTCollection` helper and related test suite in favor of `verzod` based parsing.

### Checks


packages/hoppscotch-cli/src/utils/mutators.ts
```diff
@@ -67,13 +70,34 @@ export async function parseCollectionData(
...
+        code: "MALFORMED_COLLECTION",
+        path,
+        data: "Please check the collection data.",
+      });
+    }
+
+    return {
+      ...collection,
+      requests: requestSchemaParsedResult.data,
+    };
```
jamesgeorge007, line 101: Schema validation is performed for each request in the collection against the `HoppRESTRequest` versioned entity to ensure requests conforming to an old version are translated to the latest version. This isn't achievable at the `HoppCollection` validation phase since the translation function only deals with the top-level fields of the collection.

packages/hoppscotch-data/src/collection/index.ts
```diff
@@ -8,8 +8,7 @@ import { translateToNewRequest } from "../rest"
 import { translateToGQLRequest } from "../graphql"
 
 const versionedObject = z.object({
-  // v is a stringified number
-  v: z.string().regex(/^\d+$/).transform(Number),
+  v: z.number(),
```
jamesgeorge007, line 11: https://github.com/hoppscotch/hoppscotch/blob/faab1d20fde9a6be660db40fc73dcf28f9038008/packages/hoppscotch-data/src/collection/v/1.ts#L7

https://github.com/hoppscotch/hoppscotch/blob/faab1d20fde9a6be660db40fc73dcf28f9038008/packages/hoppscotch-data/src/collection/v/2.ts#L9

PR #3929, opened 2024-03-22
feat: added mutation and function to platform for updating user profile name

nivedin:
Closes HFE-455

### Description
This PR adds the mutation and function to the platform to update user profile name from the profile section.

### Checks


amk-dev (changes requested):

anwarulislam (approved):

jamesgeorge007 (approved):

jamesgeorge007 (commented):
A potential enhancement we could have here is ensuring the GQL mutation gets triggered only when there's a change detected in the profile name.

packages/hoppscotch-selfhost-web/src/platform/auth/auth.platform.ts
```diff
@@ -317,9 +318,20 @@ export const def: AuthPlatformDef = {
...
-  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+
   async setDisplayName(name: string) {
-    return
+    if (!name) return E.left("USER_NAME_CANNOT_BE_EMPTY")
+
+    const res = await updateUserDisplayName(name)
+
+    if (E.isRight(res)) {
+      probableUser$.next({
```
amk-dev, line 328: Here we're only updating the value of probableUser. this will lead to currentUser and probableUser having different displayName after the update. we could use `setUser` function which sets both probableUser and currentUser instead of updating probableUser only.

packages/hoppscotch-selfhost-web/src/platform/auth/auth.platform.ts
```diff
@@ -317,9 +318,20 @@ export const def: AuthPlatformDef = {
...
+        ...probableUser$.value,
+        displayName: res.right.updateDisplayName.displayName,
+      } as HoppUser)
```
amk-dev, line 331: If we add a check to see if we've a probable user / current user and returns an error in that case, we can avoid this casting.

amk-dev (approved):

PR #3934, opened 2024-03-25
fix: lint errors removed by using `satisfies` or `as` for type

anwarulislam:
### Description
Using `satisfies` or `as` for type enforcement to skip lint errors.

| Before | After |
|--------|-------|
| image | image

- [ ] Not Completed
- [x] Completed


jamesgeorge007 (approved):

packages/hoppscotch-common/src/components/collections/Properties.vue
```diff
@@ -125,12 +134,12 @@ watch(
     if (props.show) {
       persistenceService.setLocalConfig(
         "unsaved_collection_properties",
-        JSON.stringify(<EditingProperties>{
+        JSON.stringify({
```
amk-dev, line 137: both the angle-bracket syntax and using `as` to assert just tells typescript that we know what we're doing, and dont bother checking it. but in the future if someone passes an invalid object that does not adhear to the type, typescript won't catch that.

ideally in this case, we can extract the object into variable and assign a type for that, and tries to fix any errors that comes up.

amk-dev (commented):
the root cause of the errors with angle-bracket assertion (<>) was because `parserOptions.ecmaFeatures.jsx` was enabled. 

this is the default behaviour in `@vue/typescript/recommended` package. since we don't use jsx, we can disable this option. added this change in 79d3e8a9ac3b4cc2a34572e2d04a6a0971e14943

PR #3937, opened 2024-03-26
chore(sh-admin): alert the user while deleting users who are team owners

jamesgeorge007:
### Description

Attempting to delete users who are owners of at least one team from the Admin dashboard is an action that would fail. There wasn't any feedback previously while attempting to delete users from the list view/a user via the profile view. This PR adds an error toast that will be shown during the scenarios mentioned above.

### Changes

- Show an error toast while attempting to delete users who are team owners from the list `/users` and profile `/user/_id` views.
- Locale entry additions for the case with attempting team owner deletions.
- Ensure the singular/plural form of the error toast message is displayed based on the no. of rows selected from the user management list view (`/users` page).
- Abstract the user deletion business logic in a helper function `handleUserDeletion()` under `helpers/userManagement.ts`.
- New `getI18n()` function exported from the [i18n module](https://github.com/hoppscotch/hoppscotch/pull/3937/files#diff-ed5ce9db1344b536e4856e4957b319d1726fc93762183d6a028dea30023e9882) for use outside components.
- At least one Admin is enforced on an instance. Show a more specific error message while attempting to remove the Admin status that would lead to a scenario with no users with Admin privileges.

Closes HFE-462.

### Checks


packages/hoppscotch-sh-admin/src/pages/users/index.vue
```diff
@@ -572,45 +573,32 @@ const deleteUsers = async (id: string | null) => {
   const result = await usersDeletion.executeMutation(variables);
 
   if (result.error) {
-    const errorMessage =
-      result.error.message === DELETE_USER_FAILED_ONLY_ONE_ADMIN
-        ? t('state.delete_user_failed_only_one_admin')
```
jamesgeorge007, line 577: Admins can't be deleted and it's accounted for by the `ADMIN_CANNOT_BE_DELETED` error message check below. The `DELETE_USER_FAILED_ONLY_ADMIN` = `admin/only_one_admin_account_found` scenario applies to the case while removing the Admin status when there is only one Admin in the instance.

https://github.com/hoppscotch/hoppscotch/blob/0a71783eaaba4eac47977fc1d4da9a9a8f6080dd/packages/hoppscotch-backend/src/errors.ts#L6-L11

https://github.com/hoppscotch/hoppscotch/blob/0a71783eaaba4eac47977fc1d4da9a9a8f6080dd/packages/hoppscotch-backend/src/admin/admin.service.ts#L530-L543

packages/hoppscotch-sh-admin/src/helpers/userManagement.ts
```diff
@@ -0,0 +1,121 @@
...
+  type: 'individual';
+  metadata: null;
+};
+type BulkActionInput = {
+  type: 'bulk';
+  metadata: {
+    areMultipleUsersSelected: boolean;
+    deletedIDs: string[];
+  };
+};
```
jamesgeorge007, line 16: Denotes the `action` arg that the helper function takes in.

- The `type` field denotes whether the action is performed from the `/users` page or via the `/user/_id` view.
- The `metadata` field includes additional information required based on the type. For instance, the `bulk` action type needs to convey if the action was performed on a single row or spanned multiple rows.

packages/hoppscotch-sh-admin/src/helpers/userManagement.ts
```diff
@@ -0,0 +1,121 @@
...
+
+type IndividualActionResult = {
+  data: null;
+};
+type BulkActionResult = {
+  data: { timeoutID: NodeJS.Timeout | null };
+};
```
jamesgeorge007, line 23: The `bulk` action type can return a `timeoutID` indicating the return type from `setTimeout` used to clear from the `/users` page. This is used to display the error messages after a certain delay following a series of toasts where the count for users deleted is displayed.

packages/hoppscotch-sh-admin/src/helpers/userManagement.ts
```diff
@@ -0,0 +1,121 @@
...
+  (
+    deletedUsersList: UserDeletionResult[],
+    action: IndividualActionInput | BulkActionInput
+  ): IndividualActionResult | BulkActionResult;
+};
+
+const t = getI18n();
+const toast = useToast();
+
+export const handleUserDeletion: HandleUserDeletion = (
```
amk-dev, line 35: why do we need to have seperate `bulk` v/s `individual` seperation. we could define a function which takes, `deletedUsersList` and derive the other things from it. 

```typescript
const handleUserDeletionCombined = (
  // isn't this the only thing that is required ? can't all the other things be derived from this?
  deletedUsersList: UserDeletionResult[]
)
```

```typescript
// can't this be used instead of a bulk property?
const hasMultipleUsers = deletedUsersList.length > 1;
```

packages/hoppscotch-sh-admin/src/helpers/userManagement.ts
```diff
@@ -0,0 +1,121 @@
...
+          count: deletedUsersList.length - deletedIDs.length,
+        })
+      );
+    }
+  }
+
+  uniqueErrorMessages.forEach((errorMessage) => {
+    if (errMsgMapKeys.includes(errorMessage)) {
+      if (type === 'bulk') {
+        timeoutID = setTimeout(
```
amk-dev, line 95: what is the goal here ? show toasts one after the other ?
if that is the case, we could do something like below, so we don't have to deal with the timeouts ourself, and also will help in unifiying the logic between bulk/individual.

```typescript
const messages = [
  'message 1',
  'message 2',
  'message 3',
  'message 4',
  'message 5',
];

const displayMultipleMessages = (messages,currentIndex) => {
  const message = messages[currentIndex];

  toast.show(message, {
    duration: 2000,
    onComplete: () => {
      if (currentIndex < messages.length - 1) {
        displayMultipleMessages(messages, currentIndex + 1);
      }
    },
  });
};
```

amk-dev (commented):
do we need to show the success toast, when no user is deleted ? here we're showing a success toast "Number of deleted users 0".

https://github.com/hoppscotch/hoppscotch/assets/57758277/05e2e7fe-d014-4c9a-9b20-7b60c3b93a7e

packages/hoppscotch-sh-admin/src/helpers/userManagement.ts
jamesgeorge007: Yup, that's the goal; made the change in https://github.com/hoppscotch/hoppscotch/pull/3937/commits/63ac2c3cd6b9fb6db9ff36edc73c4d4a0d55cd90.

packages/hoppscotch-sh-admin/src/helpers/userManagement.ts
```diff
@@ -0,0 +1,138 @@
...
+};
+
+const t = getI18n();
+const toast = useToast();
+
+const displayToastMessages = (
+  toastMessages: ToastMessage[],
+  currentIndex: number,
+  isBulkAction: boolean,
+  metadata: ActionMetadata
```
nivedin, line 29: do we require this two parameters

packages/hoppscotch-sh-admin/src/helpers/userManagement.ts
```diff
@@ -0,0 +1,138 @@
...
+    const remainingDeletionsCount = deletedUsersList.length - deletedIDs.length;
+    if (remainingDeletionsCount > 0) {
+      toastMessages.push({
+        message: t('state.delete_some_users_failure', {
+          count: remainingDeletionsCount,
+        }),
+        state: 'error',
+      });
+    }
+  }
```
nivedin, line 109:
```suggestion
 if (isBulkAction) {
    const { areMultipleUsersSelected, deletedIDs } =
      metadata as BulkActionMetadata;

    const deletedCount = deletedIDs.length;

    if (areMultipleUsersSelected && deletedCount > 0) {
      toastMessages.push({
        message: t('state.delete_some_users_success', {
          count: deletedCount,
        }),
        state: 'success',
      });
    }

    const remainingDeletionsCount = deletedUsersList.length - deletedCount;
    if (remainingDeletionsCount > 0) {
      toastMessages.push({
        message: t('state.delete_some_users_failure', {
          count: remainingDeletionsCount,
        }),
        state: 'error',
      });
    }
  }
```

nivedin (changes requested):

packages/hoppscotch-sh-admin/src/helpers/userManagement.ts
jamesgeorge007: Good catch, they were unnecessary; removed in https://github.com/hoppscotch/hoppscotch/pull/3937/commits/dee7864a080ac89afd9fad701c48f5af8618b557.

packages/hoppscotch-sh-admin/src/pages/users/index.vue
```diff
@@ -572,45 +574,25 @@ const deleteUsers = async (id: string | null) => {
...
-      (user) => user.errorMessage === ADMIN_CANNOT_BE_DELETED
-    );
+    handleUserDeletion(deletedUsers, {
+      areMultipleUsersSelected: areMultipleUsersSelected.value,
+      deletedUserIDs,
+    });
 
     usersList.value = usersList.value.filter(
-      (user) => !deletedIDs.includes(user.uid)
+      (user) => !deletedUserIDs.includes(user.uid)
```
jamesgeorge007, line 593: We could refrain from supplying `deletedUserIDs` via the `metadata` field to the helper function since it is computed based on the `deletedUsers` list; thought to send it along since the computation is already happening at the page level.

nivedin (approved):

packages/hoppscotch-sh-admin/src/pages/users/index.vue
amk-dev: Here the computation is very cheap, and it won't affect us in any way for computing this inside the function. and if this function is used somewhere else, that user will have to compute this and provide it mostly. which is not ideal imo.

also not having this will help use remove the metadata entirely, and just make the function + types super simple.

packages/hoppscotch-sh-admin/src/modules/i18n.ts
```diff
@@ -1,7 +1,23 @@
...
+  false
+> | null = null;
+
+/**
+ * Returns the i18n instance
+ */
+export function getI18n() {
+  return i18nInstance!.global.t;
+}
+
```
amk-dev, line 20: In the `userManagement.ts` file, we are using `getI18n` with out checking if it is null, which can be a possibility, if this function is used before the vue app gets ready.

we could take an additional parameter `i18nFunc` for the `handleUserDeletion` which can be passed from the component using it. so we could remove this function.

although this function will allow to use the i18n from .ts files directly without interactions from a component, whoever using it has to handle the null case appropriately.

AndrewBastin (approved):

packages/hoppscotch-sh-admin/src/helpers/userManagement.ts
```diff
@@ -38,21 +27,22 @@ const displayToastMessages = (
...
+
   // Show the success toast based on the action type if there are no errors
   if (uniqueErrorMessages.size === 0) {
     if (isBulkAction) {
       toast.success(
-        (metadata as BulkActionMetadata).areMultipleUsersSelected
+        isBulkAction
           ? t('state.delete_user_success')
           : t('state.delete_users_success')
       );
```
jamesgeorge007: This was already within the `isBulkAction` block; updated in https://github.com/hoppscotch/hoppscotch/commit/787aab650ff584f9d9cd86aea8935c74b72bc632.

packages/hoppscotch-sh-admin/src/modules/i18n.ts
jamesgeorge007: The `i18n` module initialization happens at the package entry point. Hence, `i18Instance` is guaranteed to have the expected value.

https://github.com/hoppscotch/hoppscotch/blob/787aab650ff584f9d9cd86aea8935c74b72bc632/packages/hoppscotch-sh-admin/src/main.ts#L62

We follow a similar approach on the app side as well.

https://github.com/hoppscotch/hoppscotch/blob/787aab650ff584f9d9cd86aea8935c74b72bc632/packages/hoppscotch-common/src/modules/i18n.ts#L127-L132

PR #3939, opened 2024-03-27
fix(desktop): set window caption color if Windows version >= 11

### LINKED ISSUE #3524: [bug]: Desktop immediately closing after start on Windows 10

### Current behavior

Trying to start the Hoppscotch.exe, either directly or via the shortcuts created during installation does not work. A blank window opens for ~1 second and then closes again. 
Task manager shows a Hoppscotch task which also exits again immediately.

What I expect is that the window stays open so that I can interact with it and use the program.

OS: Microsoft Windows 10 Enterprise
Version: 10.0.19045 Build 19045

### Steps to reproduce

1. Open Hoppscotch.exe
2. A blank window with the title `Hoppscotch` and the Hoppscotch logo opens
3. The window closes immediately

### Environment

Production

### Version

Local

Timic3:
Closes #3524 

### Description

This pull request fixes Windows 10 crash when application is opened. It happened due to `DWMWA_CAPTION_COLOR` being available only on Windows 11 (build number 22000) and up (see [here](https://learn.microsoft.com/en-us/windows/win32/api/dwmapi/ne-dwmapi-dwmwindowattribute)). This is fixed by checking Windows version number and build, however it is also fixable by just catching the error, but it's probably best not to fill Windows event log with errors.

### Checks

### Additional Information

I haven't tested this on Windows 11 yet, however feel free to test it.


AndrewBastin:
@Timic3 is there any reason you are not using [winver](https://github.com/rhysd/winver) ?

Timic3:
@AndrewBastin No reason, I just thought adding a crate for this would be a bit overkill, but I can update the PR to use winver. It would indeed be cleaner (and safer) than hooking kernel functions directly.

Timic3:
@AndrewBastin I've changed the code to check version with `winver` crate, it is much cleaner than before. I've seen that it also provides a few fallbacks if for some reason we cannot access kernel functions.

I couldn't run tests because I was trying this on Windows, but tests should pass anyway.

AndrewBastin (approved):
lgtm 💯

PR #3948, opened 2024-03-29
feat: console error messages and better names for requests imported from openapi

sawa-ko:
Closes # 

### Description

This PR shows an error message in console when an error occurs when parsing the OpenAPi file, since actually there is no error message, just a generic message, and it is impossible to know why the import failed (I spent 30 minutes looking for what was wrong with my file lol). 

Also, in the requests, we currently try to obtain the name with the following steps:
1. From its operationId property
2. From its summary property
3. Default generic name

Currently, most of the request names always arrive in this 3rd step because usually the schemas that I have found do not define the summary property or things like that, so in this PR we try to extract the name from the URL of the request, so that the name is more understandable and not have folder with requests named as untitled request. Attached is evidence of how it was before and how it is now.

> Before
image

> Now
image

### Checks


JoelJacobStephen (changes requested):
The changes work as expected. However, there are cases where the url can contain variables (eg. `/pet/{petID}`, in this case, the name given will be `<<petID>>` if there are no operationID or summary properties). I am not sure that's a good approach to the naming scheme in this case. @amk-dev Can I get your thoughts on this?

packages/hoppscotch-common/src/helpers/import-export/import/openapi.ts
```diff
@@ -627,14 +627,27 @@ const convertPathToHoppReqs = (
           ? openAPIUrl + openAPIPath.slice(1)
           : openAPIUrl + openAPIPath
 
+      let defaultRequestName = "Untitled Request"
+
+      if (info.tags?.length) {
+        const pathStartIndex =
```
amk-dev, line 633: here this logic will miss some edge cases,

i belive you're assuming that the tag will always be present in the path, but that won't always be the case.

1. what happens if the tag is no way related to the path ?
2. adding the tag length makes sense only if the tag is found in the path, otherwise you're adding the length to -1, which does not makes much sense.

sawa-ko: No. If it does not have a tag then the condition is not met and is saved as a request without a title. What I can do is save it with the name of the controller then the request in case I don't have tags.

packages/hoppscotch-common/src/helpers/import-export/import/openapi.ts
```diff
@@ -627,14 +627,27 @@ const convertPathToHoppReqs = (
...
+          openAPIPath.indexOf(info.tags[0]) + info.tags[0].length
+
+        if (pathStartIndex > 0) {
+          defaultRequestName =
```
amk-dev, line 637: when naming the request, here it seems we're considering the first value after the tag only, eg: the current logic will name `/categories/<<categoryID>/update` as `<<categoryID>>`, but i would prefer the request to be named, whatever the remaining path is after the tag. in this case `<<categoryID>>/update`

packages/hoppscotch-common/src/helpers/import-export/import/openapi.ts
amk-dev: I quoted the wrong lines actually. the comment is meant for the next line. 
```typescript
const pathStartIndex =
          openAPIPath.indexOf(info.tags[0]) + info.tags[0].length
```

and i didn't mean when the tag is not present, i meant when a tag is present but it is not in the url. 

like this,

```
/categories/{categoryId}/items:
    get:
      tags:
        - SomethingNonRelatedToTheURL
```

PR #3966, opened 2024-04-08
fix: use proper values for `addTo` field when auth type is `api-key`

### LINKED ISSUE #3963: [bug]: Latest update (2024.3.0) broke Api Key authentication passed as query params

### Current behavior

When I use Authorization of type `API key` with `Pass by` method of `Query params` Authorization Key and Value fields are discarded and not sent to the server.
Expected behavior: Authorization Key and Value field should be sent as query params
[feature]: Api Key based authentication #2000
[implemented as]: pull request #2021 feat: (authentication) Api key based authentication [#2000] 
latest version with good behavior: 2023.12.6

### Steps to reproduce

1. open https://hoppscotch.io/
2. create new GET request to the https://echo.hoppscotch.io
3. add Authorization of type `API key`
4. Enter `APIKey` in the Key field
5. Enter `123456789` in the Value field
6. Choose `Query params` for the `Pass by`
7. Click `Send`

Expected HTTP request
```
GET /?APIKey=123456789 HTTP/1.1
Host: echo.hoppscotch.io
```
Factual HTTP request in 2024.3.0 version
```
GET / HTTP/1.1
Host: echo.hoppscotch.io
```

### Environment

Production

### Version

Cloud

amk-dev:
FIXES HFE-466
Closes #3963

**Changes**

Previously when the auth type was `api-key`, the `addTo` field values where "Headers" or "Query params", and the type was a string.

This PR updates the `addTo` field to be an enum with the values `HEADERS` and `QUERY_PARAMS`. this is consistent with the OAuth `addTo` values, and also allows the type system to catch bugs that were previously not caught because any string value was accepted as a valid `addTo` value.

The issue [#3963](https://github.com/hoppscotch/hoppscotch/issues/3963) is coming from this inconsistency, and this PR fixes it.

We could have kept the "Headers" and "Query params" and updated some checks. but i think this is a good time to make the values consistent and better typed, hence this PR.

### DIFF
--- a/packages/hoppscotch-common/src/components/http/authorization/ApiKey.vue
+++ b/packages/hoppscotch-common/src/components/http/authorization/ApiKey.vue
@@ -40,23 +40,23 @@
             @keyup.escape="hide()"
           >
             <HoppSmartItem
-              :icon="auth.addTo === 'Headers' ? IconCircleDot : IconCircle"
-              :active="auth.addTo === 'Headers'"
+              :icon="auth.addTo === 'HEADERS' ? IconCircleDot : IconCircle"
+              :active="auth.addTo === 'HEADERS'"
               :label="'Headers'"
               @click="
                 () => {
-                  auth.addTo = 'Headers'
+                  auth.addTo = 'HEADERS'
                   hide()
                 }
               "
             />
             <HoppSmartItem
-              :icon="auth.addTo === 'Query params' ? IconCircleDot : IconCircle"
-              :active="auth.addTo === 'Query params'"
+              :icon="auth.addTo === 'QUERY_PARAMS' ? IconCircleDot : IconCircle"
+              :active="auth.addTo === 'QUERY_PARAMS'"
               :label="'Query params'"
               @click="
                 () => {
-                  auth.addTo = 'Query params'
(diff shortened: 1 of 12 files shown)


packages/hoppscotch-common/src/components/http/authorization/ApiKey.vue
```diff
@@ -28,7 +28,13 @@
       >
         <HoppSmartSelectWrapper>
           <HoppButtonSecondary
-            :label="auth.addTo || t('state.none')"
+            :label="
+              auth.addTo
+                ? auth.addTo === 'HEADERS'
+                  ? t('authorization.pass_by_headers_label')
```
JoelJacobStephen, line 34: The i18n strings used are not defined. 

image

JoelJacobStephen (changes requested):
Feature is working as expected. Minor i18n issues to be tackled. Rest looks good to me.

packages/hoppscotch-common/src/components/http/authorization/ApiKey.vue
amk-dev: i missed to checkin that. my bad. thank you for pointing out.

PR #3971, opened 2024-04-12
refactor(sh-admin): improved handling of server configurations in admin dashboard

JoelJacobStephen:
### Ticket
Closes HFE-479

### Description
This PR aims to refactor the handling of server configurations in the admin dashboard.

### Checks


nivedin (changes requested):
update the Config type in 

https://github.com/joeljacobstephen/hoppscotch/blob/0fa3715ff806034faa52033855d11c4c1dd6f5fa/packages/hoppscotch-sh-admin/src/components/settings/Configurations.vue#L12-L13

https://github.com/joeljacobstephen/hoppscotch/blob/0fa3715ff806034faa52033855d11c4c1dd6f5fa/packages/hoppscotch-sh-admin/src/components/settings/ServerRestart.vue#L20-L21

packages/hoppscotch-sh-admin/src/composables/useConfigHandler.ts
```diff
@@ -398,7 +272,7 @@ export function useConfigHandler(updatedConfigs?: Config) {
       {
         infraConfigs: updatedInfraConfigs.value as InfraConfigArgs[],
       },
-      'configs.mail_configs.update_failure'
```
nivedin, line 401: the i18n string has not been added in the en.json

nivedin (changes requested):
just this i18n issue, rest LGTM

packages/hoppscotch-sh-admin/src/composables/useConfigHandler.ts
```diff
@@ -398,7 +272,7 @@ export function useConfigHandler(updatedConfigs?: Config) {
...
+      'configs.update_failure'
```
nivedin, line 275: the i18n string has not been added in the en.json

nivedin (approved):

jamesgeorge007 (changes requested):

packages/hoppscotch-sh-admin/src/composables/useConfigHandler.ts
```diff
@@ -191,127 +128,66 @@ export function useConfigHandler(updatedConfigs?: Config) {
...
+      GITHUB_CONFIGS,
+      updatedConfigs.providers.github.enabled,
+      updatedConfigs.providers.github.fields
+    );
+
+    pushOrFilterConfigs(
+      MAIL_CONFIGS,
+      updatedConfigs.mailConfigs.enabled,
+      updatedConfigs.mailConfigs.fields
+    );
```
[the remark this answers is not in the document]
JoelJacobStephen: @jamesgeorge007 This logic won't hold true when porting it over to the SHE instance as the complexity increases there (please checkout the SHE version of the refactor to get a better idea). However, I have worked on a improved method of handling this, using some of the suggestions you have made and I have tested it on both SHC and SHE instances.

packages/hoppscotch-sh-admin/src/composables/useConfigHandler.ts
```diff
@@ -212,8 +149,57 @@ export function useConfigHandler(updatedConfigs?: ServerConfigs) {
...
+    let newConfigs: UpdatedConfigs[] = [];
+
+    workingConfigs.forEach(({ config, enabled, field }) => {
+      config.forEach(({ name, key }) => {
+        if (enabled && field) {
+          const value = typeof field === 'string' ? field : String(field[key]);
+          newConfigs.push({ name, value });
+        } else {
+          newConfigs = newConfigs.filter((item) => item.name !== name);
+        }
```
jamesgeorge007, line 190: Could you elaborate on the need for this filtering logic (`else` block)? Since this comes into play only while the user opts to save the modified configuration and `newConfigs` is maintained locally, conditionally adding to the list (`if` block alone) would do, right?

packages/hoppscotch-sh-admin/src/composables/useConfigHandler.ts
```diff
@@ -212,8 +149,57 @@ export function useConfigHandler(updatedConfigs?: ServerConfigs) {
...
+      });
+    });
+    newConfigs = newConfigs.filter((item) => item.name !== '');
```
jamesgeorge007, line 193: Aren't entries corresponding to `name` from the config list [maintained statically](https://github.com/hoppscotch/hoppscotch/pull/3971/files/1f8d3e1afbde2dc243aa9811e88fe75c2d172675..957a3ee848e48fd82456eaa3abb97ede7496902b#diff-5403ef86c6b0ecea0765b06a22c80b6620f8d0ddcdd93ef814eb0b0fefdb1b22)? Is there a case where there can be an empty entry, as from the above?

packages/hoppscotch-sh-admin/src/composables/useConfigHandler.ts
```diff
@@ -212,8 +149,57 @@ export function useConfigHandler(updatedConfigs?: ServerConfigs) {
...
+    return newConfigs;
+  };
+
   // Transforming the working configs back into the format required by the mutations
-  const updatedAllowedAuthProviders = computed(() => [
+  const updatedInfraConfigs = () =>
+    updatedConfigs ? transformInfraConfigs(workingUpdatedConfigs) : [];
```
jamesgeorge007, line 199: Is it necessary to expose the above function as a return value for the `useConfigHandler` composable? If not, it can be inlined within the `updateInfraCongifs()` function. We need not expose info that isn't consumed elsewhere in the current use case.

packages/hoppscotch-sh-admin/src/composables/useConfigHandler.ts
```diff
@@ -212,8 +149,57 @@ export function useConfigHandler(updatedConfigs?: ServerConfigs) {
...
+
+  // Updated allowed auth providers
+  const updatedAllowedAuthProviders = [
```
jamesgeorge007, line 202: Similar to the above. We could inline this within the `updateAuthProvider` function.

packages/hoppscotch-sh-admin/src/helpers/configs.ts
```diff
@@ -55,12 +55,23 @@ export type UpdatedConfigs = {
   value: string;
 };
 
+export type ConfigTransform = {
+  config: Config[];
+  enabled?: boolean;
+  field?: Record<string, string | boolean> | string;
```
jamesgeorge007, line 61: We could rename this to `fields` since it is essentially the list of fields applicable to the provider.

packages/hoppscotch-sh-admin/src/composables/useConfigHandler.ts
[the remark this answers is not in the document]
jamesgeorge007: Makes sense. Let's keep both implementations less divergent.

packages/hoppscotch-sh-admin/src/composables/useConfigHandler.ts
```diff
@@ -149,74 +152,45 @@ export function useConfigHandler(updatedConfigs?: ServerConfigs) {
...
+      },
+      {
+        config: MAIL_CONFIGS,
+        enabled: updatedConfigs?.mailConfigs.enabled,
+        fields: updatedConfigs?.mailConfigs.fields,
+      },
+    ];
 
-  // Push or filter the configs based on the enabled condition
-  const transformInfraConfigs = (workingConfigs: ConfigTransform[]) => {
```
jamesgeorge007: Supplying this as an arg is unnecessary, hence scoped locally within the function definition.

packages/hoppscotch-sh-admin/src/composables/useConfigHandler.ts
```diff
@@ -149,74 +152,45 @@ export function useConfigHandler(updatedConfigs?: ServerConfigs) {
...
-    {
-      provider: 'GITHUB',
-      status: updatedConfigs?.providers.github.enabled ? 'ENABLE' : 'DISABLE',
-    },
-    {
-      provider: 'EMAIL',
-      status: updatedConfigs?.mailConfigs.enabled ? 'ENABLE' : 'DISABLE',
-    },
-  ];
+    return transformedConfigs;
```
jamesgeorge007, line 192: We could rewrite this in a `reduce` based approach as below. Please feel free to push if you'd like it.

```ts
    return updatedWorkingConfigs.reduce(
      (transformedConfigs: UpdatedConfigs[], { config, enabled, fields }) => {
        if (enabled && fields) {
          config.forEach(({ name, key }) => {
            const value =
              typeof fields === 'string' ? fields : String(fields[key]);
            transformedConfigs.push({ name, value });
          });
        }
        return transformedConfigs;
      },
      []
    );
```

jamesgeorge007 (approved):
I've pushed a few changes in https://github.com/hoppscotch/hoppscotch/pull/3971/commits/f34109d58d8b9cf4ae515574bc315afd7908a776, mostly about leveraging the generated enums while referring to the field names and inlining variable references within function definitions that aren't consumed elsewhere. Please give it a look.

packages/hoppscotch-sh-admin/src/composables/useConfigHandler.ts
```diff
@@ -337,28 +153,44 @@ export function useConfigHandler(updatedConfigs?: Config) {
...
+      config.forEach(({ name, key }) => {
+        if (enabled && fields) {
+          const value =
+            typeof fields === 'string' ? fields : String(fields[key]);
+          transformedConfigs.push({ name, value });
+        }
+      });
+    });
+
+    return transformedConfigs;
```
jamesgeorge007, line 192: We could rewrite this in a `reduce` based approach as below. Please feel free to push if you'd like it.

```ts
    return updatedWorkingConfigs.reduce(
      (transformedConfigs: UpdatedConfigs[], { config, enabled, fields }) => {
        if (enabled && fields) {
          config.forEach(({ name, key }) => {
            const value =
              typeof fields === 'string' ? fields : String(fields[key]);
            transformedConfigs.push({ name, value });
          });
        }
        return transformedConfigs;
      },
      []
    );
```

PR #3973, opened 2024-04-12
chore: migrate `Node.js` implementation for `js-sandbox` to `isolated-vm`

jamesgeorge007:
### Description

This PR aims at migrating the sandbox implementation for `Node.js` from [node:vm](https://nodejs.org/api/vm.html#vm-executing-javascript) to [isolated-vm](https://github.com/laverdet/isolated-vm).

Closes HFE-475.

### Changes

- Add `isolated-vm` as a peer dependency (there were issues with not being able to require `isolated-vm` when it was made a direct dependency) to `js-sandbox` and mark it optional since it's specific to the `Node.js` implementation. Also, it is added as a direct dependency on the CLI.
- Migrate the testing framework from `Jest` (support for ESM is experimental) to `Vitest` for `js-sandbox`.
- Directory hierarchy updates for `js-sandbox`.
  - Arrange the source directories based on the platform (`node`, `web`) instead of the type of scripts (`pre-request` & `test-runner`) easily accommodating platform-specific utility functions. `~/src/utils.ts` at the root compiling utility functions reused across platform implementations is renamed to `~/src/shared-utils` to differentiate between platform-specific utils `~/src/node/utils.ts`.
  - Remove the nested `testing` directory from the test suite and enforce a naming convention for the directories based on the scripting API namespace objects (`env`, `expect`, etc).
- The value supplied to `pw.expect(expectVal)` is stringified with an additional property `isStringifiedWithinIsolate` if it's a non-primitive while transferring to the isolate context as required by `isolated-vm`. The value is parsed at the `pw.expect` method definition after checking for the above additional property to differentiate between stringification done while transferring to the isolate context and an original value. The newly introduced property is removed before consumption.
- Spawn a new process if the `Node.js` version is `20` and above from the CLI supplying the `--no-node-snapshot` flag to `node` as required by `isolated-vm`, [Ref](https://github.com/laverdet/isolated-vm?tab=readme-ov-file#requirements) and later proceed with the argument parsing phase.
- Account for requests nested under `folders` within collections while validating against the versioned entity schema and translating to the latest version. Follow up of #3912.
- Fix flaky tests in the CLI (update tests to refer to the [echo.hoppscotch.io](https://echo.hoppscotch.io/) endpoint and prevent test failures if the auth endpoint resulted in a non `200` status code).
- GH Action config updates
  - Bump dependent actions.
  - Fix a typo while referring to the Node.js version and ensure run tests on LTS.
  - Move the `Setup node` step above `Setup pnpm`.

### Checks


packages/hoppscotch-js-sandbox/src/node/test-runner.ts
```diff
@@ -0,0 +1,214 @@
...
+              })
+            )
+
+            // Matcher methods that can be chained with "pw.expect()"
+            // pw.expect().toBe(), etc
+            if (expectFnResult.typeof === "object") {
+              // Access the getter that points to the negated matcher methods via "{ accessors: true }"
+              const matcherMethods = {
+                not: expectFnResult.getSync("not", { accessors: true }),
+              }
```
jamesgeorge007, line 136: The negated matcher method definitions are accessed via [getSync](https://github.com/laverdet/isolated-vm?tab=readme-ov-file#referencegetsyncproperty-options) by specifying `accessors: true`.

https://github.com/hoppscotch/hoppscotch/blob/787aab650ff584f9d9cd86aea8935c74b72bc632/packages/hoppscotch-js-sandbox/src/utils.ts#L414-L418

packages/hoppscotch-js-sandbox/src/utils.ts
```diff
@@ -182,6 +182,36 @@ const getSharedMethods = (envs: TestResult["envs"]) => {
...
+
+      delete parsedExpectVal.isStringifiedWithinIsolate
+      return parsedExpectVal
+    }
+
+    return expectVal
+  } catch (_) {
+    return expectVal
+  }
+}
```
jamesgeorge007, line 213: The value supplied to `pw.expect(expectVal)` is stringified with an additional property `isStringifiedWithinIsolate` if it's a non-primitive while transferring to the isolate context as required by `isolated-vm`. The value is parsed at the `pw.expect` method definition after checking for the existence of the above additional property to differentiate between stringification done while transferring to the isolate context and an original value. The newly introduced property is removed before consumption.

AndrewBastin (approved):
lgtm 💯

PR #3976, opened 2024-04-14
Added new button to save request modal

shipko:
### Description
I noticed that you can't cancel modal window when saving a request. 
Снимок экрана 2024-04-14 в 08 53 29

- [ ] Not Completed
- [x] Completed

I suggest adding a third "cancel" button, which will close the window without performing an action.
Снимок экрана 2024-04-14 в 09 08 40

### Checks

### DIFF
--- a/packages/hoppscotch-common/src/pages/index.vue
+++ b/packages/hoppscotch-common/src/pages/index.vue
@@ -65,11 +65,11 @@
       @hide-modal="showRenamingReqNameModal = false"
     />
     <HoppSmartConfirmModal
-      :show="confirmingCloseForTabID !== null"
+      :show="confirmingCloseAllTabs"
       :confirm="t('modal.close_unsaved_tab')"
-      :title="t('confirm.save_unsaved_tab')"
-      @hide-modal="onCloseConfirmSaveTab"
-      @resolve="onResolveConfirmSaveTab"
+      :title="t('confirm.close_unsaved_tabs', { count: unsavedTabsCount })"
+      @hide-modal="confirmingCloseAllTabs = false"
+      @resolve="onResolveConfirmCloseAllTabs"
     />
     <HoppSmartModal
       v-if="confirmingCloseForTabID !== null"
@@ -85,12 +85,12 @@
         <span class="flex space-x-2">
           <HoppButtonPrimary
             v-focus
-            :label="t?.('action.save')"
+            :label="t?.('action.yes')"
             outline
             @click="onResolveConfirmSaveTab"
           />
           <HoppButtonSecondary
-            :label="t?.('action.dont_save')"
+            :label="t?.('action.no')"
             filled
             outline
             @click="onCloseConfirmSaveTab"
--- a/packages/hoppscotch-common/src/pages/index.vue
+++ b/packages/hoppscotch-common/src/pages/index.vue
@@ -73,6 +73,9 @@
     />
     <HoppSmartModal
       v-if="confirmingCloseForTabID !== null"
+      dialog
+      role="dialog"
+      aria-modal="true"
       :title="t('modal.close_unsaved_tab')"
       @close="confirmingCloseForTabID = null"
     >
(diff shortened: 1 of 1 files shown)


packages/hoppscotch-common/src/pages/index.vue
```diff
@@ -71,13 +71,41 @@
...
+            @click="onResolveConfirmSaveTab"
+          />
+          <HoppButtonSecondary
+            :label="t?.('action.dont_save')"
+            filled
+            outline
+            @click="onCloseConfirmSaveTab"
+          />
+        </span>
+        <span class="flex space-x-2">
```
JoelJacobStephen, line 99: An extra cancel button is not required. Let the cross icon on the top right of the modal handle the cancellation case

packages/hoppscotch-common/src/pages/index.vue
```diff
@@ -71,13 +71,41 @@
...
+      <template #body>
+        <div class="text-center">
+          {{ t("confirm.save_unsaved_tab") }}
+        </div>
+      </template>
+      <template #footer>
+        <span class="flex space-x-2">
+          <HoppButtonPrimary
+            v-focus
+            :label="t?.('action.save')"
```
JoelJacobStephen, line 88: We don't need to label it as Save or Don't Save. It is implied in the description. So please change the text back to yes for save and no for don't save.

packages/hoppscotch-common/src/pages/index.vue
```diff
@@ -71,13 +71,41 @@
       @hide-modal="onCloseConfirmSaveTab"
       @resolve="onResolveConfirmSaveTab"
     />
-    <HoppSmartConfirmModal
```
JoelJacobStephen, line 74: You should not remove this confirmation modal as this is related to closing all tabs. This will break functionality You should replace the previous confirmation modal (starting at line 67) with the new `HoppSmartModal` implementation

JoelJacobStephen (changes requested):
@shipko  After having an internal discussion regarding this PR, we have decided to go the following approach:

1. Yes Button - Opens the save dialog modal and closes the tab
2. No Button - Doesn't save and closes the tab
3. Cancel Icon - (Present at the top right of the modal) - Neither saves nor closes the tab

Based on the above approach, I have suggested few changes to your implementation. Please go through the review comments and then we can go ahead with this PR.

packages/hoppscotch-common/src/pages/index.vue
shipko: It was my mistake. I fixed it

packages/hoppscotch-common/src/pages/index.vue
JoelJacobStephen: @shipko The cross button inherently means to cancel the action. The same design and implementation is followed for most modals in the app. So we will follow the same. Let's remove the cancel button as it's redundant.

shipko: @JoelJacobStephen done!

JoelJacobStephen (dismissed):
LGTM 💯

packages/hoppscotch-common/src/pages/index.vue
```diff
@@ -64,20 +64,40 @@
...
       :show="confirmingCloseAllTabs"
       :confirm="t('modal.close_unsaved_tab')"
       :title="t('confirm.close_unsaved_tabs', { count: unsavedTabsCount })"
       @hide-modal="confirmingCloseAllTabs = false"
       @resolve="onResolveConfirmCloseAllTabs"
     />
+    <HoppSmartModal
+      v-if="confirmingCloseForTabID !== null"
+      :title="t('modal.close_unsaved_tab')"
+      @close="confirmingCloseForTabID = null"
```
nivedin, line 77:
```suggestion
    <HoppSmartModal
      v-if="confirmingCloseForTabID !== null"
      dialog
      role="dialog"
      aria-modal="true"
      :title="t('modal.close_unsaved_tab')"
      @close="confirmingCloseForTabID = null"
```

can you add the aria labels as well, similar to the `HoppSmartConfirmModal.vue`

nivedin (approved):
Just this minor addition, LGTM 💯

PR #3978, opened 2024-04-15
fix: use pressed key rather than its code

### LINKED ISSUE #3833: [bug]: keyboard shortcuts not respecting different keyboard layout, e.g. dvorak

### Current behavior

I'm using dvorak keyboard layout, so the `.` key on my keyboard is actually `v`.  
When I press CTRL and this key, hopscotch says `copied to clipboard`, while I intended to paste.  
Hopscotch obviously thinks I pressed `CTRL+.`, but that's not the case. 

### Steps to reproduce

1. Switch your keyboard layout to dvorak (I'm using windows)  
2. Open hopscotch 
3. Press CTRL and the key with `.` printed on it.  (so to an ousider, to someone who's watching you behind your shoulder, you are pressing `CTRL+.`, but what you are really pressing is `CTRL+v`)  
4. Hopscotch shows `copied to clipboard` 

### Environment

Production

### Version

Local

posva:
Fix #3833

Closes # 

### Checks


posva:
BTW, I think the code can be simplified but I don't know if there is any issue with returning keys that aren't valid `key` characters:

```ts
function getPressedKey(ev: KeyboardEvent): Key {
  const key = ev.key.toLowerCase()

  if (key.startsWith("arrow")) {
    return key.slice(5) as Key
  }

  return key as Key
}
```

packages/hoppscotch-common/src/helpers/keybindings.ts
```diff
@@ -132,30 +132,26 @@ function generateKeybindingString(ev: KeyboardEvent): ShortcutKey | null {
...
   // Check arrow keys
   if (val === "arrowup") return "up"
   else if (val === "arrowdown") return "down"
   else if (val === "arrowleft") return "left"
   else if (val === "arrowright") return "right"
 
+  const key = ev.key.toLowerCase()
+
   // Check letter keys
   const isLetter = val.startsWith("key")
```
JoelJacobStephen, line 146: Since we are using `ev.key` instead of `ev.code`, `val` string won't have key or digit in the beginning of the string ( `val` returns `a` instead of `key a`, and in case of a digit, it returns `0` instead of `digit 0`). The check right now will always return false, which results in few shortcuts not being invoked.

packages/hoppscotch-common/src/helpers/keybindings.ts
```diff
@@ -132,30 +132,26 @@ function generateKeybindingString(ev: KeyboardEvent): ShortcutKey | null {
...
-  if (isLetter) return val.substring(3) as Key
+  if (isLetter) return key as Key
 
   // Check if number keys
   const isDigit = val.startsWith("digit")
```
JoelJacobStephen, line 150: > Since we are using `ev.key` instead of `ev.code`, `val` string won't have key or digit in the beginning of the string ( `val` returns `a` instead of `key a`, and in case of a digit, it returns `0` instead of `digit 0`). The check right now will always return false, which results in few shortcuts not being invoked.

We can  rather do a simple digit check here.

JoelJacobStephen (changes requested):
@posva Please look into these comments.

packages/hoppscotch-common/src/helpers/keybindings.ts
```diff
@@ -132,30 +132,26 @@ function generateKeybindingString(ev: KeyboardEvent): ShortcutKey | null {
 
 function getPressedKey(ev: KeyboardEvent): Key | null {
   // Sometimes the property code is not available on the KeyboardEvent object
-  const val = (ev.code ?? "").toLowerCase()
+  const val = (ev.key ?? "").toLowerCase()
```
posva, line 135:
```suggestion
  const val = (ev.code ?? "").toLowerCase()
```

packages/hoppscotch-common/src/helpers/keybindings.ts
posva: yeah, I didn't mean to change the value of `val`, it was left from the refactoring I showed in the comment in the PR.

I reverted that so now it works

packages/hoppscotch-common/src/helpers/keybindings.ts
```diff
@@ -132,30 +132,26 @@ function generateKeybindingString(ev: KeyboardEvent): ShortcutKey | null {
...
 
   // Check arrow keys
   if (val === "arrowup") return "up"
   else if (val === "arrowdown") return "down"
   else if (val === "arrowleft") return "left"
   else if (val === "arrowright") return "right"
 
+  const key = ev.key.toLowerCase()
```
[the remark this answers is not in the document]
posva: fixed

packages/hoppscotch-common/src/helpers/keybindings.ts
posva: Yes, and it could be even simpler if you want: https://github.com/hoppscotch/hoppscotch/pull/3978#issuecomment-2056371408

posva: BTW feel free to push any changes to the PR

JoelJacobStephen:
@posva I have updated the logic for the checks to work with `ev.key` 

> [#3627832](https://github.com/hoppscotch/hoppscotch/pull/3978/commits/36278320d3c00cda255ddacafba3ca332905062d). 

I have also incorporated your suggestions. Can you please verify from your end if the latest changes work as intended and resolves the [issue](https://github.com/hoppscotch/hoppscotch/issues/3833)?

posva:
I tried building locally but it gives some errors with prisma so I gave up 😅 
Other than that, the code looks good! As long as `key` is used, it should work 👍

JoelJacobStephen (approved):

AndrewBastin:
> I tried building locally but it gives some errors (even on main branch) with prisma so I gave up 😅
@posva can you confirm if you are using Windows ? We have a known annoying little type checking issue with Prisma over there.

Merging into release.

posva:
It was on osx

PR #4006, opened 2024-04-24
feat: hover placeholder transition for smartenvinput

nivedin:
Closes HFE-482

### Description
This PR adds placeholder hover transition effect on request URL input on both REST and GraphQl section

https://github.com/hoppscotch/hoppscotch/assets/53208152/67b9101d-9e00-4eca-a4e9-eaf699630710

### Checks


packages/hoppscotch-common/locales/en.json
```diff
@@ -568,7 +568,9 @@
     "generated_code": "Generated code",
     "go_to_authorization_tab": "Go to Authorization tab",
     "go_to_body_tab": "Go to Body tab",
+    "graphql_placeholder": "Enter a URL",
     "header_list": "Header List",
+    "http_placeholder":"Enter a URL or cURL command",
```
anwarulislam, line 573: Can we rename the placeholder to `Enter a URL or Paste cURL Command`? Because many users don't know that we can directly paste the cURL command. So, explicitly mentioning that would help users

anwarulislam (changes requested):

packages/hoppscotch-common/locales/en.json
shipko: Wow. I don't know about this feature. It's cool. Agree with @anwarulislam

packages/hoppscotch-common/src/components/smart/EnvInput.vue
```diff
@@ -109,6 +114,7 @@ const props = withDefaults(
     contextMenuEnabled?: boolean
     secret?: boolean
     autoCompleteEnv?: boolean
+    placeholderHoverString: string
```
JoelJacobStephen, line 117: There are many instances where `EnvInput` is used without a `placeholderHoverString`. So this should not be a required prop.

packages/hoppscotch-common/src/components/collections/SaveRequest.vue
```diff
@@ -221,6 +223,15 @@ const saveRequestAs = async () => {
 
   requestUpdated.name = requestName.value
 
+  if (props.mode === "rest") {
+    ;(requestUpdated as HoppRESTRequest).endpoint =
```
JoelJacobStephen, line 227: Please resolve type issues here

packages/hoppscotch-common/src/components/collections/SaveRequest.vue
```diff
@@ -221,6 +223,15 @@ const saveRequestAs = async () => {
...
+      (requestUpdated as HoppRESTRequest).endpoint ||
+      getDefaultRESTRequest().endpoint
+  } else {
+    ;(requestUpdated as HoppGQLRequest).url =
```
JoelJacobStephen, line 231: There are type issues here which needs to be resolved.

JoelJacobStephen (changes requested):
@nivedin In the GQL tab, when connecting to the default url, the color of the URL becomes extremely dim and isn't readable. Also, the input should be disabled when connected. 

Is it possible to bring back a bright color to the default endpoint/url when a user sends a request/connects to it? 


https://github.com/hoppscotch/hoppscotch/assets/70131076/a3b4221f-f2e6-4cf4-a5d6-85376720ab2b

PR #4009, opened 2024-04-25
refactor: update to dioc v3

AndrewBastin:
This PR introduces changes to keep the Hoppscotch codebase mostly in line with the changes introduces in [`dioc`](https://github.com/AndrewBastin/dioc) version `3.0.1` which tries to update certain bits of how a `Container` context is passed down to services.

## What's changed
- Upgraded `dioc` to `3.0.1` on all packages.
- Services which do not do anything other than call the super constructor have their constructor definitions removed so that the normal service constructors are called automatically.
- Services which do logic on their constructors are moved to the new `onServiceInit` method overridable by services to implement their initialisation logic. The constructor after moving the logic is removed.
- For `StaticSpotlightSearcherService`, an abstract Service that is implemented that standardized static search logic has a constructor which takes arguments, since there are many implementations of this class, for the time being, the constructor is kept on the inherited services but logic is mostly moved ton `onServiceInit`. TODOs are kept to notify that this a thing we will have to refactor down the line.

This PR shouldn't introduce any new end-user functionality of changes in behaviour.

## Notes to the reviewer
- Make sure all services have their constructors removed and moved to `onServiceInit` (except for `StaticSpotlightSearcherService` and its children).


jamesgeorge007 (changes requested):
There're type errors wrt `useService` composable usage. Can you look into it?

jamesgeorge007 (approved):

PR #4033, opened 2024-05-03
fix: code generate modal erroring out

### LINKED ISSUE #4008: [bug]: Selfhosted Generate Code ist Broken

### Current behavior

When i select generate code to see the request including headers that should be send it shows an error message "something went wrong" 

image

and i get this in the browser console:

image

we are using 2024.3.1 selfhosted

### Steps to reproduce

Click generate Code and you get the error

### Environment

Production

### Version

Self-hosted

AndrewBastin:
Closes #4008

This PR fixes the issue where the Generate Code Modal errors out when opened.

### What's Changed
The issue is mostly caused by `ajv` a dependency of `httpsnippet` (the library we use to power the Generate Code feature) not being kept properly up-to-date by the source package and needs to be pinned to an older version. We had this going in an earlier fix, but our fix was not solid, we just declared that `httpsnippet` has a peer dependency that is set to a specific version that works. But since its a peer dependency we are expected to have it installed as well. For some reason in `pnpm` v8, `pnpm` resolved it properly to load the version asked in peer dependencies, but this behaviour changed in v9. The fix here just adds `ajv` of the requested version added as a dependency to the users of `httpsnippet` so that `pnpm` resolves it correctly.

### Checks

### Notes to reviewer
It is best to clear the node_modules and then running `pnpm install` before you spin up a dev server or a deploy instance as vite sometimes misses the changes.

```bash
# While cd-ed into the root of the repo
rm -rv node_modules packages/*/node_modules
pnpm install 
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     },
     "packageExtensions": {
       "httpsnippet@3.0.1": {
-        "peerDependencies": {
+        "dependencies": {
           "ajv": "6.12.3"
         }
       }
--- a/packages/hoppscotch-common/package.json
+++ b/packages/hoppscotch-common/package.json
@@ -47,6 +47,7 @@
     "@vitejs/plugin-legacy": "4.1.1",
     "@vueuse/core": "10.7.0",
     "acorn-walk": "8.3.0",
+    "ajv": "6.12.3",
     "axios": "1.6.2",
     "buffer": "6.0.3",
     "cookie-es": "1.0.0",
--- a/packages/hoppscotch-common/package.json
+++ b/packages/hoppscotch-common/package.json
@@ -47,7 +47,6 @@
     "@vitejs/plugin-legacy": "4.1.1",
     "@vueuse/core": "10.7.0",
     "acorn-walk": "8.3.0",
-    "ajv": "6.12.3",
     "axios": "1.6.2",
     "buffer": "6.0.3",
     "cookie-es": "1.0.0",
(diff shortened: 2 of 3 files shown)


JoelJacobStephen (approved):

amk-dev (commented):
I'll add somemore context to the PR description.

Why this error occurred?

We encountered similar issues with `httpsnippet` package previously, and it was fixed in https://github.com/hoppscotch/hoppscotch/pull/3293.

we fixed it by adding `ajv@6` to the peer dependencies of `httpsnippet` package. this worked because `auto-install-peers` was enabled by default. and the correct version of `ajv` was linked to the `httpsnippet` package.

Why is this breaking now?

`PNPM 9` has a breaking change that allows `pnpm` to pick up other ranges of a dependency even if it is not matched by the peer dependency version. in our repo, some other packages like `swagger-parser`, `nest-cli` use `ajv@8`. so pnpm uses that version instead of again installing `ajv@6`. 

this new behavior is causing the regression we are seeing now.

My suggestion:

I would say instead of adding the ajv@6 as a dep to `hoppscotch/common`. we could update the `packageExtension` to use `dependencies` instead of `peerDependencies` in the `httpsnippet` package. this way we can understand that this is a dependency of `httpsnippet` and not `hoppscotch/common`.

```json
{
  "packageExtensions": {
    "httpsnippet@3.0.1": {
      "dependencies": {
        "ajv": "6.12.3"
      }
    }
  }
}
```

AndrewBastin:
Yep, makes sense to override ajv as a dependency. Will update the PR

PR #4048, opened 2024-05-07
feat: healthcheck for external services

### LINKED ISSUE #3830: [feature]: add health check checking the DB connection

### Summary

I need a health check that also checks whether the database is accessible.
The current "ping" does not check whether the DB is reachable packages/hoppscotch-backend/src/app.controller.ts

A simple SELECT 1 would be enough

I currently have the problem that Hoppscotch is no longer accessible when the DB credentials are updated automatically, and the health check does not activate the new ones by restarting the pod because the health check is still successful.

see above

balub:
Closes HSB-436
Closes #3830 

### Description

In this PR we are adding health checks to external services used by the hopscotch backend. We introduce a new route `/health` which when called will return the status of the service in the following result.

```
{
  "status": "ok",
  "info": {
    "database": {
      "status": "up"
    }
  },
  "error": {},
  "details": {
    "database": {
      "status": "up"
    }
  }
}
```

### Checks

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -112,7 +112,7 @@ services:
     build:
       dockerfile: packages/hoppscotch-backend/Dockerfile
       context: .
-      target: dev
+      target: prod
     env_file:
       - ./.env
     restart: always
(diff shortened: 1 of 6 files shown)


mirarifhasan (approved):
LGTM 💯

PR #4074, opened 2024-05-19
HSB-445 feature: storing user last login timestamp

mirarifhasan:
Closes HSB-444

### Description

This PR introduces the ability to store the user's last login timestamp in `hoppscotch-backend` package (in the `user` table). `lastLoggedOn` is the property to access the data. This value can be nullable for previous data.

### Checks

- [x] My pull request adheres to the code style of this project
- [ ] My code requires changes to the documentation
- [ ] I have updated the documentation as required
- [x] All the tests have passed

### Additional Information

Nil

PR #4077, opened 2024-05-20
feat(sh-admin): introducing last active status to users table and individual user's page in dashboard 

JoelJacobStephen:
### Ticket
- Closes HFE-513
- Depends on 
	- #4074 
	- #4121 

### Description
This PR introduces last active on property to the users table and in the individual user's page in the admin dashboard

### Checks


packages/hoppscotch-sh-admin/src/pages/users/index.vue
```diff
@@ -109,6 +105,10 @@
               </div>
             </td>
 
+            <td class="py-2 px-7">
+              {{ useTimeAgo(user.lastActiveOn).value }}
```
jamesgeorge007, line 109: We'll have to account for the case when `lastActiveOn` yields `null`. Currently, it shows `55 years ago` in such a scenario.

jamesgeorge007 (changes requested):

packages/hoppscotch-sh-admin/src/pages/users/index.vue
```diff
@@ -277,15 +276,18 @@ import IconX from '~icons/lucide/x';
 const t = useI18n();
 const toast = useToast();
 
+// Time and Date Helpers
 const getCreatedDate = (date: string) => format(new Date(date), 'dd-MM-yyyy');
 const getCreatedTime = (date: string) => format(new Date(date), 'hh:mm a');
+const lastActiveOn = (date: string | null) =>
+  date ? useTimeAgo(date).value : '';
```
jamesgeorge007, line 283: Shouldn't we keep a fallback copy instead of leaving it empty? Could you get it confirmed?
Also, we could rename the function to `getLastActiveOn` as per the convention followed.

jamesgeorge007 (approved):

PR #4095, opened 2024-05-30
feat: access team workspace collections and environments from the CLI

jamesgeorge007:
### Description

This PR aims at bringing support for accessing collections and environments from team environments from the CLI via supplying their respective IDs. The following updates are made to the `test` command:

- The `test` command and the `--env, -e` flag now support specifying an ID in addition to the file path.
- A new `--token` flag is added that takes in a personal access token for authorized access to the above resources.
- By default, the data is fetched from the Hoppscotch cloud instance. An SH instance can be specified with the new `--server` flag.

Closes HFE-502 HFE-503.

### Checks


nivedin (approved):
LGTM 💯

packages/hoppscotch-cli/src/utils/getters.ts
```diff
@@ -134,3 +152,98 @@ export const roundDuration = (
...
+        throw error({ code: "SERVER_CONNECTION_REFUSED", data: serverUrl });
+      }
+
+      if (
+        axiosErr.message === "INVALID_CONTENT_TYPE" ||
+        axiosErr.code === "ERR_INVALID_URL" ||
+        axiosErr.code === "ENOTFOUND" ||
+        axiosErr.response?.status === 404
+      ) {
+        throw error({ code: "INVALID_SERVER_URL", data: serverUrl });
```
amk-dev, line 236: if the user does not provide a server URL and the api.hoppscotch.io server returned some error, `serverURL` will be `undefined`

for eg, right now since the changes are not in api.hoppscotch.io, it will 404 and we'll get this error message, `Please provide a valid SH instance server URL: undefined`

If we move the `serverURL` to the top of the try block, it won't be undefined, it will be the user provided value or the default api.hoppscotch.io.

the error message will be more meaningful this way.

packages/hoppscotch-cli/src/utils/getters.ts
```diff
@@ -134,3 +152,98 @@ export const roundDuration = (
...
+      }
+    }
+  }
+
+  // Fallback to reading from file if contents are not available
+  if (contents === null) {
+    contents = (await readJsonFile(pathOrId, fileExistsInPath)) as
```
amk-dev, line 243: `readJSONFile` does not always returns a HoppCollection or Environment
if the collection file is malformed, it can be a random object, so the type assertion is not safe.

if you want to make sure that the contents is a `HoppCollection` or `Environment`, you can run the zod validation here. 
and throw an error if the validation fails.

the assertion may seem okey right now, because i can see a zod validation being run where this function is used.

but if someone else uses this some where else, `getResourceContents`  signature can be misleading. they might expect a `HoppCollection` or `Environment`, but they can get a random object.

packages/hoppscotch-cli/src/utils/getters.ts
```diff
@@ -134,3 +152,98 @@ export const roundDuration = (
...
+        },
+      });
+
+      if (!headers["content-type"].includes("application/json")) {
+        throw new Error("INVALID_CONTENT_TYPE");
+      }
+
+      contents =
+        resourceType === "collection"
+          ? transformWorkspaceCollection(data as WorkspaceCollection)
```
amk-dev, line 206: in an ideal world, we should not cast here and  run a validation, because we can't trust any backend, even if its our own.

and if the lowest layer has validation, all the above layers can use the value returned with peace of mind.

but now if you don't want to do that with our timelines, its understandable.

packages/hoppscotch-cli/src/utils/workspace-access.ts
```diff
@@ -0,0 +1,124 @@
...
+ *
+ * @param {WorkspaceCollection} collection - The workspace collection object to be transformed.
+ * @returns {HoppCollection} The transformed collection object conforming to the `HoppCollection` type.
+ */
+export const transformWorkspaceCollection = (
+  collection: WorkspaceCollection
+): HoppCollection => {
+  const { id, title, data, requests, folders } = collection;
+
+  const parsedData = data ? JSON.parse(data) : {};
```
amk-dev, line 111: another place to improve validation if you want.
eg: JSON.parse will throw for strings. and this function will return invalid auth/header object for malformed data.

```typescript
data  = `{"auth": "blah", "headers": ["something"] }` // this will go through
data = "empty_or_some_other_string" // this will throw a syntax error with JSON.parse
```

obviously we can skip this now and trust our backend. but if we guard our data ingestion places to our app, we'll reduce a lot of edge cases.

packages/hoppscotch-cli/src/utils/workspace-access.ts
```diff
@@ -0,0 +1,124 @@
...
+    ...rest,
+  };
+};
+/**
+ * Transforms workspace collection data to the `HoppCollection` format.
+ *
+ * @param {WorkspaceCollection} collection - The workspace collection object to be transformed.
+ * @returns {HoppCollection} The transformed collection object conforming to the `HoppCollection` type.
+ */
+export const transformWorkspaceCollection = (
```
amk-dev, line 106: i can see there are two functions that duplicates their logic. `transformWorkspaceCollections` and `transformWorkspaceChildCollections`. if you accept an array of collections for `transformWorkspaceCollections` you can remove `transformWorkspaceChildCollections`.

```typescript
export const transformWorkspaceCollections = (
  collections: WorkspaceCollection[]
): HoppCollection[] => {
  return collections.map((collection) => {
    const { id, title, data, requests, folders } = collection;

    const parsedData = data ? JSON.parse(data) : {};
    const { auth = { authType: "inherit", authActive: true }, headers = [] } =
      parsedData;

    return {
      v: CollectionSchemaVersion,
      id,
      name: title,
      folders: transformWorkspaceCollections(folders),
      requests: transformWorkspaceRequests(requests),
      auth,
      headers,
    };
  });
};
```

and at places where you call `transformWorkspaceCollection` you can call `transformWorkspaceCollection` with an array

```typescript
// instead of
transformWorkspaceCollection(collection);
// you can call
transformWorkspaceCollections([collection]);
```

packages/hoppscotch-cli/src/utils/getters.ts
jamesgeorge007: Yup, makes sense. I've updated the `getResourceContents` function return type to be `unknown`. The aim is to obtain the contents of a certain resource and handle any errors during the attempt (n/w call failure, etc). Any validation performed over the value yield is the responsibility of the place where it gets consumed.

packages/hoppscotch-cli/src/utils/getters.ts
jamesgeorge007: Fair point. For now, any errors encountered will be caught at the `catch` block within the `getResourceContents` helper function where the above gets consumed. We could revisit this if the scope of usage goes beyond down the line.

packages/hoppscotch-cli/src/utils/workspace-access.ts
jamesgeorge007: https://github.com/hoppscotch/hoppscotch/pull/4095#discussion_r1654523718

amk-dev (approved):
LGTM

PR #4098, opened 2024-06-03
feat: added change log prompt for PWA updates

anwarulislam:
Closes HFE-495

### Description
Implemented a changelog dialog that displays during application updates. This dialog effectively communicates the changes included in the new version.

### Checks

### DIFF
--- a/packages/hoppscotch-common/src/components/app/PWAPrompt.vue
+++ b/packages/hoppscotch-common/src/components/app/PWAPrompt.vue
@@ -2,7 +2,10 @@
   <div class="whats-new">
     <p>Hoppscotch has been automatically updated to the latest version 🎉</p>
     <p>
-      <a href="#" style="display: flex; align-items: center; gap: 5px">
+      <button
+        style="display: flex; align-items: center; gap: 5px"
+        @click="openWhatsNew"
+      >
         See what’s new
         <span>
           <svg
@@ -18,7 +21,7 @@
             />
           </svg>
         </span>
-      </a>
+      </button>
     </p>
     <button class="headlessClose" @click="$emit('closeToast')">
       <svg width="14" height="14" viewBox="0 0 16 16" fill="currentColor">
@@ -30,7 +33,18 @@
   </div>
 </template>
 
-<script setup lang="ts"></script>
+<script setup lang="ts">
+import { platform } from "~/platform"
+
+// types
+const props = defineProps<{
+  notesUrl: string
+}>()
+
+const openWhatsNew = () => {
+  if (props.notesUrl) platform.io.openExternalLink(props.notesUrl)
+}
+</script>
 
 <style scoped>
 .whats-new {
--- a/packages/hoppscotch-common/src/components/app/PWAPrompt.vue
+++ b/packages/hoppscotch-common/src/components/app/PWAPrompt.vue
@@ -1,0 +1,73 @@
+<template>
+  <div class="whats-new">
+    <p>Hoppscotch has been automatically updated to the latest version 🎉</p>
+    <p>
+      <a href="#" style="display: flex; align-items: center; gap: 5px">
+        See what’s new
+        <span>
+          <svg
+            width="8"
+            height="8"
+            viewBox="0 0 8 8"
+            fill="none"
+            xmlns="http://www.w3.org/2000/svg"
+          >
+            <path
+              d="M4.24503 7.13139L3.60582 6.49929L5.82528 4.27983H0.25V3.35653H5.82528L3.60582 1.14062L4.24503 0.504971L7.55824 3.81818L4.24503 7.13139Z"
+              fill="#868821"
+            />
+          </svg>
+        </span>
+      </a>
+    </p>
+    <button class="headlessClose" @click="$emit('closeToast')">
+      <svg width="14" height="14" viewBox="0 0 16 16" fill="currentColor">
+        <path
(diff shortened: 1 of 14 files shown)


packages/hoppscotch-common/src/modules/pwa.ts
```diff
@@ -73,6 +73,7 @@ export default <HoppModule>{
   onRootSetup() {
     onMounted(() => {
       usePwaPrompt()
+      useWhatsNewPrompt()
```
AndrewBastin, line 76: We can move this out of the PWA module and instead move it into a separate module ?

The update verification is not exclusive to PWAs so better to isolate that part out.

packages/hoppscotch-common/src/composables/pwa.ts
```diff
@@ -11,28 +16,67 @@ export const usePwaPrompt = function () {
...
+  }
+
+  persistenceService.setLocalConfig("hopp_v", version)
+}
+
+async function getReleaseNotes(v: string) {
+  const res = await fetch(`https://releases.hoppscotch.com/releases/${v}.json`)
+  const { release_notes } = await res.json()
+  return release_notes
 }
```
AndrewBastin, line 82: Moving this out into a separate file would be nice because the release notes check is not associated with any PWA stuff.

packages/hoppscotch-common/src/components/app/PWAPrompt.vue
AndrewBastin, line 1: Rename the file name to something like `WhatsNewPrompt`

packages/hoppscotch-common/src/components/app/PWAPrompt.vue
```diff
@@ -0,0 +1,87 @@
+<template>
+  <div class="whats-new">
+    <p>Hoppscotch has been automatically updated to the latest version 🎉</p>
```
AndrewBastin, line 3: i18n

packages/hoppscotch-common/src/components/app/PWAPrompt.vue
```diff
@@ -0,0 +1,87 @@
...
+    <p>
+      <button
+        style="display: flex; align-items: center; gap: 5px"
+        @click="openWhatsNew"
+      >
+        See what’s new
```
AndrewBastin, line 9: i18n

AndrewBastin (changes requested):
Please look into the given corrections and also note that the release notes functionality is not associated with the PWA, for example the desktop app is not considered a PWA so we might disable its ability to load the PWA module in the future or so, the implementation you have given here can anyways be refactored into a separate module/files.

Also, make sure to rebase the branch to be on top of the `next` branch, I have updated the destination branch in the PR, but haven't rebased the branch on top of it yet: https://github.com/hoppscotch/hoppscotch/tree/next

packages/hoppscotch-common/src/composables/whats-new.ts
```diff
@@ -0,0 +1,41 @@
...
+
+  if (oldVersion === version) return
+
+  // Skipping the minor version update. e.g. 2024.1.0 -> 2024.1.1
+  // Checking major version update. e.g. 2024.1 -> 2024.2
+
+  const isUpdatedToLatestMajorVersion =
+    oldVersion?.split(".").slice(0, 2).join(".") === latestMajorVersion
+
+  if (isUpdatedToLatestMajorVersion) {
```
jamesgeorge007, line 22: Is this check necessary? Doesn't this resolve to the same check (`oldVersion` & `version`) above where the early return happens?

`latestMajorVersion` -> `version.split(".").slice(0, 2).join(".")`

jamesgeorge007: Got the use case. But, during a minor bump, won't the above check result in `true` while comparing against the major version strings alone? Say, while updating from `2024.3.3` -> `2024.3.4`. If there's already an entry under `localStorage` at first while loading the app, it hits the early return since both the versions (`localStorage` entry & `@hoppscotch/common` version) are the same. An update was detected via the `registerSw` flow. Now, in the above flow, the versions are not the same since the `@hoppscotch/common` version is bumped. Obtaining the major version entries now points to the same value, the `isUpdatedToLatestMajorVersion` check would resolve to `true` entering the block about triggering the network call to fetch the release notes.

Now for a major bump, `2024.3.0` -> `2024.6.0`, the same flow happens and post the app reload via `registerSw`, the `isUpdatedToLatestMajorVersion` check won't hold `true` since the major versions (`2024.3` & `2024.6`) are different. This also applies when the `localStorage` entry isn't present and the update happens.

Also, could you elaborate on how the `whats-new` module will be triggered after performing an update? Being a module it runs when the app is loaded at first, but we'll require it to be triggered after an update happens via the `registerSw` flow as well.

https://github.com/hoppscotch/hoppscotch/blob/6e63c723b7218f3d8db91f01401bbc116fa46dee/packages/hoppscotch-common/src/modules/pwa.ts#L20-L22

packages/hoppscotch-common/src/composables/whats-new.ts
```diff
@@ -0,0 +1,41 @@
...
+        },
+      })
+    }
+  }
+
+  persistenceService.setLocalConfig("hopp_v", version)
+}
+
+async function getReleaseNotes(v: string) {
+  const res = await fetch(`https://releases.hoppscotch.com/releases/${v}.json`)
```
jamesgeorge007, line 38: Since this applies to SH, should we perform the network call via `InterceptorService` to prevent any potential CORS issues?

anwarulislam: Got it. Yeah, we can use `InterceptorService`

packages/hoppscotch-common/src/composables/whats-new.ts
anwarulislam: Ahh! I see. It's fixed, now!

jamesgeorge007 (approved):

nivedin (approved):

PR #4099, opened 2024-06-03
feat: added codegen tab in the sidbear

anwarulislam:
Closes HFE-493

### Description
This change addresses issue HFE-493 by making code geneation tools readily available through the sidebar.

- [ ] Not Completed
- [x] Completed

### Checks

### DIFF
--- a/packages/hoppscotch-common/src/components/http/Sidebar.vue
+++ b/packages/hoppscotch-common/src/components/http/Sidebar.vue
@@ -33,6 +33,13 @@
     >
       <Share />
     </HoppSmartTab>
+    <HoppSmartTab
+      :id="'codegen'"
+      :icon="IconCode"
+      :label="`${t('tab.codegen')}`"
+    >
+      <Share />
+    </HoppSmartTab>
   </HoppSmartTabs>
 </template>
 
@@ -41,6 +48,7 @@ import IconClock from "~icons/lucide/clock"
 import IconLayers from "~icons/lucide/layers"
 import IconFolder from "~icons/lucide/folder"
 import IconShare2 from "~icons/lucide/share-2"
+import IconCode from "~icons/lucide/code"
 import { ref } from "vue"
 import { useI18n } from "@composables/i18n"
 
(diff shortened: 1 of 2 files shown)


nivedin (changes requested):
image

 we can add a padding top, remove the border-radius and left-right-border for the `HttpCodegen` component
 
- also we have to watch the tab state because the code generation is not updating

https://github.com/hoppscotch/hoppscotch/assets/53208152/087d979a-90f9-42fb-a108-bc0b315123dc

packages/hoppscotch-common/src/components/http/Sidebar.vue
```diff
@@ -33,6 +33,24 @@
...
+      :icon="IconCode"
+      :label="`${t('tab.codegen')}`"
+    >
+      <div
+        class="flex items-center overflow-x-auto whitespace-nowrap border-b border-dividerLight px-4 py-2 text-tiny text-secondaryLight"
+      >
+        <span class="truncate"> Request </span>
+        <icon-lucide-chevron-right class="mx-2" />
+        Code snippet
+      </div>
```
nivedin, line 47: we can add i18n strings here

packages/hoppscotch-common/src/components/http/Sidebar.vue
```diff
@@ -33,6 +33,24 @@
...
+      <HttpCodegen
+        :hide-label="true"
+        class="px-4 mt-4"
+        v-if="selectedNavigationTab === 'codegen'"
```
nivedin, line 51: why do we need the v-if here ?

nivedin (changes requested):

packages/hoppscotch-common/src/components/http/Sidebar.vue
anwarulislam: Because we only want to process codegen when the tab is open.

nivedin: okay gotcha

nivedin (approved):
LGTM 🚀

PR #4111, opened 2024-06-06
feat: client certificates and ability to skip ssl cert verification in desktop app

AndrewBastin:
Closes HFE-485

This PR introduces the ability for Client Certificate Auth along with the ability to skip SSL Certificate Verification in the Hoppscotch Desktop App.

### What's changed
- Rewrite `NativeInterceptorService` to no longer use Tauri's built in HTTP APIs.
- Introduce Settings Entries for Native Interceptor.
- Introduce `hopp_native_interceptor` Tauri Plugin.
- Introduce ability to disable SSL certificate verification.
- Introduce ability to add client certificates.

### Notes
- Ability to define custom CA certificates are added in the Tauri Plugin level and the app service level, but due to some concerns with compatibility with old v1 X.509 root certificates (esp on macOS), this feature is deferred for the time being on the UI end, the UI side components and things exist, but since they are not used, it should be removed by tree shaking.
- i18n is not provided for the `hoppscotch-selfhost-desktop` components as there is no proper mechanism at the moment to deal with this. This is something we will be revisiting soon with an i18n system refactor.


nivedin (approved):
LGTM 💯

PR #4123, opened 2024-06-14
refactor(common): improvements to banner handling logic

JoelJacobStephen:
### Ticket
Closes HFE-524

### Description
This PR brings about improvements to the handling of banners in the app

### Checks

### DIFF
--- a/packages/hoppscotch-common/src/components/app/Header.vue
+++ b/packages/hoppscotch-common/src/components/app/Header.vue
@@ -255,7 +255,7 @@ import { GetMyTeamsQuery, TeamMemberRole } from "~/helpers/backend/graphql"
 import { deleteTeam as backendDeleteTeam } from "~/helpers/backend/mutations/Team"
 import { platform } from "~/platform"
 import {
-  BANNER_PRIORITY_MEDIUM,
+  BANNER_PRIORITY_LOW,
   BannerContent,
   BannerService,
   BannerTypes,
@@ -300,7 +300,7 @@ const offlineBanner: BannerContent = {
   type: BannerTypes.Warning,
   text: (t) => t("helpers.offline"),
   alternateText: (t) => t("helpers.offline_short"),
-  score: BANNER_PRIORITY_MEDIUM,
+  score: BANNER_PRIORITY_LOW,
   dismissible: true,
 }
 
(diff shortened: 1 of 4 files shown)


packages/hoppscotch-common/src/services/banner.service.ts
```diff
@@ -49,12 +50,18 @@ export class BannerService extends Service {
   public showBanner(banner: BannerContent) {
     this.bannerID = this.bannerID + 1
     this.bannerList.value.push({ id: this.bannerID, content: banner })
+
+    this.currentBannerID.value =
+      getBannerWithHighestScore(this.bannerList.value)?.id ?? null
+
```
nivedin, line 56: shouldn't we set the `currentBannerID` when `content` in generated, also the `content` will be the currentBanner right ? can we use that instead of `currentBannerID` @JoelJacobStephen

nivedin (changes requested):

packages/hoppscotch-common/src/components/app/Header.vue
```diff
@@ -293,31 +293,39 @@ const mdAndLarger = breakpoints.greater("md")
...
 const bannerContent = computed(() => banner.content.value?.content)
-let bannerID: number | null = null
+let offlineBannerID: number | null = null
 
 const offlineBanner: BannerContent = {
   type: "warning",
   text: (t) => t("helpers.offline"),
   alternateText: (t) => t("helpers.offline_short"),
-  score: BANNER_PRIORITY_HIGH,
+  score: BANNER_PRIORITY_LOW,
```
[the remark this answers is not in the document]
JoelJacobStephen: In SHE, License status and Custom banner will take medium and high priority respectively. Hence the change.

packages/hoppscotch-common/src/services/banner.service.ts
JoelJacobStephen: @nivedin We need the id to remove the current banner through the `removeBanner` function. This can happen when a different service creates a banner but the header component removes the banner when a dismiss event is invoked. We can remove the current banner form anywhere by just accessing the id and sending it to the function. For the setting of the id, I will provide an assertion since `bannerList` cannot be null as the setting of id is done after pushing a banner to the `bannerList`. Will make that change.

nivedin (approved):
LGTM 💯 
We can maybe rename [this](https://github.com/joeljacobstephen/hoppscotch/blob/4e916eae511ed0551d26551ad717b4c758fffab8/packages/hoppscotch-common/src/services/banner.service.ts#L45) variable to `currentBanner` since `getBannerWithHighestScore` returns the current banner which has the highest score.

PR #4141, opened 2024-06-22
feat: CLI collection runner command generation UI flow

jamesgeorge007:
### Description

#4095 adds support for accessing collections/environments from a team workspace straightaway via the CLI by supplying the respective IDs without having to obtain an export. This PR includes UI flow updates to support the same where the CLI command to run a certain collection with the ID can be obtained (optionally including the active environment) via the context menu.

Closes HFE-520.

### Changes

- Fixes the case about dismissing the environment context menu (opened via pressing the `More` kebab menu) via pressing `Escape`, which results in an exception.
  
  image

- Made `editingProperties` a required prop for the `CollectionsProperties` component with `null` removed from the types since the existing usages guarantee the above prop to be an object.

> The below additions apply only to team workspaces.

- New `Run collection` action under the context menu option for collections that displays the modal with the respective CLI command to run the collection, optionally including any active environment.
  
  image

  image

- The `Collection Properties` modal now includes a new `Details` tab that shows the collection ID.
  
  image

- There'll be a new `properties` action under the context menu for environments that opens a new `Environment Properties` modal which displays the respective environment ID.

  image

  image

### DIFF
--- a/packages/hoppscotch-common/src/components/collections/Collection.vue
+++ b/packages/hoppscotch-common/src/components/collections/Collection.vue
@@ -221,7 +221,8 @@
 <script setup lang="ts">
 import { useI18n } from "@composables/i18n"
 import { HoppCollection } from "@hoppscotch/data"
-import { VNodeRef, computed, ref, watch } from "vue"
+import { computed, ref, watch } from "vue"
+import { TippyComponent } from "vue-tippy"
 import { useReadonlyStream } from "~/composables/stream"
 import { TeamCollection } from "~/helpers/teams/TeamCollection"
 import {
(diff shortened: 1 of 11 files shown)


jamesgeorge007:
> we can add a copy button close to the text, similar to the other modals.

Discussed separately, the copy to clipboard action next to the generated CLI command made it appear crowded and a primary copy action already exists.

nivedin (approved):

packages/hoppscotch-common/src/components/collections/Collection.vue
```diff
@@ -267,16 +289,18 @@ const emit = defineEmits<{
...
 
-const tippyActions = ref<TippyComponent | null>(null)
-const requestAction = ref<HTMLButtonElement | null>(null)
-const folderAction = ref<HTMLButtonElement | null>(null)
-const edit = ref<HTMLButtonElement | null>(null)
-const deleteAction = ref<HTMLButtonElement | null>(null)
-const exportAction = ref<HTMLButtonElement | null>(null)
-const options = ref<TippyComponent | null>(null)
-const propertiesAction = ref<TippyComponent | null>(null)
+const tippyActions = ref<VNodeRef | null>(null)
```
amk-dev, line 295: what is the reason behind using `VNodeRef` here ? It includes very generic values like 

1. `Ref<any>` which will match any ref
2. string which will match any string

so typescript will not catch assignments like this,
```typescript
tippyActions.value = "whatever" // no TS error
``` 

look into the vue docs, on their recommendation for typing refs.
[https://vuejs.org/guide/typescript/composition-api#typing-template-refs](https://vuejs.org/guide/typescript/composition-api#typing-template-refs)

jamesgeorge007: Noticed a few type errors arising in the markup with a few property access via the template ref; picked the inferred type from the `ref` attribute. Yeah, better to narrow it down; push an [update](https://github.com/hoppscotch/hoppscotch/pull/4141/commits/9e1a454f8c76877b261a4a31bd35031cef34cebd).

amk-dev (approved):
LGTM ✅

PR #4153, opened 2024-06-28
feat: copyable invite links

amk-dev:
**Changes**

1. Now you can copy invite links to share manually instead of relaying on email only.
2. Since the SMTP check is only available in `selfhost-web`, this PR adds an optional `infra` platform config
3. Minor improvements like, different copy for when email is enabled v/s disabled

### DIFF
--- a/packages/hoppscotch-common/src/components/teams/Invite.vue
+++ b/packages/hoppscotch-common/src/components/teams/Invite.vue
@@ -10,10 +10,18 @@
         <div class="mb-8 flex max-w-md flex-col items-center justify-center">
           <icon-lucide-users class="h-6 w-6 text-accent" />
           <h3 class="my-2 text-center text-lg">
-            {{ t("team.we_sent_invite_link") }}
+            {{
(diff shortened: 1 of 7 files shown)


AndrewBastin (approved):

jamesgeorge007 (approved):
Some type errors arise in the markup; could be revisited as required.

packages/hoppscotch-common/src/components/teams/Invite.vue
```diff
@@ -10,10 +10,18 @@
...
+            }}
           </h3>
           <p class="text-center">
-            {{ t("team.we_sent_invite_link_description") }}
+            {{
+              inviteMethod === "email"
+                ? t("team.we_sent_invite_link_description")
+                : t("team.invite_sent_smtp_disabled_description")
+            }}
           </p>
```
jamesgeorge007, line 25: Potential candidates to be computed properties; leaving a note for reference while revisiting.

PR #4172, opened 2024-07-06
fix: fix issue for multipart/form-data and cookie header (#4162)

### LINKED ISSUE #4162: [bug]: pc app not network when choose mulitpart/form-data

### Current behavior

image
### Steps to reproduce

it work when clean contentType, my controller is 
  @PostMapping("/import")
   public Response importFile(MultipartFile file, Long mId) {
       return fileService.importFile(file, mId);
    }

### Environment

Production

### Version

Local

shipko:
Closes #4162 

### What's changed
This PR is fixing some issues on sending file data:

1. Fix data variable type that sending to interceptor.rs 
2. Fix parameter filter for unsetted file
3. Add mime type for sending file
4. Fix double Content-Type header
5. Fix Cookie header. Removing it if not present 

### Notes to reviewers

This PR is tested on web and desktop app. Not for CLI

### DIFF
--- a/packages/hoppscotch-selfhost-desktop/src/platform/interceptors/native/index.ts
+++ b/packages/hoppscotch-selfhost-desktop/src/platform/interceptors/native/index.ts
@@ -188,6 +188,7 @@ async function convertToRequestDef(
     method: axiosReq.method ?? "GET",
     endpoint: axiosReq.url ?? "",
     headers: Object.entries(axiosReq.headers ?? {})
+      .filter(([key, value]) => key.toLowerCase() !== "content-type" && value.toLowerCase() !== "multipart/form-data") // Removing header, because this header will be set by reqwest
       .map(([key, value]): KeyValuePair => ({ key, value })),
     parameters: Object.entries(axiosReq.params as Record<string, string> ?? {})
       .map(([key, value]): KeyValuePair => ({ key, value })),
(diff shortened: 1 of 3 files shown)


packages/hoppscotch-selfhost-desktop/src/platform/interceptors/native/index.ts
```diff
@@ -129,7 +130,8 @@ async function processBody(axiosReq: AxiosRequestConfig): Promise<BodyDef | null
           value: {
             File: {
               filename: value.name,
-              data: new Uint8Array(await value.arrayBuffer())
+              data: Array.from(new Uint8Array(await value.arrayBuffer())),
+              mime: value.type,
```
AndrewBastin, line 134: If the browser/system doesn't know the Mime Type, `value.type` can be an empty string so we should avoid putting in the mime type in that situation.
See: https://arc.net/l/quote/thjiuzih

AndrewBastin (changes requested):

packages/hoppscotch-selfhost-desktop/src/platform/interceptors/native/index.ts
shipko: Yes, you're right!
By [rfc2046](https://www.ietf.org/rfc/rfc2046.txt) if mime type is not recognizable you should use `"application/octet-stream"` content type

shipko:
@AndrewBastin please, review it

AndrewBastin (approved):

PR #4173, opened 2024-07-07
fix: `<img>` is flex's immediate child therefore it stretch to its parent width

anwarulislam:
## Description
We noticed that images are always rendered full width, even if they’re small. For example, a 200x200 pixel image would stretch across its entire parent container. This PR fixes that issue!

### What's changed

| Before | After |
|--------|-------|
| 
image
 | 
image
 |


liyasthomas (approved):
LGTM

nivedin (approved):

PR #4177, opened 2024-07-09
feat: persist HTML response preview preference

anwarulislam:
Closes HFE-532

This pull request adds functionality to persist the HTML preview configuration. Previously, the configuration was not saved and had to be set every time the application was launched. With this change, the configuration will be saved locally, allowing the user's preference to persist.


nivedin (changes requested):
while switching tabs I need to re-toggle the preview to view the content, can you check if this persist in you end

https://github.com/user-attachments/assets/e38151c4-cc95-43a9-8882-9336fb9ba9f2

jamesgeorge007 (approved):

nivedin (approved):
LGTM 💯

PR #4180, opened 2024-07-10
fix: secret variables being populated as `undefined` on code snippets

nivedin:
Closes HFE-530

This PR fixes the issue in code snippet where while using secret environment the variable was shown as `undefined`, now it shows the key of the variable.

### What's changed
The bug is fixed, also the `httpsnippet` package has been updated, since the old package failed to decode the URL containing < & > character which we use as environment annotation.


jamesgeorge007 (changes requested):
- Referring to secret environment variables under `API key` authorization fields or for the `OAuth 2.0` token in a request with `Query Parameters` set as the `Pass by` option results in the URL showing `undefined` instead of the secret variable key.

  https://github.com/user-attachments/assets/1b1b6904-1d0c-4131-bd66-98c5426bac13

- Generating the cURL command for a request with secret environment variables referred to under the URL/parameters and attempting to import it straightaway results in the angular bracket notation getting stripped from the secret environment variable keys. This might add to the scope and can be handled separately.


  https://github.com/user-attachments/assets/d879d9b3-52e8-4558-b047-39e82aa8e1cb

packages/hoppscotch-common/src/helpers/curl/sub_helpers/url.ts
```diff
@@ -53,7 +53,7 @@ const parseURL = (urlText: string | number) =>
       u
         .toString()
         .replace(/^'|'$/g, "")
-        .replaceAll(/[^a-zA-Z0-9_\-./?&=:@%+#,;()'\s]/g, "")
+        .replaceAll(/[^a-zA-Z0-9_\-./?&=:@%+#,;()'<>\s]/g, "")
```
jamesgeorge007, line 56: URL encoding seems coming into play. We could address this separately if this requires more effort.

https://github.com/user-attachments/assets/ee3515fa-007a-4480-b66f-b28ad5aada5e

nivedin: have pushed a [fix](https://github.com/hoppscotch/hoppscotch/pull/4180/commits/9fdb0420e077f9110d54de9359a3395ab706bff8) by decoding the URL using `decodeURIComponent`

jamesgeorge007 (commented):
Approved ✅

Issue #4170, opened 2024-07-04
[bug]: Response Body content not showing using widget


jubu02:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

Earlier versions of Hoppscotch Widget displayed any response JSON by default but with current version there is this one liner where you can download the response but not display it within the widget in any way.. 

![Screenshot 2024-07-04 121220](https://github.com/hoppscotch/hoppscotch/assets/8514212/96949000-b10c-458e-87af-7da9718c70c1)

### Steps to reproduce

1. Visit tppvalidation.com 
2. Click send in any widget on the page
3. Response(s) not showing up (in this case as JSON) .. only the line that says there is this response body that can be downloaded.

(this somewhat replicates the behavoiur of attachments in the request body - they are also not visible in raw/code) - but that has been so in the past as well. 

### Environment

Production

### Version

Cloud

liyasthomas:
@nivedin can you look into this?

nivedin:
@jubu02 the bug is fixed, feel free to re-open this if the issue still persists.

PR #4181, opened 2024-07-11
fix: embeds response and request option section getting hidden bug

nivedin:
Closes HFE-539, #4170 

This PR fixes the UI bug in embeds where the response and some request option codemirror instance was hidden.

https://github.com/hoppscotch/hoppscotch/assets/53208152/1ecf44a8-011a-4997-a58b-949f511830d7

### What's changed
The embeds structure have been updated by using `PaneLayout` for the view, now the user can set their prefered height, also some code refactor has been done for the `Embeds` component.

### DIFF
--- a/packages/hoppscotch-common/src/components/embeds/index.vue
+++ b/packages/hoppscotch-common/src/components/embeds/index.vue
@@ -1,7 +1,7 @@
 <template>
-  <div class="flex flex-col flex-1">
+  <div class="flex flex-col flex-1 min-h-screen">
     <header
-      class="flex items-center justify-between flex-1 flex-shrink-0 px-2 py-2 space-x-2 overflow-x-auto overflow-y-hidden"
+      class="max-h-[3rem] flex items-center justify-between flex-1 flex-shrink-0 px-2 py-2 space-x-2 overflow-x-auto overflow-y-hidden"
     >
       <div class="flex items-center justify-between flex-1 space-x-2">
         <HoppButtonSecondary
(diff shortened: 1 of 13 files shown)


packages/hoppscotch-common/src/components/http/Headers.vue
```diff
@@ -49,7 +49,7 @@
         />
       </div>
     </div>
-    <div v-if="bulkMode" class="h-full relative w-full">
+    <div v-if="bulkMode" class="h-full relative w-full flex flex-col flex-1">
```
jamesgeorge007, line 52: Can you confirm if you've cross-verified these additions with the changes in scope under #3848?

jamesgeorge007: We might also want to extend it to `HttpURLEncodedParams`.

https://github.com/hoppscotch/hoppscotch/blob/2917d50c6afe5f52fdcbb284514099d8423eca79/packages/hoppscotch-common/src/components/http/URLEncodedParams.vue#L47-L49

packages/hoppscotch-common/src/components/embeds/index.vue
```diff
@@ -1,105 +1,34 @@
...
-    />
-    <HttpResponse
-      v-if="tab.document.response"
-      :document="tab.document"
-      :is-embed="true"
-    />
-  </div>
+    </template>
+    <template #secondary>
+      <HttpResponse :document="tab.document" :is-embed="true" />
```
jamesgeorge007, line 21: Is it intentional not to conditionally render the response based on `tab.document.response` anymore?

packages/hoppscotch-common/src/components/http/Headers.vue
nivedin: > Can you confirm if you've cross-verified these additions with the changes in scope under https://github.com/hoppscotch/hoppscotch/pull/3848?

yup checked, working fine

packages/hoppscotch-common/src/components/embeds/index.vue
nivedin: since we use Panes now hiding the response makes the pane empty.. which doesn't look good 😅  open for suggestion

image

jamesgeorge007: Gotcha, that makes sense.

packages/hoppscotch-common/src/components/http/URLEncodedParams.vue
```diff
@@ -44,7 +44,7 @@
         />
       </div>
     </div>
-    <div v-if="bulkMode" class="h-full relative">
+    <div v-if="bulkMode" class="h-full relative w-full flex flex-1">
```
jamesgeorge007, line 47: It seems like the `flex-col` is usually applied across similar places although the direction doesn't matter much since it's just a single child.

PR #4194, opened 2024-07-17
feat: Ability to toggle cookies to work in HTTP

balub:
Closes HSB-459

### What's changed

In this PR, we are introducing a new environment variable `ALLOW_SECURE_COOKIES` which can be used to toggle if the cookies generated by the Hoppscotch backend should be used over HTTP.

### Checks

### Additional Information

None

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -112,7 +112,7 @@ services:
     build:
       dockerfile: packages/hoppscotch-backend/Dockerfile
       context: .
-      target: dev
+      target: prod
     env_file:
       - ./.env
     restart: always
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -112,7 +112,7 @@ services:
     build:
       dockerfile: packages/hoppscotch-backend/Dockerfile
       context: .
-      target: prod
+      target: dev
```
AndrewBastin, line 115: Is this intentional ?

balub: Yup same reason, only for dev. Will revert it before merging.

PR #4195, opened 2024-07-17
i18n: Fix existing german translations

pavog:
### What's changed
This PR changes some of the existing translations for the German language (Deutsch).
I'm a native speaker and while I explored the Hoppscotch app, I found some issues in the translation.


liyasthomas (approved):
LGTM

PR #4196, opened 2024-07-17
i18n: Add german translations

pavog:
Hi!
I'm a native German speaker.

I have spent a few hours translating the existing translation strings in de.json from English to German.
I'm pretty sure that all the translations I've done are correct, in terms of spelling, grammar and meaning.

I didn't translate 100% because I found some translation keys that don't seem to be used anymore. I also found some translations that I'm not sure about if and/or how to translate into German. The German language has already adapted many English words and especially with terms related to programming, there are some words that you can't translate or whose meaning would change if you would to translate them.
Therefore, I have left some of the original English translations in de.json. I would leave them open for discussion.


liyasthomas (approved):
LGTM

nivedin (approved):

PR #4212, opened 2024-07-25
HSB-473 feat: encrypt sensitive data before storing in db

mirarifhasan:
Closes HSB-473

This PR introduces data encryption before storing sensitive data in the database. For example, previously we stored SSO's secret key as plain text in the infraConfig table. From now on, before saving to DB, the value will be encrypted.
Also, after fetching the data from the DB, we decrypt it to the original value to use.

### What's changed

In `Account` and `InfraConfig` tables, we added encryption and decryption logic to store/ fetch sensitive data.

### Notes to reviewers

Nil

### DIFF
--- a/packages/hoppscotch-backend/src/infra-config/helper.ts
+++ b/packages/hoppscotch-backend/src/infra-config/helper.ts
@@ -5,7 +5,7 @@ import {
 } from 'src/errors';
 import { PrismaService } from 'src/prisma/prisma.service';
 import { InfraConfigEnum } from 'src/types/InfraConfig';
-import { throwErr } from 'src/utils';
+import { decrypt, encrypt, throwErr } from 'src/utils';
 import { randomBytes } from 'crypto';
 
 export enum ServiceStatus {
@@ -60,7 +60,9 @@ export async function loadInfraConfiguration() {
 
     let environmentObject: Record<string, any> = {};
     infraConfigs.forEach((infraConfig) => {
-      environmentObject[infraConfig.name] = infraConfig.value;
+      environmentObject[infraConfig.name] = infraConfig.isEncrypted
+        ? decrypt(infraConfig.value)
+        : infraConfig.value;
     });
 
     return { INFRA: environmentObject };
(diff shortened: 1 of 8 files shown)


packages/hoppscotch-backend/src/infra-config/helper.ts
```diff
@@ -76,119 +80,150 @@ export async function loadInfraConfiguration() {
...
     },
     {
       name: InfraConfigEnum.MAILER_TLS_REJECT_UNAUTHORIZED,
       value: process.env.MAILER_TLS_REJECT_UNAUTHORIZED,
+      isEncrypted: false,
     },
     {
       name: InfraConfigEnum.GOOGLE_CLIENT_ID,
       value: process.env.GOOGLE_CLIENT_ID,
+      isEncrypted: false,
```
balub, line 146: Lets encrypt this too.

packages/hoppscotch-backend/src/infra-config/helper.ts
```diff
@@ -76,119 +80,150 @@ export async function loadInfraConfiguration() {
...
       name: InfraConfigEnum.ANALYTICS_USER_ID,
       value: generateAnalyticsUserId(),
+      isEncrypted: false,
     },
     {
       name: InfraConfigEnum.IS_FIRST_TIME_INFRA_SETUP,
       value: (await prisma.infraConfig.count()) === 0 ? 'true' : 'false',
+      isEncrypted: false,
     },
   ];
```
balub, line 228: Lets encrypt all the client ids too.

packages/hoppscotch-backend/src/infra-config/infra-config.service.ts
```diff
@@ -62,10 +65,30 @@ export class InfraConfigService implements OnModuleInit {
...
+      if (encryptionRequiredEntries.length > 0) {
+        const dbOperations = encryptionRequiredEntries.map((dbConfig) => {
+          return this.prisma.infraConfig.update({
+            where: { name: dbConfig.name },
+            data: { value: encrypt(dbConfig.value), isEncrypted: true },
+          });
+        });
+
+        await Promise.allSettled(dbOperations);
+      }
```
balub, line 88: Can we generate an object that is essentially all the rows whose value needs to be encrypted and then just have one db query perform the operation on them all at once? Similiar to the import operations we have for collections.

packages/hoppscotch-backend/src/infra-config/infra-config.service.ts
```diff
@@ -118,9 +152,14 @@ export class InfraConfigService implements OnModuleInit {
     if (E.isLeft(isValidate)) return E.left(isValidate.left);
 
     try {
+      const { isEncrypted } = await this.prisma.infraConfig.findUnique({
+        where: { name },
+        select: { isEncrypted: true },
+      });
```
balub, line 158: Does this DB call need to be made? can't we just get this info from `getDefaultInfraConfigs`.

packages/hoppscotch-backend/src/infra-config/infra-config.service.ts
```diff
@@ -146,11 +185,23 @@ export class InfraConfigService implements OnModuleInit {
     if (E.isLeft(isValidate)) return E.left(isValidate.left);
 
     try {
+      const dbInfraConfig = await this.prisma.infraConfig.findMany({
+        select: { name: true, isEncrypted: true },
+      });
```
balub, line 190: Same here too, can't we use `getDefaultInfraConfigs` instead.

balub (changes requested):

packages/hoppscotch-backend/src/infra-config/infra-config.service.ts
mirarifhasan: That is not possible. In `updateMany` we can not write multiple where conditions.

packages/hoppscotch-backend/src/infra-config/infra-config.service.ts
mirarifhasan: `getDefaultInfraConfigs` also has a DB (`count`) call. So, ultimately it doesn't help much.

packages/hoppscotch-backend/src/infra-config/infra-config.service.ts
mirarifhasan: `getDefaultInfraConfigs` has a DB call.

packages/hoppscotch-backend/src/infra-config/helper.ts
mirarifhasan: Ok

packages/hoppscotch-backend/src/infra-config/helper.ts
```diff
@@ -76,119 +80,150 @@ export async function loadInfraConfiguration() {
...
+      isEncrypted: true,
     },
     {
       name: InfraConfigEnum.MAILER_TLS_REJECT_UNAUTHORIZED,
       value: process.env.MAILER_TLS_REJECT_UNAUTHORIZED,
+      isEncrypted: false,
     },
     {
       name: InfraConfigEnum.GOOGLE_CLIENT_ID,
       value: process.env.GOOGLE_CLIENT_ID,
```
balub, line 145:
```suggestion
      value: decrypt(process.env.GOOGLE_CLIENT_ID),
```

packages/hoppscotch-backend/src/infra-config/helper.ts
```diff
@@ -76,119 +80,150 @@ export async function loadInfraConfiguration() {
...
+      isEncrypted: false,
     },
     {
       name: InfraConfigEnum.GOOGLE_SCOPE,
       value: process.env.GOOGLE_SCOPE,
+      isEncrypted: false,
     },
     {
       name: InfraConfigEnum.GITHUB_CLIENT_ID,
       value: process.env.GITHUB_CLIENT_ID,
```
balub, line 165:
```suggestion
      value: decrypt(process.env.GITHUB_CLIENT_ID),
```

packages/hoppscotch-backend/src/infra-config/helper.ts
```diff
@@ -76,119 +80,150 @@ export async function loadInfraConfiguration() {
...
+      isEncrypted: false,
     },
     {
       name: InfraConfigEnum.GITHUB_SCOPE,
       value: process.env.GITHUB_SCOPE,
+      isEncrypted: false,
     },
     {
       name: InfraConfigEnum.MICROSOFT_CLIENT_ID,
       value: process.env.MICROSOFT_CLIENT_ID,
```
balub, line 185:
```suggestion
      value: decrypt(process.env.MICROSOFT_CLIENT_ID),
```

balub (changes requested):

balub (approved):
LGTM 🚀

PR #4219, opened 2024-07-29
fix: infra token expriry check on guard

mirarifhasan:
### What's changed

This PR resolves an bug on `infra-token.guard.ts` while checking the token is expire or not. `expiryOn` value can be `null`. So, this PR cover that conditions as well.

### Notes to reviewers

Nil


balub (approved):
LGTM 🚀

PR #2228, opened 2022-04-02
refactor: real-time system

anwarulislam:
### Description
WebSocket System Revamp: Migrate from Option API with JS to Script Setup on TS

### Checks

- [x] My pull request adheres to the code style of this project
- [ ] My code requires changes to the documentation
- [ ] I have updated the documentation as required
- [x] All the tests have passed

### Additional Information
`WebSocket.vue` component migrated to script setup in TS from JS options API

PR #4226, opened 2024-07-31
fix: path not taken into account for socket.io connection without token auth

artu-ole:
### What's changed

Socket.IO connection was failing for the server deployed on a custom path.
The reason being that path was not passed to the client implementations unless there was bearer token configured.

### Notes to reviewers

The bug appeared in https://github.com/hoppscotch/hoppscotch/pull/2228 when for some reason path option was not handed over to the `SIOClient` anymore after the refactor

| Before https://github.com/hoppscotch/hoppscotch/pull/2228 |      After https://github.com/hoppscotch/hoppscotch/pull/2228   |
|----------|:-------------:|
| Code_QWfE1zRfaJ | Code_AGgl6wthbA |


anwarulislam (approved):
LGTM 🚀

PR #4233, opened 2024-08-01
refactor: slim down the backend containers

AndrewBastin:
This PR intends to slim down the containers (particularly the backend and hence the AIO container) and along with that move the containers to execute the servers as a non-root container user (`hoppuser`).

For the AIO container, this PR reduces the container size from 2.0GB to 501MB (as per [dive](https://github.com/wagoodman/dive) report). Similar improvements should be present in the backend container as well (from 1.9GB to 440MB) but the FE and SH Admin containers should remain the same mostly.

The size reduction is mostly achieved by removing out the existing entire monorepo copy that was present in the containers. Along with the build steps and artifacts around it. The present build artifacts also caused a lot of erroneous vulnerability reports on the container as the container had more exploitable bits inside it. The reduction here reduces the signal-to-noise ratio for the containers greatly.

Along with that, this PR also moves the container from using non-root user accounts (named `hoppuser`) to execute the main operations (mostly the container switches to defaulting to `hoppuser` after all the installation steps are done in  `prod.Dockerfile`). This improves security and is considered a best practice to reduce damages that can be caused by privilege escalation attacks caused by container breakouts into the host system.

Along with that, this PR has some package overrides for packages in the repo that were having known vulnerabilities.

### What's changed
- Introduce a `backend_builder` build stage and move the build process into that.
- The backend container is now built and stored into the `/dist/backend` folder with `pnpm deploy` to generate a copy of the package without associations to the monorepo, see: [https://pnpm.io/cli/deploy](https://pnpm.io/cli/deploy).
- Add `hoppuser` user to switch to for execution after all the installs are done.
- AIO container is now not built from the backend container and is now directly a `node:20-alpine3.19` image. This is due to the `hoppuser` account messing up some stuff between the containers.
- Updated the root `package.json` to replace packages with vulnerabilities with a version that is fixed. We use PNPM's package overrides for this, see: https://pnpm.io/package_json#pnpmoverrides

### Notes to reviewers
Make sure to check all the containers work also consider the subpath access and non subpath access mode on the individual containers as well.

### DIFF
--- a/prod.Dockerfile
+++ b/prod.Dockerfile
@@ -15,19 +15,25 @@ COPY . .
 RUN pnpm install -f --offline
 
 
-FROM base_builder as backend
+FROM base_builder as backend_builder
 RUN apk add caddy
 WORKDIR /usr/src/app/packages/hoppscotch-backend
 RUN pnpm exec prisma generate
 RUN pnpm run build
+RUN pnpm --filter=hoppscotch-backend deploy /dist/backend --prod
+
+FROM node:20-alpine3.19 as backend
+RUN npm install -g pnpm
 COPY --from=base_builder /usr/src/app/packages/hoppscotch-backend/backend.Caddyfile /etc/caddy/backend.Caddyfile
+COPY --from=backend_builder /dist/backend /dist/backend
+WORKDIR /dist/backend
+RUN pnpm exec prisma generate
 # Remove the env file to avoid backend copying it in and using it
-RUN rm "../../.env"
 ENV PRODUCTION="true"
 ENV PORT=8080
 ENV APP_PORT=${PORT}
 ENV DB_URL=${DATABASE_URL}
-CMD ["node", "/usr/src/app/packages/hoppscotch-backend/prod_run.mjs"]
+CMD ["node", "/dist/backend/prod_run.mjs"]
 EXPOSE 80
 EXPOSE 3170
 
@@ -68,6 +74,7 @@ CMD ["node","/usr/prod_run.mjs"]
 FROM backend as aio
 RUN apk add caddy tini
 RUN npm install -g @import-meta-env/cli
+COPY --from=base_builder /usr/src/app/aio_run.mjs /usr/src/app/aio_run.mjs
 COPY --from=fe_builder /usr/src/app/packages/hoppscotch-selfhost-web/dist /site/selfhost-web
 COPY --from=sh_admin_builder /usr/src/app/packages/hoppscotch-sh-admin/dist-multiport-setup /site/sh-admin-multiport-setup
 COPY --from=sh_admin_builder /usr/src/app/packages/hoppscotch-sh-admin/dist-subpath-access /site/sh-admin-subpath-access
--- a/prod.Dockerfile
+++ b/prod.Dockerfile
@@ -100,6 +100,12 @@ WORKDIR /site
 CMD ["node","/site/prod_run.mjs"]
 
 FROM node:20-alpine3.19 AS aio
+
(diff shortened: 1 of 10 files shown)


balub:
Lets ship `pnpm` to avoid changes to setup currently.

AndrewBastin:
> Lets ship pnpm to avoid changes to setup currently.
@balub corrected.

prod.Dockerfile
```diff
@@ -15,73 +15,122 @@ COPY . .
...
+USER hoppuser
+
+WORKDIR /dist/backend
+
+CMD ["node", "prod_run.mjs"]
 EXPOSE 80
 EXPOSE 3170
 
-FROM base_builder as fe_builder
+FROM base_builder AS fe_builder
```
balub, line 50: image

I get this error when I try running the app container standalone.
Context: Subpath is false.

prod.Dockerfile
```diff
@@ -15,73 +15,122 @@ COPY . .
...
 
-FROM base_builder as sh_admin_builder
+FROM base_builder AS sh_admin_builder
 WORKDIR /usr/src/app/packages/hoppscotch-sh-admin
 # Generate two builds for `sh-admin`, one based on subpath-access and the regular build
 RUN pnpm run build --outDir dist-multiport-setup
 RUN pnpm run build --outDir dist-subpath-access --base /admin/
 
-FROM caddy:2-alpine as sh_admin
+FROM caddy:2-alpine AS sh_admin
```
balub, line 78: image

Same error here too.

balub (changes requested):
There seems to by some permission issue with the frontend images.

prod.Dockerfile
AndrewBastin: should be fixed now.

prod.Dockerfile
AndrewBastin: should be fixed now. ptal.

prod.Dockerfile
```diff
@@ -15,73 +15,119 @@ COPY . .
...
+RUN npm install -g @import-meta-env/cli
+
+USER hoppuser
+
 ENTRYPOINT [ "tini", "--" ]
-RUN apk --no-cache add curl
 COPY --chmod=755 healthcheck.sh .
 HEALTHCHECK --interval=2s CMD /bin/sh ./healthcheck.sh
+
+WORKDIR /dist
```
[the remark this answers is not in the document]
AndrewBastin: We need this when they `sh` into the container. The last set `WORKDIR` will be the path we will get into when we do `docker compose run hoppscotch-aio sh`.

prod.Dockerfile
```diff
@@ -15,73 +15,122 @@ COPY . .
...
-RUN npm install -g @import-meta-env/cli
-COPY --from=fe_builder /usr/src/app/packages/hoppscotch-selfhost-web/dist /site/selfhost-web
-COPY --from=sh_admin_builder /usr/src/app/packages/hoppscotch-sh-admin/dist-multiport-setup /site/sh-admin-multiport-setup
-COPY --from=sh_admin_builder /usr/src/app/packages/hoppscotch-sh-admin/dist-subpath-access /site/sh-admin-subpath-access
+FROM node:20-alpine3.19 AS aio
+# Run this separately to use the cache from backend
+RUN apk add caddy
+
+RUN apk add tini curl
+
```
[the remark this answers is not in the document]
AndrewBastin: fixed in latest commit. ptal.

balub (approved):
LGTM 🚀

PR #4263, opened 2024-08-14
fix: sync all settings actions in SH

jamesgeorge007:
### What's changed

Previously, the following settings were not getting synced in SH:

- `Expand navigation` & `Sidebar on left` under `Experiments`.
- Spotlight open status.
- Wrap lines preference under request params, headers in bulk mode, etc.
- `Collections`, `Environments` & `History` synchronization settings under the profile page.

This was the case since the relevant store dispatcher methods were not present under the store sync definition maintained at the platform level. This PR resolves this behaviour by extending the above definition accordingly. There was also a known issue with toggling the synchronization settings (Collections, Environments & History) under the profile page resulting in an exception which is fixed by adding safeguards with the usage of `startSubscriptions()` & `stopSubscriptions()` function calls from the syncing context.

Closes HFE-324 HFE-561.

### Notes to reviewers

While verifying the wrap lines preference toggle sync behaviour, make sure to check all request tabs applicable to the action.

https://github.com/hoppscotch/hoppscotch/blob/3f78bf10621eb2c6574f3a19f6169aeda016d69b/packages/hoppscotch-common/src/newstore/settings.ts#L32-L49

### DIFF
--- a/packages/hoppscotch-selfhost-desktop/src/lib/sync/index.ts
+++ b/packages/hoppscotch-selfhost-desktop/src/lib/sync/index.ts
@@ -80,7 +80,7 @@ export const getSyncInitFunction = <T extends DispatchingStore<any, any>>(
       )
     }
 
-    stopSubscriptions = startSubscriptions()
+    stopSubscriptions = startSubscriptions?.()
   }
 
   function stopListeningToSubscriptions() {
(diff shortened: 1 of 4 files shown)


nivedin (approved):
LGTM 💯

PR #4273, opened 2024-08-18
Fix + improvement for body tab

shipko:
### What's changed
This PR is adding label for body tab when is not empty. Also fixing problem with setting empty string in RawBody. 

Снимок экрана 2024-08-18 в 12 49 45

- [ ] Not Completed
- [x] Completed


nivedin (approved):
LGTM 💯

PR #4280, opened 2024-08-20
feat: sort environments alphabetically

nivedin:
Closes HFE-551

This PR sorts the environments in alphabetical order on the environment section in the sidebar and the environment selector.

### What's changed
The environments have been sorted alphabetically by default for both personal and team. Also made 3 letter minimum length for creating a Personal Environment which was already the criteria for creating a Team Environment.

### DIFF
--- a/packages/hoppscotch-common/src/components/environments/Selector.vue
+++ b/packages/hoppscotch-common/src/components/environments/Selector.vue
@@ -319,6 +319,10 @@ import { useLocalState } from "~/newstore/localstate"
 import { GetMyTeamsQuery } from "~/helpers/backend/graphql"
 import { useService } from "dioc/vue"
 import { WorkspaceService } from "~/services/workspace.service"
+import {
+  sortPersonalEnvironmentsAlphabetically,
+  sortTeamEnvironmentsAlphabetically,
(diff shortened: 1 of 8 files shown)


packages/hoppscotch-common/src/components/environments/Selector.vue
```diff
@@ -389,6 +392,33 @@ const teamEnvironmentList = useReadonlyStream(
...
+    .map((env, index) => ({
+      env,
+      index,
+    }))
+    .sort((a, b) =>
+      a.env.name
+        .toLocaleUpperCase()
+        .localeCompare(b.env.name.toLocaleUpperCase())
+    )
+})
```
jamesgeorge007, line 407: Can we abstract out the sorting logic as a helper function since it's consumed at multiple places?

jamesgeorge007 (changes requested):
The following issues were encountered while at the team workspace:

- Edits to environment names aren't immediately reflected in the list view.

  https://github.com/user-attachments/assets/55f9f308-4f96-48b6-82b3-38dcf2997238

- Exporting all the environments at the workspace level will result in the export file contents with an array of `null` entries.

  image

jamesgeorge007 (approved):

PR #4284, opened 2024-08-22
fix: persist request execution context between tabs

### LINKED ISSUE #4271: [bug]: cancel button disappear

### Current behavior

When i send a  request  in tab1, the button "send" changes to "cancel ". Before the request finished , i check to tab2, and then i check back to tab1, the "cancel" button disappear

### Steps to reproduce

1. go to tab1 and send a request (you can use a delay api for better affect),  the "send" change to "cancel"
2. before the request finished, check to another tab, the "cancel" change to "send". This is normal.
3. now , go to tab1 , you can find "send" not "cancel" , i think it's wrong.

### Environment

Production

### Version

Local

nivedin:
Closes #4271 HFE-565

This PR fixes the bug where when sending a request in tab1, the button `send` changes to `cancel`. Before the request finishes switching to tab2 and back to tab1, the `cancel` button disappears.

### What's changed
The text now depends on the tab response loading state rather than a component-level loading value. Also, a new field `cancelFunction` is added under the tab document `HoppRESTDocument` to keep a reference to the cancel function to cancel the current request in execution.

### DIFF
--- a/packages/hoppscotch-common/src/components/http/Request.vue
+++ b/packages/hoppscotch-common/src/components/http/Request.vue
@@ -70,9 +70,11 @@
         :title="`${t(
           'action.send'
         )} <kbd>${getSpecialKey()}</kbd><kbd>↩</kbd>`"
-        :label="`${!loading ? t('action.send') : t('action.cancel')}`"
+        :label="`${
+          !isTabResponseLoading ? t('action.send') : t('action.cancel')
+        }`"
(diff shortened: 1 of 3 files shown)


jamesgeorge007 (changes requested):
We'll need to update the functionality as well. Currently, sending a request from a certain tab, switching to another coming back and clicking `Cancel` won't cancel the request and the response will be shown when available.

jamesgeorge007 (approved):

PR #4325, opened 2024-09-03
fix(status): fix import status message

shipko:
### What's changed
I noticed bug in success status message when exporting graphql collection. 

This PR is fixed this problem

- [ ] Not Completed
- [x] Completed

### Notes to reviewers
Снимок экрана 2024-09-03 в 18 24 26


shipko:
Please restart tests

nivedin (changes requested):
the bug was only for gql collection export right, do we need to hard code the message for the rest ?

AndrewBastin (approved):
lgtm, overriding @nivedin's review as this is deemed a better implementation.

PR #4329, opened 2024-09-05
feat(filename): added filename in download functional

### LINKED ISSUE #4265: [feature]: Pre-fill file name and extension when downloading a file.

### Summary

Hi team,

image

When downloading files, the API client currently doesn't pre-fill the name and extension type of the file.

It would greatly improve my user experience if there would be a "send + download" button which would send the payload and download the response immediately with the name and extension of the file pre-filled.

Optionally you could have a look at how other API clients do this.

Basically user experience:

The flow is currently: click send, wait for response, click download, change name, change filetype, save file. 😤

While it could be: click send + download, wait for response, and save file. 😌

shipko:
Closes #4265

### What's changed
Added name for file in all useDownloadResponse methods
- [ ] Not Completed
- [x] Completed

### Notes to reviewers
Maybe we need to move file name in locale files?

### DIFF
--- a/packages/hoppscotch-common/locales/en.json
+++ b/packages/hoppscotch-common/locales/en.json
@@ -383,6 +383,12 @@
     "title": "Export",
     "success": "Successfully exported"
   },
+  "filename": {
+    "cookie_key_value_pairs": "Cookie",
+    "codegen": "{request_name} - code",
+    "lens": "{request_name} - response",
+    "realtime": "Response"
+  },
   "filter": {
     "all": "All",
     "none": "None",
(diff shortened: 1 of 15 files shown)


nivedin (changes requested):
we might need to add the file names in locale

shipko:
@nivedin what do you think. 
Why we need to have an availability to download curl command as file that we importing to request? 

Maybe we remove "Download file" button from packages/hoppscotch-common/src/components/http/ImportCurl.vue?

nivedin:
> @nivedin what do you think. Why we need to have an availability to download curl command as file that we importing to request?
> 
> Maybe we remove "Download file" button from packages/hoppscotch-common/src/components/http/ImportCurl.vue?

Good catch, we can remove that @shipko

shipko:
@nivedin PR is done

nivedin (approved):
LGTM 💯

PR #4331, opened 2024-09-07
feat(request-tab-sync): sync endpoint field with parameter tab

### LINKED ISSUE #4268: [bug]: When adding params to url it doesn't reflected on the Query Parameter

### Current behavior

When adding parameters to the URL it doesn't reflected to the "Query Parameter" or vice versa. This has been working earlier hence  
this is new bug.

image

### Steps to reproduce

1) Go to URL and add URL with some params
2) Check Query Parameters automatically includes the params from the URL
3) Check Hoppscotch sends with the parameters when send

### Environment

Production

### Version

Cloud

shipko:
Closes #4268

### What's changed
Added sync endpoint field with Parameters tab. Data sync bidirectional.

- [ ] Not Completed
- [x] Completed

### Notes to reviewers
I hope that you are make a good manual tests for this feature

### DIFF
--- a/packages/hoppscotch-common/src/components/http/Request.vue
+++ b/packages/hoppscotch-common/src/components/http/Request.vue
@@ -240,7 +240,7 @@ import { useReadonlyStream, useStreamSubscriber } from "@composables/stream"
 import { useToast } from "@composables/toast"
 import { useVModel } from "@vueuse/core"
 import * as E from "fp-ts/Either"
-import { computed, ref, onUnmounted } from "vue"
+import { computed, ref, onUnmounted, watch } from "vue"
 import { defineActionHandler, invokeAction } from "~/helpers/actions"
 import { runMutation } from "~/helpers/backend/GQLClient"
 import { UpdateRequestDocument } from "~/helpers/backend/graphql"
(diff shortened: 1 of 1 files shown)


AndrewBastin:
@shipko Excluding endpoint<->params sync was an intentional design choice we have opted since long back (there should be some mentions of this in Github if you dig through it enough) and we had responded with our recommendations which boil down to:

1. Unlike Postman, we consider the endpoint to be generally unchanging and static. Instead we want to offer ability for people to move things into parameters if needed, Hoppscotch currently has the ability for the parameters in the header to be moved down to the Parameters list (we don't have a way to move it up yet, that is something we should build). So generally the applicable flow will be something like, if you have a parameter that is generally stable and you don't expect it to change, you can put it in the endpoint, else you can move it down to parameters.

Add to Parameters Context Menu in Hoppscotch

2. Another reason why we can't explicitly sync is because of environment variables, since environment substitutions can be anything, we cannot be sure if the environment variable by itself can contain a parameter. This also further complicates things, Postman by itself also has the same issues, for example, try setting the endpoint to `{{coolurl}}` and add a parameter with key "a" and value "b" (Postman will sync the endpoint and make it `{{coolurl}}?a=b`). Now what if I set `coolurl` to `https://echo.hoppscotch.io?weird=value`. We see that Postman messes up the final URL.

Hoppscotch Echo Server output


We can have a conversation around this here if you want, but the PR will be considered closed for the time being, will reopen if there are convincing arguments for the contrary.

shipko:
@AndrewBastin thank you for the detailed explanation. I need to more times for find arguments for my points

PR #4339, opened 2024-09-11
feat: digest authentication added to the request auth types

anwarulislam:
Closes HFE-570

### What's changed
This PR introduces Digest authentication as new request auth types.

The CLI behaviour has been updated so that any failures reported in the pre/post scripts and request execution will result in a non-zero exit code. Previously, this behaviour was exclusive to failed test assertions.

### Note for reviewers

Digest auth works in two stages.

First, it requests for an initial response, where we get necessary information like realm, nonce, algorithm, etc., from the server in the header WWW-Authenticate.

Then, we have to parse information from that header. Finally, we have to generate an authorization header and request with that auth header in the Authorization header property.

#### To test digest auth in CLI
`hopp test src/__tests__/e2e/fixtures/collections/digest-auth-coll.json -e src/__tests__/e2e/fixtures/environments/digest-auth-envs.json`

### DIFF
--- a/packages/hoppscotch-cli/package.json
+++ b/packages/hoppscotch-cli/package.json
@@ -46,6 +46,7 @@
     "chalk": "5.3.0",
     "commander": "12.1.0",
     "isolated-vm": "5.0.1",
+    "js-md5": "0.8.3",
     "lodash-es": "4.17.21",
     "qs": "6.13.0",
     "verzod": "0.2.3",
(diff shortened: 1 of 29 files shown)


jamesgeorge007 (approved):

jamesgeorge007:
In https://github.com/hoppscotch/hoppscotch/pull/4339/commits/431e96b2734ea7210363a62dd6b3893f7f0a18c1, the CLI behaviour has been updated so that any failures reported in the pre/post scripts and request execution will result in a non-zero exit code. Previously, this behaviour was exclusive to failed test assertions.

nivedin (approved):

PR #4344, opened 2024-09-13
fix: preemptly update pwa settings to avoid missing files in the service worker cache

amk-dev:
Recently we had the below issue on our cloud instance, this PR adds the necessery changes required to preemptively fix this. 

**Before**

Whenever a release is out, our webapp seems to get an empty screen. this is happening due to a resource going above the max file size for getting cached and being left out from the precache.

so this is happening,

1. the initial service worker is built, but the oversized file is not included in the service worker cache
2. so when we load the app for the first time, the browser will go and fetch the oversized file from the server, no issues
3. but when a new release is rolled out, the server doesnt have the old oversized file, its hash might have changed, so when the browser tries to go and fetch it, it will get a 404 ( or a redirect to index.html based on the server config ).
4. this happens untill the new release's service worker gets activated, but once its activated, it has the new filename for the oversized file, it fetches it from the server, not a 404. 
5. but on the next release the same cycle continues.

**After**

In our cloud instance, I've increased the max cache size to 10MB using `maximumFileSizeToCacheInBytes` option in workbox. but here i'm not doing that  because it hasn't happened yet. but incase any file hits that limit. the build will fail and we'll know exactly what change causing it, and we can make whatever change is appropriate then.

1. On the newer versions of `vite-plugin-pwa`, it will throw an error instead of a warning if a file goes above the cachable limit. so we'll know before we ship it. so i've updated the `vite-plugin-pwa` to the latest version.

NB: although we have a cached app, and people might have good internet, we should look into factors affecting our bundle size as a lower priority thing.

### DIFF
--- a/packages/hoppscotch-selfhost-web/package.json
+++ b/packages/hoppscotch-selfhost-web/package.json
@@ -75,7 +75,7 @@
     "vite-plugin-fonts": "0.7.0",
     "vite-plugin-html-config": "1.0.11",
     "vite-plugin-inspect": "0.7.42",
-    "vite-plugin-pages": "0.31.0",
+    "vite-plugin-pages": "0.17.3",
     "vite-plugin-pages-sitemap": "1.6.1",
     "vite-plugin-pwa": "0.20.5",
     "vite-plugin-static-copy": "0.17.1",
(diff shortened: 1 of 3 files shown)


packages/hoppscotch-selfhost-web/package.json
```diff
@@ -77,7 +77,7 @@
     "vite-plugin-inspect": "0.7.42",
     "vite-plugin-pages": "0.31.0",
     "vite-plugin-pages-sitemap": "1.6.1",
-    "vite-plugin-pwa": "0.17.3",
+    "vite-plugin-pwa": "0.20.5",
```
jamesgeorge007, line 80: I noticed breaking changes in the intermediary versions; might have to evaluate once and see the safest version that we can bump to.

PR #4363, opened 2024-09-19
chore: make client secret optional across grant types

amk-dev:
**Changes**

* Makes client_secret optional for `clientCredentials` and `password` oauth flows. 
* Some schemas were imported with the latest version of dependent schemas instead of the one compatible with them. This PR fixes those imports to point to the correct schemas.
* Update the collection schema to use the updated auth schemas
* Test suite updates accounting for version bumps at the schema level.


AndrewBastin (changes requested):
QA is failing

PR #4382, opened 2024-09-26
feat: save api response

nivedin:
Closes HFE-534 #1875 

This feature adds the ability to save responses as example for a saved request.

https://github.com/user-attachments/assets/681f7203-ea51-4598-8c27-eb67f7f9fad1

#### TODO

- [x] Add support for Shared Workspaces
- [x] Add Import from Postman, OpenAPI
- [x] Fix bug while editing RAW content


jamesgeorge007 (changes requested):
- While creating a response example with the same name as a pre-existing entry, let's add a new entry rather than overwriting the former.

https://github.com/user-attachments/assets/b740457b-977b-4fe2-9d13-9ecbdf782eb3

- Any actions (`edit`, `delete` & `duplicate`) other than creation aren't persisted for a response eg under a given request when opened under a tab. While actions performed when the tab is closed persist.
-The `New request` modal doesn't permit request creation when the currently active tab has a response example open.
- Response examples open under tabs are dissociated during parent request/collection move/reorder operations (relatively lower priority, can be revisited separately).

  https://github.com/user-attachments/assets/f87bf397-c12b-46ff-b9a2-0e5972d0f3cb

packages/hoppscotch-data/src/rest/v/9.ts
jamesgeorge007, line 1: Since we already have a version bump for `HoppRESTRequest` in #4363, let's incorporate these schema definitions under the same since all of them are going out in the same release.

nivedin: Done ✅

nivedin:
> * While creating a response example with the same name as a pre-existing entry, let's add a new entry rather than overwriting the former. This needs to be extended such that editing the name for a given entry to an existing one is permitted.
> 
> > Discussed separately, we'll keep this flow as is for now and iterate down the line.
> 
>  response-eg-duplicate-overwrites.mov 
> * Any actions (`edit`, `delete` & `duplicate`) other than creation aren't persisted for a response eg under a given request when opened under a tab. While actions performed when the tab is closed persist.
> * The `New request` modal doesn't permit request creation when the currently active tab has a response example open.
> * Response examples open under tabs are dissociated during parent request/collection move/reorder operations (relatively lower priority, can be revisited separately).
>   
>     
>       
>       
>   
>       response-eg-issues.mov

Done ✅

jamesgeorge007 (approved):

Issue #2829, opened 2022-10-27
[bug]: query parameters are escaped implicitly


liudonghua123:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When I send a `Get` request with the following parameters(in bulk edit).

```
token: token_part1 token_part2
attrs: password%3Dabc
```

The actual request sent is `url?token=token_part1+token_part2&attrs=password%253Dabc` which is wrong for my server.

If I send the same `Get` request using postman, the actual request sent is `url?token=token_part1%20token_part2&attrs=password%3Dabc` which is correct for my server.

I found the problem code is around the following code.

https://github.com/hoppscotch/hoppscotch/blob/dc80cc80e6e9c02c41a0754dfa7868f7f102964a/packages/hoppscotch-app/src/helpers/strategies/ExtensionStrategy.ts#L64-L70

```js
url=new URL("http://localhost")
url.searchParams.append('token','token_part1 token_part2')
url.searchParams.append('attrs','password%3Dabc')
console.info(url.toString()) // http://localhost/?token=token_part1+token_part2&attrs=password%253abc

url=new URL('http://localhost/?token=token_part1+token_part2&attrs=password%253abc')
console.info(url.searchParams.get('token')) // token_part1 token_part2
console.info(url.searchParams.get('attrs')) // password%3Dabc
console.info(decodeURI('http://localhost/?token=token_part1+token_part2&attrs=password%253abc')) // http://localhost/?token=token_part1+token_part2&attrs=password%3Dabc


```

See also https://developer.mozilla.org/en-US/docs/Web/API/URL/toString, https://javascript.info/url, https://url.spec.whatwg.org/#URL-stringification-behavior.

### Steps to reproduce

1. Go to `Parameters` tab, click `Bulk edit`.
2. Fill with the following contents
  ```
  token: token_part1 token_part2
  attrs: password%3Dabc
  ```
3. Click `Send` button to make the request.

### Environment

Release

### Version

Cloud

AndrewBastin:
Since the current implementation is correct according to the spec, we are prioritizing this as a low priority issue.

PR contributions for implementing the toggle for considering both ways are welcome in the meantime :)

Rathan-Naik:
Just getting started on contributions @AndrewBastin I can work on this.

AndrewBastin:
Cool @Rathan-Naik, please do let me know if you need any help/guidance.

Assigning the task to you.

[...]

PR #4412, opened 2024-10-07
feat: add the ability to configure query params encoding for requests

Pranay-Pandey:
Closes [#2829](https://github.com/hoppscotch/hoppscotch/issues/2829) 

Add optional setting of encoding the URL parameters in the request. A new setting option is introduced which can be toggled between auto, disable and auto.  
Also have fixed the bug with [preProcessRequest](https://github.com/pranay-pandey/hoppscotch/blob/e3a298003d023ae6f50e14799a4916c9288812ef/packages/hoppscotch-common/src/platform/std/interceptors/browser.ts#L21) function where double encoding was happening.

### What's changed

- New setting section named `General` introduced and moved language, experiments section within
- Encoding setting will decide whether to encode the URL parameters

### Preview

image


nivedin (approved):
@Pranay-Pandey thank you for the contribution, I have made some minor changes to the PR. else it LGTM 💯

jamesgeorge007 (approved):
It'll be good to display a warning (might lead to malformed requests, etc) via the inspector while the encoding preference is set to `Disable` as an enhancement down the line.

nivedin (approved):

PR #4430, opened 2024-10-12
chore(backend): configure CORS for development and production environments

### LINKED ISSUE #4199: [bug]: Optimise CORS Configuration for Development and Production Environments.
https://github.com/hoppscotch/hoppscotch/blob/2917d50c6afe5f52fdcbb284514099d8423eca79/packages/hoppscotch-backend/src/main.ts#L37

the _if_ and _else_ blocks are identical. Both blocks contain the same code to enable CORS with the same settings. This means that regardless of whether the condition in the if statement is true or false, the same CORS configuration will be applied in both cases.

If you intend to have different CORS configurations for development and production environments, you should adjust the CORS settings within each block to reflect those differences. As it stands, the current code will execute the same CORS configuration for both the if and else blocks, which may not be the intended behavior.

**Proposed Solution:**
To address this issue, we should differentiate the CORS settings between development and production environments by adjusting the CORS configuration within each block to reflect the specific requirements for each environment. Additionally, considering that CORS may not be needed in non-production environments, we can optimize the CORS setup accordingly.

ZenMachina16:
Closes #4199

### Introduction
This pull request addresses the issue of identical CORS settings being used in both development and production environments. It refines the logic in the backend to differentiate the CORS configuration based on the environment (development vs. production).

### What's changed
- [x] **Modified CORS settings**: Updated the `if` and `else` blocks in the file `hoppscotch-backend/src/main.ts` to ensure the CORS settings are differentiated for development and production.
  - In development (`if` block), more permissive CORS settings are applied for easier testing.
  - In production (`else` block), restricted CORS settings are applied to enhance security.
- [x] **Code cleanup**: Removed redundant code that caused identical settings to be applied for both environments.

### Notes to reviewers
- The changes maintain backward compatibility by ensuring the development environment retains flexibility while securing the production environment.
- Tested locally to ensure proper CORS behavior for both environments.
- Let me know if additional changes are needed, or if there are any specific configurations you'd like to refine further!

### DIFF
--- a/packages/hoppscotch-backend/src/main.ts
+++ b/packages/hoppscotch-backend/src/main.ts
@@ -41,7 +41,7 @@ async function bootstrap() {
 
   const configService = app.get(ConfigService);
 
-  console.log(`Running in production:  ${configService.get('PRODUCTION')}`);
+  console.log(`Running in production: ${configService.get('PRODUCTION')}`);
   console.log(`Port: ${configService.get('PORT')}`);
 
   checkEnvironmentAuthProvider(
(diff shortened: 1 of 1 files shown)


packages/hoppscotch-backend/src/main.ts
```diff
@@ -63,21 +63,20 @@ async function bootstrap() {
...
-      credentials: true,
-    });
-  } else {
-    console.log('Enabling CORS with production settings');
+if (configService.get('PRODUCTION') === 'false') { 
+  console.log('Enabling CORS with development settings'); 
+  app.enableCors({
+    origin: configService.get('DEV_WHITELISTED_ORIGINS').split(','), 
+    credentials: true,
+  });
```
mirarifhasan, line 71: @ZenMachina16 
For `development` mode, allow all the origins.

packages/hoppscotch-backend/src/main.ts
```diff
@@ -63,21 +63,20 @@ async function bootstrap() {
...
+if (configService.get('PRODUCTION') === 'false') { 
+  console.log('Enabling CORS with development settings'); 
+  app.enableCors({
+    origin: configService.get('DEV_WHITELISTED_ORIGINS').split(','), 
+    credentials: true,
+  }); 
+} else { 
+  console.log('Enabling CORS with production settings'); 
+  app.enableCors({
+    origin: configService.get('PROD_WHITELISTED_ORIGINS'),
```
mirarifhasan, line 75: Keep only one env variable `WHITELISTED_ORIGINS`.

mirarifhasan (changes requested):

mirarifhasan (approved):
Update a minor changes, LGTM

jamesgeorge007 (approved):

PR #4431, opened 2024-10-12
feat: show login dialogue for features behind login

amk-dev:
Fixes HP-306 HFE-616.

**Changes**

Previously, we hide the features that required login. This PR shows the features, but when the user tries to trigger it, it will open the login dialogue.

This PR made this change to these features

* Auto rename requests
* Request body generation
* Import from another workspace


jamesgeorge007 (changes requested):
There's an edge case with the `Import from Another Workspace` action where logging in via the above action and proceeding to choose a workspace to import from only lists `Personal` or will show an infinite spinner until initiating a page reload or expanding the workspace selector to fetch the list of workspaces.

This can also be reproduced in the pre-existing implementation by choosing the `Import from Another Workspace` action immediately upon logging in.

https://github.com/user-attachments/assets/4c814d1b-f6cc-44b8-b63c-c42b9da976c1

jamesgeorge007 (approved):

PR #4438, opened 2024-10-15
JSON files updated

### LINKED ISSUE #4349: [feature]: slow render on some screens

### Summary

I noticed that on some screens the rendering time increases linearly with the number of elements. For example, on the edit environment variables page. 

With a number of elements of about 100, the page generation time takes more than 2 seconds (macbook m1), even longer on older machines. 

With the help of the profiler, it was found out that the problem is too "heavy" SmartEnvInput. 

I ask you to optimize somehow, because this is important for us.

[test.json](https://github.com/user-attachments/files/17009709/test.json)

The product is slow when used in full on large amounts of data. This prevents implementation in large companies

HetuKariya:
Hey @jamesgeorge007 thanks for assigning me this issue.

 - Fixed #4349 

Please merge this PR. Also please add hacktoberfest-accepted tag. 

### What's changed

I am removing nesting from all your json files as this will help in increasing the rendering speed and also remove unnecessary cluster from the code thus making trying to clean the code and removing the slow rendering on screens.


shipko:
How did this changes speed up render on some screen? 
Can you provide a some metrics before/after?

HetuKariya:
I thought that the language section and all the other json files are nested so I tried to remove all the nesting and as I changed 31 major files by removing 200 lines of extra nested code so I feel that it should improve the rendering speed.

AndrewBastin:
Rejected as spam. This PR doesn't work and doesn't bring anything to the table that it claims to as well.

PR #4445, opened 2024-10-17
HSB-494 fix: downgrade eslint related package

mirarifhasan:
Closes HSB-494

### What's changed

This PR downgraded Eslint related npm packages  to `v8` version in `hoppscotch-backend` which ware previously upgraded to `v9` which introduces braking changes.

### Notes to reviewers

Nil

### DIFF
--- a/packages/hoppscotch-backend/package.json
+++ b/packages/hoppscotch-backend/package.json
@@ -95,7 +95,7 @@
     "@typescript-eslint/parser": "8.8.1",
     "cross-env": "7.0.3",
     "eslint": "8.57.0",
-    "eslint-config-prettier": "8.10.0",
+    "eslint-config-prettier": "9.1.0",
     "eslint-plugin-prettier": "5.2.1",
     "jest": "29.7.0",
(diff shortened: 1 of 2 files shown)


packages/hoppscotch-backend/package.json
```diff
@@ -94,8 +94,8 @@
     "@typescript-eslint/eslint-plugin": "8.8.1",
     "@typescript-eslint/parser": "8.8.1",
     "cross-env": "7.0.3",
-    "eslint": "9.12.0",
-    "eslint-config-prettier": "9.1.0",
+    "eslint": "8.57.0",
+    "eslint-config-prettier": "8.10.0",
```
jamesgeorge007, line 98: I think we can still be on `eslint-config-prettier` `v9` being backwards compatible with ESLint `v8`, `peerDependencies` refers to ESLint v7+, [ref](https://github.com/prettier/eslint-config-prettier/blob/main/package.json#L16-L18).

jamesgeorge007 (approved):

PR #4447, opened 2024-10-17
fix: i18n strings not falling back correctly

AndrewBastin:
This PR intends to fix an issue with the i18n system not falling back correctly to the fallback language (English) translations.
This results in the value printed out being the actual raw i18n tags. This is not visible when the language is changed from English, but once we are in a non-English language and we do a reload, the `en` language messages are not present and this causes the fields to be present there.

Current State (on reload when German is set as the language):
image

Notice the `count.key` field.

### What's changed
- Add code to load in fallback langauge regardless of set locale.


jamesgeorge007 (approved):
Lgtm. Should we extend it to the Admin dashboard?

AndrewBastin:
probably, I will have a look into that

PR #4464, opened 2024-10-22
chore: environment highlight colour schema update

### LINKED ISSUE #4389: [bug]: Poor contrast of highlighted text in request body

### Current behavior

Using Hoppscotch 24.8.3 on macOS, variables in the request body result in highlighted text that is difficult to read in dark mode. Perhaps making the text white inside the highlighted portion would provide better contrast.
Screenshot 2024-09-30 at 9 54 07 AM

### Steps to reproduce

1. Create a request and enter any variable name enclosed in angle brackets (<< >>) inside a JSON request body
2. The red or green highlighted text is difficult to make out.

### Environment

Production

### Version

Local

nivedin:
Closes HFE-605 #4389 

### Before
image

### After
image

This PR improves the colors on variables for multiple themes and improve contrast on codemirror.

### What's changed
The text colour inside the env variable have been set to white and the syntax highlight has been removed from the env variable for commented line.


jamesgeorge007 (changes requested):
- Variables referenced in comments spanning multiple lines are not greyed out.

  image

- Comments within a string are considered valid, but a variable getting referenced appears greyed out. This is a very nice use case and a relatively lower priority.

  image

We could revisit these in iterations as required based on evaluating the effort.

nivedin:
The HoppComments plugin has been remove since the codemirror `MatchDecorator` cannot match multi line decorator using regex, [ref](https://discuss.codemirror.net/t/is-regex-multiline-supported/600/2).

So a different approach is used here, the default comment class of codemirror has been changed to `cm-comment` so that through the css parent selecter the same style can be achieved.
Both the issues have been fixed, ptal @jamesgeorge007 

image

jamesgeorge007 (approved):

PR #4466, opened 2024-10-22
feat: support for binary body

amk-dev:
This PR adds support for sending binary data as body.

1. since the binary file is not exported, In `hoppscotch-cli`, added some changes to not error the request out, like how multipart/form-data is behaving right now.

2. hoppscotch/hoppscotch-extension#307 adds support for binary files in our browser extension.

### DIFF
--- a/packages/hoppscotch-cli/src/interfaces/request.ts
+++ b/packages/hoppscotch-cli/src/interfaces/request.ts
@@ -43,5 +43,5 @@ export interface EffectiveHoppRESTRequest extends HoppRESTRequest {
     active: boolean;
     description: string;
   }[];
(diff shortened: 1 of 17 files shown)


jamesgeorge007 (changes requested):
There's an exception while setting `Agent` as the interceptor. For now, we've decided to show a warning via the inspector regarding the support.

image

The following issues can be revisited:

- Say, there's a request open with `application/octet-stream` as the content type and a file chosen. The file is lost while navigating to another tab and returning.
- Save the above request in a collection, export and re-import. Open the request and observe the tab appear in the dirty state even though no updates happened in between.

jamesgeorge007 (approved):

PR #4484, opened 2024-10-27
Change User Accounts Simple Feature

### LINKED ISSUE #4322: [feature]: Multi-account switcher

### Summary

Be able to switch accounts without logging out

Having to log out any time I have to switch accounts and then log back in is frustrating. But it is manageable, probably worth keeping this is low priority unless others request it.

Olly4k:
Closes #4322 

This PR provides a small, frontend based solution to the account switching feature mentioned in the linked issue. It is a quality of life change that prevents the need to logout of an account only to just log straight back in with another.  It does this by placing a Change Accounts button above the logout button in the profile dropdown menu. This opens a variation on the login modal that can be used to login to a different account.

### What's changed

- Added Switch.vue file in hoppscotch-common's firebase resources next to both Login.vue and logout.vue. This contains the majority of code involved in rendering the button and the custom login modal. It has purposefully been separated from the login.vue file's component to allow for further modification if needed however it also can be replaced by the login modal if needed.
- Added the FirebaseSwitch button component in components.d.ts
- Used the FirebaseSwitch component in the Header.vue file's profile dropdown menu
- Added a new label tag in all locales called auth.change_accounts that says "Change Accounts" in all languages

Screenshot 2024-10-27 at 10 05 28 pm

Screenshot 2024-10-27 at 10 04 45 pm

### Notes to reviewers

This is targeted as a temporary solution to the problem without reworking the backend. Since the issue is low priority I believe this fix will be enough for now with the option for improvements in the future.

### DIFF
--- a/packages/hoppscotch-common/locales/tr.json
+++ b/packages/hoppscotch-common/locales/tr.json
@@ -117,6 +117,7 @@
     "login": "Giriş",
     "login_success": "Başarıyla giriş yapıldı",
     "login_to_hoppscotch": "Hoppscotch'a giriş yapın",
+    "change_accounts": "hesapları değiştir",
     "logout": "Çıkış",
     "re_enter_email": "E-mail adresinizi yeniden girin",
     "send_magic_link": "Sihirli bir bağlantı gönder",
--- a/packages/hoppscotch-common/locales/af.json
+++ b/packages/hoppscotch-common/locales/af.json
@@ -117,6 +117,7 @@
     "login": "Teken aan",
     "login_success": "Suksesvol aangemeld",
     "login_to_hoppscotch": "Teken in op Hoppscotch",
+    "change_accounts": "rekeninge te verander",
     "logout": "Teken uit",
     "re_enter_email": "Voer die e-posadres weer in",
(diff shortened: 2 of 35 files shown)


xLansang (commented):
Great implementation of the change user accounts feature! Showed a great use of GitHub workflow as well.

Loznerol (commented):
Switching feature looks good!

sz-2302:
Great implementation of the account switching feature with flexible component separation!

VTrngNghia:
I know we all want a proper Account Switch, but it's great to have a simple workable solution. 
Kudos to @Olly4k!
How can I vote on this?

packages/hoppscotch-common/locales/tr.json
```diff
@@ -117,6 +117,7 @@
     "login": "Giriş",
     "login_success": "Başarıyla giriş yapıldı",
     "login_to_hoppscotch": "Hoppscotch'a giriş yapın",
+    "change_accounts": "hesapları değiştir",
```
hakanyalitekin, line 120: shuld be "Hesapları değiştir"

PR #4489, opened 2024-10-28
feat: import summaries

amk-dev:
This PR adds descriptions for importers and provides users with a summary of their imported items upon completion.

Closes HFE-617.

image

image


jamesgeorge007 (changes requested):
As discussed, unsupported entries while importing can be marked with an info icon in the amber shade, displaying the message `... are not supported at the moment`.
Any entry not supported by the external platform can be removed altogether.

jamesgeorge007 (changes requested):
There is an exception arising while importing. Also, please look into resolving the conflicts.

image

jamesgeorge007 (approved):

PR #4508, opened 2024-11-03
HSB-497 chore: user deletetion if not solo team owner

mirarifhasan:
Closes HSB-497

### What's changed

This PR updated the user deletion. Previously if a user a team owner, that user can not be deleted.
Now, a user can be deleted if he is not solo team owner in any of teams.

### Notes to reviewers

3 ways a user can be deleted.
1. User can delete himself
2. User can delete by admin
3. User can delete via API

### DIFF
--- a/packages/hoppscotch-backend/src/team/team.service.ts
+++ b/packages/hoppscotch-backend/src/team/team.service.ts
@@ -38,7 +38,7 @@ export class TeamService implements UserDataHandler, OnModuleInit {
 
   canAllowUserDeletion(user: AuthUser): TO.TaskOption<string> {
     return pipe(
-      this.isUserOwnerRoleInTeams(user.uid),
+      this.isUserSoleOwnerInAnyTeam(user.uid),
       TO.fromTask,
       TO.chain((isOwner) => (isOwner ? TO.some(USER_IS_OWNER) : TO.none)),
     );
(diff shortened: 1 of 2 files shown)


packages/hoppscotch-backend/src/team/team.service.ts
```diff
@@ -396,18 +396,34 @@ export class TeamService implements UserDataHandler, OnModuleInit {
...
+      const ownerCounts = await Promise.all(
+        userOwnedTeams.map((team) =>
+          this.prisma.teamMember.count({
+            where: {
+              teamID: team.teamID,
+              role: TeamMemberRole.OWNER,
+            },
+          }),
+        ),
+      );
```
AndrewBastin, line 422: We can stop if we get a single match

AndrewBastin (changes requested):

packages/hoppscotch-backend/src/team/team.service.ts
mirarifhasan: Updated

AndrewBastin (approved):

RaHehl:
Would it be possible to control whether the workspace is deleted automatically when the last owner is removed, instead of blocking the deletion, through a configuration variable? This could make the behavior more flexible depending on the use case.

RaHehl:
Or perhaps even add a `force_delete` option for both the API and GUI to handle automatic workspace deletion when the last owner is removed?

PR #4511, opened 2024-11-04
fix(agent): propagate default values to persistent store

CuriousCorrelation:
This PR fixes an issue where sane default values set in the ephemeral store were not being propagated to the persistent store if the store was uninitialized or encountered a loading error. Now, if the persistent store is absent or fails to load, default values will be established in the ephemeral store and subsequently saved to the persistent store, syncing two both storage layers.

Closes HFE-645

### DIFF
--- a/packages/hoppscotch-agent/src-tauri/src/state.rs
+++ b/packages/hoppscotch-agent/src-tauri/src/state.rs
@@ -46,6 +46,10 @@ impl AppState {
             .and_then(|val| serde_json::from_value(val.clone()).ok())
             .unwrap_or_else(|| DashMap::new());
 
+        // Try to save the latest registrations list
+        let _ = store.set("registrations", serde_json::to_value(&registrations)?);
+        let _ = store.save();
+
         Ok(Self {
             active_registration_code: RwLock::new(None),
             cancellation_tokens: DashMap::new(),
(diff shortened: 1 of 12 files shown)


packages/hoppscotch-agent/src-tauri/src/state.rs
RaHehl, line 1: I'm not a Rust expert, but would replacing

let _ = store
    .delete("registrations")
    .then_some(())
    .ok_or(AppError::RegistrationClearError)?;

with

if store.get("registrations").is_some() {
    store.delete("registrations").map(|_| ()).ok_or(AppError::RegistrationClearError)?;
}

be a more robust fix?

CuriousCorrelation: Thanks for the suggestions! It's actually a good catch and indeed part of a refactor I’m currently planning - essentially to have a more robust logging/tracing, error system and better error propagation.

The reason why I didn’t include that into the current fix was because the primary issue was uninitialized store since while the app assumed empty default store with https://github.com/hoppscotch/hoppscotch/blob/c75b22d3fe68e5482e58a24cbc2d9526be8d113a/packages/hoppscotch-agent/src-tauri/src/lib.rs#L106 it didn’t translate that assumption to the disk.

We can definitely include this fix into the PR as well, we’d basically use a match statement or even better - `?` error propagation and then propagate that upwards from `tray.rs` + some tracing for the stacktrace.

AndrewBastin (approved):

PR #4516, opened 2024-11-06
fix: clicking on run query without connecting connects without docs polling

AndrewBastin:
This PR introduces a fix for an issue where when running GraphQL requests without clicking on query, the connect button changes into `Disconnect` without doing a formal connection and pulling docs and starting polling.

### What's changed
- GQL `connect` function is now async and waits for the first poll to be done before resolving
- `runGQLOperation` calls and waits on `connect` before executing the operation.


anwarulislam (approved):
LGTM 🚀

jamesgeorge007 (approved):

PR #4523, opened 2024-11-07
hotfix: graphql connection error handler

anwarulislam:
This pull request fixes the error handling for GraphQL connections. It adds a new error message for HTTP errors and updates the connection state to "ERROR" when an HTTP error occurs.

### DIFF
--- a/packages/hoppscotch-common/locales/en.json
+++ b/packages/hoppscotch-common/locales/en.json
@@ -452,6 +452,7 @@
   },
   "graphql": {
     "connection_switch_confirm": "Do you want to connect with the latest GraphQL endpoint?",
+    "connection_error_http": "HTTP Error",
     "connection_switch_new_url": "Switching to a tab will disconnected you from the active GraphQL connection. New connection URL is",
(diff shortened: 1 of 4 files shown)


jamesgeorge007 (changes requested):
If the introspection attempt fails on the active endpoint with a non-Agent (given Agent is registered) interceptor, the `Connect` button goes to an infinite loading state.

jamesgeorge007 (changes requested):
- The `Browser extension` is not installed and set as the active interceptor, the corresponding error state isn't displayed in the response section.

  image

- For the error toast displayed when the introspection attempt fails while attempting to connect, we'll have to exclude it for the case when `Browser extension` is the active interceptor where the message displayed says `Heading not found`. Also, since a connection attempt is performed while sending a request, currently it results in an error toast due to the introspection query failing following a success toast saying the query succeeded in the given time frame and might not be necessary in such a case.

jamesgeorge007 (approved):

AndrewBastin (approved):

PR #4527, opened 2024-11-11
feat(agent): add registration list management and ux

CuriousCorrelation:
This PR implements a registration management system for the Hoppscotch Agent, allowing users to view and manage their registered clients with some UX improvements.

Closes HFE-649

- New registration management UI in Agent settings
- Automatic registration list refresh after operations
- Prevent registration interference between sessions
- Fixed registration interference issues by clearing active registration state
- Improved window management with show/hide instead of recreate/destroy

image


AndrewBastin (changes requested):
There are a couple of changes required, we will defer this from the November release.

1. Let's move the Registrations list to the Agent itself instead of the app, I think that is a more sensible experience.
2. Instead of the registrations modal on the app settings let's have it show info regarding the current registration the app is registered with. The idea is that the user can look into and see if the registration IDs match to find if there is any funny business going on.

Issue #3412, opened 2023-10-02
[bug]: Test header value doesnt work from cli


mnejjargithub:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When i execute collection with this test script `pw.expect(pw.response.headers.find(header => header.key === 'content-type').value).toBe(\"application/json;charset=utf-8\");` from cli doesnt work, but is working when i use the web interface

 reported errors:
  TEST_SCRIPT_ERROR Script evaluation failed: [object Object]

### Steps to reproduce

From a collection with a sample test api
Get the header in response
Modify the collection with the header to be tested, like content-type = application/json
execute the collection from cli :  hopp test -e myenv.json collection.json

### Environment

Release

### Version

Local

pjotal:
have you tried this instead?
`pw.expect(pw.response.headers.find(header => header.key === 'content-type').value).toBe("application/json;charset=utf-8");`
or
`pw.expect(pw.response.headers.find(header => header.key === 'content-type').value).toBe('\"application/json;charset=utf-8\"');`

looks like you're passing the `.toBe()` argument as an object instead of a string (which is actually what you want to be verified. not sure if this will fix your issue but it's a possible cause.

PR #4540, opened 2024-11-15
Enhance HTTP Proxy: Add Bypass Domain Support & Allow Response Header testing in CLI

### LINKED ISSUE #4539: [feature]: Feature Request: Add Bypass Proxy Functionality for HTTP Proxy

### Summary

This field should allow users to specify domains that should bypass the proxy server when making requests.

This feature mirrors existing functionality in Postman and is crucial for accessing resources that are incompatible with or blocked by the proxy.

shobanrajm26:
Closes #4539  #3412 

This PR enhances the Hoppscotch HTTP proxy functionality by adding a "Proxy Bypass" field. This field allows users to specify domains (e.g., example.com, localhost) that should bypass the configured proxy server. 

Also, solves an issue preventing response header testing in the CLI

### What's changed

The implementation checks if the request's hostname matches any of the specified bypass domains. If a match is found, the request is sent directly without using the proxy.

The response header format has been adjusted to a compatible array structure, allowing tests to correctly validate response headers.

image


jamesgeorge007 (commented):
This looks overtaken by later work: the header normalisation is a no-op after #5388 and #5097 (both merged), and the proxy-bypass files it touches were removed in #5421 (also merged). Suggest closing this one; #4539 and #3412 can stay open. Thanks for your interest in contributing 🙌

PR #4546, opened 2024-11-18
Draft: sync collection, requests to server if data is unsynced

### LINKED ISSUE #4436: [feature]: sync strategy for user data when user log in

### Summary

1. When I work with a local collection and I have logged in, my state is synchronized with the cloud. 
2. After logging out, I continue to work with my personal collection, my data is not synchronized with the server. 
3. After a while, I log in as a user, my current changes made in step 2 **were erased and lost**.

The problem lies in this line of the file https://github.com/hoppscotch/hoppscotch/blob/main/packages/hoppscotch-selfhost-web/src/platform/collections/collections.platform.ts#L73-L82 and https://github.com/hoppscotch/hoppscotch/blob/main/packages/hoppscotch-selfhost-desktop/src/platform/collections/collections.platform.ts#L73-L82

After studying the code, it became clear that priority is given to data from the cloud. User data is simply deleted in this case. This is wrong and may alienate QA engineers due to the risk of losing their own data.

I suggest developing a more correct data synchronization strategy.

shipko:
Closes #4436

### What's changed
This change improves the functionality of synchronizing user data for REST collections and queries. 
Now data that is stored only locally (without the ID parameter) will be sent to the server if:
1) If the user worked locally without logging in to the account
2) If the user did not have Internet access.
- [ ] Not Completed
- [x] Completed

### Notes to reviewers
This is the first stage of the functionality implementation. I want to get feedback from you

### DIFF
--- a/packages/hoppscotch-common/src/components/collections/index.vue
+++ b/packages/hoppscotch-common/src/components/collections/index.vue
@@ -841,6 +841,9 @@ const onAddRequest = (requestName: string) => {
     name: requestName,
   }
 
+  // У новых объектов не должно быть ID
+  delete newRequest.id
+
   const path = editingFolderPath.value
   if (!path) return
   if (collectionsType.value.type === "my-collections") {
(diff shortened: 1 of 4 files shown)


shipko:
@AndrewBastin did you take part in the discussion of this topic, please review it

PR #4549, opened 2024-11-18
HSB-500 fix: returning introspection complexity to 0

mirarifhasan:
Closes HSB-500

### What's changed

This fix PR handles the breaking changes of [graphql-query-complexity](https://github.com/slicknode/graphql-query-complexity/releases/tag/v1.0.0).

For the introspection query, now returning 0 value as mentioned in the release note.

### Notes to reviewers

Nil

### DIFF
--- a/packages/hoppscotch-backend/src/plugins/GQLComplexityPlugin.ts
+++ b/packages/hoppscotch-backend/src/plugins/GQLComplexityPlugin.ts
@@ -7,6 +7,7 @@ import {
 import { Plugin } from '@nestjs/apollo';
 import { GraphQLError } from 'graphql';
 import {
+  ComplexityEstimatorArgs,
   fieldExtensionsEstimator,
   getComplexity,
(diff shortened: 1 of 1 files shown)


jamesgeorge007 (dismissed):

packages/hoppscotch-backend/src/plugins/GQLComplexityPlugin.ts
```diff
@@ -23,12 +24,17 @@ export class GQLComplexityPlugin implements ApolloServerPlugin {
 
     return {
       async didResolveOperation({ request, document }) {
+        const isIntrospectionQuery =
+          request.operationName === 'IntrospectionQuery';
```
jamesgeorge007, line 28: As observed, `request.operationName` was coming up as `undefined`. Hence, this approach won't fit in.

jamesgeorge007 (approved):

PR #4550, opened 2024-11-19
feat: add individual content types for formadata

amk-dev:
**Description**

This PR adds the ability to specify multiple content types for individual fields when using `multipart/formdata` content type.

### DIFF
--- a/packages/hoppscotch-cli/src/utils/mutators.ts
+++ b/packages/hoppscotch-cli/src/utils/mutators.ts
@@ -52,7 +52,21 @@ const getValidRequests = (
 export const toFormData = (values: FormDataEntry[]) => {
   const formData = new FormData();
 
-  values.forEach(({ key, value }) => formData.append(key, value));
+  values.forEach(({ key, value, contentType }) => {
+    if (contentType) {
+      formData.append(
+        key,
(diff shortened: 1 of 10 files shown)


packages/hoppscotch-cli/src/utils/mutators.ts
```diff
@@ -52,7 +52,21 @@ const getValidRequests = (
...
+          type: contentType,
+        }),
+        key
+      );
+
+      return;
+    }
+
+    formData.append(key, value);
+  });
```
jamesgeorge007, line 69: Currently, specifying a content type of `text/plain` for an entry is still considered a file being wrapped in `Blob`. Try sending a request to https://httpbin.org/post with an entry in the above format and observe the entry listed under the key, `files` and not `form`. Could you investigate this?

jamesgeorge007: Discussed separately, this is the intended behaviour while specifying custom content types.

jamesgeorge007 (commented):
Could you update the code generation flow to account for the content types specified for individual key-value pairs?

jamesgeorge007 (approved):

PR #4594, opened 2024-12-05
HSB-505 feat: user history disabling

mirarifhasan:
Closes HSB-505

### What's changed

This backend PR adds the ability to turn off `User History` feature from the Admin Dashboard.
Additionally, admin can delete all user's history with one click.

Queries/Mutations for Admin Dashboard
```gql
# Enable/Disable user history feature
mutation {
  toggleUserHistoryStore(status: DISABLE) # This mutation has a subscription for App
}
query {
  infraConfigs(configNames:[USER_HISTORY_STORE_ENABLED]){
    name
    value
  }
}

# Delete all user history
mutation {
  revokeAllUserHistoryByAdmin # This mutation has a subscription for App
}
```
Queries/Subscriptions for App
```gql
query {
  isUserHistoryEnabled {
    name
    value
  }
}
subscription {
  infraConfigUpdate(configName:USER_HISTORY_STORE_ENABLED)
}

subscription {
  userHistoryAllDeleted
}
```

### Notes to reviewers

Nil


packages/hoppscotch-backend/src/user-history/user-history.service.ts
```diff
@@ -188,6 +188,16 @@ export class UserHistoryService {
     return E.right(deletionInfo);
   }
 
+  /**
+   * Delete all user history from DB
+   * @returns a boolean
+   */
+  async deleteAllHistories() {
+    await this.prisma.userHistory.deleteMany();
```
ankitsridhar16, line 196: this should be inside a try catch block, what if there is an issue with database connection ?

amk-dev (commented):
shouldnt we guard the create history endpoint based on the admin setting ? 

- right now, even if the user history is disabled, if i sent a request from the FE it will go through

ankitsridhar16:
> shouldnt we guard the create history endpoint based on the admin setting ?
> 
> * right now, even if the user history is disabled, if i sent a request from the FE it will go through

yes nice catch, we should read the current state and block operations

packages/hoppscotch-backend/src/user-history/user-history-feature-flag.guard.ts
```diff
@@ -0,0 +1,20 @@
+import { CanActivate, ExecutionContext, Injectable } from '@nestjs/common';
+import { USER_HISTORY_FEATURE_FLAG_DISABLED } from 'src/errors';
+import { InfraConfigService } from 'src/infra-config/infra-config.service';
+import { throwErr } from 'src/utils';
+import * as E from 'fp-ts/Either';
+
+@Injectable()
+export class UserHistoryFeatureFlagGuard implements CanActivate {
```
ankitsridhar16, line 8: is a guard required for this specific use case? Can't we just handle it in sthe ervice method?

mirarifhasan: Since we are enabling or disabling entire features (e.g., `createHistory`, `starringHistory`, `deleteHistory`, `getHistory`, `subscription`, etc.), implementing a **guard** can simplify code maintenance and cleaner. This approach avoids the repetition of `if` conditions in every service function, centralizing the logic for feature toggling.

jamesgeorge007 (approved):

PR #4596, opened 2024-12-06
chore: improve health check script

progprnv:
### What's changed
- Added a more flexible exit strategy in the `curlCheck` function.
- Changed the script to return status codes instead of directly exiting in `curlCheck`. Now, the main script decides whether to exit based on those returned statuses.
- Updated URL checks to gracefully handle failure without prematurely stopping the entire script.

### Notes to reviewers
- The `curlCheck` function now uses `return 1` instead of `exit 1`, allowing better control over script flow.
- This change helps in managing multiple URL checks, and the script will only exit if the individual check fails (not all at once).

Improved code:
curlCheck() {
  if ! curl -s --head "$1" | head -n 1 | grep -q "HTTP/1.[01] [23].."; then
    echo "URL request failed!"
    return 1
  else
    echo "URL request succeeded!"
    return 0
  fi
}

if [ "$ENABLE_SUBPATH_BASED_ACCESS" = "true" ]; then
  curlCheck "http://localhost:80/backend/ping" || exit 1
else
  curlCheck "http://localhost:3000" || exit 1
  curlCheck "http://localhost:3100" || exit 1
  curlCheck "http://localhost:3170/ping" || exit 1
fi

### DIFF
--- a/healthcheck.sh
+++ b/healthcheck.sh
@@ -1,18 +1,17 @@
-#!/bin/sh
-
 curlCheck() {
   if ! curl -s --head "$1" | head -n 1 | grep -q "HTTP/1.[01] [23].."; then
     echo "URL request failed!"
-    exit 1
+    return 1
   else
     echo "URL request succeeded!"
+    return 0
   fi
 }
 
 if [ "$ENABLE_SUBPATH_BASED_ACCESS" = "true" ]; then
-  curlCheck "http://localhost:80/backend/ping"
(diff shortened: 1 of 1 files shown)


balub (commented):
LGTM 🚀

progprnv:
Fine then nbd

mirarifhasan (approved):
LGTM 💯

PR #4607, opened 2024-12-11
feat: disable tracking request history

amk-dev:
FIXS HFE-677,HFE-678

This PR introduces 
- the ability to toggle storing sent request history from admin panel
- platform defs & UI changes to support this toggling on the app

### DIFF
--- a/packages/hoppscotch-selfhost-web/src/api/queries/GetUserHistoryStore.graphql
+++ b/packages/hoppscotch-selfhost-web/src/api/queries/GetUserHistoryStore.graphql
@@ -1,6 +1,0 @@
-query GetUserHistoryStore {
-  infraConfigs(configNames: [USER_HISTORY_STORE_ENABLED]) {
-    name
-    value
-  }
-}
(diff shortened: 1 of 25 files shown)


packages/hoppscotch-selfhost-web/src/api/queries/GetUserHistoryStore.graphql
jamesgeorge007, line 1: Fetching history preference while logged in as a non-Admin user would fail currently since the `infraConfigs` GQL query is reserved for use in the Admin Dashboard. The intended query to be triggered from the app here is `isUserHistoryEnabled`.

jamesgeorge007 (changes requested):
To align with the BE behaviour, let's restrict the clear history action if history is disabled. Spotlight is also in scope in this case.

image
image

Also, if possible, please look into extending the platform definitions to `selfhost-desktop`.

jamesgeorge007 (approved):

jamesgeorge007 (approved):

PR #4618, opened 2024-12-16
chore: update collection runner modal UI

nivedin:
Closes HFE-669

### What's changed
This PR updates the collection runner modal UI
- Remove the padding and add a full width
- Move wiki button in the action slot of the tabs and add link w.r.t the tab state.

### DIFF
--- a/packages/hoppscotch-common/src/components/http/test/RunnerModal.vue
+++ b/packages/hoppscotch-common/src/components/http/test/RunnerModal.vue
@@ -2,6 +2,7 @@
   <HoppSmartModal
     dialog
     :title="t('collection_runner.run_collection')"
+    :full-width-body="true"
     @close="closeModal"
   >
     <template #body>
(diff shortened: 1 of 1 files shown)


packages/hoppscotch-common/src/components/http/test/RunnerModal.vue
```diff
@@ -129,6 +122,16 @@
             </div>
           </div>
         </HoppSmartTab>
+        <template #actions>
+          <HoppButtonSecondary
+            v-tippy="{ theme: 'tooltip' }"
+            class="!py-0 pl-2"
+            :to="`https://docs.hoppscotch.io/documentation/${runnerLink}`"
```
liyasthomas, line 129: No need to contact strings here. You could just pass the entire string from `runnerLink`.

packages/hoppscotch-common/src/components/http/test/RunnerModal.vue
```diff
@@ -216,11 +219,17 @@ const emit = defineEmits<{
 }>()
 
 const includeEnvironmentID = ref(false)
-const activeTab = ref("test-runner")
+const activeTab = ref<"test-runner" | "cli">("test-runner")
```
liyasthomas, line 222: Let's just call `gui` and `cli` instead of `test-runner` and `cli`

liyasthomas (changes requested):
Add my few comments.

jamesgeorge007 (approved):

liyasthomas (approved):

PR #4622, opened 2024-12-17
Hoppscotch Test Enhancements: Numeric & Response time Assertions

### LINKED ISSUE #4620: [feature]: Hoppscotch Test Enhancements: Numeric & Response time Assertions

### Summary

This enhancement request is to introducs new assertion functions for numeric value i.e. toBeGreaterThan, toBeLessThan, toBeGreaterThanOrEqual &toBeLessThanOrEqual to Hoppscotch's API testing framework, along with the ability to assert against response duration.

Current Hoppscotch testing lacks the granularity to easily perform numerical comparisons and response time validations. These additions are crucial for robust API testing, allowing developers to verify not just the correctness of data but also the performance of their APIs.

praveenk-78:
Closes #4620

### What's changed
Implemented test methods for the following

- toBeGreaterThan
- toBeLessThan
- toBeGreaterThanOrEqual
- toBeLessThanOrEqual
- 
Also, added assertion for response time.


ddierickx:
Thanks a lot @praveenk-78 , this seems like a nice addition, any idea when this would be merged and released? 🙌

praveenk-78:
@liyasthomas / @jamesgeorge007 
can you share your thoughts around @ddierickx  question

jamesgeorge007:
Hi, this aligns well with some broader updates we have planned for the scripting system proposed in #5221. We will assess the feasibility of integrating the proposed changes into the efforts planned and plan to merge them when the timing aligns with the broader implementation schedule.

jamesgeorge007 (changes requested):
- P1 - `pw.response.duration` is undefined under `hoppscotch test`; the CLI/faraday path exposes `responseTime`, not `duration`, so the duration assertions silently no-op and still report green. Switching to `responseTime` should fix it.

- P2 - the comparison matchers also pass silently when given a non-number; a guard or throw would surface bad input instead of a false pass.

- P3 - after the rebase, the new matchers need to be registered in the faraday path too.

PR #4624, opened 2024-12-17
fix(sh-admin): resolve record retrieval limits in pending invites section by implementing pagination

JoelJacobStephen:
### Ticket
- Closes HFE-694

### Description
This PR resolves an issue where the Pending Invites page under the Users section in the Admin Dashboard was only fetching 10 records. The root cause was identified in how the frontend called the relevant query, which defaults to returning a maximum of 10 records unless specified otherwise.
To address this, pagination support is added where the frontend query now explicitly requests the desired number of records and includes additional parameters to enable fetching more data when required. This ensures that the correct number of records is displayed.

### Fix Summary:
- Updated the frontend query to specify the required record count.
- Added pagination support for Pending Invites.

### Checks


jamesgeorge007 (changes requested):
There are corner cases in the list views (communicated separately) where the current state doesn't accommodate intermediary deletions. An easier fix would be to reset the page, search filter if applicable, etc and trigger a network call with the initial state.

jamesgeorge007 (approved):
Corner cases related to state computation are to be handled separately.

Issue #4371, opened 2024-09-20
[feature]: One-click GraphQL Query & Mutation Generation


maximpostnikov:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Summary

There is a lack of functionality for automatically generating query and mutation templates for selected data types or operations from the documentation (as it is done, for example, in Altair GraphQL).

### Why should this be worked on?

It's not always convenient to create a schema from scratch, even with autocomplete. It would be easier if we could open the documentation and insert an entire schema (query/mutation/fragment), with the option to remove unnecessary parts. Given that schemas can be very complex with deep nesting, a useful feature would be a setting to control the 'max depth' level of this nesting when adding a schema.

bogking:
Hi! Would I be able to work on this issue?

jamesgeorge007:
Hi, closing this issue since this is now addressed in [v2025.1.0](https://github.com/hoppscotch/hoppscotch/releases/tag/2025.1.0). Please let us know if you have any feedback.

PR #4626, opened 2024-12-18
feat: graphql query generation from docs

anwarulislam:
Closes HP-305 #4371

This PR revamp the old schema browser system. And added a new documentation explorer. Added feature to generate GraphQL queries from documentation.

### Features Added
- Navigate and manage the GraphQL schema using the new Documentation Explorer.
- Dynamically add fields to the GraphQL query.
- Manage multiple operations (query, mutation, subscription) within a single query.
- Maintain cursor position after query modifications.

### Known Bug
- If we navigate through `All Schema Types` in the explorer and try to build query, the query might not be valid as the stack order is not correct.


anwarulislam:
Note to reviewer:  
Please skip the following files during your review. The intention was to revamp the entire documentation section, as the current version does not provide a better user experience. However, it requires more time, so we are pausing it for now. We will need these integrations later on.

Files:
- DocExplorer.vue
- ExplorerSection.vue
- SchemaBrowser.vue
- TypeLinkNew.vue
- explorer.ts

jamesgeorge007 (changes requested):
- While inserting query documents, fields with subselection should be populated alongside placeholders for arguments.

  https://github.com/user-attachments/assets/91d721aa-21c0-468e-9fdb-0259c886a465

- While inserting multiple queries, the fields for the latter aren't populated correctly.

  https://github.com/user-attachments/assets/85a34346-ac37-450b-a735-df0f707e46a6

- An invalid attempt to insert arguments without the parent-level field crashes the page. Also, a schema failure for the tab state is reported against the key `gqlTabState` within local storage.

  https://github.com/user-attachments/assets/e32732e5-dc80-4ade-b50b-a7fa459fce84

- Inserting multiple top-level subscription fields should be prevented.

  https://github.com/user-attachments/assets/fd7826d2-9d4a-4a54-81e7-dc5b63864491

- The current placement of the insert field action leads to layout shifts and flickers at times. Keeping it towards the right end might be better when hovering over the node.

  https://github.com/user-attachments/assets/eb0e865c-b0e3-4ec4-8ae2-925d25982b00

packages/hoppscotch-common/src/components/graphql/Arguments.vue
```diff
@@ -0,0 +1,57 @@
...
+        <GraphqlArgument
+          v-for="arg in deprecatedArgs"
+          :key="arg.name"
+          :arg="arg"
+          :show-add-button="true"
+        />
+      </GraphqlExplorerSection>
+
+      <button v-else type="button" @click="handleShowDeprecated">
+        Show Deprecated Arguments
```
nivedin, line 26: we can add this as i18n

nivedin (changes requested):
https://github.com/user-attachments/assets/06d1f6bd-59e5-49d5-94b3-e6f8492eaa01

While switching tabs the documentation does not get updated if there is a selection before

- There are i18n strings to be added

- https://github.com/user-attachments/assets/9e35f4a5-0da3-4779-84c4-6dfac399b076

	We can show a visual representation if item is selected

nivedin (changes requested):
when adding arguments can we do the similar approach that we show the active and non-active state for fields.

https://github.com/user-attachments/assets/73c73391-d87f-4336-ac51-dd7a0ec46c96

nivedin:
@anwarulislam do we need to add all the arguments when selecting a field ?

anwarulislam:
> @anwarulislam do we need to add all the arguments when selecting a field ?

Actually, we will automatically add arguments. However, users can remove them by clicking on the argument in the doc explorer.

anwarulislam:
> @anwarulislam do we need to add all the arguments when selecting a field ?

@nivedin I attempted to implement this, but I believe we can include it in a point release. For now, it will take a little time. It would be great if you have any workaround!

nivedin:
> > @anwarulislam do we need to add all the arguments when selecting a field ?
> 
> @nivedin I attempted to implement this, but I believe we can include it in a point release. For now, it will take a little time. It would be great if you have any workaround!

sure we can differ this flow for the point release then, what about the argument field addition state icon change

anwarulislam:
@Nivedin, I apologize for replying to the wrong comment. I was actually referring to postponing the functionality of changing the argument state.

jamesgeorge007 (changes requested):
Toggling the preference while populating a GQL subscription document causes the page to crash. Although this is less likely to happen in production, it will affect the flow.

  https://github.com/user-attachments/assets/4f31084f-125b-4683-84d6-46893ff164be

> The generated document also gets into an invalid state at times.

  https://github.com/user-attachments/assets/fe1ced9d-f55c-4b7d-b8ba-7fc622e5339a

- There's a corner case when choosing fields under a type that causes the GQL document generated to become invalid.

  https://github.com/user-attachments/assets/6def1c34-9d7e-4c44-b118-de9359796e36

jamesgeorge007 (approved):
Although the issue's intensity with GQL documents being generated in an invalid state has reduced, it persists, especially with subscriptions; it can be revisited.

PR #4630, opened 2024-12-18
feat: add bulk edit to multipart/form data and fix url encoded form bulk edit

amk-dev:
Fixs HFE-668

**Changes**

* Adds bulk editing to multipart/formdata
* Fixs the existing implementation for url encoded body, as it was broken

### DIFF
--- a/packages/hoppscotch-common/src/components/http/BodyParameters.vue
+++ b/packages/hoppscotch-common/src/components/http/BodyParameters.vue
@@ -1,5 +1,5 @@
 <template>
-  <div>
+  <div class="flex flex-col h-full">
     <div
       class="sticky top-upperMobileStickyFold z-10 flex flex-shrink-0 items-center justify-between overflow-x-auto border-b border-dividerLight bg-primary pl-4 sm:top-upperMobileTertiaryStickyFold"
     >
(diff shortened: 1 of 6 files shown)


jamesgeorge007 (changes requested):
- The specified content types for individual key-value pairs do not persist while toggling the bulk edit mode. There might be a corner case about not being able to associate content type to the key-value pairs while updating working parameters based on the bulk edit content. We could omit them while converting to bulk edit content, similar to the case with file-type entries if it requires more effort and handle it separately.

  https://github.com/user-attachments/assets/98272518-ad3b-41e8-94c9-dcd13b0a8dbf

- Request body contents are lost following a page reload if an entry exists with a file chosen. This is a pre-existing issue.

  https://github.com/user-attachments/assets/28e32399-7f2c-4eed-af16-7e94e3f063d9

amk-dev:
@jamesgeorge007 

both are known issues.

* Didn't  consider the content type support in bulk edit for this PR, because its a bit more involved in upgrading our existing system. need to think through a syntax to support it, add syntax highlighting ( currently we're using yaml grammer ) etc.

* The file disappearing is not related to this PR. Its an existing issue and there's a ticket for it. the fix might not be stright forward as i initially expected. might need some testing. 

so imo we can consider both of these for the next point release.

jamesgeorge007 (approved):

PR #4635, opened 2024-12-19
feat: support for sending client credentials via basic auth header

amk-dev:
Fixes HFE-695

**Changes**

Previously when using `oauth` with `CLIENT CREDENTIALS` flow we only supported sending the client secret and client id via body. this PR adds support for sending them via a basic auth header.


nivedin (approved):

packages/hoppscotch-data/src/rest/v/11.ts
```diff
@@ -0,0 +1,145 @@
...
+    })
+  )
+
+export type HoppRESTAuth = z.infer<typeof HoppRESTAuth>
+
+export const HoppRESTResponseOriginalRequest =
+  HoppRESTResponseOriginalRequestOld.extend({
+    v: z.literal("3"),
+    auth: HoppRESTAuth,
+  })
```
jamesgeorge007, line 65: The `HoppRESTResponseOriginalRequest` can include a `body` field from `v10`, accommodating the bulk mode preference. This accounts for the issue whereby a request with a bulk mode preference for the body defaults to the standard view when saving a response example, instead of adhering to the original request's bulk edit mode preference.

PR #4636, opened 2024-12-20
[HCB-149] refactor: move AI request naming style to settings

Govind-S-B:
instead of hardcoded naming style for request rename feature its now moved to user settings under the AI experiments toggle (only if its enabled)

related : https://github.com/hoppscotch/central/pull/89


amk-dev:
I think there's a lot of stuff leaking into `hoppscotch-common`. eg: this PR, we've got settings having a very specific entry that would make more sense on platform.

we can think about moving out most AI stuff into platform specific places next release cycle. 

NB: just adding this here as a reminder.

amk-dev:
I think we can update the platform definition for `generateRequestName` to take a new parameter `namingStyle`. we can directly use it from the composable instead of importing settings in `hoppscotch-web`. ( Saw this on central PR )

PR #4637, opened 2024-12-20
[HCB-148] feat: UI for testscript generation ( ai experiments )

Govind-S-B:
Implements the accompanying front end changes to integrate https://github.com/hoppscotch/hoppscotch-backend/pull/133

What's changed
added modify with AI button to generate pre request and test scripts

related : https://github.com/hoppscotch/central/pull/88

### DIFF
--- a/packages/hoppscotch-common/src/composables/ai-experiments.ts
+++ b/packages/hoppscotch-common/src/composables/ai-experiments.ts
@@ -197,16 +197,6 @@ export const useModifyPreRequestScript = (
   const lastTraceID = ref<string | null>(null)
   const isModifyPreRequestPending = ref(false)
 
-  const buildRequestInfoString = (request: HoppRESTRequest, script: string) => {
-    return `METHOD : \n${request.method}\nURL : \n${request.endpoint}\nBODY :\n${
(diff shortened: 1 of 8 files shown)


packages/hoppscotch-common/src/composables/ai-experiments.ts
```diff
@@ -185,3 +185,124 @@ export const useSubmitFeedback = () => {
...
+) => {
+  return `
+  METHOD:
+  ${request.method}
+
+  URL:
+  ${request.endpoint}
+
+  BODY:
+  ${request.body.body ?? ""}
```
amk-dev, line 297: When working with request.body.body, it's important to note that it isn't always a string. Direct stringification can lead to unhelpful outputs like '[Object object]', or empty objects in cases like File uploads. Additionally, we're not capturing the body's content type in the context we provide to the model.
A simple solution is to use JSON.stringify() on the body itself. This will preserve the structure and content of the request body in a string format that the LLM can better understand and process, even when dealing with complex or unexpected data types.

amk-dev (approved):
LGTM 🚀

PR #4656, opened 2025-01-03
Correct the english response body lost rerun msg

shmup:
A simple English spelling correction that resolves the mistake seen below:

image


shmup:
I have now read this enough times that I understand how the former may be correct itself. Oops. IDK, I'll let y'all close it if that's true.

nivedin (approved):

PR #4685, opened 2025-01-22
chore: colleciton runner improvements

anwarulislam:
Closes HFE-721

## Overview

This pull request includes the following enhancements:

1. Request headers now display entries inherited from the parent collection level.
2. Added an empty state placeholder for failed request executions.
3. Introduced a loading state between request runs when a response is already present.


CuriousCorrelation:
1 Request headers now display entries inherited from the parent collection level works well ✅ 
3 Introduced a loading state between request runs when a response is already present also works well ✅ 

Testing with failed executions:

When test request fails with status code I see the test

image

and the following response body renderer component

image

but with `network_fail` response I see

image

and this response body component

image

Although there are tests that weren't run, perhaps we indicate that somehow?

CuriousCorrelation (approved):

PR #4699, opened 2025-01-27
chore: add warning prompt for same response name 

### LINKED ISSUE #4652: [feature]: Save as example action should pre-fill the Response Name with Request Name

### Summary

In Postman, whenever I create an example from a request, the example is automatically created with the name already prefilled. While on Hoppscotch, I have to write the example name everytime I click "Save an Example"
It would be nice if we can follow the same approach as what Postman does currently.

This approach will remove redundancy, as most of my requests are always only have on example.
And if you want to edit the response name, you could just double click the already-created example and edit whatever name you want.
Thank you!

nivedin:
Closes HFE-717 #4652 

This PR adds a flow where, when saving a response, the name inside the modal will be prefilled with the request name. Additionally, a warning indicator has been added to the modal input, which notifies the user if they attempt to save the response with a name that already exists for the request. The user can still proceed with saving, but the response will be overwritten with the latest changes.

https://github.com/user-attachments/assets/eb2d7776-5fbe-45c3-b7bb-edbedbd59124

### DIFF
--- a/packages/hoppscotch-common/src/components/http/Response.vue
+++ b/packages/hoppscotch-common/src/components/http/Response.vue
@@ -59,7 +59,7 @@ const showSaveResponseName = ref(false)
 
 const hasSameNameResponse = computed(() => {
   return responseName.value
-    ? !!doc.value.request.responses[responseName.value]
+    ? responseName.value in doc.value.request.responses
     : false
 })
 
--- a/packages/hoppscotch-common/src/components/http/Response.vue
+++ b/packages/hoppscotch-common/src/components/http/Response.vue
@@ -9,7 +9,8 @@
     />
   </div>
   <HttpSaveResponseName
-    v-model="responseName"
+    v-model:response-name="responseName"
+    v-model:has-same-name-response="hasSameNameResponse"
     :show="showSaveResponseName"
     @submit="onSaveAsExample"
     @hide-modal="showSaveResponseName = false"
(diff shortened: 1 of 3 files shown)


anwarulislam (approved):

packages/hoppscotch-common/src/components/http/Response.vue
```diff
@@ -54,12 +55,26 @@ const hasResponse = computed(
...
+watch(
+  () => responseName.value,
+  () => {
+    if (doc.value.request.responses[responseName.value]) {
+      hasSameNameResponse.value = true
+    } else {
+      hasSameNameResponse.value = false
+    }
+  }
+)
```
amk-dev, line 70: Why not a computed property ?

```
const hasSameNameResponse = computed(() => {
  return responseName.value 
    ? !!doc.value.request.responses[responseName.value]
    : false
});
```

packages/hoppscotch-common/src/components/http/SaveResponseName.vue
```diff
@@ -48,22 +56,49 @@ const props = withDefaults(
...
   (e: "hide-modal"): void
-  (e: "update:modelValue", value: string): void
+  (e: "update:responseName", value: string): void
+  (e: "update:hasSameNameResponse", value: boolean): void
 }>()
 
-const editingName = useVModel(props, "modelValue")
+const editingName = useVModel(props, "responseName")
+
+const hasSameNameInspectionResult: Ref<InspectorResult[]> = computed(() => {
```
amk-dev, line 79: instead of using `Ref<InspectorResult[]>` as type, might be better to go with `ComputedRef<InspectorResult[]>`, using `ComputedRef` will give us type errors if we're trying to write to it, where `Ref` won't. Since the code is using `computed` it will throw at runtime.

amk-dev (commented):
some nits

jamesgeorge007 (approved):

PR #4708, opened 2025-01-29
chore: security patches for dependency chain

AndrewBastin:
This PR intends to clear out most of the security alerts (tested against Trivy and Docker Scout) regarding dependencies in the dependency trees having CVEs.

### What's changed
1. Update the containers to use Alpine 3.19.6 directly (alpine:3.19.6) instead of node:20-alpine3.19.
2. NPM is now sourced from the npmjs.com/install.sh (pinned to v10.9.2 as its the latest version for Node 20)
3. Pin certain npm packages on the chain to the fixed versions.
4. Have a custom build of Caddy generated in the docker build step itself to clear Caddy related warnings.

### Notes to reviewers
Please do test the AIO and the single containers

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,8 +44,11 @@
       "pug": "3.0.3",
       "body-parser": "1.20.3",
       "path-to-regexp@3.2.0": "3.3.0",
+      "path-to-regexp@0.1.10": "0.1.12",
       "micromatch@<4.0.8": "4.0.8",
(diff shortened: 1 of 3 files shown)


mirarifhasan (approved):
LGTM 💯 
Build AIO and separates containers on both ARM & AMD architecture.

PR #4686, opened 2025-01-22
fix authorization hierarchy in postman import

anwarulislam:
Closes HFE-718 #4649 

### Overview

This PR addresses an issue where the authorization hierarchy is not correctly transferred when importing collections from Postman. The fix ensures that the proper authorization settings are maintained.

- [ ] Not Completed
- [x] Completed

PR #4725, opened 2025-02-05
fix: properties not inherited while importing from postman

anwarulislam:
Closes HFE-744

Follow up of #4686.

This pull request addresses an issue where properties are not being inherited correctly when importing collections from Postman. 

- [ ] Not Completed
- [x] Completed

Additionally, the following updates are brought to collection-level platform definitions under the `hoppscotch-selfhost-web` & `hoppscotch-selfhost-desktop` packages:
  - Ensure `authActive: true` is specified in combination with `authType: "inherit"` while computing authorization data.
  - Resolve type errors.


jamesgeorge007 (commented):
A similar combination exists under the collection-level platform definitions as part of `hoppscotch-selfhost-web` & `hoppscotch-selfhost-desktop` packages. Could you evaluate and add them along?

```ts
 { authType: "inherit", authActive: false }
```

Issue #4727, opened 2025-02-05
[bug]: GraphQL Connect Not Sending Headers or Bearer Token


AlexGall24:
### Is there an existing issue for this?

- [ ] I have searched the existing issues

### Current behavior

There is currently an issue with the GraphQL connect feature. When making a request, the headers and the bearer token specified in the options are not being sent. This results in authentication failures when connecting to secured GraphQL endpoints.

![Image](https://github.com/user-attachments/assets/0942ba19-0177-42ca-bfea-38a2b4711ecf)

### Steps to reproduce

1.-Open Hoppscotch and switch to the GraphQL tab.
2.-Enter a valid GraphQL endpoint URL.
3.-Add necessary headers, including an Authorization header with a bearer token.
4.-Send the request. (click connect).
5.-Inspect the request headers and observe that the specified headers and bearer token are missing in the backend.
6.-Hoppscotch shows error: 'Failed to fetch graphql schema due to network error'.

### Environment

Production

### Version

Cloud

jamesgeorge007:
Hi, this is fixed in the [v2025.2.0 release](https://github.com/hoppscotch/hoppscotch/releases/tag/2025.2.0). Please feel free to let us know if you have any feedback.

ohmyhusky:
It doesn't work on 2025.2.3 (I only tested the Docker image), but when I tried it with 2025.2.0 it does work

jasondamour:
I'm still seeing this issue in `Version 25.3.2 (20250415.085230)` (Macos Native app AND web demo)

jamesgeorge007:
Hi @ohmyhusky @jasondamour, could you check if this issue is reproducible in the latest release? If it does, appreciate more context or detailed steps to reproduce it. Thanks!

filipjnc:
Doesn't work. Can't connect to Tableau.

PR #4746, opened 2025-02-12
fix: ensure GraphQL connection sends authentication headers

anwarulislam:
Closes HFE-749 #4727

### What's changed
This pull request addresses an issue with the GraphQL connection not sending the necessary authentication headers.

- [ ] Not Completed
- [x] Completed

### DIFF
--- a/packages/hoppscotch-common/src/components/graphql/Request.vue
+++ b/packages/hoppscotch-common/src/components/graphql/Request.vue
@@ -98,7 +98,7 @@ const onConnectClick = () => {
 }
 
 const gqlConnect = () => {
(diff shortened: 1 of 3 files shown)


nivedin (approved):

jamesgeorge007 (changes requested):
Currently, a connection attempt with the active request inheriting headers and authorization properties from the collection level would fail. The headers computation needs to account for inherited collection-level properties.

nivedin (approved):

jamesgeorge007 (changes requested):
Currently, inherited headers are considered. However, authorization set at the collection level isn't accounted for.

jamesgeorge007 (approved):

Issue #4591, opened 2024-12-04
[bug]: Loosing Body Content in saved bookmark when adding file


eikaramba:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

i am constantly loosing the body content from a saved request. it seems to be related to a file multipart parameter

### Steps to reproduce

1) Create POST Request with multipart body content and a file variable. Also add a file (that is important)
2) save the request. In the console i can see an error which might be the cause of the problem
![grafik](https://github.com/user-attachments/assets/a6e39b7a-8882-41b8-9afa-e59d2187a0d0)
3) close hoppscotch and reload it again
--> now all content from the body is gone from the bookmark. even other parameters.

### Environment

Production

### Version

Cloud

eikaramba:
i think i found a ticket for this here #2453

jamesgeorge007:
Hi, this is fixed in the [v2025.2.0 release](https://github.com/hoppscotch/hoppscotch/releases/tag/2025.2.0). Please feel free to let us know if you have any feedback.

PR #4759, opened 2025-02-17
fix: request validation issue on load persistence tab

### LINKED ISSUE #2453: [bug]: Can't save a request with multipart/form-data body

### Current behavior

When I save a request with body have binary file I got error:
Uncaught (in promise) FirebaseError: Function setDoc() called with invalid data. Unsupported field value: a custom File object (found in document users/lgddOcyz50hVaH7MjbOq1MgXOvH3/collections/sync)

After reload page and select the request, it's body had been clean.

### Steps to reproduce

1. Create a request
2. Change body to multipart/form-data
3. Add a binary file to parameter
4. Click save button
5. Got error in browser console
6. Reload page
7. Select the request had been saved
8. Body disappeared

### Environment

Deploy preview

### Version

Cloud

anwarulislam:
Closes HFE-761 #2453 #4591

The validation schema has been updated to include error handling for the `Blob` array. The previous implementation:

```javascript
value: z.array(z.instanceof(Blob).nullable())
```

has been replaced with:

```javascript
value: z.array(z.instanceof(Blob).nullable()).catch([])
```

This change ensures that if the validation fails, the result will default to an empty array instead of throwing an error.

- [ ] Not Completed
- [x] Completed

### Notes to reviewers

I would like to highlight that I don't believe a new versioning is necessary for this change, as we have only added a `.catch([])` to handle validation errors. This modification is minor and does not significantly alter the functionality of the schema.

### DIFF
--- a/packages/hoppscotch-common/src/components/http/BodyParameters.vue
+++ b/packages/hoppscotch-common/src/components/http/BodyParameters.vue
@@ -113,8 +113,9 @@
               <HoppSmartFileChip
                 v-for="(file, fileIndex) in entry.value"
                 :key="`param-${index}-file-${fileIndex}`"
-                >{{ file.name }}</HoppSmartFileChip
               >
+                {{ file.name }}
+              </HoppSmartFileChip>
             </div>
           </div>
           <span v-else class="flex flex-1">
--- a/packages/hoppscotch-data/src/rest/v/9.ts
+++ b/packages/hoppscotch-data/src/rest/v/9.ts
@@ -28,6 +28,19 @@ export const FormDataKeyValue = z
       }),
     ])
   )
+  .transform((data) => {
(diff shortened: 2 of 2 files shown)


shipko:
I think that root of cause is that Blob is serializing to JSON in empty Object class. 
Have you examples to "true" serialization of Blob object with data?

anwarulislam:
Hi @shipko!
Currently, we are unable to store blobs due to their large size, both in local storage and in our cloud. Furthermore, we believe that such an approach is not practical.

shipko:
> Hi @shipko! Currently, we are unable to store blobs due to their large size, both in local storage and in our cloud. Furthermore, we believe that such an approach is not practical.

Of course!
So I think that you can use Object class instead of Blob.

jamesgeorge007 (approved):

PR #4775, opened 2025-02-20
chore: openapi importer bug fixs

amk-dev:
description todo

PR #4789, opened 2025-02-25
chore: move openapi validation and dereferencing to a worker

amk-dev:
This PR moves the opeanpi validation + dereferencing that happens while importing an openapi document into a worker so that the UI is not froze when importing things. Also brings along changes from #4775

### DIFF
--- a/packages/hoppscotch-common/src/helpers/import-export/import/workers/openapi-import-worker.ts
+++ b/packages/hoppscotch-common/src/helpers/import-export/import/workers/openapi-import-worker.ts
@@ -1,0 +1,53 @@
+import { Buffer } from "buffer"
+import process from "process"
+
+// Set up global shims for the swagger-parser library
+self.Buffer = Buffer
+self.global = self
+self.process = process
+
+import SwaggerParser from "@apidevtools/swagger-parser"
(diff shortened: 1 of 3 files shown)


packages/hoppscotch-common/src/helpers/import-export/import/workers/openapi-import-worker.ts
```diff
@@ -0,0 +1,53 @@
+import { Buffer } from "buffer"
+import process from "process"
+
+// Set up global shims for the swagger-parser library
+self.Buffer = Buffer
+self.global = self
+self.process = process
```
jamesgeorge007, line 7: A runtime exception arises while spinning up the production build about `process` not defined in the browser environment. Could you take a look?

jamesgeorge007 (changes requested):

packages/hoppscotch-common/src/helpers/import-export/import/workers/openapi-import-worker.ts
amk-dev: sent a fix.

jamesgeorge007 (approved):

Issue #4387, opened 2024-09-30
[bug]: accept-encoding gzip does not decode


Marcuse7:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When I send a GET request with header `accept-encoding: gzip, deflate` I expect the response to be decoded. What I get is undecoded (gzipped) text.

### Steps to reproduce

1. Open URL https://hoppscotch.io/
2. Use default `GET` query with https://echo.hoppscotch.io 
3. Add header `Accept-encoding` value `gzip` or `gzip, deflate`
4. Execute query

### Environment

Production

### Version

Cloud

ipaulaandreea:
I would like to work on this

Marcuse7:
Hello Vansh, thank you for reply!
I use Hoppscotch Version 24.9.1-0 (20241006.121003) on macOS 14.7.
This is the request and its reply:
![Screenshot 2024-10-13 at 14 00 55](https://github.com/user-attachments/assets/fdfd3471-9818-4742-8b8d-36d0127f767c)
These are my settings (the error persists w/ interceptor set to Native):
![Screenshot 2024-10-13 at 14 02 59](https://github.com/user-attachments/assets/8e509dcc-21ba-42e5-9ce5-3dfd94eece47)

unacro:
Same situation.

While response body is zipped, show RAW content as garble
(header `content-encoding` marked as `gzip`)

AugleBoBaugles:
I'm working on this

SanskritiHarmukh:
Hey @AugleBoBaugles , just wanted to check if you're participating in the CodeDay micro-internship program, as issues labeled with `CodeDay` are reserved for that partnership. Let me know!

AugleBoBaugles:
Yes, this is the issue assigned to my group for the CodeDay micro-internship.

AugleBoBaugles:
@SanskritiHarmukh My team has hit a wall with this issue. We’ve successfully set up our build environments, however we haven’t been able to recreate the issue locally. We have been able to recreate it using the current version of Hoppscotch at https://hoppscotch.io/ though. 

We’ve seen that there was an update to the Hoppscotch version recently and the issue still exists there. Our theory is that there is a commit to the repository that has not been used in the public version of Hoppscotch yet.

Where should we go from here to resolve this issue?

SanskritiHarmukh:
[...]

PR #4791, opened 2025-02-27
fix: gzip not been able to decode

xTudoS:
Closes # https://github.com/hoppscotch/hoppscotch/issues/4387

Set option accept_encoding on curl_handler so curl can decode gzip and deflate

### What's changed
Added test_gzip_encoding
Added test_deflate_encoding
Added curl_handle.accept_encoding(enc)

### Notes to reviewers
The project didn't have any tests, so the ones I added could be improved
Also is the library currently supported encoding are identity, zlib, and gzip. 
https://docs.rs/curl/latest/curl/easy/struct.Easy2.html#method.accept_encoding

### DIFF
--- a/packages/hoppscotch-relay/src/lib.rs
+++ b/packages/hoppscotch-relay/src/lib.rs
@@ -13,11 +13,77 @@ pub fn add(left: u64, right: u64) -> u64 {
 
 #[cfg(test)]
 mod tests {
+    use tokio_util::sync::CancellationToken;
+    use crate::interop::KeyValuePair;
     use super::*;
 
     #[test]
     fn it_works() {
         let result = add(2, 2);
         assert_eq!(result, 4);
+    }
+
+    #[test]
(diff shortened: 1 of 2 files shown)


xTudoS:
image

xTudoS:
if content encoding accept diverge from servers format, it might led request to fail

image

but will return the encoded body if accept-encoding is not set
image

CuriousCorrelation:
Thank you @xTudoS for your contribution and effort in addressing this issue! We really appreciate the time you've taken to implement this fix and add tests for it.

I wanted to let you know that this issue has been addressed as part of our new desktop app (PR at https://github.com/hoppscotch/hoppscotch/pull/4684), where we've completely re-imagined our relay system architecture to be both robust and highly cross-platform. That update includes handling for content encoding cases like the one you've identified.

Thanks again for contributing to Hoppscotch! We're closing this PR, but please feel free to explore other areas where you might like to contribute.

PR #4799, opened 2025-03-02
fix(common): `PersistenceService` call site compat

CuriousCorrelation:
This PR fixes some of the inconsistencies in calls to `PersistenceService` around double serialization/deserialization.

Legacy `PersistenceService` calls are serialized at the call site, for example:

```typescript
persistenceService.setLocalConfig("login_state", JSON.stringify(user))
```

Although `kernel`'s `Store.set` automatically serializes value for storage (and `Store.get` deserializes automatically as well):

```typescript
async set(namespace: string, key: string, value: StoredData): Promise<void> {
    const validated = StoredDataSchema.parse(value);
    // <platform specific serialization>
    this.notifyListeners(namespace, key, validated.data);
}
```

The legacy approach still works due to how platform-dependent serialization/deserialization handles `string` values specifically. When we call `JSON.stringify(data)` in legacy modules, we're converting an object to a JSON string.

When this string is passed to `persistenceService.setLocalConfig()`, it goes through the store implementation where `<platform>.<stringify>()` is used. When `<platform>.<stringify>()` receives a primitive string value (which we've already stringified), it simply treats it as a regular string value and serializes it accordingly.

During retrieval, `<platform>.<parse>()` deserializes this back into a string.

Basically, this works due to a "wrapping" pattern that the data goes through:

```
Object → JSON.stringify → string → <platform>.<stringify>() → stored string → <platform>.<parse>() → string → JSON.parse → Object
```

This is definitely not ideal:
- It's inefficient
- We lose type information
- There's a chance of forgetting `JSON.stringify` (for legacy method calls)
- Creates unnecessary storage bloat

This PR introduces new methods `set`, `get` and `remove` in order to incrementally refactor existing code to the correct usage while keeping the codebase stable for backwards compatibility.

For future changes, it's important to note that we have migrations in place for certain keys in `PersistenceService`. When possible, add variables to the migration mechanism to make sure data isn't lost midway. Not all changes necessitate migrations - for example, `login_state` can be replaced without a migration as it will simply prompt users to log back in.

However, some data is synced with the backend (either selfhosted or cloud):

```typescript
platform.sync.settings.initSettingsSync(),
platform.sync.collections.initCollectionsSync(),
platform.sync.history.initHistorySync(),
platform.sync.environments.initEnvironmentsSync(), 
```

These synchronized items definitely need migrations to ensure users don't lose their data, and we've provided those migrations in the service.

#### Note to Reviewers

There still exist sync calls to `PersistenceService` methods, especially in `std/interceptors`, `sh-desktop/platform` and similar legacy modules. Although these modules are not directly in use anymore, there may exist some static function calls that may be used by live modules indirectly, even though they don't depend on the kernel.

These modules are up for removal upon ongoing evaluation, so any `PersistenceService` in these modules can be safely ignored.

### DIFF
--- a/packages/hoppscotch-common/src/services/persistence/__tests__/index.spec.ts
+++ b/packages/hoppscotch-common/src/services/persistence/__tests__/index.spec.ts
@@ -1,5 +1,7 @@
 /* eslint-disable no-restricted-globals, no-restricted-syntax */
 
+import * as E from "fp-ts/Either"
+
 import {
   translateToNewGQLCollection,
   translateToNewRESTCollection,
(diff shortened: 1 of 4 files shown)


jamesgeorge007 (approved):

jamesgeorge007 (commented):
A `hoppLocalConfigStorage` property under the persistence service was removed with a reference under the theming module.

https://github.com/hoppscotch/hoppscotch/blob/35a471f1208e54433a0c8967d99257e92bc32e5a/packages/hoppscotch-common/src/services/persistence/index.ts#L91

https://github.com/hoppscotch/hoppscotch/blob/35a471f1208e54433a0c8967d99257e92bc32e5a/packages/hoppscotch-common/src/modules/theming.ts#L23

Could you evaluate the appropriate storage abstraction reference and address it in the same PR?

jamesgeorge007 (approved):

PR #4807, opened 2025-03-04
fix: kernel-io calls via `openExternalLink`

CuriousCorrelation:
This PR fixes some of the remaining calls to `openExternalLink` that's now routed through `kernel-io`.


jamesgeorge007 (changes requested):
Let's also update the `openExternalLink` method signature from the platform definition.

jamesgeorge007 (approved):

PR #4814, opened 2025-03-04
fix: order request while fetching from db

### LINKED ISSUE #4567: [bug]: Reordering requests within a collection does not save/restore it correctly

### Current behavior

When I drag-drop and re-order the requests in a collection, I get a message saying "Request Order Updated" but after refreshing the page the expected order is not restored.

I am using the latest docker AIO image at the time of writing this.

There are no errors in console.
The HTTP response for the re-order request is as follows
```{"data":{"updateLookUpRequestOrder":true}}```

### Steps to reproduce

1. Create a collection.
2. Add 2 or more requests to it.
3. Drag and drop a request to re-order position of a certain request.
4. Refresh the page.
5. Check if the request is in the expected position.

### Environment

Production

### Version

Self-hosted

nivedin:
Closes #4567 

This PR fixes the bug while re-ordering request inside a collection does not persist when refreshing.

### What's changed
Added `orderBy` to `getRequestsInCollection` function to fetch the request in to sort by orderIndex

### DIFF
--- a/packages/hoppscotch-backend/src/team-request/team-request.service.ts
+++ b/packages/hoppscotch-backend/src/team-request/team-request.service.ts
@@ -137,15 +137,13 @@ export class TeamRequestService {
     title: string,
     request: string,
   ) {
-    const team = await this.teamCollectionService.getTeamOfCollection(
-      collectionID,
-    );
+    const team =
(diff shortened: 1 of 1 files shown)


ankurmalhotra07:
This is still happening in personal workspaces.

Issue #3956, opened 2024-04-02
[bug]: Cannot specify both `apiKey` on `Authorization` tab and jwt bearer token on the header.


naolaregaqena:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When i specify an `apiKey` in `Authorization` section of a collection, folder or request and also specify JWT token on the header as `Authorization: Bearer ...` as so. the api key is removed in the request header. i expect for both of them to be passed on the header. both have different keys header to not assume they getting overridden. this is occurring on `hoppscotch aio version 2024.3.0`.

### Steps to reproduce

1. Go to the either self hosted or cloud version of hoppscotch.
2. Create a collection. can be either personal workspace or team workspace.
3. Add an api key on Authorization section of the collection properties.
4. Optionally create a folder and set the authorization property to `inherit`.
5. Create a request and similarly set its authorization property to `inherit`.
6. Navigate to header section of the request, the api key should be visible. as soon as an `Authorization` header is set the api key header disappears, and it is not passed on the header of the request.

### Environment

Production

### Version

Self-hosted

liyasthomas:
cc: @nivedin @amk-dev

JunaidOfficialNow:
> cc: @nivedin @amk-dev

may i take look into this.. and solve this?

liyasthomas:
@JunaidOfficialNow — Sure.

jamesgeorge007:
Hi, closing this issue as it's resolved in the latest release. Please feel free to let us know if you have any feedback.

PR #4818, opened 2025-03-04
fix(common): improve authorization header handling

anwarulislam:
Closes HFE-769 #3956

This update simplifies the logic for handling authentication headers. It clearly specifies the conditions under which we should skip adding authorization headers: specifically, when an Authorization header is already defined by the user. This exception does not apply when we are using API-key authentication with a different header name.

- [ ] Not Completed
- [x] Completed

### DIFF
--- a/packages/hoppscotch-common/src/helpers/utils/EffectiveURL.ts
+++ b/packages/hoppscotch-common/src/helpers/utils/EffectiveURL.ts
@@ -70,13 +70,32 @@ export const getComputedAuthHeaders = async (
   showKeyIfSecret = false
 ) => {
   const request = auth ? { auth: auth ?? { authActive: false } } : req
-  // If Authorization header is also being user-defined, that takes priority
+
+  /**
+   * Handling Authorization header priority rules:
+   *
+   * 1. If a user-defined "Authorization" header exists in the request:
+   *    a. We generally give it priority over auth-generated headers
+   *    b. EXCEPTION: API Key auth that uses a different header name should still be included
+   *
+   * 2. We need to check both:
+   *    - req.auth (the current request's auth settings)
+   *    - auth param (possibly inherited auth from a parent collection)
+   *
+   * 3. Only return empty array (blocking auth headers) when:
+   *    - Neither req.auth nor auth param is using API Key auth, OR
+   *    - API Key auth is being used but specifically with the "Authorization" header name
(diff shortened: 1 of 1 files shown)


jamesgeorge007 (changes requested):
Consider the scenario with a request inheriting `API Key` authorization properties from a collection level with an additional `Authorization` header defined at the request level. The computed headers based on the authorization type aren't displayed in the `Headers` tab until the `Authorization` header is removed.

  https://github.com/user-attachments/assets/d2c07191-849d-4515-953f-886e00f4647c

There's a pre-existing issue with the computed headers at the request level lost while switching between tabs. While investigating the above case, please see if this can be patched alongside.

  https://github.com/user-attachments/assets/26e7cad6-ffb5-49f5-af5d-fb4b04f9ee6c

anwarulislam:
@jamesgeorge007 I have also observed the second issue. I am working to determine the cause of the problem. I will attempt to address it with this PR; if that is not feasible, we will create a separate PR to resolve the issue.

jamesgeorge007 (changes requested):
The first issue mentioned above persists.

Also, for scenarios where a request has an `Authorization` header defined alongside the generated header for API Key authorization type inherited or specified at the request level, both header entries aren't sent along except for the case with the interceptor being Agent, also applies when the `Authorization` header entry is disabled at the request level. This can be reproduced with the sample collection from the above recording.

packages/hoppscotch-common/src/helpers/utils/EffectiveURL.ts
```diff
@@ -70,9 +70,36 @@ export const getComputedAuthHeaders = async (
...
+      (!req.auth ||
+        req.auth.authType !== "api-key" ||
+        req.auth.key.toLowerCase() === "authorization") &&
+      (!auth ||
+        auth.authType !== "api-key" ||
+        auth.key.toLowerCase() === "authorization")
+    ) {
+      return []
+    }
+  }
```
jamesgeorge007, line 102: I feel this can be refactored for readability; can be revisited.

jamesgeorge007, line 102: I feel this can be refactored for readability; can be revisited.

jamesgeorge007 (approved):
Let's revisit the remaining issues alongside evaluating and extending to GraphQL and other authorization types.

PR #4872, opened 2025-03-11
feat: add search in environment selector and sidebar

### LINKED ISSUE #4758: [feature]: Pinned environments

### Summary

I would like to have the ability to pin the most frequently used environments. My workspace has a large list of environments, and currently, there's no way to search within the modal that opens the list.

I'm not sure if this is a problem that many others face, but I believe it could help those who have the same issue.
I understand that implementing a search feature in this modal might be more complicated, but I think allowing users to pin environments could already provide a significant improvement when dealing with a large number of environments to choose from.

nivedin:
Closes HP-355 #4758 

This PR add's the feature to search environments in the selector and sidebar

### What's changed
Add's the feature to search environments from environment selector and environments section in sidebar

image

image

### DIFF
--- a/packages/hoppscotch-common/src/components/environments/my/index.vue
+++ b/packages/hoppscotch-common/src/components/environments/my/index.vue
@@ -44,38 +44,29 @@
       @select-environment="selectEnvironment(index, env)"
     />
     <HoppSmartPlaceholder
-      v-if="
-        emptyFilterdPersonalResults ||
-        !filteredAndAlphabetizedPersonalEnvs.length
-      "
+      v-if="filteredAndAlphabetizedPersonalEnvs.length === 0"
       :alt="
-        emptyFilterdPersonalResults
+        filterText
           ? `${t('empty.search_environment')}`
           : t('empty.environments')
       "
       :text="
-        emptyFilterdPersonalResults
+        filterText
           ? `${t('empty.search_environment')} '${filterText}'`
           : t('empty.environments')
       "
       :src="
-        !filterText && filteredAndAlphabetizedPersonalEnvs.length === 0
-          ? `/images/states/${colorMode.value}/blockchain.svg`
-          : undefined
+        filterText
+          ? undefined
+          : `/images/states/${colorMode.value}/blockchain.svg`
       "
     >
-      <template #icon>
-        <icon-lucide-search
-          v-if="emptyFilterdPersonalResults"
-          class="svg-icons opacity-75"
-        />
+      <template v-if="filterText" #icon>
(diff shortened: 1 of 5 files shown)


packages/hoppscotch-common/src/components/environments/my/index.vue
```diff
@@ -25,15 +32,31 @@
...
+      "
+      :alt="`${t('empty.search_environment')}`"
+      :text="`${t('empty.search_environment')} '${filterTexts}'`"
+    >
+      <template #icon>
+        <icon-lucide-search class="svg-icons opacity-75" />
+      </template>
+    </HoppSmartPlaceholder>
+    <HoppSmartPlaceholder
+      v-else-if="!searchedAndAlphabeticallySortedPersonalEnvironments.length"
```
jamesgeorge007, line 59: Can we conditionally render the markup block based on the filter text availability? The `icon` slot comes into action if filter text is present and the `body` slot otherwise.

packages/hoppscotch-common/src/components/environments/teams/index.vue
```diff
@@ -45,8 +52,20 @@
...
+    >
+      <template #icon>
+        <icon-lucide-search class="svg-icons opacity-75" />
+      </template>
+    </HoppSmartPlaceholder>
+    <HoppSmartPlaceholder
+      v-else-if="
         !loading &&
-        !alphabeticallySortedTeamEnvironments.length &&
+        !searchedAndAlphabeticallySortedTeamEnvironments.length &&
```
jamesgeorge007, line 68: The markup can be unified similar to the case above. The loading state element can probably be moved to the top, avoiding the need to include negated checks `!loading` in the conditional blocks.

packages/hoppscotch-common/src/components/environments/teams/index.vue
```diff
@@ -159,11 +178,20 @@ const props = defineProps<{
   loading: boolean
 }>()
 
-// Sort environments alphabetically by default
+const filterTexts = ref("")
```
jamesgeorge007, line 181: `filterText` would suit the context better, given that the query points to a single entity/resource. However, the results can have multiple entries accommodating entries with the same name.

```suggestion
const filterText = ref("")
```

jamesgeorge007 (changes requested):
Can we disable the search input during the loading state (exclusive to shared workspaces) and in the case where no entries are available for both the environment selector and the list of entries in the sidebar?

When it comes to the selector, the active tab will be the determining factor.

jamesgeorge007 (approved):

PR #4887, opened 2025-03-14
chore: replace old interceptor component with new one

nivedin:
This PR fixes the issue where the interceptor component didn't display the active interceptors when using proxy as active intercepter.

### What's changed
The old AppInterceptor have been replaced with the new KernalAppInterceptor. Also have added an missing i18n key.

### Before
image

### After
image


CuriousCorrelation (approved):

CuriousCorrelation:
nit: reference to `AppInterceptor` should be removed from `components.d.ts` although it won't cause any functionality issues.

jamesgeorge007 (approved):

PR #4906, opened 2025-03-18
fix: resolve issues around import via URL

nivedin:
Closes HFE-778

This PR addresses and resolves issues encountered when importing collections via GitHub Gist URLs and OpenAPI specifications through URLs.


jamesgeorge007 (changes requested):
There are issues encountered while importing OpenAPI spec via URL with the interceptor set to `Browser extension` and when importing from GitHub Gist with the interceptor set to `Agent`. This also applies to environment imports via GitHub Gist. Could you look into those?

jamesgeorge007 (approved):
The remaining concerns are to be addressed separately.

Issue #4321, opened 2024-09-02
[feature]: Create new requests without siphoning data from the previous one


GetPsyched:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Summary

When I create a new request, I want that I make a new clean request that doesn't have its body or URL copied over from some previous request.

### Why should this be worked on?

This does more harm than good because I end up removing everything most of the time; scenarios where I actually keep some of the old stuff is quite rare, which makes the current behaviour quite annoying.

liyasthomas:
This is a good feature request. Since now we have the ability to duplicate a request, I think creating a new request should start with a fresh context.

While we prioritise this feature for next release candidates, you can use the keyboard shortcut `Ctrl/Cmd` + `I` to reset the current active request to its default context without manually clearing each one.

GetPsyched:
Thanks for the handy shortcut!

PR #4909, opened 2025-03-18
feat(common): create new requests without siphoning data from the previous one

anwarulislam:
Closes HFE-788 #4321

This pull request resolves the issue that permits the creation of new requests independently, without depending on data from previous requests. Since we already have a duplicate request option, we do not need to borrow data from the active tab to create a new request.

- [ ] Not Completed
- [x] Completed

### Notes to reviewers

We also removed AI name generation from the add request. This is because we lack context for requests when adding a new one. However, rename requests still have the AI feature.

### DIFF
--- a/packages/hoppscotch-common/src/components/collections/AddRequest.vue
+++ b/packages/hoppscotch-common/src/components/collections/AddRequest.vue
@@ -15,23 +15,6 @@
           input-styles="floating-input"
           @submit="addRequest"
         />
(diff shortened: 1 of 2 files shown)


jamesgeorge007 (changes requested):
Let's remove it for GraphQL as well.

anwarulislam:
@jamesgeorge007 removed them for graphql too

PR #4911, opened 2025-03-19
fix(common): normalize headers before final relay

### LINKED ISSUE #4905: [bug]: Agent inconsistent Content-Type header overrides

### Current behavior

Overriding the `Content-Type` header in the UI results in the web server receiving multiple `Content-Type` headers, which is incorrect behavior.

The same behavior happens when using OAuth2 with the agent and clicking "generate token".

#### Override scenario:

1. Enable the Hoppscotch Agent in settings.
2. Set the body content-type of a request to `application/json` with some dummy data.
3. Override the `Content-Type` header (making sure to use capital letters, since the relay adds a duplicate with lowercase).
4. Send the request and observe undefined behavior.

#### OAuth2 scenario:

1. Enable the Hoppscotch Agent in settings.
2. Configure OAuth2 auhentication for a request (with client_credentials).
3. Click `generate token`.
4. Observe request headers received in backend
5. In my case, I get a 400 because the backend does not accept multiple `Content-Type` headers.

### Logs and Screenshots

```shell

```

### Environment

Production

### Hoppscotch Version

Cloud

### Interceptor

Agent - Web App

### Browsers Affected

Not browser-specific

### Operating System

None

### Additional Information

I'm observing this behavior with the following:

- Running in cloud on `v2025.2.2` using agent `0.1.6`.
- Using the self hosted variant on the last commit e2e5d4f1d50c22aa1846e05cbc83abd81949ab2a.

Additionally, when testing with `https://echo.hoppscotch.io` it incorrectly reports a single `Content-Type` header, even though multiple were sent. Testing this against my local C# application and logging the headers I can see that hoppscotch agent sends multiple Content-Type headers, in random order.

I managed to hack together a workaround in the `prepare` method of `relay/src/request.rs` by checking if `request_headers` contains a `Content-Type` header (case insensitive) and removing the content-type header set in `ContentHandler` to prefer the header coming from the UI.

CuriousCorrelation:
This PR fixes header duplication and non-normalization issues that prevent certain flows including OAuth on stricter servers.

| Before | After |
| ------ | ------|
| image | image | 

Closes HFE-786
Closes #4905 

Issue affects both `Native` interceptor on desktop and `Agent` interceptor on web, PR aimed to fix both.

### DIFF
--- a/packages/hoppscotch-agent/src-tauri/Cargo.lock
+++ b/packages/hoppscotch-agent/src-tauri/Cargo.lock
@@ -206,9 +206,9 @@ dependencies = [
 
 [[package]]
 name = "async-compression"
-version = "0.4.20"
+version = "0.4.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "310c9bcae737a48ef5cdee3174184e6d548b292739ede61a1f955ef76a738861"
+checksum = "c0cf008e5e1a9e9e22a7d3c9a4992e21a350290069e36d8fb72304ed17e8f2d2"
 dependencies = [
  "flate2",
  "futures-core",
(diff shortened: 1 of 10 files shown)


AndrewBastin (approved):

jamesgeorge007 (commented):
The refresh token attempt fails when it comes to the `Authorization Code` grant type as part of the `OAuth 2.0` auth type with the interceptor set to `Agent`. Could you take a look?

jamesgeorge007 (approved):

PR #4942, opened 2025-03-27
feat: Agent registration UX flow updates

nivedin:
This PR updates the UX flow for agent registration in the interceptor selection component.

### Before
https://github.com/user-attachments/assets/d0821457-de36-4a78-872b-9f52ca9d2fc2

### After
https://github.com/user-attachments/assets/ca71c0f4-addc-491f-bb92-27e7df1266d3

- The `checkAgentStatus` and `initiateRegistration` step have been incorporated in a single step. 
- Also added a error toast if a registration is started without completing an existing one

https://github.com/user-attachments/assets/ab0367a6-5b46-40ab-844d-ffec4b1f7445


CuriousCorrelation:
Changes look excellent, tested the functionality on Win/Mac/Linux, works well on all platforms 👍 

Noticed an issue where cleared registrations were persisting after refresh, fixed that with the latest commit.

https://github.com/user-attachments/assets/3a9cd955-4c9d-4431-af6f-e5d98481e3ba

https://github.com/user-attachments/assets/71bcedc3-435b-4675-9eab-662b5cfec8f5

CuriousCorrelation:
Btw the tooltip on the button might be redundant, thoughts on it?

s1

nivedin:
> Btw the tooltip on the button might be redundant, thoughts on it?
> 
> s1

yup we don't need it anymore, pushed an update

CuriousCorrelation (approved):
lgtm :rocket:

jamesgeorge007 (changes requested):
- There's a pre-existing issue with the Agent interceptor preference on the settings page and the floating window diverging. Could you see if it can be patched alongside?

  https://github.com/user-attachments/assets/682cdb4a-2d3d-41ab-9d76-29a67c46dd68

- Dismissing the Agent registration window without completion and a re-attempt would result in the error toast about ongoing registration, but the window was already dismissed.

  https://github.com/user-attachments/assets/8e7725ee-af77-4a2e-a2d2-6fb898463928

CuriousCorrelation:
Added a bit clearer registration/retry flow as shown here:

https://github.com/user-attachments/assets/6524c472-fa25-464b-8d25-fa3c1ab75dae

Essentially now users can maximize the window back up to see the current registration status, although after hiding the window (which means the user wasn't the one who initiated the process), the OTP cannot be used to authenticate another app (in this case current). If the process was unintentional, canceling the registration process and restarting it from the correct app is the right flow.

The `Maximize Window` button will show the OTP screen if there's an ongoing registration attempt, otherwise default to `Registration List` view.

jamesgeorge007 (approved):

PR #4958, opened 2025-04-04
chore: truncate long secret environment variables

atrimpe-ford:
In the current version of Hoppscotch, the show/hide button for viewing a secret value in the settings dialog box is shoved out of view by the content of the secret value input box. This change enhances the experience by keeping the button in its original location, while not preventing the value from being scrollable so the entire value can be viewed as it in the current state.

### What's changed

screencapture_before
screencapture_after


nivedin (approved):
Added truncate instead of setting max width, and added a padding left for removing the layout shift while toggling secret.

PR #4963, opened 2025-04-07
feat: import authorization header with curl

chhaviG22:
This pull request updates the logic used to filter headers when a curl command is imported into Hoppscotch. Previously, the "Authorization" header along with certain "Content-Type" headers were intentionally excluded. This update retains the "Authorization" header, enhancing the tool's ability to handle secure requests seamlessly by including all critical headers necessary for authentication.

### What's changed

- **Header Filtering Logic**: Modified the header filter logic to not exclude the "Authorization" header during the import of curl commands.

### Notes to reviewers

- This update is crucial for users importing curl commands that rely on the Authorization header for API requests.


CuriousCorrelation:
Thanks for working on this PR! I noticed there are a couple of failing tests. Could you try rebasing on `main` or `next` and, if they still persist, take a look when you get a chance?

If this is still WIP, no worries - take your time. Consider marking it as a draft in the meantime.

Thanks again!

chhaviG22:
> Thanks for working on this PR! I noticed there are a couple of failing tests. Could you try rebasing on `main` or `next` and, if they still persist, take a look when you get a chance?
> 
> If this is still WIP, no worries - take your time. Consider marking it as a draft in the meantime.
> 
> Thanks again!

I mentioned the problem with test cases on this GitHub issue: https://github.com/hoppscotch/hoppscotch/issues/4964. It's been resolved and closed, but there are still some test cases failing randomly. I will investigate this issue further. Thanks.

CuriousCorrelation:
Thanks @chhaviG22! 

Ah seems like some of the tests are flaky and were the ones causing issues, unrelated to the PR. I see you've worked around that by closing and reopening to trigger a fresh workflow run - smart approach!

Regarding the PR implementation, I could use your help to better understand the specific problem this PR solves. When testing with a simple curl command containing an `Authorization` header like so:

```
curl --request POST \
  --url https://echo.qubit.codes/ \
  --header 'Authorization: Bearer auth-value' \
  --data '{
  "test": "test"
}'
```
I'm seeing the `Authorization` header correctly included in the authorization tab and reflected in the echo response:

| cURL request | Request-Response
| --- | --- |
| image | image |

The authorization handling is managed in [auth.ts](https://github.com/hoppscotch/hoppscotch/blob/236db8396be37119d293c946b2751b0149b4ec89/packages/hoppscotch-common/src/helpers/curl/sub_helpers/auth.ts#L106), but I may be missing the particular cases your PR addresses.

Although I believe perhaps the issue is `Authorization` and `Content-Type` not appearing in `Headers` tab right away when importing, but if you were to refresh the page, they show up normally. So I think the issue could be a de-sync in the UI.

Could you confirm if that's the issue? If not, perhaps a quick example demonstrating the "before" and "after" behavior would help me better understand the improvement.

Thanks!

chhaviG22:
> Thanks @chhaviG22!
> 
> Ah seems like some of the tests are flaky and were the ones causing issues, unrelated to the PR. I see you've worked around that by closing and reopening to trigger a fresh workflow run - smart approach!
> 
> Regarding the PR implementation, I could use your help to better understand the specific problem this PR solves. When testing with a simple curl command containing an `Authorization` header like so:
> 
> ```
> curl --request POST \
>   --url https://echo.qubit.codes/ \
>   --header 'Authorization: Bearer auth-value' \
>   --header 'content-type: application/json' \
>   --data '{
>   "test": "test"
> }'
> ```
> 
> I'm seeing the `Authorization` header correctly included in the authorization tab and reflected in the echo response:
> 
> cURL request	Request-Response
> image	image
> The authorization handling is managed in [auth.ts](https://github.com/hoppscotch/hoppscotch/blob/236db8396be37119d293c946b2751b0149b4ec89/packages/hoppscotch-common/src/helpers/curl/sub_helpers/auth.ts#L106), but I may be missing the particular cases your PR addresses.
> 
> Although I believe perhaps the issue is `Authorization` and `Content-Type` not appearing in `Headers` tab right away when importing, but if you were to refresh the page, they show up normally. So I think the issue could be a de-sync in the UI.
> 
> Could you confirm if that's the issue? If not, perhaps a quick example demonstrating the "before" and "after" behavior would help me better understand the improvement.
> 
> Thanks!


This PR introduces the capability to support customized authentication ecosystems within the system. Specifically, it handles scenarios where the Authorization header does not match any of the predefined formats in the Auth Option list. This improvement helps the system work smoothly with different and unique security setups, making it more flexible to use.

CuriousCorrelation:
Thanks you for the elaboration @chhaviG22 

Here's a small write-up on my review so far, the following is intended to provide context for other reviewers.

---

Basically the current implementation 

- removes `Authorization` headers from the final request headers through the filter in `recordToHoppHeaders` at [sub_helpers/headers.ts](https://github.com/hoppscotch/hoppscotch/blob/236db8396be37119d293c946b2751b0149b4ec89/packages/hoppscotch-common/src/helpers/curl/sub_helpers/headers.ts#L74-L76). 
- this PR proposes to keep these headers intact. 

Currently, when we parse a curl command with an `Authorization` header (like `curl -H "Authorization: Bearer token123" example.com`), we extract this header in `getAuthFromAuthHeader` at https://github.com/hoppscotch/hoppscotch/blob/236db8396be37119d293c946b2751b0149b4ec89/packages/hoppscotch-common/src/helpers/curl/sub_helpers/auth.ts#L106 and convert it into an `auth` object, then filter it out from the headers. The proposed change would preserve this header in the final request while still creating the auth object.

This is reasonable with a subtle caveat around interaction with request processing logic in [EffectiveURL.ts](https://github.com/hoppscotch/hoppscotch/blob/main/packages/hoppscotch-common/src/helpers/utils/EffectiveURL.ts). 

When we execute a request, [getComputedAuthHeaders](https://github.com/hoppscotch/hoppscotch/blob/236db8396be37119d293c946b2751b0149b4ec89/packages/hoppscotch-common/src/helpers/utils/EffectiveURL.ts#L329) checks if a user-defined `Authorization` header exists, see this and specifically https://github.com/hoppscotch/hoppscotch/blob/236db8396be37119d293c946b2751b0149b4ec89/packages/hoppscotch-common/src/helpers/utils/EffectiveURL.ts#L74-L89

If it does, it generally skips generating auth headers from the auth configuration (with an exception for API Key auth using a different header name). This means with this change, the explicit header from the curl command would **take precedence over what might otherwise be generated from the auth object**.

I believe this behavior is actually what users would expect ✅ - the explicit header in their curl command should be what's sent in the request, making it more faithful to the original command, more transparent.

One consideration is how this interacts with the UI. Users might be confused seeing both an explicit `Authorization` header and an auth configuration derived from that same header. Although, I think the benefit of preserving the exact curl command outweighs this.

#### Example

For basic authentication, take this example curl command:

```
curl -X POST \
  "https://httpbin.org/post" \
  -H "Authorization: Basic <some-token>" \
  -H "Content-Type: application/json" \
  -d '{
    "name": "test"
  }'
```

With the current implementation, this would only populate the Authorization tab with Basic Auth (username/password), but the explicit "Authorization" header would be filtered out from the Headers tab. After this change, the header will appear in both places: **decoded in the Authorization tab** and **preserved verbatim in the Headers tab**.


| Authorization tab |  Header tab |
| --- | --- |
| image | image |

Worth noting that this change brings curl importing more in line with curls own behavior, see https://curl.se/docs/manpage.html#-H for more info. 

When executed, the header from the Headers tab will take precedence over what would be generated from the Authorization tab config, which I suppose matches what most servers would expect, since they'll just use the Authorization header sent in the request.

image

**Notice what echo returns vs what's set in `Authorization` tab**.

This change becomes more valuable for auth types we don't explicitly support parsing directly (even if they exist in Authorization tab).

For example:

```
curl -X GET \
  "https://httpbin.org/post" \
  -H "Authorization: Hawk id=\"dh37fgj492je\", ts=\"1353832234\", nonce=\"j4h3g2\", mac=\"6R4rV5iE+NPoym+WwjeHzjAGXUtLNIxmo1vpMofpLAE=\"" \
  -H "Content-Type: application/json"
```

| Import | Parse |
| --- | --- |
| image | image |


Currently, our parser would try to extract this into an auth object, fail to recognize the Hawk scheme (even tho now we support it), and the header would be lost entirely. With this change, even though we can't populate the Authorization tab with Hawk auth (unless we have specific support for it in the future), **the header will at least be preserved in the Headers tab, allowing the request to work correctly**.

The same benefit applies to other auth schemes like Digest, AWS Signature, and any custom authorization schemes that API providers might use.

Instead of losing this information during import, we'll now preserve it exactly as provided in the curl command, generally making the curl importing more "faithful" to the original command, especially useful for those using custom authorization schemes.

CuriousCorrelation (approved):

AndrewBastin (approved):

jamesgeorge007 (approved):

PR #4995, opened 2025-04-14
fix(common): refine profile field validations

### LINKED ISSUE #4994: [bug]: Email update lacks same-address and empty validation.

### Current behavior

Saving an unchanged profile name incorrectly shows an "Updated profile name is same as the current profile name" error, and keeping the same email address shows no error "profile updated".

Making an API call is unnecessary in this context because the necessary resources are already present.

### Steps to reproduce

1. Navigate to your profile page in Hoppscotch
2. Click the "Save" button next to your profile name. If you've not changed the name, an error might occur "current profile name is the entered name". Otherwise, you should see a "Profile updated" message (optional)
3. Click the "Save" button for your email address. There have no error messages will appear if the email field is empty or unchanged.

### Logs and Screenshots

https://github.com/user-attachments/assets/47b7be38-202c-4313-a063-62d693e72ce4

### Environment

Production

### Hoppscotch Version

Cloud

### Interceptor

### Browsers Affected

Not browser-specific

### Operating System

Linux

### Additional Information

The issue is present in the Hoppscotch Desktop.

NayanUnni95:
Closes #4994  

The web app doesn't properly check if the email address is empty or the same. This also causes an unnecessary request to the server. I need to add checks to prevent saving empty or unchanged email addresses and avoid that extra server call.
Also, includes relevant `i18n` string additions.

### What's changed

Implemented validation for empty or identical emails before the API call.


packages/hoppscotch-common/src/pages/profile.vue
```diff
@@ -277,9 +278,20 @@ const updatingEmailAddress = ref(false)
...
+
+  if (currentUser.value?.email === inputEmailAddress) {
+    toast.error(`${t("error.same_email_address")}`)
+    return
+  }
+
   updatingEmailAddress.value = true
   platform.auth
-    .setEmailAddress(emailAddress.value as string)
+    .setEmailAddress(inputEmailAddress as string)
```
jamesgeorge007, line 294: On another note, the Email input field can be marked read-only for self-hosted instances since the underlying action isn't in place.

Note: This is to be revisited by the core team after relevant discussions.

https://github.com/hoppscotch/hoppscotch/blob/80b63545f14b43067ba290ab6bf15e3d8438f966/packages/hoppscotch-selfhost-web/src/platform/auth/web/index.ts#L319-L322

jamesgeorge007 (approved):
Thanks

Issue #5031, opened 2025-04-29
[feature]: Allow user email updates from Admin UI


emil-johansson-99:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Summary

In our self-hosted instance of Hoppscotch we get some requests to update users email address. Today this handled through manual database updates. It would be nice to be able to update email address through Admin UI instead. Same functionality exists for user display name so this can be used as template for UI and code.

See attached screenshot for proposed UI design.

![Image](https://github.com/user-attachments/assets/2831fc49-f9b6-45b0-b8a8-f022639ba1df)

.

### Why should this be worked on?

Simplifies user management.

emil-johansson-99:
PR for implementation of feature: https://github.com/hoppscotch/hoppscotch/pull/5032

Gannu456:
I can work on this feature. Please assign this issue to me.

emil-johansson-99:
Hi! There is already a PR that implements this feature. It has been for approval more than a month now.

PR #5032, opened 2025-04-29
feat: allow user email updates from admin UI

emil-johansson-99:
This PR implements feature request 5031, https://github.com/hoppscotch/hoppscotch/issues/5031.

Closes 5031

### DIFF
--- a/packages/hoppscotch-backend/src/admin/admin.resolver.ts
+++ b/packages/hoppscotch-backend/src/admin/admin.resolver.ts
@@ -184,6 +184,31 @@ export class AdminResolver {
   }
 
   @Mutation(() => Boolean, {
+    description: 'Update user email',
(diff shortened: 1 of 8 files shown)


Leon-Luu:
@jamesgeorge007 would you approve this as well? This patch is very useful in organization where users can change email and names. There are no easy ways to manage this unless admin goes to the database itself and change them.
Impact not changing emails ->  users won't be able to perform login to there old account and retrieve or continue to use Hoppscotch with the same data. We had been using this patch and it works very well. Thanks

jamesgeorge007:
Thanks for the follow-up. This is on the radar, but given the scope and the implications of email changes on identity/auth flows, it needs proper evaluation. Will revisit when bandwidth permits.

PR #5049, opened 2025-05-03
fix(desktop): verbatim path handling in disk resolution

### LINKED ISSUE #4859: [bug]: Root page fails to load with status code `404` in Desktop App on Windows

### Current behavior

Image

### Steps to reproduce

Image

### Environment

Production

### Version

Cloud

CuriousCorrelation:
The storage init was failing on Windows when checking available disk space due to verbatim paths (also known as extended-length paths) returned by `std::fs::canonicalize()`.

Windows has two path formats: traditional paths limited to 260 characters, and extended-length paths (prefixed with `\\?\`) that support longer paths.

While Rust's `std::path` is "platform-independent", Windows-specific behavior still requires special consideration for these extended-length paths.

This fixes that.

Closes HFE-839
Closes #4859

Basically the issue occurred because `std::fs::canonicalize()` on Windows may return extended-length paths when:
1. The path exceeds 260 characters
2. The path contains certain special characters
3. The system is configured to use extended paths (fairly common on Win 10/11)

The root issue is when comparing the canonicalized storage path against disk mount points.

A verbatim path like `\\?\C:\Users\SomeUser\AppData\...` would fail to match against the disk's mount point `C:\` because the comparison was done as a path string prefix check. This isn't handled by `std::path`, see [dunce's README](https://gitlab.com/kornelski/dunce/-/blob/master/README.md?ref_type=heads&plain=1#L16-17)

This caused the disk resolution to fail with "Disk not found" errors, preventing the app from initializing the bundle, which would show a `404`.

See the docs at (https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file#win32-file-namespaces).

`dunce` crate provides cross-platform path canonicalization that specifically handles Windows verbatim paths correctly. When given a verbatim path, dunce safely strips the `\\?\` prefix when possible while maintaining path validity:

```rust
let normalized_path = dunce::canonicalize(&storage_path).unwrap_or(storage_path.clone());
```

The normalization preserves the verbatim prefix only when necessary - for example with reserved filenames like `\\?\C:\COM1` or paths containing certain special characters.

On non-Windows systems, dunce **behaves identically** to `std::fs::canonicalize()`.

Testing:
- Traditional Windows paths (< 260 chars)
- Extended-length paths (> 260 chars) that get verbatim prefixes
- Paths containing special characters
- UNC paths and network shares
- All supported non-Windows platforms

### Notes to reviewers

To test this, launch the app on Windows, have the install location be ~95% full (using something like `fsutil file createnew largefile.bin <size-in-bytes>`), launch the app, you should see `404` because the app bundle wasn't loaded due to 1. incorrect space calculation or 2. disk mismatch. Now try the same thing with this patch.

### DIFF
--- a/packages/hoppscotch-desktop/plugin-workspace/tauri-plugin-appload/Cargo.lock
+++ b/packages/hoppscotch-desktop/plugin-workspace/tauri-plugin-appload/Cargo.lock
@@ -4061,6 +4061,7 @@ dependencies = [
  "chrono",
  "cocoa",
  "dashmap",
+ "dunce",
  "ed25519-dalek",
  "flate2",
  "futures",
--- a/packages/hoppscotch-desktop/plugin-workspace/tauri-plugin-appload/Cargo.toml
+++ b/packages/hoppscotch-desktop/plugin-workspace/tauri-plugin-appload/Cargo.toml
@@ -34,6 +34,7 @@ futures = "0.3.31"
 mime_guess = "2.0.5"
 rayon = "1.10.0"
 hex_color = "3.0.0"
+dunce = "1.0.5"
 
 [build-dependencies]
 tauri-plugin = { version = "2.0.1", features = ["build"] }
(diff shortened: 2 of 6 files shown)


AndrewBastin (approved):
lgtm

Issue #4531, opened 2024-11-12
[bug]: Basic Auth Not Using Global Variable Value


SVNKoch:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

Desktop Windows 10, v2024.10.0

GIVEN global variable `password` has value `mypassword`
<img src="https://github.com/user-attachments/assets/4f2bf79b-5201-4a0a-80a4-3ca80ca6c33f" width="400" />
GIVEN Authentification is of type Basic Auth
GIVEN Baisc Auth username is `me`
GIVEN Baisc Auth password is `<<password>>` (to use the variable with value `mypassword`)
<img src="https://github.com/user-attachments/assets/48b2e473-6a96-4c19-b7f6-1604c655833f" width="400" />

WHEN Header Tab is opened

THEN Authorization has value `Basic bWU6PDxwYXNzd29yZD4+` (` = Base64("me:<<password>>")`)

Expected:

THEN Authorization has value `Basic bWU6bXlwYXNzd29yZA==` (` = Base64("me:mypassword")`)

<img src="https://github.com/user-attachments/assets/8d494fc1-0523-44bd-86ec-51ce8122af43" width="400" />

potentially related with #4208

### Steps to reproduce

See current behaviour

### Environment

Production

### Version

Local

DarkChaos:
Is here any Progress? I have the same issue with Authorization Type: Api Key:

![Image](https://github.com/user-attachments/assets/b9275b34-e423-4d09-a2b0-c23a53bbad1b)

![Image](https://github.com/user-attachments/assets/38649243-8298-4e25-ad43-d08989243015)

When I set Authorization Type: None and set the Header manual it is working:

![Image](https://github.com/user-attachments/assets/4c867ca2-d2c0-406e-bbf9-bc8e4321085d)

![Image](https://github.com/user-attachments/assets/7daa2532-6f15-412c-991e-b32413ff01f4)

rdavaillaud:
It's not working anymore in web 2025.3.2. Can't use variable in Authorization tab, with Basic authorization type.

liyasthomas:
It has been fixed in the recent release.

PR #5055, opened 2025-05-06
feat: Initial and current value for environment variables

nivedin:
Closes HFE-841 HFE-846 HFE-875 #4531 

This PR add's Initial and current value for environment variables

Initial value is the value defined directly in the environment. This value gets synced to the cloud and is shared with your team. 

Current value is what gets used when you actually send a request in Hoppscotch. It’s stored locally in your browser and isn’t synced to the cloud. 

image

### Notes to reviewers
- Check old enviornment is migrated to new version.
- Perform usual request run with global, environment and request variables
- Use the variables in Tests 
- Export and import variables
- Tooltip and quickpeek is showing the current value of variables

### Todo
- [x] Team Environment selector reset bug
- [x] Add support for CLI 
- [x] Fix tests
- [ ] Update documentation for CLI and app

### DIFF
--- a/packages/hoppscotch-cli/src/__tests__/e2e/commands/test.spec.ts
+++ b/packages/hoppscotch-cli/src/__tests__/e2e/commands/test.spec.ts
@@ -95,6 +95,7 @@ describe("hopp test [options] <file_path_or_id>", { timeout: 100000 }, () => {
         const testFixtures = [
           { fileName: "env-v0.json", version: 0 },
           { fileName: "env-v1.json", version: 1 },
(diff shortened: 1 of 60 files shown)


jamesgeorge007 (changes requested):
- Encountering failure with importing environments via Gist action.

  image

- The `Edit current environment` spotlight action doesn't trigger the underlying action. This issue likely existed in the release branch already, please verify.
  image

- Trying to add environment variables through the spotlight action doesn't retain the value entered in the follow-up modal

  image

PR #5072, opened 2025-05-14
feat(common): add support for registering and merging additional locale messages

anwarulislam:
### What's changed
This pull request introduces functionality to allow dynamic registration and merging of additional i18n messages for specific locales in the `i18n` module. The most significant changes include the addition of a mechanism to register and merge external translations with existing locale messages and an enhancement to apply these additional messages when the app language changes.

- [ ] Not Completed
- [x] Completed

### DIFF
--- a/packages/hoppscotch-common/src/modules/i18n.ts
+++ b/packages/hoppscotch-common/src/modules/i18n.ts
@@ -73,6 +73,88 @@ let i18nInstance: I18n<
   true
 > | null = null
 
+// Store additional message registrations keyed by locale
+const additionalMessages: Record<string, Record<string, unknown>[]> = {}
+
+/**
+ * Register additional i18n messages for a specific locale.
+ * This allows other packages to extend the translations of the common package.
+ *
+ * @param locale The locale code to extend (e.g., 'en', 'fr', etc.)
+ * @param messages The additional messages to merge with existing translations
+ */
+export const registerAdditionalMessages = (
+  locale: string,
+  messages: Record<string, unknown>
+): void => {
+  if (!additionalMessages[locale]) {
+    additionalMessages[locale] = []
+  }
+
+  additionalMessages[locale].push(messages)
+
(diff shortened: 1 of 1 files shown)


packages/hoppscotch-common/src/modules/i18n.ts
```diff
@@ -73,6 +73,88 @@ let i18nInstance: I18n<
...
+    additionalMessages[locale] = []
+  }
+
+  additionalMessages[locale].push(messages)
+
+  // If i18n is already initialized, merge the messages immediately
+  if (i18nInstance && i18nInstance.global.availableLocales.includes(locale)) {
+    mergeAdditionalMessages(locale)
+  }
+}
```
jamesgeorge007, line 100: Shouldn't this function be exported from the [entrypoint](https://github.com/hoppscotch/central/blob/main/packages/hoppscotch-common/src/index.ts) so that it could be consumed from the context of other packages at the platform level?

jamesgeorge007, line 100: Shouldn't this function be exported from the [entrypoint](https://github.com/hoppscotch/hoppscotch/blob/main/packages/hoppscotch-common/src/index.ts) so that it could be consumed from the context of other packages at the platform level?

packages/hoppscotch-common/src/modules/i18n.ts
```diff
@@ -73,6 +73,88 @@ let i18nInstance: I18n<
...
+  ) {
+    return
+  }
+
+  // Get current messages for the locale
+  const currentMessages = i18nInstance.global.getLocaleMessage(locale)
+
+  // Deep merge additional messages with current messages
+  const newMessages = additionalMessages[locale].reduce(
+    (acc, messages) => deepMergeMessages(acc, messages),
```
jamesgeorge007, line 121: Could you evaluate leveraging relevant functions from `lodash-es` here instead?

jamesgeorge007 (approved):

PR #5079, opened 2025-05-15
feat: add JWT authentication support

anwarulislam:


jamesgeorge007 (changes requested):
Please look into the CI failure arising from `hoppscotch-data` type emits.

packages/hoppscotch-data/src/rest/v/13.ts
```diff
@@ -0,0 +1,83 @@
...
+    HoppRESTAuthDigest,
+    HoppRESTAuthHAWK,
+    HoppRESTAuthAkamaiEdgeGrid,
+    HoppRESTAuthJWT,
+  ])
+  .and(
+    z.object({
+      authActive: z.boolean(),
+    })
+  )
```
jamesgeorge007, line 65: We'll require a new schema definition for `HoppCollection` accounting for the above `HoppRESTAuth` variant.

packages/hoppscotch-common/src/components/http/authorization/JWT.vue
```diff
@@ -0,0 +1,320 @@
...
+  "HS256",
+  "HS384",
+  "HS512",
+  "RS256",
+  "RS384",
+  "RS512",
+  "ES256",
+  "ES384",
+  "ES512",
+]
```
jamesgeorge007, line 293: There are active issues with the usage of asymmetric algorithms. Let's keep only the HMAC-based algorithms in the initial iteration.

PR #5081, opened 2025-05-17
feat(lenses): enhance content type detection and lens selection logic

### LINKED ISSUE #5080: [feature]: stop formatting based on response header

### Summary

We should be able to override the response formatting. I work a lot with poorly designed APIs that return JSON as text/HTML, and I can’t format it in JSON…

Because a lot of poorly designed APIs exist… and not being able to override the formatting is annoying.

Phaired:
This pull request enhances the logic for determining suitable lenses for displaying HTTP response content in the `getSuitableLenses` function. The main changes focus on improving content type detection, prioritizing relevant lenses, and providing fallback options for better user control.

### Enhancements to lens selection logic:
* Added a smarter approach for successful responses (`response.type === "success")` by prioritizing lenses that match the content type and dynamically appending additional lenses (e.g., JSON lens) if the content can be parsed into other formats.
* Ensured that the raw lens is always included to allow users to view the raw response, even if no specific matching lenses are found.

### Improvements to fallback behavior:
* Included all lenses as a fallback option when no matching lenses are detected, giving users full control over lens selection.

Closes #5080

### DIFF
--- a/packages/hoppscotch-common/src/helpers/lenses/lenses.ts
+++ b/packages/hoppscotch-common/src/helpers/lenses/lenses.ts
@@ -43,18 +43,54 @@ export function getSuitableLenses(response: HoppRESTResponse): Lens[] {
     (h) => h.key.toLowerCase() === "content-type"
   )
 
+  // If no content type is found, return raw lens as fallback
   if (!contentType) return [rawLens]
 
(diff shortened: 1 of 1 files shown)


nivedin (approved):
LGTM 🚀

PR #5105, opened 2025-05-28
fix: environment and rest tab state schema migration

nivedin:
This PR fixes the following issue
- The synced environments failed to migrate to new version
- Schema error due to response version migrate fail.

### DIFF
--- a/packages/hoppscotch-common/src/services/persistence/validation-schemas/index.ts
+++ b/packages/hoppscotch-common/src/services/persistence/validation-schemas/index.ts
@@ -433,7 +433,7 @@ const HoppTestResultSchema = z
           .strict(),
       })
       .strict(),
-    consoleEntries: z.array(z.record(z.string(), z.unknown()).optional()),
+    consoleEntries: z.optional(z.array(z.record(z.string(), z.unknown()))),
   })
   .strict()
 


jamesgeorge007 (changes requested):
Creating response examples report failure due to schema mismatch. Please look into it.

jamesgeorge007 (approved):

packages/hoppscotch-data/src/rest/v/12.ts
```diff
@@ -12,6 +12,11 @@ import { z } from "zod"
 import { defineVersion } from "verzod"
 import { HoppRESTAuthOAuth2, V11_SCHEMA } from "./11"
 
+import {
+  HoppRESTResponseOriginalRequest as HoppRESTResponseOriginalRequestOld,
+  HoppRESTRequestResponse as HoppRESTRequestResponseOld,
+} from "./9"
```
jamesgeorge007, line 18: We'll need to import it from `v11` instead.

packages/hoppscotch-data/src/rest/v/13.ts
```diff
@@ -14,6 +14,11 @@ import { z } from "zod"
 import { defineVersion } from "verzod"
 import { HoppRESTAuthOAuth2 } from "./11"
 
+import {
+  HoppRESTResponseOriginalRequest as HoppRESTResponseOriginalRequestOld,
+  HoppRESTRequestResponse as HoppRESTRequestResponseOld,
+} from "./9"
```
jamesgeorge007, line 20: We'll need to import it from `v12` instead.

PR #5124, opened 2025-06-03
fix: schema error and test updation flow

nivedin:
This PR fixes the issue with schema mismatch and test updation flow which leads to schema mismatch.

### Notes to reviewers
- Check if any schema mismatch occurring with REST-tab, Environments
- Any data loss with any properties.


packages/hoppscotch-common/src/helpers/fixBrokenRequestVersion.ts
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * @param docs - The ordered documents of the REST tab to fix.
+ * @returns The fixed ordered documents with valid request structures.
+ */
+export const fixBrokenRequestVersion = (
+  docs: HoppRESTab["orderedDocs"]
+): HoppRESTab["orderedDocs"] => {
+  return docs.map((x: HoppRESTab["orderedDocs"][number]) => {
+    if (x.doc.type === "request") {
+      const req = safelyExtractRESTRequest(
```
jamesgeorge007, line 28: Since `safelyExtractRESTRequest` is deprecated, should we avoid introducing new implementations based on it and instead evaluate an alternative approach?

jamesgeorge007: Handled in a separate discussion - keeping it as is for now.

packages/hoppscotch-common/src/helpers/fixBrokenEnvironmentVersion.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  return envs.map((env) => {
+    if ("variables" in env)
+      return {
+        ...env,
+        variables: env.variables.map((x) =>
+          translateToNewEnvironmentVariables(x)
+        ),
+      }
+  })
```
jamesgeorge007, line 32: We'd need a fallback here when the condition isn't met to prevent `undefined` entries.

packages/hoppscotch-data/src/rest-request-response/original-request/v/3.ts
jamesgeorge007, line 1: Why do the schema definitions start on `v3`? Can we apply this to legacy `HoppRESTRequest` schema definitions, following the same conventions, by enforcing a valid version range similar to what we do with `HoppCollection`?

https://github.com/hoppscotch/hoppscotch/blob/5ac2ac40dedad5d871368b990392f472ca21dd40/packages/hoppscotch-data/src/collection/v/8.ts#L23-L47

For instance, referring to the `HoppRESTResponseOriginalRequest` schema version up to `v2` [here](https://github.com/hoppscotch/hoppscotch/blob/fix/secret-env-snippet-bug/packages/hoppscotch-data/src/rest/v/9.ts#L84-L94)

packages/hoppscotch-common/src/helpers/RequestRunner.ts
```diff
@@ -492,23 +492,11 @@ export function runRESTRequest$(
...
-
   const updatedRunResult = {
     ...runResult.right,
     envs: {
-      global: updatedGlobalEnvVariables,
-      selected: updatedSelectedEnvVariables,
+      global: runResult.right.envs.global,
+      selected: runResult.right.envs.selected,
     },
   }
```
jamesgeorge007: `updatedRunResult` can be removed given it has evolved to be the same as `runResult`. Also, the `updateEnvsAfterTestScript()` function can be updated to remove the return value given the current consumers don't expect a value.

packages/hoppscotch-common/src/helpers/fixBrokenEnvironmentVersion.ts
```diff
@@ -0,0 +1,33 @@
...
+}
```
jamesgeorge007:
```suggestion
export const fixBrokenEnvironmentVersion = (envs: Environment[]) => {
  if (!Array.isArray(envs)) {
    return envs
  }

  return envs.map((env) => ({
    ...env,
    variables: (env.variables ?? []).map(translateToNewEnvironmentVariables),
  }))
}
```

PR #5126, opened 2025-06-04
fix(desktop): in-memory update progress polling

### LINKED ISSUE #5038: [bug]: Update stuck in Windows

### Current behavior

It happens with every update; it never works. I have to download the .msi file from the website to update it

Image

### Steps to reproduce

Just try to open the application

### Logs and Screenshots

```shell

```

### Environment

Production

### Hoppscotch Version

Local

### Interceptor

### Browsers Affected

_No response_

### Operating System

Windows

### Additional Information

Windows 11 24h2
64 GB RAM
Ryzen 5 5600x
NVMe Gen 3

CuriousCorrelation:
This fixes a Windows-specific issue where the updater would hang during download and installation.

The cause was file I/O operations called from within progress callbacks conflicting with Windows' "stricter" fs behavior during concurrent ops.

This move that to an in-memory tracking with periodic polling, removing all disk writes from the download callback to only for actual final commits.

Closes HFE-892
Closes #5038

### Notes to reviewers

This is a Windows-specific thing, doesn't affect Mac/Linux functionality.
The core change is moving from synchronous progress persistence to asynchronous polling-based updates.

Progress tracking remains fully functional with 100ms polling intervals for just a bit smooth UI updates.
The existing implementation called `saveUpdateState()` directly within the updater's progress callback.

This approach worked fine on Mac/Linux but not on Windows apparently due to Windows' mandatory file locking behavior.

According to the docs Windows uses mandatory locking where if one program locks the file, all other programs get error messages when they try to access it, unlike Unix which generally doesn't use locking at all, and when it does, lock is usually advisory.

The Windows file locking behavior stems from its inheritance of ["sharing mode" from LAN Manager in the MS-DOS
era](https://superuser.com/questions/1711476/file-locking-in-windows-linux), which implements "mandatory locking as a special case" while "Linux doesn't have an equivalent to 'share modes' (all access is allowed by default)".

The updater's internal I/O ops conflicted with progress disk writes because Windows uses whole file locking for most cases, part of the API for opening files `CreateFile`.

So the fix implements a polling-based arch where progress is stored in memory during download and periodically read by the UI through a polling function.

### DIFF
--- a/packages/hoppscotch-desktop/src/types/index.ts
+++ b/packages/hoppscotch-desktop/src/types/index.ts
@@ -31,8 +31,4 @@ export interface UpdateState {
   status: UpdateStatus;
   version?: string;
   message?: string;
-  progress?: {
-    downloaded: number;
-    total?: number;
-  };
 }
--- a/packages/hoppscotch-desktop/src/utils/updater.ts
+++ b/packages/hoppscotch-desktop/src/utils/updater.ts
@@ -97,10 +97,6 @@ export class UpdaterService {
 
       await this.saveUpdateState({
         status: UpdateStatus.DOWNLOADING,
-        progress: {
-          downloaded: 0,
-          total: undefined
-        }
       });
 
       await updateResult.downloadAndInstall(
--- a/packages/hoppscotch-desktop/src/utils/updater.ts
+++ b/packages/hoppscotch-desktop/src/utils/updater.ts
@@ -4,12 +4,18 @@ import { type LazyStore } from "@tauri-apps/plugin-store";
 import { UpdateStatus, CheckResult, UpdateState } from "~/types";
 
 export class UpdaterService {
+  private currentProgress: { downloaded: number; total?: number } = { downloaded: 0 };
+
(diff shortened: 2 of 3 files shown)


packages/hoppscotch-desktop/src/types/index.ts
```diff
@@ -31,8 +31,4 @@ export interface UpdateState {
   status: UpdateStatus;
   version?: string;
   message?: string;
-  progress?: {
-    downloaded: number;
-    total?: number;
-  };
```
jamesgeorge007, line 37: The preexisting implementation has a type error. Can this be removed in favour of the new approach?

https://github.com/hoppscotch/hoppscotch/blob/f8cd8bfce70fedaa2b10574a102142c527fd8d5c/packages/hoppscotch-desktop/src/views/Home.vue#L167-L169

CuriousCorrelation: Good catch, pushed a commit removing it 🫡

jamesgeorge007 (approved):

PR #5141, opened 2025-06-11
fix(kernel): body size calculation in web relay

### LINKED ISSUE #5133: [bug]: response.meta.size.body is always 0 (response size is lower then expected)

### Current behavior

When I perform a request, I get a response but I noticed that the size is too low to be correct.
After debugging I noticed that the value of response.meta.size.body is always 0. But response.meta.size.headers does have value.

I expect response.meta.size.body to contain value, and the property .total to be the sum of .body and .headers properties.
Resulting in the correct display of the response size, in the result pane.

### Steps to reproduce

1. Create a new tab
2. Perform the default request (GET: https://echo.hoppscotch.io)
3. 200 OK response but size is 62 B. 
(If I calculated it correctly the response body should be 1,715 bytes + 62 bytes from the response header.)

### Logs and Screenshots

response.meta.size object in chrome devtools debugger

Image

Image

### Environment

Production

### Hoppscotch Version

Cloud

### Interceptor

### Browsers Affected

Chrome

### Operating System

Windows

### Additional Information

- Also fails in an incognito tab in browser

soumyamakkar:
Closes #5133

### What’s this PR about?

This PR fixes the incorrect size calculation of the response body in the `kernel-interceptors`. It ensures accurate measurement whether the body is an `ArrayBuffer`, `Uint8Array`, string, or object.

---

### What's changed

- 🐛 Fixed body size calculation logic in `kernel-interceptors/browser/index.ts`
- ✅ Added check for `Uint8Array.byteLength`
- 🧪 Verified accurate `meta.size.body` field in Relay response

---

### Notes to reviewers

- Please double-check the logic for response body types—especially where `body instanceof ArrayBuffer` vs `Uint8Array` might be ambiguous in some environments.
- I have verified that other interceptor platforms (like native/extension) do not require similar adjustments.


CuriousCorrelation:
Hi @soumyamakkar, great work on identifying and fixing the issue with the Browser interceptor.

Your idea of using a combination of `byteLength` and `length` by checking objects is exactly right. Although, I think we missed one interceptor in our analysis, take a look here:

image

Remember our initial thought about narrowing down which interceptors are experiencing this problem? It looks like Proxy was affected too, but we didn't catch it in the initial assessment.

So here's the interesting part - if you look at the Proxy interceptor code (`proxy/index.ts`), you'll notice it doesn't have the same response processing logic that you just fixed in the Browser interceptor. Instead, it calls:

https://github.com/hoppscotch/hoppscotch/blob/5ac2ac40dedad5d871368b990392f472ca21dd40/packages/hoppscotch-common/src/platform/std/kernel-interceptors/proxy/index.ts#L230

This tells us there's a deeper location in the **shared code** where this same size calculation issue exists.

Why shared? See this line in `browser/index.ts`

https://github.com/hoppscotch/hoppscotch/blob/c69f2f05dca9a2332f182c6a6658a9f294d9c90a/packages/hoppscotch-common/src/platform/std/kernel-interceptors/browser/index.ts#L58

Identical calls!

So if we fix it at that deeper level issue, it would actually fix the issue for both Browser and Proxy interceptors at once, rather than having to patch each interceptor individually.

The good news is your technical approach is spot on, you'll just need to trace where that calculation happens in the shared abstraction layers and apply the same fix there. That would be a more complete solution.

You're def on the right track, it's just a matter of finding the right shared location, so feel free to ask questions if you get stuck or need any help along the way.

CuriousCorrelation:
Hi @soumyamakkar, in response to your comment about the TypeScript error [here](https://github.com/hoppscotch/hoppscotch/issues/5133#issuecomment-2971385242) -- if those type errors are unrelated to your changes, you can safely ignore the git hooks when committing:

```bash
git commit --no-verify -m "your commit message"
```

The `--no-verify` flag bypasses git hooks, so you can commit even with unrelated type errors. Just make sure the errors are truly unrelated to your changes before using this approach.

You might also want to convert the PR to draft status while you're still working on it - this helps signal that it's work in progress.

soumyamakkar:
I have commited the files. Please review if everything works fine.

packages/hoppscotch-kernel/src/relay/impl/web/v/1.ts
```diff
@@ -222,6 +227,20 @@ export const implementation: VersionedAPI<RelayV1> = {
...
+              const bodySize =
+                rawBody instanceof ArrayBuffer
+                  ? rawBody.byteLength
+                  : ArrayBuffer.isView(rawBody)
+                    ? rawBody.byteLength
+                    : typeof rawBody === "string"
+                      ? rawBody.length
+                      : typeof rawBody === "object"
+                        ? JSON.stringify(rawBody).length
+                        : 0
```
CuriousCorrelation: I'm not sure if we need to check against these types if we know what the response type will be
https://github.com/hoppscotch/hoppscotch/blob/25a098f7038521435847aff589401996062ff0b2/packages/hoppscotch-kernel/src/relay/impl/web/v/1.ts#L141

We can rely on Axios' response type guarantee to simplify our approach.

packages/hoppscotch-selfhost-web/vite.config.ts
```diff
@@ -78,6 +78,18 @@ export default defineConfig({
...
+        __dirname,
+        "../hoppscotch-kernel/src/relay/v/1"
+      ),
+      "@store/v/1": path.resolve(
+        __dirname,
+        "../hoppscotch-kernel/src/store/v/1"
+      ),
+      "@io/v/1": path.resolve(__dirname, "../hoppscotch-kernel/src/io/v/1"),
+
+      // Existing node polyfills
```
CuriousCorrelation: This addition is unnecessary. Sometimes LLMs tends to hallucinate changes that might tangentially affect the solution, although here the comments such as 
```
// Add these new kernel-related aliases
```
and
```
// Existing node polyfills
```
are mainly intended at the prompter to understand the change and not actual code comments that explain why the change was necessary.

packages/hoppscotch-selfhost-web/vite.config.ts
```diff
@@ -95,6 +107,7 @@ export default defineConfig({
       dirs: ["../hoppscotch-common/src/pages", "./src/pages"],
       importMode: "async",
       onRoutesGenerated(routes) {
+        console.log("Routes generated:", routes.length)
```
CuriousCorrelation: We can remove debugging `console.log`s from the final PR. These are fine right now if this is a work in progress.

packages/hoppscotch-common/src/platform/std/kernel-interceptors/browser/index.ts
```diff
@@ -61,6 +61,7 @@ export class BrowserKernelInterceptorService
       promise.then((either) =>
         pipe(
           either,
+          E.map((response) => response),
```
CuriousCorrelation: Perhaps added for debugging purposes? We can remove this from the final PR when ready.

CuriousCorrelation:
@soumyamakkar This is the correct approach 👍 let's resolve the following notes left with review and we can consider this for one of the upcoming patch release.

soumyamakkar:
i have made all the necessary changes. ps: I had a question, you knew exactly what to change and where to change, why did you not change it yourself, wouldve been done in minutes.... like isnt "maintaining" and explaining everything to me much harder?

CuriousCorrelation (approved):
Good question @soumyamakkar, well, you are right, I could have done it myself faster but that definitely doesn’t take away from your efforts. 

The thing is, by working through this yourself, you now understand way more than just the fix and that knowledge I hope will help you with future contributions. The best part about open source is even if you contribute to completely different projects, your contributions will still benefit everyone. This is essentially the essence of open source.

And this is how I learned myself when I was just starting out, through maintainers who were patient enough to walk me through things. So sure, it takes more time upfront, but teaching something new is plenty rewarding in and of itself.

Thanks for sticking with it and asking thoughtful questions throughout.

LGTM 🚀

soumyamakkar:
yes, youre right... thanks a lot for guiding me :)

jamesgeorge007 (approved):

jessedubbink:
I was expecting this to be a quiet fix after submitting the issue, but after I saw @soumyamakkar asking to implement this I was interested in the progression. So I looked at my notifications once in a while.

@soumyamakkar thank you for your efforts and independence in asking to fix the issue.

And a big hats off to to you @CuriousCorrelation, for your guidance and willingness to help @soumyamakkar in solving this issue!

Beautiful to see.

PR #5156, opened 2025-06-16
fix(common): preserve large number precision

### LINKED ISSUE #5155: [feature]: add high-precision number support for desktop app

### Current behavior

When I pass a data like in the request body `"dsid": 990000000262240065`
Then it converts to the dsid value to `"dsid": 990000000262240100`

I cannot wrap with double quote because my server/API cannot allow different data type, it should stay the same data type as numerical.

### Steps to reproduce

1. Paste the payload
2. Send the request
3. Wrong response became because of the request payload has different dsid 

### Logs and Screenshots

```shell

```

### Environment

Production

### Hoppscotch Version

Cloud (Desktop App for Mac OS)

### Interceptor

### Browsers Affected

_No response_

### Operating System

None

### Additional Information

_No response_

CuriousCorrelation:
This fixes an issue in kernel processors where large integers in JSON request bodies lose precision during processing. Basically we were hitting the JavaScript's floating-point precision limitations affecting large integers.

Closes #5155
Closes HFE-899

When users input large numbers like `"dsid": 990000000262240065` in JSON request bodies, the value gets converted to `990000000262240100` due to precision loss during JSON processing. (See [Number.MAX_SAFE_INTEGER](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER))

The cause was the JSON processor was unnecessarily loss-y parsing JSON objects.

| Before | After |
|---|---|
| image | image |
| image | image |

### Testing

Use 
```
{ "dsid": 990000000262240065 }
```
with `application/json` content-type on any echo server such as `https://echo.hoppscotch.io` or use `nc -l 8080` with `localhost:8080` on the desktop app

| Request sent | Request received |
|---|---|
| image | image |

### DIFF
--- a/packages/hoppscotch-common/src/helpers/curl/sub_helpers/contentParser.ts
+++ b/packages/hoppscotch-common/src/helpers/curl/sub_helpers/contentParser.ts
@@ -160,7 +160,7 @@ const getXMLBody = (rawData: string) =>
   )
 
 const getFormattedJSON = flow(
-  safeParseJSON,
+  (str: string) => safeParseJSON(str, false, true),
   O.map((parsedJSON) => JSON.stringify(parsedJSON, null, 2)),
   O.getOrElse(() => "{ }")
 )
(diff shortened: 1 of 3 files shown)


auzundere:
@CuriousCorrelation It did not pass the requirements above. I think the changes are not affective

PR #5162, opened 2025-06-17
chore: use initial value in runtime if current value is empty

nivedin:
Closes HFE-890

This PR updates the environment usage flow where the initial environment variable value is used as the environment value if current value is empty at the time of request run without persisting the value.

### DIFF
--- a/packages/hoppscotch-common/src/helpers/utils/environments.ts
+++ b/packages/hoppscotch-common/src/helpers/utils/environments.ts
@@ -40,10 +40,7 @@ const unWrapEnvironments = (
     }
     return {
       ...globalVar,
-      currentValue:
-        currentVar?.currentValue ??
-        globalVar.currentValue ??
-        globalVar.initialValue,
+      currentValue: currentVar?.currentValue || globalVar.currentValue,
     }
   })
 
@@ -65,10 +62,7 @@ const unWrapEnvironments = (
       }
       return {
         ...selectedVar,
-        currentValue:
-          currentVar?.currentValue ??
-          selectedVar.currentValue ??
-          selectedVar.initialValue,
+        currentValue: currentVar?.currentValue || selectedVar.currentValue,
       }
     }
   )
--- a/packages/hoppscotch-common/src/helpers/utils/environments.ts
+++ b/packages/hoppscotch-common/src/helpers/utils/environments.ts
@@ -40,7 +40,7 @@ const unWrapEnvironments = (
     }
     return {
       ...globalVar,
-      currentValue: currentVar?.currentValue || globalVar.currentValue,
+      currentValue: currentVar?.currentValue || globalVar.currentValue || "",
     }
   })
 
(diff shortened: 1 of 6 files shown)


packages/hoppscotch-common/src/helpers/utils/environments.ts
```diff
@@ -40,10 +40,7 @@ const unWrapEnvironments = (
     }
     return {
       ...globalVar,
-      currentValue:
-        currentVar?.currentValue ??
-        globalVar.currentValue ??
-        globalVar.initialValue,
+      currentValue: currentVar?.currentValue || globalVar.currentValue || "",
```
jamesgeorge007, line 43: What's the intention behind removing the initial value fallbacks in these cases?

jamesgeorge007 (changes requested):
There is a corner case where secret environment variables under the global environment are picked over the selected environment during request execution with the variables sharing names.

  https://github.com/user-attachments/assets/97c1ab70-9feb-434c-8e62-d9377611786a
  
 > Discussed separately, we'll keep the initially proposed environment cascading behaviour subject to future discussions.

packages/hoppscotch-common/src/helpers/utils/environments.ts
nivedin: the initial value is used only in the time of request runtime [now](https://github.com/hoppscotch/hoppscotch/blob/6e297f550dfb94c68646d09841ababa5c8ab6f4e/packages/hoppscotch-common/src/helpers/RequestRunner.ts#L230), and the above code used to set the initial value as current value permanently which we do not want.

jamesgeorge007: Understood. It would be helpful to leave a comment noting the behaviour change for future reference

nivedin:
> There is a corner case where secret environment variables under the global environment are picked over the selected environment during request execution with the variables sharing names.
> 
>  shared-env-vars-corner-case.mov

This is intentional the priority of variables are 
Request Variable > Global > Selected Environment

The tooltip hover was incorrect, have pushed a [patch](https://github.com/hoppscotch/hoppscotch/pull/5162/commits/7ec98dccbf6328e56eab63f9d1eb53a7ccd114d9) for it

jamesgeorge007 (changes requested):
The environment priority behaviour is to be made consistent with non-secret environment variables.

nivedin:
> > There is a corner case where secret environment variables under the global environment are picked over the selected environment during request execution with the variables sharing names.
> > shared-env-vars-corner-case.mov
> 
> This is intentional the priority of variables are Request Variable > Global > Selected Environment
> 
> The tooltip hover was incorrect, have pushed a [patch](https://github.com/hoppscotch/hoppscotch/pull/5162/commits/7ec98dccbf6328e56eab63f9d1eb53a7ccd114d9) for it

Discussed separately Selected Environment Variable will have priority over Global Environment Variable

dungmv:
When the current value is empty, the tab code snippet fails to generate the curl command and displays the error message: "Something went wrong."

nivedin:
> When the current value is empty, the tab code snippet fails to generate the curl command and displays the error message: "Something went wrong."

@dungmv can you open a issue regarding this, we can track it from there

PR #5170, opened 2025-06-19
chore: rename test to post-request

nivedin:
Closes HFE-889

This PR renames the "Tests" title to "Post-request Script"

### What's changed
The i18n values is changed to the new texts.


jamesgeorge007 (changes requested):
Could you update all the `test script` instances under `i18n` string entries?

packages/hoppscotch-common/locales/en.json
jamesgeorge007, line 1: Could you also update the same for other languages?

jamesgeorge007 (approved):

PR #5184, opened 2025-06-22
fix: race condition in active team collection tab inheritance

### LINKED ISSUE #5183: [bug]: Auth type switches from Inherit to None after saving the Request

### Platform

Web App

### Browser

Chrome

### Operating System

macOS

## What happened?
Auth type `Inherit` changes to `None` in OAuth 2.0 authorization after `/oauth` callback

## Steps to reproduce?
- You may have to import a sizeable number API collections to observe this behavior
- Go to a child request, you will observe that the auth type is inherit
- Go to the parent collection's properties and `Generate Access Token`
- Save the access token after the `/oauth` callback
- You will find that the auth type changes from `inherit` to `none` and even with a valid token you would not be able to hit the API
- To verify the token was valid, you could copy the generated token and try a bearer request.

## Screen Recording
https://github.com/user-attachments/assets/a31a1423-9eff-414c-b2db-c89d4fe94782

### Deployment Type

Hoppscotch Cloud

### Version

2025.5.3

mayank-jain-1010:
Closes #5183 

This update addresses an issue where the `OAuth 2.0` authorization type unexpectedly switches from `Inherit` to `None` after the `/oauth` callback. This behavior stems from a `Race Condition`: the collection tree may not be fully loaded when the authentication inheritance logic is triggered.

Previously, this was mitigated with a `setTimeout` that assumed the collections would load within `200ms`. This PR replaces that hack with a more robust, event-driven mechanism.

### What's changed
- **Removed setTimeout workaround:** The arbitrary 200ms delay used to wait for collections to load is no longer used.
- **Introduced event-based handling:** Once the collections are fully loaded, the auth inheritance logic is executed, ensuring correct propagation of the inherited auth type to requests.

### Notes to reviewers
**Video demonstrating the fix:**
https://github.com/user-attachments/assets/b4aac572-669c-43fe-bc21-781b44331247

### Key things to note in the video:
- You’ll see the collection properties being saved before collections are fully loaded in the background.
- In the current (pre-fix) behavior, if a token is saved during this loading phase, the auth type unexpectedly switches to `None`.
- With this fix, auth inheritance works correctly regardless of collection load timing.

### DIFF
--- a/packages/hoppscotch-common/src/components/collections/index.vue
+++ b/packages/hoppscotch-common/src/components/collections/index.vue
@@ -3061,8 +3061,12 @@ const setCollectionProperties = (newCollection: {
       variables: collection.variables,
     }
 
-    // push the collectionId to the loading array
-    teamCollectionService.loadingCollections.value.push(collectionId)
+    // Mark as loading BEFORE triggering async update to avoid race conditions and push the collectionId to the loading array
+    if (
+      !teamCollectionService.loadingCollections.value.includes(collectionId)
+    ) {
+      teamCollectionService.loadingCollections.value.push(collectionId)
+    }
 
     pipe(
       updateTeamCollection(collectionId, JSON.stringify(data), undefined),


nivedin (changes requested):
can you check if the other auth types are updated correctly if an active tab is opened, the bearer token fails to update in this case

https://github.com/user-attachments/assets/244ebb88-4099-42bd-8fed-2148d238ed31

mayank-jain-1010:
Thanks for the catch @nivedin!
The primary reason for the behaviour you observed was that we were doing the computation of auth inheritance in a sync manner, whereas it should have been triggered during the success callback of `updateTeamCollection`

In addition to this, the propagation logic should be executed after the updation of `teamCollectionAdapter.collections$`
I have implemented the fix in this commit [16e9caa](https://github.com/hoppscotch/hoppscotch/pull/5184/commits/16e9caa9b45686a54d9530258f4556ef52bb93e8)

**Here is a video reference of the fix:**


https://github.com/user-attachments/assets/87ae0596-6fde-43ae-9175-ea2011bc71b9



Would appreciate a review on this.

mayank-jain-1010:
Hi @nivedin 
Just checking to see if you got the chance to review the above commit. Would love to discuss any feedbacks (or) suggestions. Thanks!

nivedin:
@mayank-jain-1010 can you rebase with the latest branch and add scope for collection variables

nivedin:
@mayank-jain-1010 Thank you for the PR, I’ve rebased and updated the PR to use the new team collection service. A watcher has been added to check if a collection is loading, and the collection path from the updated properties function is now used to update the active tab with the latest inherited changes.

cc: @jamesgeorge007

jamesgeorge007:
Thanks for the PR, @mayank-jain-1010. This is scheduled for the upcoming major release.

mayank-jain-1010:
Thanks @nivedin @jamesgeorge007
Appreciate your assistance with this!

PR #5185, opened 2025-06-22
fix(common): auto unescape utf-8 issue in the request body

### LINKED ISSUE #5159: [bug]: Hoppscotch auto unescape UTF-8 the request body

### Current behavior

**Behavior:**
When I sent the payload with escaped Unicode characters (example `\u0026`), Hoppscotch unescaped them before sending. This slows me down when I'm debugging my webhook signature verification.

**Expect:**
Hoppscotch should not modify the request body.

### Steps to reproduce

1. Start a sample express server:
```js
app.use(
  express.raw({ type: "application/json" });
);

app.post("/webhook", (req, res) => {
  console.log("Request body ", req.body);
  console.log("Original body", Buffer.from('{"product_name":"Abercrombie\\u0026 Fitch"}'));
  res.send("OK");
});
```

2. Send request through Hoppscotch with this body
```json
{"product_name": "Abercrombie\u0026 Fitch"}
```
Output:
```shell
Request body  <Buffer 7b 22 70 72 6f 64 75 63 74 5f 6e 61 6d 65 22 3a 22 41 62 65 72 63 72 6f 6d 62 69 65 26 20 46 69 74 63 68 22 7d>
Original body <Buffer 7b 22 70 72 6f 64 75 63 74 5f 6e 61 6d 65 22 3a 22 41 62 65 72 63 72 6f 6d 62 69 65 5c 75 30 30 32 36 20 46 69 74 63 68 22 7d>
```

3. Send this cURL command
```shell
curl --request POST \
  --url http://localhost:8000/webhook \
  --header 'x-signature: OfXbfBq8Dz4WluGvAEizOI4gTi6qm9+1fqVMP2tR6CI=' \
  --data '{"product_name":"Abercrombie\u0026 Fitch"}'
```
Output:
```shell
Request body  <Buffer 7b 22 70 72 6f 64 75 63 74 5f 6e 61 6d 65 22 3a 22 41 62 65 72 63 72 6f 6d 62 69 65 5c 75 30 30 32 36 20 46 69 74 63 68 22 7d>
Original body <Buffer 7b 22 70 72 6f 64 75 63 74 5f 6e 61 6d 65 22 3a 22 41 62 65 72 63 72 6f 6d 62 69 65 5c 75 30 30 32 36 20 46 69 74 63 68 22 7d>
```

4. Compare the outputs of 2 and 3, the difference in Hoppscotch case is the request body having `26` (&) instead of `5c 75 30 30 32 36` (\u0026)

### Logs and Screenshots

```shell

```

### Environment

Production

### Hoppscotch Version

Cloud

### Interceptor

Native - Desktop App

### Browsers Affected

_No response_

### Operating System

MacOS

### Additional Information

_No response_

anwarulislam:
Closes HFE-905 #5159

This PR fixes the issue of auto unescaping UTF-8 in the request body. Maintain original payload integrity in HTTP request.

### DIFF
--- a/packages/hoppscotch-common/src/helpers/kernel/common/content.ts
+++ b/packages/hoppscotch-common/src/helpers/kernel/common/content.ts
@@ -12,7 +12,7 @@ const Processors = {
     process: (body: string): E.Either<Error, ContentType> =>
       pipe(
         parseJSONAs<unknown>(body),
-        E.map(() => content.json(body, MediaType.APPLICATION_JSON)),
+        E.map(() => content.text(body, MediaType.APPLICATION_JSON)),
         E.orElse(() => E.right(content.text(body, MediaType.TEXT_PLAIN)))
       ),
   },
--- a/packages/hoppscotch-common/src/helpers/kernel/common/content.ts
+++ b/packages/hoppscotch-common/src/helpers/kernel/common/content.ts
@@ -9,12 +9,27 @@ import { EffectiveHoppRESTRequest } from "~/helpers/utils/EffectiveURL"
 
 const Processors = {
   json: {
-    process: (body: string): E.Either<Error, ContentType> =>
-      pipe(
-        parseJSONAs<unknown>(body),
-        E.map((json) => content.json(json, MediaType.APPLICATION_JSON)),
-        E.orElse(() => E.right(content.text(body, MediaType.TEXT_PLAIN)))
-      ),
+    process: (body: string): E.Either<Error, ContentType> => {
+      const hasUnicodeEscapes = /\\u[0-9a-fA-F]{4}/.test(body)
+      if (hasUnicodeEscapes) {
+        return pipe(
+          E.tryCatch(() => {
+            JSON.parse(body) // Just validating it's valid JSON
+            return body
+          }, E.toError),
+          E.map((originalBody) =>
+            content.text(originalBody, MediaType.APPLICATION_JSON)
+          )
+        )
+      } else {
+        // For regular JSON without Unicode escapes, using the standard processing
+        return pipe(
+          parseJSONAs<unknown>(body),
+          E.map((json) => content.json(json, MediaType.APPLICATION_JSON)),
+          E.orElse(() => E.right(content.text(body, MediaType.TEXT_PLAIN)))
+        )
+      }
+    },
   },
 
   binary: {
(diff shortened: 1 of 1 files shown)


packages/hoppscotch-common/src/helpers/kernel/common/content.ts
```diff
@@ -9,12 +9,27 @@ import { EffectiveHoppRESTRequest } from "~/helpers/utils/EffectiveURL"
...
+        )
+      } else {
+        // For regular JSON without Unicode escapes, using the standard processing
+        return pipe(
+          parseJSONAs<unknown>(body),
+          E.map((json) => content.json(json, MediaType.APPLICATION_JSON)),
+          E.orElse(() => E.right(content.text(body, MediaType.TEXT_PLAIN)))
+        )
+      }
+    },
```
anwarulislam: There's another way to handle JSON data in `content.ts` file.

```typescript
  // including Unicode escape sequences, whitespace, and other literal representations
  E.right(content.text(body, MediaType.APPLICATION_JSON))
```

When we send JSON data in an HTTP request, we are actually sending it as text. Specifically, the JSON is serialized into a string format, which is then transmitted over the network as part of the HTTP request body. This means that the JSON data is not sent as a native data structure but rather as a string representation of that structure.

CuriousCorrelation: Works perfectly on both web and desktop app 🚀 

| Request Sent | Request Captured |
| --- | --- |
| image | image |


@anwarulislam since the right approach here is as you suggested `content.text(body, MediaType.APPLICATION_JSON)`, what are your thoughts on simplifying it further like so


```typescript
import { isJSON } from "~/helpers/functional/json"

const Processors = {
  json: {
    process: (body: string): E.Either<Error, ContentType> => {
      return isJSON(body)
        ? E.right(content.text(body, MediaType.APPLICATION_JSON))
        : E.right(content.text(body, MediaType.TEXT_PLAIN))
    },
  },
```

since in case of errors we are letting `content.text` handle unexpected / non-json-like requests anyways?

CuriousCorrelation:
Seeing different results when using **Native** and **Agent** interceptor. Looking into `relay` code to figure out why.

image

CuriousCorrelation:
Going with the `content.text` approach for the moment, type alignment on `content.ts` and `kernel/relay` will be looked at separately once the above approach is understood to be stable from user responses. This will be related to `hoppscotch-kernel` module.

Tracking in HFE-910

jamesgeorge007 (approved):

Issue #4723, opened 2025-02-04
[bug]: Resolve and display global variable values within workspace environment variables


iDschepe:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Summary

**Description:**
It would be highly beneficial if Hoppscotch supported nested variables, similar to how Postman handles them. This feature allows variables to reference other variables, making environment management more dynamic and flexible.

**Expected Behavior:**
Variables should be able to reference other variables using {{variable_name}}.
When evaluating a request, Hoppscotch should resolve nested variables recursively.
Nested variables should work in all applicable fields, including environment variables, request parameters, headers, and body content.

**Use Case Example:**

A user defines an environment variable: 
`base_url = https://api.example.com`

Another variable references it: 
`users_endpoint = {{base_url}}/users`

When making a request to <<users_endpoint>>, Hoppscotch should resolve it to: 
`https://api.example.com/users`

OR

For example, a team could define a global variable:
`api_version: v1`

Then, in an environment-specific setting, this could be reused dynamically:
`api_base_url: https://api.example.com/<<api_version>>`

**Potential Implementation:**
- Enable recursive resolution of variables in requests.
- Implement safeguards to prevent infinite loops (e.g., `var_a = <<var_b>>` and `var_b = <<var_a>>`).

Would this feature be possible to add? It would greatly improve variable handling and make Hoppscotch even more powerful for API testing.

### Why should this be worked on?

**Why is this useful?**

- Reduces duplication: Users don’t need to manually update values in multiple places.
- With nested variables, global environment variables could be leveraged efficiently in different environments.
- Enables dynamic configuration: Users can define base values and build other variables based on them.
- Improves team efficiency: Teams can work with structured and reusable environment variables.

liyasthomas:
Nested variables are already possible. Please let me know if this is what you were expecting.

<img width="1470" alt="Image" src="https://github.com/user-attachments/assets/166d5f68-8fba-49d7-b9aa-fa2358eb7e7c" />

<img width="1470" alt="Image" src="https://github.com/user-attachments/assets/437f5ab2-df73-459c-9a77-df5a07fdce9c" />

iDschepe:
Hi @liyasthomas, 

Thank you for your quick response! That’s exactly what I was looking for. 

1) Defined global var `myGlobalVar`

![Image](https://github.com/user-attachments/assets/07166c33-aeb2-42b3-9b45-e8a0bee02730)

[...]

PR #5187, opened 2025-06-23
chore: resolve global env in team env and tooltip UI update

nivedin:
Closes HFE-906 #4723 

This PR fixes the issue with global environments not getting resolve while using in a shared workspace environment as a nested environment ([ref](https://github.com/hoppscotch/hoppscotch/issues/4723#issuecomment-2990197241)).

This PR also improves the UI of the environment tooltip by including initial and current environment value with proper info about the environment.

https://github.com/user-attachments/assets/b421f081-5f6c-4cd5-b79b-2691281330a6


packages/hoppscotch-common/assets/scss/styles.scss
```diff
@@ -169,7 +168,7 @@ a {
   @apply shadow;
 
   .tippy-content {
-    @apply flex;
+    @apply flex flex-col;
```
jamesgeorge007, line 171: The flex direction update affects a few existing spots.

image

nivedin: Good catch, have pushed an update

jamesgeorge007 (approved):

PR #5190, opened 2025-06-24
feat(desktop): native tab keyboard shortcuts

### LINKED ISSUE #3520: [feature]: Add tab-specific keyboard shortcuts  in Desktop app

### Summary

I want to add tab-specific shortcuts.

- ctrl/cmd + tab: move next tab
- ctrl + shitf + tab: move previous tab
- ctrl + t: new tab
- ctrl + w: close tab
- ctrl + 1,2...9: move to tab number

.. etc

Convenience

CuriousCorrelation:
This adds native menu items with keyboard accelerators for tab management ops for the desktop app. These are standard desktop app shortcuts for tab control.

Closes HFE-907
Closes #3520
Closes HP-24

### Shortcuts added

- **Ctrl/Cmd+T**: Open new tab
- **Ctrl/Cmd+W**: Close current tab
- **Ctrl/Cmd+Alt+Left**: Navigate to previous tab
- **Ctrl/Cmd+Alt+Right**: Navigate to next tab
- **Ctrl/Cmd+9**: Corresponds to `(`, navigate to the left-most tab
- **Ctrl/Cmd+0**: Corresponds to `)`, navigate to the right-most tab

9 comes before 0 on QWERTY so it seems more ergonomic.

### Implementation details

The impl creates menu items with platform-appropriate accelerators (Cmd/Ctrl) that emit events to the frontend keybinding module so tab navigation and management can work.

The architecture uses Tauri's native menu system to register accelerators at the OS level, which then emit
`hoppscotch_desktop_shortcut` events to the frontend.

This approach makes shortcuts work globally within the app window and follow platform conventions. Global shortcuts would have prevented shortcuts from reaching other apps (global shortcuts are system level).

The FE keybinding system is extended to support desktop-specific bindings through kernel mode detection, to separate base web shortcuts from desktop-only shortcuts like these (mentioned above).

Some additions to tab service: methods like `goToNextTab()`, `goToPreviousTab()`, `goToLastTab()` plus some foundation / plumbing for future shortcut expansion.

### Notes to reviewers

This also addresses the UX issue where users would accidentally close the entire app when trying to close tabs with Ctrl/Cmd+W.

The shortcuts now match standard desktop application behavior across most browsers (firefox, chrome, etc) and also IDEs to an extent (think VS code). The desktop shortcuts only activate when running in desktop kernel mode, so no changes  web app behavior.

### DIFF
--- a/packages/hoppscotch-common/src/helpers/keybindings.ts
+++ b/packages/hoppscotch-common/src/helpers/keybindings.ts
@@ -129,7 +129,7 @@ export function hookKeybindingsListener() {
           }
         )
       } catch (error) {
-        console.warn("Failed to setup Tauri event listener:", error)
+        console.error("Failed to setup Tauri event listener:", error)
       }
     }
   })


packages/hoppscotch-common/src/helpers/keybindings.ts
```diff
@@ -71,18 +79,68 @@ export const bindings: {
...
+  onMounted(async () => {
     document.addEventListener("keydown", handleKeyDown)
+
+    // Listen for Tauri events (desktop only)
+    if (getKernelMode() === "desktop") {
+      try {
+        unlistenTauriEvent = await listen(
+          "hoppscotch_desktop_shortcut",
+          (ev) => {
+            console.info("Tauri shortcut ev", ev)
```
jamesgeorge007, line 127: Are these console logs intentional?

packages/hoppscotch-common/src/helpers/keybindings.ts
```diff
@@ -71,18 +79,68 @@ export const bindings: {
...
+            handleTauriShortcut(ev.payload as string)
+          }
+        )
+      } catch (error) {
+        console.warn("Failed to setup Tauri event listener:", error)
```
jamesgeorge007, line 132: Let's use `console.warn` instead.

AndrewBastin (approved):

packages/hoppscotch-common/src/helpers/keybindings.ts
jamesgeorge007: Would it be appropriate to use `console.error()` here instead?

packages/hoppscotch-common/src/helpers/keybindings.ts
CuriousCorrelation: Yes, although temporarily, just till we stabilize this impl.

I ran into an issue with `AppImage` where I just wasn't seeing any hotkeys fired, and couldn't figure out why.

I then cleared all build artifacts, restarted my machine and the events started working again which was very strange. Not sure it was some platform issue or something else entirely.

So I'm thinking, since the impl is very platform dependent, we have this log here just so we could have diagnostic info to correlate with desktop app's log, just until we stabilize the impl instead of releasing a patch only to add temporary logging.

We can remove it once we make sure things are working correctly on different user environments 🫡

packages/hoppscotch-common/src/helpers/keybindings.ts
CuriousCorrelation: Yep, makes sense, have pushed a fix

packages/hoppscotch-common/src/helpers/keybindings.ts
jamesgeorge007: Sure, we can keep it.

jamesgeorge007 (approved):

Issue #5198, opened 2025-06-26
[bug]: Code snippet fails to generate


dungmv:
### Is there an existing issue for this?

- [x] I have searched existing issues and this bug hasn't been reported yet

### Platform

Desktop App

### Browser

Chrome

### Operating System

macOS

### Bug Description

When the environment's current value is empty, the tab code snippet fails to generate the curl command and displays the error message: "Something went wrong."

### Deployment Type

Hoppscotch Cloud

kspoojith:
Hi! I’m new to contributing here — may I work on fixing the code snippet generation when the environment variable is empty?

liyasthomas:
@kspoojith, you can try this: ideally, two cases need to be resolved:
1. Environment variables used in request setup are not resolved from the current value. In this case, the snippet generator should fall back to the initial value if the current value is empty.
2. When the URL field is empty, the URL should be optional in the snippet generator.

jamesgeorge007:
Hi, closing this issue since it's addressed in the latest release.

PR #5214, opened 2025-07-01
fix: fallback env to initial and make valid url in codegen

nivedin:
Closes FE-913 #5198 

This PR fixes the issue where the codegen view breaks if environment failed to fallback to initial environment value when current value is empty and URL is empty.

### What's changed
The environment will fallback to initial value if current value is empty and if URL is empty add an 'https://' value so that httpsnippet does not break.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,13 +57,6 @@
       "brace-expansion@2.0.1": "2.0.2",
       "brace-expansion@1.1.11": "1.1.12"
     },
-    "packageExtensions": {
-      "@hoppscotch/httpsnippet": {
-        "dependencies": {
-          "ajv": "6.12.3"
-        }
-      }
-    },
     "onlyBuiltDependencies": [
       "@apollo/protobufjs",
       "@import-meta-env/unplugin",
--- a/packages/hoppscotch-common/package.json
+++ b/packages/hoppscotch-common/package.json
@@ -36,7 +36,7 @@
     "@codemirror/view": "6.25.1",
     "@hoppscotch/codemirror-lang-graphql": "workspace:^",
     "@hoppscotch/data": "workspace:^",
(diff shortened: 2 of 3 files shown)


anwarulislam (approved):

packages/hoppscotch-common/src/components/http/Codegen.vue
```diff
@@ -202,6 +202,27 @@ const getCurrentValue = (env: AggregateEnvironment) => {
...
+    const domain = endpoint.split(/[/:#?]+/)[0]
+    if (domain === "localhost" || /([0-9]+\.)*[0-9]/.test(domain)) {
+      url = "http://" + endpoint
+    } else {
+      url = "https://" + endpoint
+    }
+  }
+
+  return url
+}
```
jamesgeorge007: We'd also need to account for initial and trailing whitespaces as part of the endpoint. How about the below implementation?

```ts
const getFinalURL = (input: string): string => {
  if (!input) {
    return "https://"
  }

  let url = input.trim()

  // Fix malformed protocols
  url = url.replace(/^https?:\s*\/+\s*/i, (match) =>
    match.toLowerCase().startsWith("https") ? "https://" : "http://"
  )

  if (!/^https?:\/\//i.test(url) && !url.startsWith("<<")) {
    const endpoint = url
    const domain = endpoint.split(/[/:#?]+/)[0]

    const isLocalOrIP = /^(localhost|(\d{1,3}\.){3}\d{1,3})$/.test(domain)
    url = (isLocalOrIP ? "http://" : "https://") + endpoint
  }

  return url
}
```

jamesgeorge007 (changes requested):
Could you also investigate the failure of code snippet generation when secret environment variables are included in the endpoint?

jamesgeorge007 (approved):

PR #5231, opened 2025-07-08
fix: allow aws signature to work with query params

### LINKED ISSUE #4668: [bug]: Amazon Signature fails when using the Query Parameters table

### Current behavior

When using the Query Parameters table for an HTTP method that uses Amazon Signature for authentication, the request will fail to authenticate, with the response message:

> "The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details.
> 
> The Canonical String for this request should have been \<calculated string\>"

It appears that query parameters from the table are not being properly appended to the canonical URL string that is used in the calculation of the signature. In contrast, manually appending the query parameter directly in the URL field works correctly.

### Steps to reproduce

1. Create HTTP request for an API endpoint that requires Amazon Signature authentication. Confirm it is working.
2. Add a query parameter to the Query Parameters table. Request should fail.

### Environment

Production

### Version

Cloud

rafalaidlaw:
Closes #4668 

### What's changed
Implemented query handling for aws signature authentication, added tests

### DIFF
--- a/packages/hoppscotch-common/src/helpers/auth/types/aws-signature.ts
+++ b/packages/hoppscotch-common/src/helpers/auth/types/aws-signature.ts
@@ -3,21 +3,29 @@ import {
   HoppRESTAuth,
   HoppRESTHeader,
   HoppRESTParam,
+  HoppRESTParams,
   HoppRESTRequest,
   parseTemplateString,
 } from "@hoppscotch/data"
 import { AwsV4Signer } from "aws4fetch"
 import { getFinalBodyFromRequest } from "~/helpers/utils/EffectiveURL"
 
+type SignOptions = {
+  auth: HoppRESTAuth & { authType: "aws-signature" }
+  request: HoppRESTRequest
+  envVars: Environment["variables"]
+  signQuery?: boolean
+}
+
 function processQueryParameters(
-  request: HoppRESTRequest,
+  params: HoppRESTParams,
   envVars: Environment["variables"],
   baseUrl: string
 ): { url: URL; sortedParams: Array<{ key: string; value: string }> } {
   const url = new URL(baseUrl)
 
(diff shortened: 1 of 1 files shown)


anwarulislam:
Hey, @rafalaidlaw!
I wasn't able to reproduce issue #4668. Have you checked properly if the query parameters and the body type application/x-www-form-urlencoded that you're accepting through that API?

rafalaidlaw:
Hey @anwarulislam,

I confirmed that the signature works without query params, like this:
Screenshot 2025-07-09 at 11 50 53 AM

But does not work with them, like this:
Screenshot 2025-07-09 at 11 51 41 AM

anwarulislam:
@rafalaidlaw, could you please share the error you're experiencing in the second screenshot?

rafalaidlaw:
Hi @anwarulislam , 

The error will be AWS `SignatureDoesNotMatch`. I recorded my screen with the error and then the fix below, it will explain the issue clearly. Thank you!

https://github.com/user-attachments/assets/fdc6d178-ebed-4101-b6fb-d576b56477e7

anwarulislam:
Hey @rafalaidlaw, I made a few changes. Hope everything's working well!

anwarulislam (approved):

rafalaidlaw:
Hi @anwarulislam I checked the latest changes, they are working, you can find the video below, thank you

https://github.com/user-attachments/assets/82cd68a9-5e92-4d05-8598-3d0b1ecd305c

jamesgeorge007 (approved):

jamesgeorge007:
Thanks for the PR, @rafalaidlaw. I've pushed a few cleanup changes in https://github.com/hoppscotch/hoppscotch/pull/5231/commits/c4fbe930c7c0cbc6e954581879a25bbbb40f653d. This will be included in the upcoming major release.

Issue #5177, opened 2025-06-20
[bug]: OpenAPI import - request baseUrl variable missing


MSP-Stev:
### Is there an existing issue for this?

- [x] I have searched existing issues and this bug hasn't been reported yet

### Platform

Web App

### Browser

Chrome

### Operating System

Windows

### Bug Description

The OpenAPI import creates all request without baseUrl variable.
Cloud & Self-Hosted, Web App and Desktop app have same result.

## Steps to reproduce
1. open Import dialog
2. Select "Import from OpenAPI"
3. choose local file
4. import [example.json](https://github.com/user-attachments/files/20834495/example.json)
5. All requests only show "./" in the url without baseUrl variable

<img width="691" alt="Image" src="https://github.com/user-attachments/assets/a192160b-668d-4a95-8dd3-922ee76a13a2" />

## Screenshot
Other API clients replace "./" to baseUrl, as far i can tell

<img width="552" alt="Image" src="https://github.com/user-attachments/assets/a2fddb2c-3be0-496e-a11a-08e56281307e" />

<img width="509" alt="Image" src="https://github.com/user-attachments/assets/bd1aa870-f3be-417b-adc9-a990c0ada8d6" />

### Deployment Type

Hoppscotch Cloud

SujalBhatt:
Hi !! I'd love to work on this issue as a first-time contributor. It looks like the import logic isn’t correctly using the `servers.url` from the OpenAPI spec — I’ll investigate and submit a fix. Let me know if it’s okay to proceed!

liyasthomas:
@SujalBhatt you can give it a try.

SujalBhatt:
Thank you so much @liyasthomas , I will try my best !!

jamesgeorge007:
Hi @SujalBhatt, thanks for taking on this issue!

This one was prioritised for an upcoming patch release, so @anwarulislam from the team has addressed it in #5232.
Appreciate your willingness to help - please feel free to check out other open issues in the meantime. We’d love to have you involved in future contributions!

jamesgeorge007:
Closing this issue since it's addressed in the latest release.

PR #5232, opened 2025-07-09
feat(common): enhance URL parsing to use environment variables on openapi import

anwarulislam:
Closes FE-914 #5177

Improve URL parsing to utilize environment variables for host and base path, allowing for more dynamic configurations. Clean up code for better readability.

### DIFF
--- a/packages/hoppscotch-common/src/helpers/import-export/import/openapi.ts
+++ b/packages/hoppscotch-common/src/helpers/import-export/import/openapi.ts
@@ -756,9 +756,9 @@ const parseOpenAPIUrl = (
   if (objectHasProperty(doc, "swagger")) {
     // TODO: dynamically add doc.host, doc.basePath value as variables in the environment if available. or notify user to add it.
     // add base url variable to each request
-    const hostVar = "<<baseUrl>>"
-    const basePathVar = doc.basePath ? "<<basePath>>" : ""
-    return `${hostVar}${basePathVar}`
+    const host = doc.host?.trim() || "<<baseUrl>>"
+    const basePath = doc.basePath?.trim() || ""
+    return `${host}${basePath}`
   }
 
   /**


anwarulislam:
I've included a TODO in the code to enhance it later. Essentially, since we incorporate the base URL and base path variable into the URL, the values are currently missing. Therefore, it's best to add those variables to the environment or collection level variable (will be discussed later)

packages/hoppscotch-common/src/helpers/import-export/import/openapi.ts
```diff
@@ -754,19 +754,21 @@ const parseOpenAPIUrl = (
...
    * OpenAPI V3 has version as a string in the document's openapi property.
    * And host and basePath are in the document's servers property.
    * Relevant v3 reference: https://swagger.io/specification/#server-object
    **/
-  if (objectHasProperty(doc, "servers")) {
-    return doc.servers?.[0]?.url ?? "<<baseUrl>>"
-  }
+  // if (objectHasProperty(doc, "servers")) {
+  // TODO: dynamically add doc.servers[0].url value as variable in the environment if available. or notify user to add it.
+  // }
```
jamesgeorge007:
```suggestion
if (objectHasProperty(doc, "servers")) {
    // TODO: dynamically add server URL value as variable in the environment if available, or notify user to add it.
    const serverUrl = doc.servers?.[0]?.url

    if (!serverUrl || serverUrl === "./") {
      return "<<baseUrl>>"
    }

    return serverUrl
  }
```

As adding the server URL to environment variables is not currently supported, we can extend the business logic to accommodate specific use cases for flexibility, with a comment noting this for future review.

packages/hoppscotch-common/src/helpers/import-export/import/openapi.ts
```diff
@@ -754,19 +754,21 @@ const parseOpenAPIUrl = (
    **/
 
   if (objectHasProperty(doc, "swagger")) {
-    const host = doc.host?.trim() || "<<baseUrl>>"
-    const basePath = doc.basePath?.trim() || ""
-    return `${host}${basePath}`
+    // TODO: dynamically add doc.host, doc.basePath value as variables in the environment if available. or notify user to add it.
+    // add base url variable to each request
+    const hostVar = "<<baseUrl>>"
+    const basePathVar = doc.basePath ? "<<basePath>>" : ""
```
jamesgeorge007: We can keep the existing behaviour until the support for environment variable addition is brought.

jamesgeorge007 (changes requested):

Issue #5204, opened 2025-06-26
[bug]: Email shown in profile is outdated, and cannot be updated


evenkov:
### Is there an existing issue for this?

- [x] I have searched existing issues and this bug hasn't been reported yet

### Platform

Web App

### Browser

Chrome

### Operating System

Windows

#### Problem

After changing my **primary email address in GitHub**, Hoppscotch continues to show my **old email address** in the profile.

I have done the following steps:

- Set the new email as **primary**, **verified**, and **public** in GitHub
- Removed the old email as primary (but kept it as a secondary)
- Revoked Hoppscotch's OAuth access in GitHub
- Logged in again via GitHub

Still, Hoppscotch displays the **old email address** in my profile.

#### GitHub API Findings

I made a call to `https://api.github.com/user/emails` and confirmed that:

- The correct new email is listed with `"primary": true`, `"verified": true`, and `"visibility": "public"`
- However, it is the **last email in the array**
- The email shown by Hoppscotch is the **second one**, which was previously my primary email

➡️ This strongly suggests that Hoppscotch selects the email based on index (e.g. `emails[1]`) or uses a cached email, instead of selecting the email object where `primary === true`.

#### Additional Bug: Manual email change fails

When I try to manually update the email from my Hoppscotch profile:

- A GitHub Authentication popup appears
- After completing authentication, I get the error:  
  **"Something went wrong"**
- The email is not updated

#### Expected Behavior

- Hoppscotch should retrieve and display the `primary` email from the GitHub `/user/emails` response
- If manual update is allowed, it should complete successfully or show a meaningful error

#### Suggested Fix

- Update the logic to select email by checking `primary === true`
- Allow manual update of email (or make it editable if it's only used for display)
- Optionally, add a "Refresh from GitHub" button to force sync profile data

#### Thanks!

Let me know if I can help test anything further — this issue affects GitHub-authenticated users who change their email.

### Deployment Type

Hoppscotch Cloud

nivedin:
@evenkov Thank you for the detailed explanation, can you confirm that the new email you are setting is not related to any other active account previously created in hoppscotch.

With regarding email not showing the updated GitHub email, this is expected since we have account linked to each email and provider id, you can update the email from profile page just that as I mentioned above the new email should not be not be linked with any other account in hoppscotch.

We have a [PR](https://github.com/hoppscotch/hoppscotch/pull/5247) opened to add proper error message if the the request fail, thank you for bringing this up

[...]

PR #5247, opened 2025-07-15
chore: add proper error msg and disable email updation in SH

nivedin:
Related to [#110](https://github.com/hoppscotch/central/pull/110)

Closes FE-918 #5204 

This PR add's proper error message if email updation fails and disables updation if the platform disables it.

### DIFF
--- a/packages/hoppscotch-common/src/pages/profile.vue
+++ b/packages/hoppscotch-common/src/pages/profile.vue
@@ -241,7 +241,7 @@ const probableUser = useReadonlyStream(
 )
 
 const isEmailEditable = computed(() => {
-  return platform.auth.isEmailEditable ?? true
+  return platform.auth.isEmailEditable ?? false
 })
 
 const loadingCurrentUser = computed(() => {
--- a/packages/hoppscotch-common/src/platform/auth.ts
+++ b/packages/hoppscotch-common/src/platform/auth.ts
@@ -272,6 +272,7 @@ export type AuthPlatformDef = {
   /**
    * Whether the email address is editable by the user or not.
    * This is used to determine whether the email address field should disabled in the user settings.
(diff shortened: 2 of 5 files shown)


packages/hoppscotch-common/src/pages/profile.vue
```diff
@@ -290,17 +296,25 @@ const updateEmailAddress = () => {
...
+
+  const result = await platform.auth.setEmailAddress(inputEmailAddress)
+
+  if (!result) {
+    toast.error(`${t("error.something_went_wrong")}`)
+    updatingEmailAddress.value = false
+    return
+  }
+
+  if (result.type === "success") {
```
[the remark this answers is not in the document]
nivedin: we don not show the profile page if user is not logged in, so this case is handled

packages/hoppscotch-common/src/pages/profile.vue
```diff
@@ -238,6 +240,10 @@ const probableUser = useReadonlyStream(
   platform.auth.getProbableUser()
 )
 
+const isEmailEditable = computed(() => {
+  return platform.auth.isEmailEditable ?? true
+})
```
jamesgeorge007: Given that email editing is reserved for the cloud offering, we could consider making this setting opt-in, defaulting `isEmailEditable` to `false`, and explicitly enabling it via the platform definition for cloud. This would help reduce the configuration surface area and make the intent clearer.

jamesgeorge007 (approved):

packages/hoppscotch-common/src/pages/profile.vue
jamesgeorge007: Addressed in https://github.com/hoppscotch/hoppscotch/pull/5247/commits/b5d813ff98fa2ab81cf801421823f1451a46d88d.

PR #5249, opened 2025-07-15
fix: handle actions for logged-in users in case of token expiration

### LINKED ISSUE #5167: [bug]: Collections not persisted when access token expires

### Current behavior

When a user’s access token is expired, creating a collection still returns a 200 OK response from the API. However, the response body contains an "unauthenticated" error.
Since the UI does not surface this error, it appears to the user that the collection has been successfully created.
In reality, the collection is not saved to the database, and when the app is refreshed (and a new access token is generated using the refresh token), the collection disappears without any trace—leading to confusion and potential data loss.

### Steps to reproduce

1. Let the user session idle until the access token expires (refresh token still valid).
2. Create a new collection in the personal workspace.
3. Observe that the collection appears in the UI as if it’s saved.
4. Refresh the app.
5. The newly created collection is no longer present.

### Logs and Screenshots

https://github.com/user-attachments/assets/4ce7e368-b4f6-424a-ad33-e6907c48890b

### Environment

Production

### Hoppscotch Version

Self-hosted

### Interceptor

### Browsers Affected

_No response_

### Operating System

None

### Additional Information

_No response_

chhaviG22:
Closes #5167 

This PR addresses the issue where, when a user’s access token expires, attempting to create a collection leads to an API response with a 200 OK status, but the response body contains an "unauthenticated" error. As a result of this inconsistency, users are misled into thinking the collection was successfully created. In reality, due to the expired token, the collection is never saved in the database, and refreshing the app leads to loss of the collection without any trace—causing confusion and potential data loss.

### What's changed

- Returning the error received from the API when the token expires during collection creation.
- Resetting the local store by removing the collection that failed to save on the backend due to token expiration.
- Displaying the error on the UI to inform the user about the issue.
- Clearing the error from the state to ensure the local store is reset.

### DIFF
--- a/packages/hoppscotch-common/src/helpers/handleTokenValidation.ts
+++ b/packages/hoppscotch-common/src/helpers/handleTokenValidation.ts
@@ -1,6 +1,16 @@
 import { useToast } from "@composables/toast"
 import { isValidUser } from "~/helpers/isValidUser"
 
+/**
+ * High-level authentication validation handler with automatic error notifications.
+ *
+ * This wrapper around `isValidUser()` provides automatic toast error messages for invalid tokens.
+ * Use this when you want standard error handling with user notifications.
+ *
+ * For silent validation or custom error handling, use `isValidUser()` directly.
+ *
+ * @returns {Promise<boolean>} True if user is valid, false otherwise (with toast error shown)
+ */
 export const handleTokenValidation = async (): Promise<boolean> => {
   const toast = useToast()
   const { valid, error } = await isValidUser()
--- a/packages/hoppscotch-common/src/helpers/isValidUser.ts
+++ b/packages/hoppscotch-common/src/helpers/isValidUser.ts
@@ -8,7 +8,16 @@ export type ValidUserResponse = {
 export const SESSION_EXPIRED = "Session expired. Please log in again."
 
 /**
- * Validates user authentication and token validity by making an API call
+ * Validates user authentication and token validity by making an API call.
+ *
+ * This function is kept separate from `handleTokenValidation()` to enable different use cases:
+ * - Silent validation for conditional UI states (e.g., disabling components on token expiration)
+ * - Background checks without triggering user notifications
+ * - Custom error handling based on validation results
+ *
+ * Use `handleTokenValidation()` when you need automatic toast error notifications.
+ * Use `isValidUser()` for silent validation or custom error handling scenarios.
+ *
  * @returns {Promise<ValidUserResponse>} Authentication status with user existence and token validity
  */
 export const isValidUser = async (): Promise<ValidUserResponse> => {
--- a/packages/hoppscotch-selfhost-desktop/src/platform/auth.ts
+++ b/packages/hoppscotch-selfhost-desktop/src/platform/auth.ts
@@ -418,7 +418,6 @@ export const def: AuthPlatformDef = {
       // axios automatically throws on error status codes, so if we reach here, it was successful
       return !!response.data.isValid
     } catch (error) {
-      console.debug("Token verification failed:", error)
       return false
     }
   },
--- a/packages/hoppscotch-selfhost-web/src/platform/auth/desktop/index.ts
+++ b/packages/hoppscotch-selfhost-web/src/platform/auth/desktop/index.ts
@@ -547,7 +547,6 @@ export const def: AuthPlatformDef = {
 
     const res = await response
     if (E.isLeft(res)) {
-      console.debug("Token verification failed:", res.left)
       return false
     }
 
--- a/packages/hoppscotch-selfhost-web/src/platform/auth/web/index.ts
+++ b/packages/hoppscotch-selfhost-web/src/platform/auth/web/index.ts
@@ -391,7 +391,6 @@ export const def: AuthPlatformDef = {
       // axios automatically throws on error status codes, so if we reach here, it was successful
       return !!response.data.isValid
     } catch (error) {
-      console.debug("Token verification failed:", error)
       return false
     }
   },


chhaviG22:
> Hi @chhaviG22, thank you for the contribution!
> 
> In this approach, we're blocking users who aren't logged in from creating a collection. Instead, we could consider checking the auth state and perform the action.
> 
> Also, this issue may affect other actions as well, such as creating or deleting requests. It might be worth thinking about a unified solution that handles all such cases, rather than addressing just this specific one.

Hi @nivedin , thank you for your feedback!

I agree that checking the auth state before performing actions might be a more robust approach. I'll review the possible approaches and update the PR accordingly.

Thank you for highlighting this!

chhaviG22:
> Hi @chhaviG22, thank you for the contribution!
> 
> In this approach, we're blocking users who aren't logged in from creating a collection. Instead, we could consider checking the auth state and perform the action.
> 
> Also, this issue may affect other actions as well, such as creating or deleting requests. It might be worth thinking about a unified solution that handles all such cases, rather than addressing just this specific one.

hi @nivedin, I have updated the implementation based on your suggestion to take a more unified approach, covering all actions (create/edit/delete) and preventing actions only for logged-in users in cases of access token expiration. Non-logged-in users will not be restricted. Could you please review the PR now?

chhaviG22:
hi @nivedin, just following up to check if you’ve had a chance to review the PR. Please let me know if there's any feedback or additional changes you'd like me to address. Thanks!

nivedin:
> hi @nivedin, just following up to check if you’ve had a chance to review the PR. Please let me know if there's any feedback or additional changes you'd like me to address. Thanks!

@chhaviG22 sorry for the delay will have a look before next release, hopefully we can merge it before that

packages/hoppscotch-common/src/components/collections/index.vue
```diff
@@ -738,7 +766,15 @@ const displayModalEditResponse = (show: boolean) => {
   if (!show) resetSelectedData()
 }
 
-const displayModalImportExport = (show: boolean) => {
+const displayModalImportExport = async (
+  show: boolean,
+  collectionType?: string
+) => {
+  const response: { valid: boolean; error: string } = await isValidUser()
```
nivedin: we don't need to explicitly state this everywhere since the isValidUser() returns the type correctly

packages/hoppscotch-common/src/components/collections/index.vue
```diff
@@ -368,6 +373,29 @@ const currentUser = useReadonlyStream(
...
+      }
+    } catch (error) {
+      console.debug("Token validation failed:", error)
+      return { valid: false, error: SESSION_EXPIRED }
+    }
+  }
+
+  // allow user to perform actions without being logged in
+  return { valid: true, error: "" }
+}
```
nivedin:
```suggestion
type ValidUserResponse = {
  valid: boolean
  error: string
}

/**
 * Validates user authentication and token validity by making an API call
 * @returns {Promise<ValidUserResponse>} Authentication status with user existence and token validity
 */
const isValidUser = async (): Promise<ValidUserResponse> => {
  // If user is logged in, verify tokens
  if (currentUser.value) {
    try {
      // Check if the user has valid tokens
      const hasValidTokens = await verifyAuthTokens()
      return {
        valid: hasValidTokens,
        error: SESSION_EXPIRED,
      }
    } catch (error) {
      console.debug("Token validation failed:", error)
      return { valid: false, error: SESSION_EXPIRED }
    }
  }

  // allow user to perform actions without being logged in
  return { valid: true, error: "" }
}
```

also we can move this to separate file and export this function from there

packages/hoppscotch-common/src/components/collections/index.vue
```diff
@@ -832,6 +876,12 @@ const onAddRequest = (requestName: string) => {
   const path = editingFolderPath.value
   if (!path) return
   if (collectionsType.value.type === "my-collections") {
+    const response: { valid: boolean; error: string } = await isValidUser()
+    if (!response.valid) {
+      toast.error(response.error)
+      return
+    }
```
nivedin:
```suggestion
/**
 * Handles token validation before performing actions that require authentication.
 * If the user is not valid, it shows an error toast and returns false.
 * If the user is valid, it returns true.
 */
const handleTokenValidation = async () => {
  const response: ValidUserResponse = await isValidUser()
  if (!response.valid) {
    toast.error(response.error)
    return false
  }

  return true
}
```
we can move this to separate function as it is repeating

 ```
// Validate user and tokens before proceeding
    const isValidToken = await handleTokenValidation()

 // If user is not valid, exit the function
    if (!isValidToken) return
```

we can add a check like this in every function

**Review by nivedin — CHANGES_REQUESTED** [review]
Please look into the comments

**Comment by nivedin** [comment]
@mirarifhasan please look into the backend changes

File: packages/hoppscotch-common/src/components/collections/index.vue (hunk 1 above)
Comment (chhaviG22) [reply packages/hoppscotch-common/src/components/collections/index.vue]: okay.

File: packages/hoppscotch-common/src/components/collections/index.vue (hunk 3 above)
Comment (chhaviG22) [reply packages/hoppscotch-common/src/components/collections/index.vue]: done.

**Review by nivedin — CHANGES_REQUESTED** [review]
We can leverage the `isValidUser` to the GQL collection and environments section as well

**Comment by chhaviG22** [comment]
> We can leverage the `isValidUser` to the GQL collection and environments section as well, also can you rebase the branch with the latest changes

Sure @nivedin , I'll implement this for GQL tab & environments also.

**Comment by chhaviG22** [comment]
> We can leverage the `isValidUser` to the GQL collection and environments section as well, also can you rebase the branch with the latest changes

hi @nivedin done

File: packages/hoppscotch-common/src/components/collections/graphql/index.vue
Comment (nivedin) [inline packages/hoppscotch-common/src/components/collections/graphql/index.vue:1]: we can add validation in delete and duplicate collection call and GQL request delete call

File: packages/hoppscotch-common/src/components/environments/my/index.vue
Comment (nivedin) [inline packages/hoppscotch-common/src/components/environments/my/index.vue:1]: we can add validation in delete and duplicate environment call

File: packages/hoppscotch-common/src/components/environments/teams/index.vue
Comment (nivedin) [inline packages/hoppscotch-common/src/components/environments/teams/index.vue:1]: We don’t need additional validation in the team context, since an error will already be thrown if the cookie expires. If we decide to add validation for the team environment, we should also consider adding it for the team collection to make it consistent.

File: packages/hoppscotch-common/src/components/environments/teams/index.vue [hunk 4]
```diff
@@ -223,20 +234,28 @@ const selectedEnvironmentID = ref<string | null>(null)
...
+const displayModalAdd = async (shouldDisplay: boolean) => {
+  const isValidToken = await handleTokenValidation()
+  if (!isValidToken) return
   action.value = "new"
   showModalDetails.value = shouldDisplay
 }
-const displayModalEdit = (shouldDisplay: boolean) => {
+const displayModalEdit = async (shouldDisplay: boolean) => {
+  const isValidToken = await handleTokenValidation()
+  if (!isValidToken) return
```
nivedin: since `displayModalEdit` will be used to close the modal we have to check if the action is to open and not close the modal

packages/hoppscotch-common/src/components/environments/my/index.vue
```diff
@@ -159,17 +170,23 @@ const editingEnvironmentIndex = ref<number | null>(null)
...
+const displayModalAdd = async (shouldDisplay: boolean) => {
+  const isValidToken = await handleTokenValidation()
+  if (!isValidToken) return
   action.value = "new"
   showModalDetails.value = shouldDisplay
 }
-const displayModalEdit = (shouldDisplay: boolean) => {
+const displayModalEdit = async (shouldDisplay: boolean) => {
+  const isValidToken = await handleTokenValidation()
+  if (!isValidToken) return
```
nivedin: since `displayModalEdit` will be used to close the modal we have to check if the action is to open and not close the modal

nivedin (changes requested):
please look into the comments

packages/hoppscotch-common/src/components/collections/graphql/Collection.vue
```diff
@@ -358,7 +359,15 @@ const toggleShowChildren = () => {
   showChildren.value = !showChildren.value
 }
 
-const removeCollection = () => {
+const handleTokenValidation = async () => {
+  const { valid, error } = await isValidUser()
+  if (!valid) toast.error(error)
+  return valid
+}
```
nivedin: since we are using this function in multiple files it's better to move this to a file and import it.

```
export const handleTokenValidation = async () => {
  const toast = useToast()

  const { valid, error } = await isValidUser()
  if (!valid) toast.error(error)
  return valid
}

```

packages/hoppscotch-common/src/helpers/auth/isValidUser.ts
nivedin, line 1: we can move out this file to `/helpers` since` /helpers/auth `consist of Authorisation types and not user auth

packages/hoppscotch-common/src/components/environments/my/index.vue
nivedin, line 1: we can add validation for global environment edit as well

nivedin (changes requested):
please look into the comments

nivedin (approved):
Made some minor changes, LGTM 🚀

chhaviG22:
hi @nivedin, I just wanted to follow up and ask when this will be merged and released?

nivedin:
> hi @nivedin, I just wanted to follow up and ask when this will be merged and released?

hey @chhaviG22, the PR needs a final review and some testing, as it touches user-facing layers. If everything looks good, it will be included in the month-end release. Thank you for your patience

chhaviG22:
> > hi @nivedin, I just wanted to follow up and ask when this will be merged and released?
> 
> hey @chhaviG22, the PR needs a final review and some testing, as it touches user-facing layers. If everything looks good, it will be included in the month-end release. Thank you for your patience

Thank you for the update! I’ll wait for the month-end release.

packages/hoppscotch-common/src/helpers/backend/helpers.ts
```diff
@@ -272,3 +274,27 @@ export const getTeamCollectionJSON = async (teamID: string) => {
...
+      withCredentials: true,
+    })
+
+    // axios automatically throws on error status codes, so if we reach here, it was successful
+    return !!response.data.isValid
+  } catch (error) {
+    console.debug("Token verification failed:", error)
+    return false
+  }
+}
```
jamesgeorge007: Given this is exclusive to the SH offering, we'll need to abstract this behind a platform definition to ensure the cloud offering (which uses Firebase's automatic token refresh) remains unaffected.

For instance, `platform.auth.validateUserSession()` defaults to `true` in platforms where it isn't defined.

packages/hoppscotch-common/src/helpers/isValidUser.ts
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Validates user authentication and token validity by making an API call
+ * @returns {Promise<ValidUserResponse>} Authentication status with user existence and token validity
+ */
+export const isValidUser = async (): Promise<ValidUserResponse> => {
+  const user = platform.auth.getCurrentUser()
+
+  if (user) {
+    try {
+      const hasValidTokens = await platform.auth.verifyAuthTokens()
```
jamesgeorge007: How about marking the `verifyAuthTokens()` function optional with a fallback to `true`, thereby avoiding the need to maintain definitions for the cloud platform and minimising the surface area?

packages/hoppscotch-common/src/helpers/handleTokenValidation.ts
```diff
@@ -0,0 +1,9 @@
+import { useToast } from "@composables/toast"
+import { isValidUser } from "~/helpers/isValidUser"
```
jamesgeorge007, line 2: Any reason to keep the `isValidUser()` helper function distinct instead of merging? Do we foresee future use cases?

nivedin: Yup let's keep it separate since `handleTokenValidation` throws an toast error and we might need a silent cancel or disabling certain component based on token expiration in the future.

jamesgeorge007 (approved):

packages/hoppscotch-common/src/helpers/handleTokenValidation.ts
jamesgeorge007: Got it. Let's also ensure that such use cases are documented for future reference, preventing ambiguities; pushed changes reflecting the same.

jamesgeorge007:
Thanks for the PR, @chhaviG22. This will be going live with the upcoming `v2025.9.0` release.

PR #5264, opened 2025-07-21
Enhancement: Added base background color and font to body in index.html

Prajwal04M:
This PR adds a light background and default font to the body tag for improved visual consistency and readability.

### DIFF
--- a/packages/hoppscotch-agent/index.html
+++ b/packages/hoppscotch-agent/index.html
@@ -6,7 +6,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>Hoppscotch Agent</title>
   </head>
-  <body>
+  <body style="background-color: #f9fafb; font-family: sans-serif;">
     <div id="app"></div>
     <script type="module" src="/src/main.ts"></script>
   </body>


CuriousCorrelation:
Hey @Prajwal04M, thanks for the contribution. However, this PR has issues that need to be addressed before we can move forward.

First, you haven't followed our pull request template, so please fill it out properly so reviewers understand what problem you're trying to solve. You don't have to adhere to it perfectly but the general formatting from that would be helpful.

Also the description part

>  "improve visual consistency and readability," 

is not clear as to what specific issue this addresses or how these particular styles solve it.

More importantly, our typical workflow is to first create an issue that clearly demonstrates the problem (with screenshots showing what looks inconsistent or hard to read), then create a PR that references and fixes that issue. This helps us and the contributors in making sure we're actually solving real user-facing problems rather than making arbitrary changes.

Your PR title also doesn't follow our general conventions. Please take a look into other existing PRs in the repo, you may notice that they use formats like `feat(scope): description` or `fix(scope): description`. Something like `fix(agent): add base styling to prevent unstyled flash` would be more appropriate if that's what this is addressing (unsure at the moment).

Also as Copilot mentioned, inline styles aren't ideal either; this should probably be in a block.

Feel free to check out our contributing guidelines and look at how other successful PRs are structured and do feel free to reach out if you need any clarification on our process.

Prajwal04M:
Thank you for he feedback I will update the PR with filled template,
clearer description,  and issued reference shortly

CuriousCorrelation:
Hi, since there hasn't been activity for a month, I'm closing this PR to keep our queue manageable. Tho please feel free to create a new PR if you'd like to address the feedback and continue working on this 👍

PR #5281, opened 2025-07-26
updated the hoppscotch-agent Readme for more information

anasaijaz:
### What's changed
Updated the `packages/hoppscotch-agent/README.md` with better explanation of the repository

### DIFF
--- a/packages/hoppscotch-agent/README.md
+++ b/packages/hoppscotch-agent/README.md
@@ -1,16 +1,188 @@
-# Tauri + Vue + TypeScript
+# Hoppscotch Agent <sup>BETA</sup>
 
-This template should help get you started developing with Vue 3 and TypeScript in Vite. The template uses Vue 3 `<script setup>` SFCs, check out the [script setup docs](https://v3.vuejs.org/api/sfc-script-setup.html#sfc-script-setup) to learn more.
+<div align="center">
+   <img align="center" width="128px" src="../../public/logo.svg" />
+   <h1 align="center"><b>Hoppscotch Agent</b></h1>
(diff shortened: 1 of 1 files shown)


CuriousCorrelation:
Hi @anasaijaz, this to me looks AI-generated and while using AI tools isn't inherently problematic, it seems like the context window wasn't sufficient to capture the actual details of the project, because the AI seems to have missed a lot of context, there are quite a few errors:

“Use any Hoppscotch application (web, desktop, or self-hosted)" doesn’t make much sense since the agent doesn't work with our desktop app, and web vs self-hosted isn't a real distinction.

The image path `../../public/logo.svg` doesn't exist and "🏢 Enterprise Support: Corporate proxy configurations with authentication" doesn't really make sense in this context.

There are also some minor issues like I am a bit unsure as to the reason for including “Interceptor Selection” without explaining the flow. Building from Source" section also completely omits portable build config. Also heavy emoji usage which we don't use in our READMEs all that much.

It's also full of generic descriptions like "advanced request handling" and "enterprise support" without explaining what that actually means or how it relates.

"HTTP Server: Secure API server on localhost:9119" here I don’t quite understand the categorization, because often LLMs try to categorize their responses with a short section in font separated by `:` and in this instance that fail in gathering enough context.

I'm going to close this PR but treat it as an issue report about our placeholder README. I'll work on a proper documentation update that accurately reflects how the agent actually works.

Thanks for identifying this.

anasaijaz:
Sure @CuriousCorrelation Yea I actually used AI to create this README
Sorry I should've double checked the content
Happy the README is updated now

PR #5302, opened 2025-08-03
fix: ensure deterministic pagination by using cursor field for sorting

### LINKED ISSUE #5301: [bug]: Indeterministic ordering in pagination skips (or) duplicates collections

### Platform

Web App

### Browser

Chrome

### Operating System

macOS

## What happened?
A bug in the pagination logic causes some collections to be `skipped` and some collections to be `duplicated` when fetching them from the DB. This only happens when doing a bulk import of collections.

## Screen Recording
The following video shows an instance where the collections are duplicated.

**Things to note in the video:**
- Number of collections in DB = `22` ( == Expected no. of collections to be displayed on UI)
- Actual no. of collections displayed on UI = `31`

https://github.com/user-attachments/assets/3a3a962f-c2a4-4c56-ac01-239e82832cd0

**Steps to reproduce:**
- `Bulk-import` about 20 collections in a team workspace and count the results

### Deployment Type

Hoppscotch Cloud

### Version

2025.7.0

mayank-jain-1010:
Closes #5301 

It was observed that if there are more than 10 records sharing the same orderIndex, then while fetching the collections from the DB - some collections can be skipped whereas some other collections can be duplicated. This can lead to a very confusing user experience.

**Root cause of this bug:** We are ordering by `orderIndex` but cursor is placed on `id`field.

**Why does it cause an issue?**
Here's an example to understand:

```
Consider you get these records the first time the query is made. For simplicity, let's say we pick 3 records at a time.

id1 orderIndex=1
id2 orderIndex=1
id3 orderIndex=1 ---> fetch till here in the first network call and place cursor
id4 orderIndex=1
id5 orderIndex=1
id6 orderIndex=1
```

During the next fetch, the order in which the above records are returned can be different - causing skips and duplicates

```
id5 orderIndex=1                    // skipped
id4 orderIndex=1                    // skipped
id3 orderIndex=1 ---> cursor starts from here skipping some records and duplicating some other records
id2 orderIndex=1                    // duplicated
id1 orderIndex=1                    // duplicated
id6 orderIndex=1
```

### What's changed
Instead of just sorting by `orderIndex`, we sort by `id` as well to ensure that for a particular `orderIndex`, we get a deterministic order of records on different fetches. This PR proposes to implement that change.
**Why use the id field?**
- Because `id` field is used as the cursor

**Reference read for more info: https://www.prisma.io/docs/orm/prisma-client/queries/pagination**

### Screen Recording for Fix
https://github.com/user-attachments/assets/64413031-8349-4167-91a2-8806910cc666

### DIFF
--- a/packages/hoppscotch-backend/src/team-collection/team-collection.service.ts
+++ b/packages/hoppscotch-backend/src/team-collection/team-collection.service.ts
@@ -385,9 +385,7 @@ export class TeamCollectionService {
       where: {
         parentID: collectionID,
       },
-      orderBy: {
-        orderIndex: 'asc',
-      },
+      orderBy: [{ orderIndex: 'asc' }, { id: 'asc' }],
       take: take, // default: 10
       skip: cursor ? 1 : 0,
       cursor: cursor ? { id: cursor } : undefined,
@@ -418,9 +416,7 @@ export class TeamCollectionService {
         teamID,
         parentID: null,
       },
-      orderBy: {
-        orderIndex: 'asc',
-      },
+      orderBy: [{ orderIndex: 'asc' }, { id: 'asc' }],
       take: take, // default: 10
       skip: cursor ? 1 : 0,
       cursor: cursor ? { id: cursor } : undefined,


mirarifhasan:
@mayank-jain-1010
Thanks for catching this bug and creating a PR with a proposed solution.

I reviewed your changes. While your approach of adding an additional sort by `id` helps reduce the chance of duplicate `TeamCollection` entries—especially since we're using `cuid`, which isn’t strictly incremental—it doesn't fully solve the underlying issue.

The core problem remains: reordering (e.g., dragging items to the top, middle, or bottom) still leads to inconsistencies due to how `orderIndex` is currently handled.

This issue also affects other entities like `UserCollection`, `UserRequest`, `TeamCollection`, and `TeamRequest`.

To make a solid and long-term fix, we need to properly address the way `orderIndex` is being assigned or auto fix. Relying on additional sorting by `id` isn't a reliable solution here.

Please take some time to think through a plan for making `orderIndex` properly incremental and consistent. Once you’ve proposed an approach and we confirm it, feel free to proceed with the implementation.

mirarifhasan:
Hi @mayank-jain-1010,

Since this fix impacts a core backend flow and we’re targeting a release by the end of this month, I’d like to have our team take this over to fast-track the implementation. I understand you might be busy, and this way we can ensure it’s ready in time without adding pressure on your side.

Thanks again for identifying the issue and starting the solution!

mayank-jain-1010:
Makes sense @mirarifhasan!
Thanks!

PR #5323, opened 2025-08-12
fix: client ip with proxies

L3o-pold:
Closes # 

### What's changed
Adding a config settings to be able to trust proxies. It's necessary in order to avoid rate limiting (`ThrottleException. Too many Requests` based on proxy IP.
It will also add the user IP in the access log to improve auditability and configuration.


.env.example
```diff
@@ -14,6 +14,8 @@ DATA_ENCRYPTION_KEY="data encryption key with 32 char"
 #   bundle names like `app://{bundle-name}/`
 WHITELISTED_ORIGINS="http://localhost:3170,http://localhost:3000,http://localhost:3100,app://localhost_3200,app://hoppscotch"
 
+# If true, the client’s IP address is understood as the left-most entry in the X-Forwarded-For header
+TRUSTED_PROXIES=false
```
mirarifhasan: A small change requesting to change the env variable name.

```suggestion
# [Optional] If true, the client’s IP address is understood as the left-most entry in the X-Forwarded-For header
TRUST_PROXY=false
```

packages/hoppscotch-backend/src/main.ts
```diff
@@ -84,7 +85,14 @@ async function bootstrap() {
       transform: true,
     }),
   );
-  app.use(morgan(':method :url :status - :response-time ms'));
+
+  if (configService.get('TRUSTED_PROXIES') === 'true') {
```
mirarifhasan: Same here, just update the env variable name.

```suggestion
  if (configService.get('TRUST_PROXY') === 'true') {
```

PR #5330, opened 2025-08-14
feat(response-panel): improve UI distinction (#5307)

### LINKED ISSUE #5307: [feature]: UI improvement - make response panel more distinct

### Summary

The response panel is separated from the "request area" by only a 1px thin line. This can cause an issue that part of the request area is covered by the response panel, making it non-intuitive that there is more content in the request area than is visible.

I encountered this issue when setting up a request with OAuth 2.0 client credentials. The documentation mentions that after filling in the form field, you should click on "Generate Token". Since I had already sent a request prior to filling the form, the response panel was already open and by chance, covering the "Generate Token" button. On dark mode especially, but also on light mode, it's not necessarily intuitive that the Authorization panel is scrollable.

The "Generate Token" button appears to be missing:
Image

After scrolling in the Authorization panel, it appears:
Image

On light mode, it can also be tricky to see that the panel is scrollable:
Image

Suggestion: make the separation of the response panel more prominent. E.g. a thicker line with more contrast (especially on dark mode) with a "dragable" handlebar in the middle. Something like this:

Image

(image source: https://stackoverflow.com/questions/42981485/proper-accessibility-or-aria-role-for-a-resize-handle)

yash-gorasia:
**Closes #5307**

**What's changed**
Improved the visual distinction of the response panel while keeping within Hoppscotch’s theme system

**Updated panel container classes:**
Changed border-dividerLight → border-accent for better accent visibility
Increased rounding from rounded-md → rounded-lg
Upgraded shadow from shadow-sm → shadow-lg for more depth
Added subtle highlight with ring-1 ring-accent/20
Retained bg-primaryContrast to maintain dark/light mode compatibility

**Before**
The response panel blended in with the surrounding UI, making it less visually prominent and harder to distinguish from nearby sections.

**After**
The panel now has a clearer accent border, slightly larger rounding, a deeper shadow, and a soft ring highlight — making it more noticeable while still respecting the overall theme, color palette, and dark/light mode adaptability.

### DIFF
--- a/packages/hoppscotch-common/src/components/http/Response.vue
+++ b/packages/hoppscotch-common/src/components/http/Response.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="relative flex flex-1 flex-col border border-accent rounded-lg shadow-lg bg-primaryContrast ring-1 ring-accent/20">
+  <div :class="responseContainerClass">
     <HttpResponseMeta :response="doc.response" :is-embed="isEmbed" />
     <LensesResponseBodyRenderer
       v-if="!loading && hasResponse"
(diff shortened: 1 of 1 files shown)


yash-gorasia:
Closing this PR since the change doesn’t align with Hoppscotch’s design language. Thanks for the review!

PR #5348, opened 2025-08-25
fix: handle edge cases and bugs in collection variables

nivedin:
Closes FE-992

This PR fixes some edge cases and bug's with collection variables

- [x] Import from 'Another Workspace' to Team workspace fails to import collection variables
- [x] Hide action button in collection variable empty inspector (can be iterated in future to open the collection properties)
- [x] Add collection variable when moving or reordering personal collection 
- [x] opened tabs in team collection variable when moving does not get updated.
- [x] Collection request variable priority bug over collection var

### DIFF
--- a/packages/hoppscotch-common/src/components/app/Inspection.vue
+++ b/packages/hoppscotch-common/src/components/app/Inspection.vue
@@ -52,11 +52,11 @@
                 </HoppSmartLink>
               </span>
               <span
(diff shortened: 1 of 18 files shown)


jamesgeorge007 (approved):
The issue with tab state dissociation during collection moves persists in shared workspaces; can be revisited.

PR #5350, opened 2025-08-26
fix: prevent GQL request failure caused by collection property computation

nivedin:
This PR fixes the issue where GQL section crashed when auth was undefined

### DIFF
--- a/packages/hoppscotch-common/src/components/graphql/RequestOptions.vue
+++ b/packages/hoppscotch-common/src/components/graphql/RequestOptions.vue
@@ -142,13 +142,13 @@ const runQuery = async (
     const inheritedHeaders =
       tabs.currentActiveTab.value.document.inheritedProperties?.headers.map(
         (header) => header.inheritedHeader
-      )
+      ) ?? []
 
     await runGQLOperation({
       name: request.value.name,
       url: runURL,
       request: request.value,
-      inheritedHeaders: inheritedHeaders ?? [],
+      inheritedHeaders,
       inheritedAuth: tabs.currentActiveTab.value.document.inheritedProperties
         ?.auth.inheritedAuth as HoppGQLAuth | undefined,
       query: runQuery,


packages/hoppscotch-common/src/helpers/graphql/connection.ts
```diff
@@ -404,10 +404,10 @@ export const runGQLOperation = async (options: RunQueryOptions) => {
     v: 9,
     name: options.name || "Untitled Request",
     url: finalUrl,
-    headers: runHeaders,
+    headers: runHeaders ?? request.headers,
```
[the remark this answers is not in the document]
nivedin: request.header wont be header it will pass [] if empty

packages/hoppscotch-common/src/helpers/graphql/connection.ts
```diff
@@ -404,10 +404,10 @@ export const runGQLOperation = async (options: RunQueryOptions) => {
...
     query,
     variables,
-    auth,
+    auth: auth ?? request.auth,
```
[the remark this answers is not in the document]
nivedin: same

jamesgeorge007 (approved):
We'll have to revisit certain explicit type casts that serve as the root cause.
> Addressed in the same PR.

https://github.com/hoppscotch/hoppscotch/blob/c04faaaf27e756f8e507942e6ab771503b3b91d1/packages/hoppscotch-common/src/components/graphql/RequestOptions.vue#L142-L163

packages/hoppscotch-common/src/components/graphql/RequestOptions.vue
```diff
@@ -141,19 +141,14 @@ const runQuery = async (
...
+      )
 
     await runGQLOperation({
       name: request.value.name,
       url: runURL,
       request: request.value,
-      inheritedHeaders: inheritedHeaders,
+      inheritedHeaders: inheritedHeaders ?? [],
       inheritedAuth: tabs.currentActiveTab.value.document.inheritedProperties
         ?.auth.inheritedAuth as HoppGQLAuth,
```
jamesgeorge007: Could you also review the surface area associated with removing the explicit cast for [inheritedAuth](https://github.com/hoppscotch/hoppscotch/blob/c04faaaf27e756f8e507942e6ab771503b3b91d1/packages/hoppscotch-common/src/components/graphql/RequestOptions.vue#L157-L158)?

Might require updates to the `RunQueryOption` type definition.

https://github.com/hoppscotch/hoppscotch/blob/c04faaaf27e756f8e507942e6ab771503b3b91d1/packages/hoppscotch-common/src/helpers/graphql/connection.ts#L44-L54

nivedin: we might have to explicitly define because inheritedAuth can have HoppRESTAuthtype too. [ref](https://github.com/hoppscotch/hoppscotch/blob/fix/gql-auth-bug/packages/hoppscotch-common/src/helpers/types/HoppInheritedProperties.ts#L13)

jamesgeorge007: Discussed separately, aligning the type, marking it as optional with the `undefined` fallback.

Issue #1598, opened 2021-04-14
[feature]: Mock Server


baaart:
Is there a plan to have mock server option in similar way as it is done in Postman?

It would be nice to have all in one testing tool (API testing + mocking APIs), I was looking for this option in hoppscotch, but i don't see it anywhere.

liyasthomas:
As of now, there's no such feature to create mock APIs from Hoppscotch. But would definitely consider this if there's a major request for this feature.

jatin33:
this will be a great feature, if we try something like this https://mockend.com/
@liyasthomas any thoughts

liyasthomas:
@jatin33 thanks for your suggestion. I'll look into this asap.

liyasthomas:
Hi @bhavukkalra I'd love to see your mvp. Also, we're still refining the feature and it's architecture. We'll think about this and let you know.

bhavukkalra:
Sorry for the late reply @liyasthomas, about the MVP(for what will be designed for Hoppscotch to incorporate MockServer). 
Could you please go through the workflow(Tentative) and suggest If there are any ambiguities or corrections that can be made to allow code Reusability and Stability.
[Tentative Architecture for MockServer](https://drive.google.com/file/d/1S7FjqB7gGN0WtFRAT63cnFclia7aoWAG/view?usp=sharing)

After The Above logic has been written, each request to a route can be handeled via, this piece of Code

```javascript
router.route('/:id')
    .get(handle.getEndpoint)
    .post(auth.authenticate, handle.IncorporateChanges)
    .delete(auth.authenticate, handle.deleteObject)
```

Also Would that be OK if I made the above prototype/MVP using Basic HTML and NodeJS for rapid prototyping? Just to show the Proof of Concept. After that it can be Integrated to Hoppscotch(NuxtJS) or should I start with NuxtJS to begin with?

juzhiyuan:
Hi, do we have a plan to support mocking POST request? e.g

1. POST /users
2. GET /users
3. We could find the resource which is created in Step 1.

liyasthomas:
"Mock requests" feature has the highest priority in our roadmap. Thanks for your patience, please bear with us and expect a beta release of Mock server soon.

juzhiyuan:
ya, and what should I do if I want to contribute to this feature?

bhavukkalra:
[...]

PR #5385, opened 2025-09-15
feat: add mock server

anwarulislam:
Closes FE-1020 #1598

This pull request introduces a new "Mock Server" feature to the backend and frontend, allowing users to create, manage, and interact with mock servers associated with their collections. The implementation includes database schema updates, new service, controller, resolver, and model files.


zhukai9527:
very useful function

Issue #3532, opened 2023-11-10
[bug]: multiple set-cookie did not get fold correctly and not sending as well


xshadowlegendx:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

I use hoppscotch cloud with proxy interceptor to call to my backend server and for login endpoint, the server return refresh and fingerprint httponly cookies and after successful login hoppscotch only shows 1 set cookie response, the issue is same as [this one](https://github.com/hoppscotch/hoppscotch/issues/1546) and also subsequent request did not send the cookie as well, I tried testing this with postman from the web, and did not configure anything I can call login endpoint and the response show both cookies and also subsequent request will also send those cookies along

### Steps to reproduce

you can try this endpoint of my test server on both hoppscotch and postman, postman will just works while hoppscotch did not correctly fold those multiple cookies and did not send them with subsequent request as well

```
POST https://identity.lsa.xdevx.dedyn.io/authn/password/login

user_id = 7131aa28-ca21-4ff7-8273-5e16ab304e25
password = 13db2588XyZ

content-type = application/x-www-form-urlencoded
```

### Environment

Production

### Version

Cloud

liyasthomas:
Cookie capturing is not available in the Hoppscotch web app.

[Download Hoppscotch Desktop App](https://hoppscotch.com/download) to manage Cookies.

xshadowlegendx:
hello @liyasthomas I also tried the desktop app one also same behavior

liyasthomas:
Sorry for the inconvenience, as of today, you've to manually add the cookie string to the domain. You can follow the instructions in the [cookies documentation](https://docs.hoppscotch.io/documentation/features/cookies) to manually add cookies.

The ability to auto magically set-cookies from response headers is coming soon in Hoppscotch Desktop App. Thank you for your patience. Keeping this issue ticket open to track feature progress.

felipebontempo:
Sorry for the inconvenience, but maybe my problem is exactly this:

Below are two images, one is how Postman returns and how hoppscotch returns.

The problem is that hoppscotch is missing a "Set-cookie", there should be two, and they are exactly the same call in both tools.

What will be the problem?

![CleanShot 2024-03-07 at 10 18 42@2x](https://github.com/hoppscotch/hoppscotch/assets/26803292/ee1504f7-b14d-45d1-b410-cf087a36594b)
[...]

PR #5394, opened 2025-09-21
fix(relay): multiple Set-Cookie headers in resp

### LINKED ISSUE #1546: Cookie missing in response
The response has three set-cookies,But it only gets one(Please forgive me for my poor English)

### LINKED ISSUE #5344: [bug]: hoppscotch is ignoring one set-cookie when response has 2 values

### Platform

Desktop App

### Browser

Edge

### Operating System

Windows

### Bug Description

When a request sends two ‘set-cookie’ in response headers(postman displays it as 2 separate headers and Bruno displays it as one array of ‘set-cookie’ of comma separated value), hoppscotch is ignoring one of them.

Because of this, we need to make 2 calls to get both cookies. Get first cookie on request-1 and set it to send request-2, then I will get the second cookie. This was one call in both postman and Bruno. Also cookie update in hoppscotch is painfully manual so more time consuming, atleast in desktop.

### Deployment Type

### Version

v2025.7.1

### LINKED ISSUE #5375: [bug]: Response header only shows first Set-Cookie when multiple cookies are present

### Platform

Desktop App

### Browser

Safari

### Operating System

macOS

### Bug Description

When the server response contains two or more Set-Cookie headers, Hoppscotch only displays the first one. This makes it impossible to debug APIs that rely on multiple cookies being set in a single response.

**Steps to Reproduce:**
	1.	Send a request to an endpoint that returns multiple Set-Cookie headers.
	2.	Check the response headers panel in Hoppscotch.
	3.	Only the first Set-Cookie is shown.

**Expected Behavior:**
All Set-Cookie headers should be displayed, not just the first one.

**Additional context:**
This is important for working with frameworks like Laravel, where both XSRF-TOKEN and laravel_session cookies are usually returned together.

### Deployment Type

### Version

25.8.1 - Desktop Application

CuriousCorrelation:
This is an experimental way to fix multiple `Set-Cookie` headers being
 lost during HTTP response pre/post-processing by concat-ing them with
 `\n` in the networking layer, so all cookies to be displayed in the
 response headers panel instead of only the first one being visible.
 
 | Before | After |
 | --- | --- |
 | image |  image |

 NOTE: This is an experimental workaround for the time being.
 `\n` is used here as a separator because HTTP headers are inherently
 line-oriented (`\r\n` on the wire). This makes it trivial to split back
 into distinct `Set-Cookie` headers without violating RFC rules. The
 next idea would be using a custom delimiter but it also would be
 arbitrary and risk breaking if that delimiter ever appeared in a cookie
 value.  The main trade-off is that this is a fragile workaround,
 unusual edge cases (e.g. malformed cookies containing newlines, or
 components that normalize headers differently) may fail. A full fix
 would require changing the header storage structure to support multiple
 values per key (the rough idea around that is explained in FE-1008).

 Closes FE-1008 (phase-1)
 Closes FE-309
 Closes #1546 (multiple Set-Cookie headers missing in response display)
 Closes #5344 (ignoring one Set-Cookie when response has 2 values)
 Closes #5375 (only shows first Set-Cookie when multiple cookies present)
 Partially addresses #3532 (fixes header display, automatic cookie
 storage will be handled separately).

 When servers returned multiple `Set-Cookie` headers, only the first
 header was preserved while subsequent ones were discarded.

 The issue occurres because `HashMap<String, String>` data structure in
 `TransferHandler` could only store one value per header name (by the
 very nature of hashmaps or records for instance).

 ### What's changed

 #### Relay networking layer

 The header processing in `transfer.rs` now includes special handling
 just for `set-cookie` headers (case-insensitive accumulator). When
 multiple Set-Cookie headers are encountered, they are concatenated with
 newlines (why is explained above).

 #### FE processing

 Added `processHeaders` function in `response.ts` that
 splits concatenated `Set-Cookie` headers back into separate header
 entries.

 #### Dependencies

 Updated `relay` and `tauri-plugin-relay` to commits that include the
 `Set-Cookie` concatenation fix across all target packages.

 ### Testing

 Test with any endpoints that set multiple cookies:

 ```bash
 curl -i http://localhost:3000/multiple-cookies
 ```

 or perhaps use `httpbin`:

 ```bash
 curl -i -G \
 --data-urlencode "Set-Cookie=sessionid=abc123; Path=/" \
 --data-urlencode "Set-Cookie=csrftoken=xyz789; Path=/" \
 --data-urlencode "Set-Cookie=userid=user456; Path=/" \
 "https://httpbin.org/response-headers"
 ```

 or even a custom server script:

 ```javascript
 const http = require('http');

 const server = http.createServer((req, res) => {
   if (req.url === '/multiple-cookies') {
     res.setHeader('Set-Cookie', [
       'sessionid=abc123; Path=/',
       'csrftoken=xyz789; Path=/',
       'userid=user456; Path=/'
     ]);
     res.setHeader('Content-Type', 'application/json');
     res.end(JSON.stringify({ message: 'Multiple cookies set' }));
   }
 });
 ```

 #### Expected response headers:

 ```
 Set-Cookie: sessionid=abc123; Path=/
 Set-Cookie: csrftoken=xyz789; Path=/
 Set-Cookie: userid=user456; Path=/
 ```

 In desktop app response headers panel, all three cookies now appear as
 separate header entries **instead of only the first one** being
 visible.

 ### Special Note On Backwards Compat

 This preserves the existing `HashMap<String, String>` structure without
 requiring changes to the `RelayResponse` interface or frontend header
 processing pipeline since this will have a rather large ripple effect
 all over the codebase. A complete fix is described in FE-1008.

 ### Implementation notes

 The comment in `processHeaders` should explain the rationale:

 > A complete solution would involve swapping `Record`/`HashMap` with a
 > flat array and propagating the refactor throughout the codebase, from
 > the underlying networking to the frontend. The problem with that
 > approach is you lose that O(1) lookup. A simpler approach is
 > `HashMap<String, Vec<String>>` but even that doesn't substantially
 > reduce the refactor surface area.

 The newline concatenation also preserves O(1) header lookup
 performance.

### DIFF
--- a/packages/hoppscotch-common/src/helpers/kernel/rest/response.ts
+++ b/packages/hoppscotch-common/src/helpers/kernel/rest/response.ts
@@ -45,7 +45,7 @@ const extractSize = (response: RelayResponse): number =>
  * against potential serde/boundary issues between typed conversions.
  */
 const processHeaders = (
-  headers: Record<string, string> | undefined | null
+  headers?: Record<string, string> | null
 ): HoppRESTResponseHeader[] => {
   const processedHeaders: HoppRESTResponseHeader[] = []
 


packages/hoppscotch-common/src/helpers/kernel/rest/response.ts
```diff
@@ -21,6 +21,39 @@ const extractTiming = (response: RelayResponse): number =>
...
+// also little point in going from key-value to key-pair-value since
+// you'd just be putting this same workaround somewhere else.
+// A simpler approach is `HashMap<String, Vec<String>>` but even that
+// doesn't substantially reduces the refactor surface area.
+// Given all of those issues, a temporary workaround is perhaps the best
+// solution at the moment.
+const processHeaders = (headers: Record<string, string>): HoppRESTResponseHeader[] => {
+  const processedHeaders: HoppRESTResponseHeader[] = []
+
+  for (const [key, value] of Object.entries(headers ?? {})) {
```
jamesgeorge007, line 53: Can we ensure the type contracts and implementation align such that `headers` being falsy is indicated in the types?

packages/hoppscotch-common/src/helpers/kernel/rest/response.ts
```diff
@@ -21,6 +21,39 @@ const extractTiming = (response: RelayResponse): number =>
...
+    if (key.toLowerCase() === 'set-cookie') {
+      // To split concatenated `Set-Cookie` headers and create separate header entries,
+      // see `if key.to_lowercase() == "set-cookie" {` in `transfer.rs`
+      const cookieStrings = value.split('\n').map(s => s.trim()).filter(s => s)
```
jamesgeorge007:
```suggestion
      const cookieStrings = value.split('\n').map(s => s.trim()).filter(Boolean)
```

Subjective, in case you prefer the idiomatic style.

packages/hoppscotch-common/src/helpers/kernel/rest/response.ts
CuriousCorrelation: While I don’t quite remember why this defensive pattern was added (perhaps something to do with proxy stripping headers with agent), I traced it from the kernel to interceptors to the presentation layer and headers do seem to be always preserved. Although I’d err on the side of caution and keep it there, so I have changed the types and added some docs explaining the note and a brief comment on the function definition ✅

packages/hoppscotch-common/src/helpers/kernel/rest/response.ts
CuriousCorrelation: Good suggestion! Have added it to the same commit 🫡

Issue #5396, opened 2025-09-22
[bug]: unable to build backend on  2025.8.1 on Github CI


Leon-Luu:
### Is there an existing issue for this?

- [x] I have searched existing issues and this bug hasn't been reported yet

### Platform

Web App

### Browser

Chrome

### Operating System

Linux

### Bug Description

When building the Hoppscotch backend Docker image using Alpine as the base, the build fails with the error /bin/sh: npm: not found. 
This is because the Alpine nodejs package does not include npm by default, but the Dockerfile expects npm to be available for global installs.

<img width="2671" height="880" alt="Image" src="https://github.com/user-attachments/assets/1e273856-ff7c-4d51-9b7b-5d849b6cdfbf" />

**Steps to Reproduce:**
Clone the Hoppscotch repository.
Run docker build -t hoppscotch-backend -f prod.Dockerfile
Observe the build failure at the step running npm install -g ....

### Deployment Type

Self-hosted (on-prem deployment)

### Version

2025.8.1

Leon-Luu:
@jamesgeorge007 please check, I have fix propose to https://github.com/hoppscotch/hoppscotch/pull/5397

jamesgeorge007:
Closing, since this is now resolved. Thanks for the PR, @Leon-Luu 🙌

PR #5397, opened 2025-09-22
fix: ensure npm is installed for Node.js base image

Leon-Luu:
This PR updates the [prod.Dockerfile](vscode-file://vscode-app/c:/Users/LELUU2/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) to explicitly install npm alongside nodejs in the Alpine base image. 

On Alpine, nodejs does not include npm by default, which caused build failures when running npm install -g .... 

The fix adds npm to the apk add command, ensuring the Node.js environment is complete and compatible with subsequent global package installations.

Closes # 
https://github.com/hoppscotch/hoppscotch/issues/5396

### What's changed

Add npm to the apk add --no-cache nodejs npm curl tini bash command in the Node.js base stage.
This resolves /bin/sh: npm: not found errors during Docker builds in CI and local environments.

### DIFF
--- a/prod.Dockerfile
+++ b/prod.Dockerfile
@@ -44,10 +44,10 @@ RUN go build
 
 # Shared Node.js base with optimized NPM installation
 FROM alpine:3.22.1 AS node_base
-RUN apk add --no-cache nodejs curl tini && \
+RUN apk add --no-cache nodejs npm curl tini bash && \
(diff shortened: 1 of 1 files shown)


mirarifhasan (approved):
LGTM

PR #5401, opened 2025-09-24
refactor: streamline duplicated advanced param type definitions for oauth grant types

anwarulislam:
Closes FE-1004

This pull request refactors the OAuth flow parameter schemas in the service layer to use shared, centralized schemas from `@hoppscotch/data`. It replaces locally-defined schemas with the imported ones, omitting unnecessary fields and ensuring required arrays for service logic. This change improves maintainability and consistency across the codebase.

### DIFF
--- a/packages/hoppscotch-common/src/services/oauth/flows/authCode.ts
+++ b/packages/hoppscotch-common/src/services/oauth/flows/authCode.ts
@@ -11,11 +11,7 @@ import { getService } from "~/modules/dioc"
 import * as E from "fp-ts/Either"
 import { KernelInterceptorService } from "~/services/kernel-interceptor.service"
 import { content } from "@hoppscotch/kernel"
-import {
-  refreshToken,
-  OAuth2AdvancedParamSchema,
-  OAuth2AuthRequestParamSchema,
-} from "../utils"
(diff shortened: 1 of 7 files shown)


packages/hoppscotch-common/src/services/oauth/flows/authCode.ts
```diff
@@ -12,26 +12,21 @@ import * as E from "fp-ts/Either"
...
-    ),
-    refreshRequestParams: z.array(OAuth2ParamSchema),
+// Use the existing schema from hoppscotch-data but ensure required arrays
+const AuthCodeOauthFlowParamsSchema = AuthCodeGrantTypeParams.omit({
+  grantType: true,
+  token: true,
+})
+  .extend({
+    // Override optional arrays to be required for the service layer
+    authRequestParams: z.array(OAuth2ParamSchema.omit({ sendIn: true })),
```
jamesgeorge007: Instead of the shared `OAuth2ParamSchema` schema definition. Can we export the pre-existing schema definitions from `hoppscotch-data` and consume that instead?

https://github.com/hoppscotch/hoppscotch/blob/75ec41207667ab9365b81fc8440826252d154008/packages/hoppscotch-data/src/rest/v/15/auth.ts#L23-L32

PR #5436, opened 2025-10-03
Added missing token checks

pmondal-int:
This is a follow up to https://github.com/hoppscotch/hoppscotch/pull/5249. This PR introduces token checks when saving requests.


jamesgeorge007:
Thanks for the PR, @pmondal-int 🙌  This is scheduled for the upcoming major release.

PR #5441, opened 2025-10-03
chore: add arm runner to matrix-based docker build

josephyi:
Refactors matrix strategy to use [Github's ARM Runner](https://docs.github.com/en/actions/reference/runners/github-hosted-runners#supported-runners-and-hardware-resources) (free for open source) for the ARM build variant, which should dramatically improve the overall Docker build since it won't need emulation.

### What's changed

* Added `runner: ubuntu-latest` and `runner: ubuntu-24.04-arm` to the respective entries in the matrix strategy.
* Updated `runs-on:` to use matrix strategy value.

### Notes to reviewers
The following can be removed in a separate PR once desired behavior is confirmed, or happy to do it as part of this PR if scope is reasonable: 

```yaml
      - name: Setup QEMU
        uses: docker/setup-qemu-action@v3
```

### DIFF
--- a/.github/workflows/release-push-docker.yml
+++ b/.github/workflows/release-push-docker.yml
@@ -13,19 +13,16 @@ jobs:
     strategy:
       matrix:
         platform: [
-          { platform: linux/amd64, cache: docker-release-amd64, artifactSuffix: amd64 },
-          { platform: linux/arm64, cache: docker-release-arm64, artifactSuffix: arm64 },
+          { runner: ubuntu-latest, platform: linux/amd64, cache: docker-release-amd64, artifactSuffix: amd64 },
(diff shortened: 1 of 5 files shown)


mirarifhasan (changes requested):
Hi @josephyi 
I think we can remove the following step from the build stage:
```
      - name: Setup QEMU
        uses: docker/setup-qemu-action@v3
``` 
Everything else looks good to me — good to merge!

josephyi:
> Hi @josephyi I think we can remove the following step from the build stage:
> 
> ```
>       - name: Setup QEMU
>         uses: docker/setup-qemu-action@v3
> ```
> 
> Everything else looks good to me — good to merge!

Done!

mirarifhasan (dismissed):
Thanks, looks good to me 🚀

@josephyi Since the `ubuntu-24.04-arm` runner is currently in the public beta stage, let's hold off merging until it reaches general availability (GA).

PR #5458, opened 2025-10-07
  feat: add reusable JavaScript functions library for pre-request scripts [Hacktoberfest]

abhishekmaniy:
## Summary
  Implements a reusable JavaScript functions library that allows users to define and store common functions for use across multiple
  pre-request scripts. This addresses the need for better script reusability, reduces code duplication, and enhances developer productivity    
   by providing a centralized function management system.

  ## Key Features
  - **Centralized Function Library**: Create, edit, and manage reusable JavaScript functions in one place
  - **Rich Code Editor**: Monaco editor with JavaScript syntax highlighting and error detection
  - **Seamless Integration**: Functions automatically available in all pre-request scripts
  - **CRUD Operations**: Full create, read, update, delete functionality with duplication support
  - **Metadata Support**: Functions include names, descriptions, and timestamps
  - **Backward Compatible**: Existing pre-request scripts continue to work unchanged

  ## Changes Made
  - Add `HoppReusableFunction` data types and validation schemas
  - Create Vue component for function management UI with split-pane layout
  - Implement store for state management with reactive updates
  - Integrate function injection into pre-request script sandbox
  - Add localization strings for new UI elements
  - Update request options to include new "Reusable Functions" tab

  ## Usage Example
  1. Create a reusable function:
     ```javascript
     function generateTimestamp() {
       return new Date().toISOString();
     }

  2. Use it in any pre-request script:
  const timestamp = generateTimestamp();
  pw.env.set('currentTimestamp', timestamp);

  Technical Implementation

  - Data Layer: Zod schemas for type validation and data integrity
  - State Management: Reactive store with dispatchers for all CRUD operations
  - UI Components: Clean Vue 3 composition API with TypeScript
  - Script Execution: Integration with existing Faraday Cage sandbox
  - Type Safety: Full TypeScript support throughout the implementation

  Test Plan

  - Function creation, editing, and deletion works correctly
  - Functions execute properly in pre-request scripts
  - Backward compatibility maintained with existing scripts
  - UI is responsive and accessible
  - All localization strings display correctly
  - Type safety verified across all components


jamesgeorge007 (changes requested):
P1 - the reusable-functions tab shows regardless of `EXPERIMENTAL_SCRIPTING_SANDBOX`, and `RequestRunner` passes the functions into the flag-off branch too, but the worker fallback there never injects them, so with the experimental sandbox off, a reusable-function call throws `ReferenceError`. Since that path is reachable and already receives the functions, it needs to actually run them; if you'd prefer to keep this experimental-only, gating the tab behind the flag would avoid the silent break for opt-out users.

P2 - the bootstrap eval's `catch` is empty (`pre-request.js`), so if one function fails to eval, the whole library drops silently and the user just sees `X is not defined` later. A line to the sandbox console would make that much easier to trace.

P3 - the branch is conflicting and needs a rebase.

PR #5461, opened 2025-10-07
 Feature: Postman-Style Nested Environment Variables

Sambhram1:
This PR implements Postman-style nested environment variables in Hoppscotch, enabling dynamic variable key construction like {{ {{country-code}}{{environment}} }}. This addresses a critical user need where maintaining 450+ separate environments is now reduced to 15 dynamic variables in a single environment configuration.

🎯 Problem Solved
User Pain Point: Postman supports nested environment variables like {{ {{country-code}}{{environment}} }} but Hoppscotch doesn't. This limitation forces users to create hundreds of separate environments instead of using dynamic templating, significantly reducing productivity and increasing configuration complexity.

Before: 450+ separate environments
After: 15 dynamic variables in 1 environment

✨ Features Added
🔄 Dual Syntax Support
Existing Hoppscotch: <<variable>> (fully backward compatible)
NEW Postman Style: {{variable}}
Mixed Syntax: <<{{country}}>> or {{<<region>>}}

### DIFF
--- a/packages/hoppscotch-backend/package.json
+++ b/packages/hoppscotch-backend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "hoppscotch-backend",
-  "version": "2025.9.1",
+  "version": "2025.9.2",
   "description": "",
   "author": "",
   "private": true,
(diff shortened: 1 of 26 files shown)


Sambhram1:
please check this pr

Issue #5358, opened 2025-08-28
[bug]: Authorization code flow not starting with PKCE


macmessa:
### Is there an existing issue for this?

- [x] I have searched existing issues and this bug hasn't been reported yet

### Platform

Web App/Desktop App

### Browser

Edge Version 134.0.3124.72

### Operating System

Windows

### Bug Description

Since lots of the apps created in our authorization server are public clients, they implement the PKCE for client authentication.
I've tried to set an authorization code flow with PKCE, so the label `Code Challenge` shows up, but with no additional fields like `Code Verifier` and `Code Challenge Method`, I think this might be triggering a form validation when I click `Generate Token`, requesting these fields (they are needed).

<img width="893" height="871" alt="Image" src="https://github.com/user-attachments/assets/19679dce-f7de-4e76-979e-db805ec449b4" />

### Deployment Type

Self-hosted (on-prem deployment)

### Version

v2025.7.1

Gandhi11:
Same problem here. I think the "algorithms" dropdown should be shown which is not the case at the moment.

Gandhi11:
In local development the dropdown is there. But not in builded web app or desktop app.

<img width="482" height="266" alt="Image" src="https://github.com/user-attachments/assets/3ad9a27a-32ef-4516-9733-0adb6eca0f07" />

Gandhi11:
@macmessa Is the collection you are trying to add this is a imported collection? On my side it seems to work when it is a completely new collection...

macmessa:
> @macmessa Is the collection you are trying to add this is a imported collection? On my side it seems to work when it is a completely new collection...

Yes, turns out it was because it was an imported collection, I had to create a new one and recreate all of the endpoints, making the import useless for me, in this case.

anwarulislam:
Hello @Gandhi11, @macmessa,

Following the discussion, it seems that the issue may not be related to the PKCE implementation. The problem pertains to the failure to display all the required fields after the import of collections. However, when a new collection is created, it functions correctly. Could you please confirm if this is accurate or provide further details?

macmessa:
> Hello @Gandhi11, @macmessa,
> 
> Following the discussion, it seems that the issue may not be related to the PKCE implementation. The problem pertains to the failure to display all the required fields after the import of collections. However, when a new collection is created, it functions correctly. Could you please confirm if this is accurate or provide further details?

Exactly, only fails when it's a n imported collection, it does not display all required fields.

[...]

PR #5480, opened 2025-10-13
fix(common): preserve PKCE and client secret in postman collection imports

anwarulislam:
Closes FE-1028 #5358 

Enhance the import functionality to support PKCE settings and include the client secret in the authentication process.


Gandhi11:
@anwarulislam The changes to the import looks good, but it doesn't seems to solve the collection that were already imported. Maybe add some kind of default to 'plain' when the key doesn't exist in the settings?

anwarulislam:
Thanks for the check and the feedback, @Gandhi11.

I have modified the code to use 'plain' as the default when the key doesn't exist. As a result, older collections will now default to 'plain' as well.

PR #4211, opened 2024-07-24
feat: duplicate REST/GraphQL collections

jamesgeorge007:
### What's changed

This PR includes the UI changes adding the ability to duplicate REST/GraphQL collections. The context menu for collections (root/child) from the tree will include a new option labelled `Duplicate collection`. This will result in a new collection entry within the same hierarchy, effectively cloning the children and retaining the collection level authorization/header properties. Additionally, while the context menu is open, the duplication action can be initiated by pressing the `d` key.

> The duplication action is disabled under personal workspace in SH for now due to a known issue with syncing and will be revisited in a later iteration.

Closes HFE-491 HFE-545 HFE-547 #2475.

> The BE changes were implemented in #4207.

Apart from that, the following issues are patched:

- The `userCollectionCreated` and `teamCollectionCreated` GQL subscriptions fired after collection create actions (import, create, duplicate) sent collection-level properties (authorization/headers) applying stringification via `JSON.stringify()` multiple times which led to an exception since FE expects it to be parsed via `JSON.parse()` in one pass.
- Bulk export of REST collections from a team workspace results in the export format not confirming to `HoppCollection`. Attempting to export collections existing with authorization/headers set and importing them will result in the respective collection properties not getting populated. This is because the export format doesn't conform to the internal data structure `HoppCollection` and while importing it fails to parse and goes ahead with the defaults. This also led to the CLI rejecting bulk collection exports from a team workspace - [HFE-545](https://linear.app/hoppscotch/issue/HFE-545/bulk-export-of-collections-from-a-team-workspace-and-re-import-doesnt)
- There was an issue specific to team workspaces on cloud, where requests were lost while importing except for the last level. This also extends to the case where child collections/requests are being created for a collection in the non-expanded state having pre-existing children and expanding will reveal only the above additions. There was a change in behaviour in the `teamCollectionAdded` GQL subscription on cloud compared to SH where it was fired post-collection create actions. Child collections at all levels get populated under the imported collection or while creating child collections they get synced straightaway via the subscription. This has been patched at the tree level ensuring collection entries received via GQL subscriptions for a parent collection that is not expanded yet doesn't get populated - [HFE-547](https://linear.app/hoppscotch/issue/HFE-547/bulk-export-of-collections-from-a-team-workspace-and-re-import-doesnt).
- GraphQL collections properties modal can now be opened via the `p` keypress while the context menu is open.

### Preview

  https://github.com/user-attachments/assets/2536c93d-d5b0-4170-846d-fb95286767f7

[...]

PR #5532, opened 2025-10-30
feat(common): mock server ui improvements

anwarulislam:
- Update active state styles for better visibility in the mock server.
- BE updates catered to improving content type handling in the mock server.
- Introduced `disableMockServerInPersonalWorkspace` platform-level feature flag.
- Remove inactive keyboard shorthand nudges from the Mock server dashboard context menu.

### DIFF
--- a/packages/hoppscotch-common/src/components/collections/Collection.vue
+++ b/packages/hoppscotch-common/src/components/collections/Collection.vue
@@ -136,8 +136,7 @@
                   @keyup.t="runCollectionAction?.$el.click()"
                   @keyup.s="sortAction?.$el.click()"
                   @keyup.m="
-                    isMockServerVisible &&
-                    mockServerAction?.$el.click()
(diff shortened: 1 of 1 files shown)


packages/hoppscotch-common/src/platform/index.ts
```diff
@@ -66,6 +66,12 @@ export type PlatformDef = {
      * Whether to show the A/B testing workspace switcher click login flow or not
      */
     workspaceSwitcherLogin?: Ref<boolean>
+
+    /**
+     * Whether to disable mock servers in personal workspaces
+     * If a value is not given, then the value is assumed to be false
+     */
+    mockServerInPersonalWorkspace?: boolean
```
jamesgeorge007: Let's follow a minimal surface area approach, perhaps calling it `disableMockServerInPersonalWorkspace` following the conventions from #4211, where only the platforms that need it to be disabled specify it explicitly, and no mention is assumed as `false`.

Issue #5525, opened 2025-10-29
[bug]: Switch workspace need refresh data


esotericman:
### Is there an existing issue for this?

- [x] I have searched existing issues and this bug hasn't been reported yet

### Platform

Web App

### Browser

Chrome

### Operating System

macOS

### Bug Description

Before switching workspaces, switch from REST to GraphQL, then switch back to REST. At this point, when switching workspaces, the corresponding workspace's collections and other data are not refreshed.

### Deployment Type

Self-hosted (on-prem deployment)

### Version

25.9.2

esotericman:
I created multiple workspace, added different collections and apis, If i switched workspace directly, It's okay, but if I clicked GraphQL and clicked back to REST, and then switched workspace, collections and apis not updated anymore. I must to refresh browser to refetch related workspace data

jamesgeorge007:
Closing since this is fixed in the [latest release](https://github.com/hoppscotch/hoppscotch/releases/tag/2025.10.1).

PR #5533, opened 2025-10-30
fix: team collection not loading on route change

nivedin:
Closes FE-1043 https://github.com/hoppscotch/hoppscotch/issues/5525

This PR fixes an issue where the collection service was not re-initialized when the route changed or when it unmounted, causing the collections to not update when the workspace was changed.

### DIFF
--- a/packages/hoppscotch-common/src/services/workspace.service.ts
+++ b/packages/hoppscotch-common/src/services/workspace.service.ts
@@ -136,10 +136,14 @@ export class WorkspaceService extends Service<WorkspaceServiceEvent> {
 
   /**
    * Checks if two workspaces are effectively equal to avoid unnecessary updates
+   *
+   * Note: Vue's watch API provides `undefined` as `oldValue` on the first callback
+   * invocation when using `{ immediate: true }`, since there is no previous value yet.
+   * This is why `oldWorkspace` has an optional type, while `newWorkspace` is always defined.
    */
   private areWorkspacesEqual(
-    newWorkspace: Workspace | undefined,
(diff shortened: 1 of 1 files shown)


packages/hoppscotch-common/src/services/workspace.service.ts
```diff
@@ -101,6 +101,61 @@ export class WorkspaceService extends Service<WorkspaceServiceEvent> {
...
+      { immediate: true }
+    )
+  }
+
+  /**
+   * Checks if two workspaces are effectively equal to avoid unnecessary updates
+   */
+  private areWorkspacesEqual(
+    newWorkspace: Workspace | undefined,
+    oldWorkspace: Workspace | undefined
```
jamesgeorge007: Is there a scenario where `newWorkspace` can be `undefined`? The inferred type contracts from the watcher don't indicate. However, there's the runtime nuance with `oldWorkspace` resolved as `undefined` during the first load.

nivedin: while TypeScript inference suggests newWorkspace can't be undefined, the runtime behaviour of watchers and potential edge cases since it's an immediate call  mean it's theoretically possible, making the defensive programming in areWorkspacesEqual

jamesgeorge007 (approved):

packages/hoppscotch-common/src/services/workspace.service.ts
jamesgeorge007: Discussed separately and decided to move forward with aligning the implementation with the type contracts.

PR #5613, opened 2025-11-24
fix(common): agent interceptor dependency compat

CuriousCorrelation:
Fixes agent interceptor registration broken by dependency update

 Closes FE-1059

 ### What's changed

 Updated `KernelInterceptorAgentStore` to work with @noble/curves v2.0
 and @scure/base v2.0 breaking changes.

 Applied `Uint8Array` wrapper to all `base16.decode()` calls used with
 Web Crypto API for type compat.

 ### Testing

 Verify agent registration flow.

 ### Notes to reviewers

 The switch from `x25519.utils.randomPrivateKey()` to
 `crypto.getRandomValues()` uses browser-native cryptographically secure
 random generation, which is appropriate for this context and removes
 dependency on @noble/hashes.

 The `.js` extension in imports is required by @noble/curves v2.0's
 ESM-only distribution.

---
## Summary by cubic
Restores agent interceptor registration after dependency upgrades (FE-1059). Updates crypto usage to work with @noble/curves v2.0 and @scure/base v2.0.

- **Bug Fixes**
  - Use crypto.getRandomValues for x25519 private key generation.
  - Wrap base16.decode outputs in Uint8Array for Web Crypto compatibility (nonce, keys).
  - Fix types for watcher change event and completeRequest parameter.

<sup>Written for commit f605e85821a9aed81e4be2bc53872bb19fdbf323. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/hoppscotch-common/src/platform/std/kernel-interceptors/agent/store.ts
+++ b/packages/hoppscotch-common/src/platform/std/kernel-interceptors/agent/store.ts
@@ -323,7 +323,6 @@ export class KernelInterceptorAgentStore extends Service {
     const sharedSecretKeyBytes = new Uint8Array(
       base16.decode(this.sharedSecretB16.value!.toUpperCase())
     )
-
     const sharedSecretKey = await window.crypto.subtle.importKey(
       "raw",
       sharedSecretKeyBytes,


packages/hoppscotch-common/src/platform/std/kernel-interceptors/agent/store.ts
```diff
@@ -246,7 +246,9 @@ export class KernelInterceptorAgentStore extends Service {
     if (typeof newAuthKey !== "string")
       throw new Error("Invalid auth key received")
 
-    const agentPublicKey = base16.decode(agentPublicKeyB16.toUpperCase())
```
jamesgeorge007, line 249: https://github.com/hoppscotch/hoppscotch/blob/51840cb5e5e0526d5e186cfcdb9b701ff056f230/packages/hoppscotch-common/src/platform/std/interceptors/agent/index.ts#L628

packages/hoppscotch-common/src/platform/std/kernel-interceptors/agent/store.ts
```diff
@@ -228,7 +228,7 @@ export class KernelInterceptorAgentStore extends Service {
   }
 
   public async verifyRegistration(otp: string): Promise<void> {
-    const myPrivateKey = x25519.utils.randomPrivateKey()
```
jamesgeorge007, line 231: https://github.com/hoppscotch/hoppscotch/blob/51840cb5e5e0526d5e186cfcdb9b701ff056f230/packages/hoppscotch-common/src/platform/std/interceptors/agent/index.ts#L611

packages/hoppscotch-common/src/platform/std/kernel-interceptors/agent/store.ts
CuriousCorrelation: The files under `std/interceptor` are up for deprecation soon, need to gather a bit more context on that front before safely deprecating those since they have a few exports that are tied to user settings via persistence (also a few helper functions). Otherwise they are safe to be left as is without further updates (since only types are used and not functionality).

packages/hoppscotch-common/src/platform/std/kernel-interceptors/agent/store.ts
CuriousCorrelation: Same as above: The files under `std/interceptor` are up for deprecation soon, they are safe to be left as is without further updates (since only types are used and not functionality).

packages/hoppscotch-common/src/platform/std/kernel-interceptors/agent/store.ts
jamesgeorge007: Discussed separately, safe to keep as is with deprecation plans.

packages/hoppscotch-common/src/platform/std/kernel-interceptors/agent/store.ts
jamesgeorge007: https://github.com/hoppscotch/hoppscotch/pull/5613#discussion_r2559194998

PR #5619, opened 2025-11-26
fix: guard tauri calls with kernel check

CuriousCorrelation:
Store wrappers were calling Tauri APIs at module load time, causing
 "Cannot read properties of undefined" errors in web mode where
 `window.__TAURI_INTERNALS__` doesn't exist.

 Added `getKernelMode()` checks with dynamic imports before all Tauri
 API usage in `store.ts`. Desktop path resolution functions remain
 exported and functional.

 Closes FE-1069

 Consolidated sub-commit:

 - ref(platform): migrate instance switching from service to platform definition

 Updated `_version.vue` to access  instance state through
 `platform.instance` with proper feature detection for
 `instanceSwitchingEnabled` and stream availability.

 Closes FE-1070

---
## Summary by cubic
Guarded all Tauri API usage with kernel-mode checks and dynamic imports to prevent web-mode crashes, and moved instance switching to the platform definition. Fixes the “Cannot read properties of undefined” error in web while keeping desktop store path functions working. Closes FE-1069 and FE-1070.

- **Bug Fixes**
  - Added getKernelMode() guards and lazy imports in common and selfhost store.ts so Tauri APIs are only used in desktop.
  - Fallback to safe store paths when join/invoke aren’t available.
  - Single init promise for dynamic imports to avoid multiple loads and race conditions.

- **Refactors**
  - Migrated instance switching from a service to platform.instance with feature detection.
  - Updated _version.vue to read connection state via platform.instance streams when available.

<sup>Written for commit b38eba161ce8c2a9aad8917dc0a46b745a95ec8f. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/hoppscotch-common/src/kernel/store.ts
+++ b/packages/hoppscotch-common/src/kernel/store.ts
@@ -18,58 +18,68 @@ let invoke:
   | undefined
 let join: ((...paths: string[]) => Promise<string>) | undefined
 
-if (getKernelMode() === "desktop") {
-  import("@tauri-apps/api/core").then((module) => {
-    invoke = module.invoke
-  })
-  import("@tauri-apps/api/path").then((module) => {
-    join = module.join
-  })
+// Single init promise to avoid multiple imports and race conditions
+let initPromise: Promise<void> | undefined
+
+const isInitd = async () => {
+  if (getKernelMode() !== "desktop") return
+
+  if (!initPromise) {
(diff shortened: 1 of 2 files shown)


CuriousCorrelation:
While the diagnosis is correct, the solution doesn't account for multiple calls to dynamic imports causing another problem. It's much simpler to use a mixed promise to handle that in one go. Latest commit should fix that.

Issue #5513, opened 2025-10-24
[bug]: environment variable with dot is not highlighted


lchan8-intuit:
### Is there an existing issue for this?

- [x] I have searched existing issues and this bug hasn't been reported yet

### Platform

Desktop App

### Browser

Chrome

### Operating System

macOS

### Bug Description

If I add an environment variable with a dot in it (e.g. "my.var1") and use it in a field as <<my.var1>>, the value is substituted correctly when the request is sent.  However, the variable doesn't show any green or red highlight and appear as just plain text.  This makes it difficult to know if it is recognized as an environment variable or not.

### Deployment Type

Self-hosted (on-prem deployment)

### Version

2025.8.1

tlawjddnjs:
Hi, I'd like to work on this issue.  
Could you please assign it to me if it's still available?

myeonbi:
I tested the same variable on the latest Web version of Hoppscotch, and <<my.var1>> is highlighted correctly as an environment variable.

So it seems that the issue may be specific to the Desktop App / self-hosted 2025.8.1 build, since the Web build already recognizes and highlights variables with dots.

<img width="1210" height="704" alt="Image" src="https://github.com/user-attachments/assets/9679b742-870e-43c4-b275-1f80af98de0a" />

Could you share:

The exact Desktop app version (from the About menu)

Whether you see the same behavior on the Web app

Any custom theme or editor settings you’re using

This will help narrow down whether the highlighting parser in the Electron bundle is outdated or behaving differently.

Thanks!

lchan8-intuit:
@myeonbi I have tried the latest hoppscotch.io on the web, which is running v2025.10.1.  While variables in dot notation are now highlighted, there is still issue.  The highlight is correct in URL and header fields, but it always shows as red for any variables in the body.

Here is a screenshot:

<img width="1157" height="433" alt="Image" src="https://github.com/user-attachments/assets/00581ca5-c0fd-4c3b-a912-6b62eeb592b0" />

myeonbi:
Thanks for the screenshot and the detailed update!

I’m currently investigating this as a follow-up after my PR (#5630).  
Thanks for pointing out what I may have missed — I’ll continue looking into it!

PR #5630, opened 2025-11-28
fix: support dots/dashes in environment variable autocomplete

myeonbi:
Closes # 

### Summary

This PR aligns the environment variable autocomplete regex with the shared pattern defined in `environment-regex.ts`.

Currently, `environment-regex.ts` defines `ENV_VAR_NAME_PATTERN` as `[a-zA-Z0-9_.-]+`, which allows dots (`.`) and dashes (`-`). However, the autocomplete logic in `EnvInput.vue` was using `\w*`, which only supports `[A-Za-z0-9_]`. This caused inconsistency where variables like `<<client.id>>` or `<<client-id>>` were not properly recognized in the autocomplete dropdown.

### What's changed

- **Updated Regex:** `tagBefore` and `validFor` regexes in `EnvInput.vue` now allow `[A-Za-z0-9_.-]`.
- **Consistency:** This ensures the autocomplete logic matches `ENV_VAR_NAME_PATTERN` from `environment-regex.ts`.
- **Behavior Change:** - *Before:* Environment variables with dots or dashes (e.g., `client.id`) did not trigger autocomplete properly.
    - *After:* These variables are now correctly recognized and autocompleted.

### Notes to reviewers

- This may be related to issue #5513 (Desktop app environment recognition), as it involves similar regex inconsistencies with special characters in variable names.

---
## Summary by cubic
Updated environment variable autocomplete to allow dots and dashes in names, matching the shared ENV_VAR_NAME_PATTERN. Tags like <<client.id>> and <<client-id>> now trigger suggestions and won’t insert duplicate >> if closing brackets already exist.

<sup>Written for commit a361018fd072d9d54826b50eb25d81cd8d9fff01. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/hoppscotch-common/src/components/smart/EnvInput.vue
+++ b/packages/hoppscotch-common/src/components/smart/EnvInput.vue
@@ -441,19 +441,37 @@ const envVars = computed(() => {
 })
 
 function envAutoCompletion(context: CompletionContext) {
-  const options = (props.autoCompleteEnvSource ?? envVars.value ?? [])
-    .map((env) => ({
-      label: env?.key ? `<<${env.key}>>` : "",
-      info: env?.currentValue ?? "",
-      apply: env?.key ? `<<${env.key}>>` : "",
-    }))
-    .filter(Boolean)
-
   const nodeBefore = syntaxTree(context.state).resolveInner(context.pos, -1)
   const textBefore = context.state.sliceDoc(nodeBefore.from, context.pos)
   const tagBefore = /<<\$?[A-Za-z0-9_.-]*$/.exec(textBefore) // Update regex to match <<$ as well
 
   if (!tagBefore && !context.explicit) return null
+
+  // Check if there's already a closing ">>" after the cursor
+  const textAfter = context.state.sliceDoc(context.pos, context.pos + 2)
+  const hasClosingBrackets = textAfter === ">>"
+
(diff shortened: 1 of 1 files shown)


anwarulislam (approved):
@jamesgeorge007 I've added a commit that fixed an existing issue where autocompleting a variable inside existing `<<>>` brackets would add extra closing brackets if the closing `>>` already exists after the cursor position.

myeonbi:
Thank you for the suggestion!

I actually took a look at the issue where environment variable references are not recognized in the raw request body editor through issue #5513, and I’m currently investigating it in more depth.

Since this seems to involve a different part of the flow, would it be better to handle it in a separate PR?  
I'm not yet sure how long a complete fix will take, but I’ll continue exploring it.

If the current PR looks good to you, I would appreciate it if it could be merged.  
Thank you again for the review and guidance!

jamesgeorge007:
> Thank you for the suggestion!
> 
> I actually took a look at the issue where environment variable references are not recognized in the raw request body editor through issue #5513, and I’m currently investigating it in more depth.
> 
> Since this seems to involve a different part of the flow, would it be better to handle it in a separate PR? I'm not yet sure how long a complete fix will take, but I’ll continue exploring it.
> 
> If the current PR looks good to you, I would appreciate it if it could be merged. Thank you again for the review and guidance!

Sure thing! Thanks for contributing. Feel free to explore and tackle the raw body editor concern separately.

PR #5634, opened 2025-11-30
fix: detect trailing comma and report error at correct position

### LINKED ISSUE #5502: [bug]: not detecting invalid JSON (trailing comma) in request body editor.

### Platform

Web App

### Browser

Chrome

### Operating System

Linux

### Bug Description

When I send a request, the server throws a JSON parse error saying the payload is invalid.
But in the editor, no error is shown in the request body, it looks perfectly valid.

After spending some time debugging, I found out the issue was caused by a trailing comma in the JSON payload.

### Example payload
```json
{
  "origins": [
    {
      "waypoint": {
        "location": {
          "latLng": {
            "latitude": 28.6330,
            "longitude": 77.2193
          }
        }
      },
      "routeModifiers": { "avoid_tolls": false }
    }
  ],
  "destinations": [
    {
      "waypoint": {
        "location": {
          "latLng": {
            "latitude": 28.5245,
            "longitude": 77.1859
          }
        }
      }
    }, //trailing comma here
  ],
  "travelMode": "DRIVE",
  "routingPreference": "TRAFFIC_AWARE"
}
```

### Why This Matters:
This can be confusing and time-wasting for developers, the request appears valid in the editor but fails after hitting the server. It’s especially annoying when debugging API issues.

### Deployment Type

Hoppscotch Cloud

### Version

v2025.9.2

tlawjddnjs:
Closes #5502

This PR fixes an issue where trailing commas in JSON objects or arrays were not detected as invalid JSON in the request body editor. The server rejected such bodies, but the editor did not highlight the error. This change updates the parser to properly detect and report trailing commas with accurate error positions.

### What's changed

- Added explicit trailing comma detection in `parseObj` and `parseArr`.
- Store the comma token range before consuming it with `expect(",")`.
- When the next token after the comma is a closing `}` or `]`, report a `SyntaxError` pointing to the comma token rather than the closing bracket/brace.
- Ensures that invalid JSON with trailing commas is now surfaced correctly in the UI.

### Notes to reviewers

- The fix is strictly scoped to JSON trailing comma handling.
- No existing behavior for valid JSON structures is affected.
- The error reporting position now aligns with VSCode-like behavior, improving the editor’s feedback accuracy.

---
## Summary by cubic
Detects trailing commas in JSON objects/arrays and highlights the comma as the error in the request body editor. Aligns UI validation with server behavior for clearer feedback.

- **Bug Fixes**
  - Added trailing comma checks in parseObj/parseArr and throw with the comma token range.
  - Store comma start/end before consuming "," to report the exact error position.
  - Unified root parsing via parseVal + EOF, replacing the object/array try-catch flow.

<sup>Written for commit ecd614e875bc69703e8e0aa47b26e225cb450fa1. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/hoppscotch-common/src/helpers/jsoncParse.ts
+++ b/packages/hoppscotch-common/src/helpers/jsoncParse.ts
@@ -140,7 +140,6 @@ function parseObj(): JSONObjectValue {
 
       // After consuming ",", if the next token is "}", this is a trailing comma
       if (kind === "}") {
-        // After consuming ",", if the next token is "}", this is a trailing comma
         throw {
           message: "Trailing comma in object is not allowed in JSON.",
           start: commaStart,


anwarulislam:
Hello @tlawjddnjs,

We support JSONC, which does not generate a warning for trailing commas after values. This feature is intentional, as it allows users to include comments and trailing commas. However, please ensure that the comma is removed prior to sending the request. If this is not occurring, you can address that issue within this PR.

tlawjddnjs:
Hello @anwarulislam, thanks for the clarification.

I fully understand that Hoppscotch intentionally supports JSONC (comments + trailing commas).
My PR does NOT attempt to change or restrict JSONC behavior in the editor.

Instead, the issue I am fixing is:

When the request is actually sent, the body is still treated as JSON.

Trailing commas in the body cause the server to reject the request (invalid JSON).

Hoppscotch currently does not warn the user nor remove the trailing comma before sending.

So the purpose of this PR is to detect a trailing comma only when parsing JSON values, to prevent sending invalid JSON to the server.

To be clear:

JSONC editing remains allowed.

Only invalid JSON syntax that breaks the outgoing request is detected.

If JSONC should remain untouched in the editor, I can update the PR so that the trailing comma check is applied only at the moment of request serialization, not during editing time.

Please let me know which direction you prefer — I am fully willing to adjust the PR to align with Hoppscotch’s JSONC design.

PR #5642, opened 2025-12-01
fix: resolve environment variable resolution in URL path Fixes #5621

### LINKED ISSUE #5621: [bug]: Environment  values not working  in url like (baseURL) when import postman environment file

### Platform

Web App

### Browser

Chrome

### Operating System

macOS

### Bug Description

Here’s an updated issue text you can paste and tweak:

***

## What happened?

When importing a Postman environment file into Hoppscotch and using it in requests, the app does not correctly resolve environment variables in the request path. The variable placeholder `<<baseUrl>>` stays highlighted in purple in the URL bar instead of being replaced with the actual value, and every request returns a 405 status code even though the backend is running correctly.[11][12]

## Steps to reproduce

1. Open the web version at https://hoppscotch.io/.  
2. Import a Postman environment JSON file that defines a `baseUrl` variable with value `http://localhost:8089`.  
3. Select the imported environment.  
4. Create or open a request whose URL uses the variable in the path, for example `<<baseUrl>>/api/...`.  
5. Send the request.  
6. Observe that:  
   - The URL field still shows `<<baseUrl>>/api/...` in purple instead of resolving to `http://localhost:8089/api/...`.  
   - The response status is always 405.

## What I expected

- Hoppscotch should replace `<<baseUrl>>` with `http://localhost:8089` in the request URL when the environment is active.  
- The request should be sent to the correct endpoint and return the same successful response that is obtained when running the equivalent curl command directly in the terminal.

## Screenshots

Screenshots attached showing the unresolved `<<baseUrl>>` in the URL bar and the environment configuration.[12][11]

## Additional context

- In previous versions of Hoppscotch this exact setup worked perfectly; the same Postman environment import and requests resolved the variables correctly and the API calls succeeded.  
- When using “Generate curl” for the same request, the generated curl command contains the correct expanded URL (`http://localhost:8089/...`), and running that curl command in the terminal works as expected, returning a successful response.  
- This suggests the environment value is loaded correctly for curl generation but not applied when actually sending the request from the web UI.  
- Backend: local server on `http://localhost:8089` that responds correctly to the same endpoint when called via curl.

Image
Image
Image
Image

### Deployment Type

Hoppscotch Cloud

### Version

2025.10.1

FaisalFaraj:
Closes #5621

Hey there! This PR fixes the issue where environment variables weren't being properly parsed when importing Postman environment files, specifically when trying to use them in URLs like {{baseURL}}.
The problem was that after importing a Postman environment, the variables would show up in Hoppscotch but wouldn't actually resolve when used in request URLs. Pretty frustrating when you're trying to switch between dev/staging/prod environments!

What's changed
* Fixed the environment variable parsing logic to properly handle Postman's environment file format
* Updated the variable interpolation mechanism to correctly resolve {{baseURL}} and other env vars in the URL field
* Added proper validation to ensure imported variables are accessible immediately after import
* Tested with multiple Postman environment files to confirm variables now work as expected

Notes to reviewers
I've tested this with a few different Postman environment exports (both simple and complex setups with nested variables). Everything seems to be working smoothly now - variables resolve in URLs right after import without needing to refresh or re-enter them manually.
The main change was in how we were mapping Postman's env structure to our internal format. Turns out we were missing a step in the transformation pipeline that was causing the variables to exist but not be properly linked for URL resolution.
Let me know if you'd like me to add any additional test cases or if there are edge cases I should look into!

---
## Summary by cubic
Fixes environment variable resolution in request URLs so placeholders imported from Postman environments resolve reliably in paths, query strings, codegen, and cURL views.

- **Bug Fixes**
  - Resolve env vars in URLs and codegen/cURL views using aggregate environments, including imported Postman variables.
  - Expand placeholders inside collection variable values; avoid circular references; prioritize environment vars over collection vars on key conflicts.
  - Fallback to initialValue when currentValue is empty across environment sources.
  - parseTemplateStringE now keeps unresolved tags (e.g., <<var>>) instead of empty strings and stops looping when no change; secret masking uses current or initial value length.
  - Filter/dedup variables after resolving placeholders, preferring real values over unresolved placeholders.
  - Editor/UX: tooltips show resolved values and allow editing for all sources; autocomplete prioritizes environment vars; highlight precedence updated; request runner logs the final URL used.

<sup>Written for commit a325d8558ecfbb37f541421ef9f1b11295dfe824. Summary will update automatically on new commits.</sup>


FaisalFaraj:
@nivedin  Hello! I hope you have a moment to review my pull request. I can't wait to see my changes merged into Hoppscotch!

PR #5658, opened 2025-12-04
feat(desktop): atomic write for registry persistence

### LINKED ISSUE #5599: [bug]: tauri_plugin_appload: Failed to initialize storage manager error=Serialization error (Appl fails to startup)

### Platform

Desktop App

### Browser

Other (specify in additional info)

### Operating System

Windows

### Bug Description

# What I did
I tried to open the desktop app by clicking the icon

# what I expected to happen
The application starts up

# what Happened
The application failed to start. The window did not open. There was no process in the taskmanager assosciated with Hoppscotch.

# Environement Details 
Edition	Windows 11 Pro
Version	25H2
Installed on	‎2025-‎07-‎11
OS build	26200.7171
Experience	Windows Feature Experience Pack 1000.26100.265.0

System handled by Company Administrator Users

# Further Details 
I checked the application logs in the path : C:\Users\my.user\AppData\Roaming\io.hoppscotch.desktop and got the following info from io.hoppscotch.desktop.log file : 
> ThreadId(01) hoppscotch_desktop_lib: Error while running Hoppscotch Desktop error=failed to initialize plugin `appload`: Configuration error: Serialization error: trailing characters at line 11 column 2

 # what I tried to fix the issue

1. searched for solutions online
2. tried removing cache by deleting C:\Users\my.user\AppData\Local\io.hoppscotch.desktop folder and restarted the app. The folder was created again, but the issue still persists. Created a copy of the original folder for future reference.
3. tried searching for any files with invalid format that might cause the serialization error.

# Version
2025.10.1 
App updates are handled by company Administator users.

### Deployment Type

### Version

2025.10.1

### LINKED ISSUE #5632: [bug]: Corrupted registry.json Prevents Hoppscotch Desktop From Starting on Windows 11

### Platform

Desktop App

### Browser

Chrome

### Operating System

Windows

### Bug Description

Hoppscotch Desktop fails to start on Windows 11 due to an automatically generated configuration file (registry.json) containing invalid JSON.
The issue occurs in the appload plugin, which crashes during deserialization because of extra trailing characters added to the file.

Environment

Operating System: Windows 11
Hoppscotch Desktop Version: v25.11.0
Mode: Standard
Config directory: C:\Users\<user>\AppData\Roaming\io.hoppscotch.desktop

Problem Description:
When launching Hoppscotch Desktop, the application immediately exits.
The logs show:

failed to initialize plugin `appload`: Configuration error: Serialization error: trailing characters at line 11 column 2

After inspecting the configuration, the generated registry.json file contained extra closing braces, resulting in invalid JSON:

{
  "version": 1,
  "servers": {
    "https://vendor": {
      "bundle_name": "hoppscotch",
      "version": "0.1.0",
      "created_at": "2025-11-20T08:32:18.686798Z",
      "last_accessed": "2025-11-20T08:32:18.686798300Z"
    }
  }
}}}

Because of this corruption, the appload plugin fails and the application cannot start.

### Deployment Type

Hoppscotch Cloud

### Version

25.11.0

CuriousCorrelation:
This implements atomic writes for registry.json using temp file plus
 rename to prevent file corruption on Windows 11.

 Closes FE-1078
 Closes #5632
 Closes #5599

 The `Registry::save` method previously wrote directly to registry.json.
 On Windows 11, several factors can interrupt or interfere with writes
 and when interrupted, the file can end up with partial new content plus
 remnants of old content, resulting in malformed JSON like trailing `}}}`
 characters.

 Users report "failed to initialize plugin appload:
 Configuration error: Serialization error: trailing characters at line 11
 column 2" preventing application startup.

 ### What's changed

 Updated `Registry::save` in `tauri-plugin-appload/src/storage/registry.rs`
 to write to a temporary file first, then atomically rename:

 ```rust
 pub async fn save(&self, layout: &StorageLayout) -> Result<()> {
     let content = serde_json::to_string_pretty(self)?;
     let path = layout.registry_path();
     let temp_path = path.with_extension("json.tmp");

     // this is where that swap happens,
     // essentially this writes to a tmp file
     tokio::fs::write(&temp_path, &content).await?;
     // and this atomically "swaps" them
     tokio::fs::rename(&temp_path, &path).await?;

     Ok(())
 }
 ```

 This pattern is already used in `store_bundle` in `manager.rs`. NTFS
 rename is atomic within the same volume, so if the write is interrupted
 the original file remains intact.

 ### Notes to reviewers

 This is **defensive programming rather than a confirmed fix** for the
 reported issues. **The exact root cause of the Windows 11 corruption
 is still uncertain**, but atomic writes via temp file plus rename is
 standard practice for config file persistence and protects against
 the identified failure modes.

 Testing should verify on all platform and Windows 11 that normal
 save/load cycles work correctly and no leftover `.json.tmp` files
 remain after successful saves.

---
## Summary by cubic
Make registry.json writes atomic to prevent Windows 11 file corruption and startup failures in the desktop appload plugin. Addresses Linear FE-1078 by saving via a temp file and atomic rename.

- **Bug Fixes**
  - Save to .json.tmp and atomically rename to registry.json to avoid partial writes.
  - Prevents malformed JSON and the “Serialization error: trailing characters” on startup.

- **Dependencies**
  - Bumped tauri-plugin-appload across desktop/common/selfhost packages to include the atomic write change.
  - Cargo lock updated accordingly.

<sup>Written for commit 180f919f6d34daff0c2b3d7964dbe11ddccaa53a. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/hoppscotch-desktop/plugin-workspace/tauri-plugin-appload/src/storage/registry.rs
+++ b/packages/hoppscotch-desktop/plugin-workspace/tauri-plugin-appload/src/storage/registry.rs
@@ -1,7 +1,6 @@
 use chrono::{DateTime, Utc};
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
-use tokio::fs;
 use url::Url;
 
 use super::error::Result;
@@ -30,13 +29,18 @@ impl Registry {
             });
         }
 
-        let content = fs::read_to_string(layout.registry_path()).await?;
+        let content = tokio::fs::read_to_string(layout.registry_path()).await?;
         Ok(serde_json::from_str(&content)?)
     }
 
     pub async fn save(&self, layout: &StorageLayout) -> Result<()> {
         let content = serde_json::to_string_pretty(self)?;
-        fs::write(layout.registry_path(), content).await?;
+        let path = layout.registry_path();
+        let temp_path = path.with_extension("json.tmp");
+
+        tokio::fs::write(&temp_path, &content).await?;
+        tokio::fs::rename(&temp_path, &path).await?;
+
         Ok(())
     }
 
--- a/packages/hoppscotch-common/package.json
+++ b/packages/hoppscotch-common/package.json
@@ -40,7 +40,7 @@
     "@hoppscotch/httpsnippet": "3.0.9",
     "@hoppscotch/js-sandbox": "workspace:^",
     "@hoppscotch/kernel": "workspace:^",
-    "@hoppscotch/plugin-appload": "github:CuriousCorrelation/tauri-plugin-appload#e05861959938b57479a1a81fa796735ebbd08c7c",
+    "@hoppscotch/plugin-appload": "github:CuriousCorrelation/tauri-plugin-appload#52744a8f35bf81b039410522efd4168bd06d4f35",
     "@hoppscotch/ui": "0.2.5",
     "@hoppscotch/vue-toasted": "0.1.0",
     "@lezer/highlight": "1.2.1",
--- a/packages/hoppscotch-desktop/package.json
+++ b/packages/hoppscotch-desktop/package.json
@@ -24,7 +24,7 @@
     "@fontsource-variable/roboto-mono": "5.2.8",
     "@hoppscotch/common": "workspace:^",
     "@hoppscotch/kernel": "workspace:^",
-    "@hoppscotch/plugin-appload": "github:CuriousCorrelation/tauri-plugin-appload#e05861959938b57479a1a81fa796735ebbd08c7c",
+    "@hoppscotch/plugin-appload": "github:CuriousCorrelation/tauri-plugin-appload#52744a8f35bf81b039410522efd4168bd06d4f35",
     "@hoppscotch/ui": "0.2.5",
     "@tauri-apps/api": "2.1.1",
     "@tauri-apps/plugin-fs": "2.0.2",
--- a/packages/hoppscotch-desktop/src-tauri/Cargo.lock
+++ b/packages/hoppscotch-desktop/src-tauri/Cargo.lock
@@ -2426,7 +2426,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.6.1",
+ "socket2 0.5.10",
  "system-configuration",
  "tokio",
  "tower-service",
@@ -2446,7 +2446,7 @@ dependencies = [
  "js-sys",
  "log",
  "wasm-bindgen",
- "windows-core 0.57.0",
+ "windows-core 0.61.2",
 ]
 
 [[package]]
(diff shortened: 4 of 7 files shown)


packages/hoppscotch-desktop/plugin-workspace/tauri-plugin-appload/src/storage/registry.rs
```diff
@@ -30,13 +29,18 @@ impl Registry {
...
     pub async fn save(&self, layout: &StorageLayout) -> Result<()> {
         let content = serde_json::to_string_pretty(self)?;
-        fs::write(layout.registry_path(), content).await?;
+        let path = layout.registry_path();
+        let temp_path = path.with_extension("json.tmp");
+
+        tokio::fs::write(&temp_path, &content).await?;
+        tokio::fs::rename(&temp_path, &path).await?;
+
         Ok(())
```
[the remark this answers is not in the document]
CuriousCorrelation: No the file remaining there is important for future debugging purposes. If a write fails, there will still be the `.json.tmp` file which can be then manually swapped. If the swap begins after writing and there's a failure mid-way, there's no way to recover the state that was supposed to be present.

In the other case, if the operator doesn't care about the prior state or if the state can be recovered without manual swapping, the `.json.tmp` will be overwritten. So the .json.tmp overwrite issue exists but the impact is low for this particular file since the chances are halved. For something like user data or credentials, we'd want more defensive handling.

packages/hoppscotch-desktop/plugin-workspace/tauri-plugin-appload/src/storage/registry.rs
```diff
@@ -30,13 +29,18 @@ impl Registry {
...
     }
 
     pub async fn save(&self, layout: &StorageLayout) -> Result<()> {
         let content = serde_json::to_string_pretty(self)?;
-        fs::write(layout.registry_path(), content).await?;
+        let path = layout.registry_path();
+        let temp_path = path.with_extension("json.tmp");
+
+        tokio::fs::write(&temp_path, &content).await?;
+        tokio::fs::rename(&temp_path, &path).await?;
```
jamesgeorge007, line 42: Would it be worth accounting for cleanup in case the rename action fails, preventing orphaned temp files?

CuriousCorrelation: Oh yeah that's similar to what copilot suggested, here's the [context](https://github.com/hoppscotch/hoppscotch/pull/5658#discussion_r2588001981), basically since json.tmp gets replaced, and there's a chance that writing to .tmp file might fail (not rename) having it linger around can be a way for the user to manually get back their state.

One the other hand if the removal fails, we got the same problem on our hand, although for the tmp file rather than config. Since the surface area for this is quite low (not user configs), it's fine to keep those as is, since removal failing will reset user configuration unknowingly, whereas user regenerating the state is done knowingly.

Like for registry.json:
```
pub struct Registry {
    version: u32,
    servers: HashMap<String, ServerEntry>,
}
```
this is server/bundle metadata. If a save fails, the user might need to re-add a server or re-download a bundle, but it's recoverable. The full state can be reconstructed through normal app usage.

jamesgeorge007: Got it 👍

PR #5729, opened 2025-12-26
fix: prevent open redirect in desktop auth callback (CWE-601)

rajnishprecogs:
Closes # 

### Notes to reviewers

---
## Summary by cubic
Prevents open redirects in the desktop auth callback by parsing redirect_uri and strictly limiting allowed hosts to loopback. Blocks malicious URLs like http://localhost.evil.com.

- **Bug Fixes**
  - Use URL parsing instead of a startsWith check.
  - Allow only http scheme and hosts: localhost, 127.0.0.1, [::1]; otherwise return 400.

<sup>Written for commit 478433e10b4c30c22a81834ac3f9ea1f355865ef. Summary will update automatically on new commits.</sup>

### DIFF
--- a/packages/hoppscotch-backend/src/auth/auth.controller.ts
+++ b/packages/hoppscotch-backend/src/auth/auth.controller.ts
@@ -204,7 +204,26 @@ export class AuthController {
     @GqlUser() user: AuthUser,
     @Query('redirect_uri') redirectUri: string,
   ) {
-    if (!redirectUri || !redirectUri.startsWith('http://localhost')) {
+    // FIX: Use proper URL parsing to prevent Open Redirect attacks (CWE-601)
+    // Previous vulnerable code used: redirectUri.startsWith('http://localhost')
+    // This was bypassable with URLs like http://localhost.evil.com
+    let parsedUrl: URL;
+    try {
+      parsedUrl = new URL(redirectUri || '');
+    } catch {
+      throwHTTPErr({
+        message: 'Invalid desktop callback URL',
+        statusCode: 400,
+      });
+      return; // Type narrowing for TypeScript
+    }
+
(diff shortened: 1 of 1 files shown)


rajnishprecogs:
## Summary
This PR fixes a security vulnerability in the desktop authentication callback endpoint that could allow attackers to steal user authentication tokens.

## Vulnerability Details

**Type:** Open Redirect (CWE-601)  
**Severity:** High  
**CVSS Score:** 7.4

### Root Cause
The [desktopAuthCallback](cci:1://file:///Users/rajnish/Documents/Documents_mac/Precogs_Code/Dec%202025/Social%20Media/hoppscotch-scan/packages/hoppscotch-backend/src/auth/auth.controller.ts:202:2-236:3) function validates the `redirect_uri` parameter using `startsWith('http://localhost')`. This is bypassable because URLs like `http://localhost.evil.com` pass the check while resolving to an attacker-controlled server.

### Exploit Scenario
1. Attacker crafts: `https://hoppscotch.io/auth/desktop?redirect_uri=http://localhost.evil.com/capture`
2. Victim clicks link while authenticated
3. Auth tokens sent to attacker's server

## Fix
Replace prefix check with proper URL parsing and strict hostname validation.

---
Found using [Precogs AI](https://precogs.ai) - AI-powered security analysis.

CuriousCorrelation:
Hi @rajnishprecogs thanks for the PR. Could you elaborate a bit more on how an attacker will craft such redirects? Looking to understand the full scope of this since this implies access to redirection mechanism that not only affects the desktop but also the web client. I'm assuming this will be via a phishing scam perhaps?

Also please take a look at the suggestions from Copilot, and see if they make sense for the objective you are going for. Thanks again.

Issue #5777, opened 2026-01-16
[bug]: typing / in the scripting will open the support menu


Leon-Luu:
### Is there an existing issue for this?

- [x] I have searched existing issues and this bug hasn't been reported yet

### Platform

Web App

### Browser

Chrome

### Operating System

Windows

### Bug Description

On the Pre/Post request script when you type / it will pops up the Support Menu and makes it impossible to make code comments etc on the scripting.. This is very annoying bug

<img width="1047" height="880" alt="Image" src="https://github.com/user-attachments/assets/5cb15be2-37bb-4e0d-957e-bab4a1aa4853" />

### Deployment Type

Hoppscotch Cloud

### Version

2025.12.1

### How to reproduce
1) Use Hoppscotch desktop or webapp (https://hoppscotch.io/)
2) Go to Pre-request Script or Post-request Script
3) Type /  => Resulting Support menu pops up
4) Expected: Developer can write code comments like / in the scripts

arun4374:
Hi, i'd like to work on this issue. Let me know if i can take it

Leon-Luu:
@jamesgeorge007 @anwarulislam is it possible to merge this on coming release?

jamesgeorge007:
> [@jamesgeorge007](https://github.com/jamesgeorge007) [@anwarulislam](https://github.com/anwarulislam) is it possible to merge this on coming release?

Could you confirm if this behaviour is specific to a certain keyboard layout, along with additional steps to reproduce?

Leon-Luu:
I had added how "How to reproduce"

jamesgeorge007:
Closing since this is now addressed in the [v2026.1.0](https://github.com/hoppscotch/hoppscotch/releases/tag/2026.1.0) release.

PR #5811, opened 2026-01-29
fix(common): prevent support menu from triggering in editors

Leon-Luu:
**What the fix does:**

Prevents the Support menu (triggered by Shift+/) from opening when typing "/" in CodeMirror or Monaco editors. The fix adds a defensive check at the action execution level to ensure that when you're typing in an editor, the keystroke is treated as text input rather than a keyboard shortcut.

Why both fixes are needed:

**Old fix** (65046526f, line 280): Attempts to prevent the binding string generation early in the event flow. When it works, it's more efficient - avoids unnecessary lookups and checks.

**New fix** : Catches cases where the old fix fails due to event capture phase timing issues. With capture: true (973572d06 ), events are intercepted very early, sometimes before CodeMirror's focus handlers complete. This causes the early check to miss some cases.

Together they provide "defense in depth" where The old fix handles most cases efficiently, while the new fix acts as a safety net for edge cases caused by the capture phase event handling. Both are needed because the capture phase can cause timing-dependent detection failures that neither fix alone can handle reliably.

Closes # 
https://github.com/hoppscotch/hoppscotch/issues/5777

### Notes to reviewers

---
## Summary by cubic
Prevents the Support menu (Shift+/) from opening when typing "/" in CodeMirror, Monaco, or input fields. Typing now behaves normally without accidental shortcut triggers.

- **Bug Fixes**
  - Added a binding-level check for "shift-/" in the keydown handler.
  - Skip the shortcut when the focused element is a CodeMirror editor, Monaco editor, or any typable element.

<sup>Written for commit c042452677cebccbaf98d04a024c48ddfeed6965. Summary will update on new commits.</sup>

### DIFF
--- a/packages/hoppscotch-common/src/services/inspection/index.ts
+++ b/packages/hoppscotch-common/src/services/inspection/index.ts
@@ -181,11 +181,18 @@ export class InspectionService extends Service {
         maxWait: 2000,
       })
 
-      const inspectorRefs = computed(() =>
-        Array.from(this.inspectors.values()).map((x) =>
-          x.getInspections(debouncedReq, debouncedRes)
+      const inspectorRefs = computed(() => {
+        if (debouncedReq.value === null) return []
+
+        return Array.from(this.inspectors.values()).map((inspector) =>
+          inspector.getInspections(
+            debouncedReq as Readonly<
+              Ref<HoppRESTRequest | HoppRESTResponseOriginalRequest>
+            >,
+            debouncedRes
+          )
         )
-      )
+      })
 
       const activeInspections = computed(() =>
         inspectorRefs.value.flatMap((x) => x?.value ?? [])


Leon-Luu:
@jamesgeorge007 another patch for https://github.com/hoppscotch/hoppscotch/issues/5777 as during the time I made the initial fix I didn't have the latest codes where https://github.com/hoppscotch/hoppscotch/commit/973572d060d59ce2c8bced9e72ebdb0b6560dd0a was missing on my codebase hence the issue is still seen on the production. This patch is working fine now.

CuriousCorrelation:
@Leon-Luu Thanks for the fix!

One small suggestion would be to consider adding a comment explaining why this check exists alongside the one in the string generation so future maintainers may get the context around the duplication.

The above suggestion isn't a blocker since `git blame` can be used to triage, it's just about convenience, so approving 👍

jamesgeorge007 (approved):
Thanks!

Issue #4880, opened 2025-03-12
[bug]: Linux Display Broken on Wayland in 2025.2.2


nathan-pisarski:
### Is there an existing issue for this?

- [x] I have searched existing issues and this bug hasn't been reported yet

### Current behavior

This is very similar to https://github.com/hoppscotch/hoppscotch/issues/3516 - on **2025.2.2**, when I launch the application I get a small white window with no other details. The terminal output is:

```bash
 INFO hoppscotch_desktop: Starting Hoppscotch Desktop...
 INFO hoppscotch_desktop_lib: Starting Hoppscotch Desktop v25.2.2
 INFO hoppscotch_desktop_lib::server: Beginning server initialization
 INFO hoppscotch_desktop_lib: Server initialization complete port=18338
 INFO hoppscotch_desktop_lib::server: Starting HTTP server address=127.0.0.1:18338 port=18338
 INFO hoppscotch_desktop_lib::server: Server bound successfully local_addr=127.0.0.1:18338
 INFO hoppscotch_desktop_lib::server: Server starting
Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...
```

![Image](https://github.com/user-attachments/assets/05790b79-9669-4b65-8398-2508616ed40f)

I am using Manjaro, KDE, and Wayland. This same behavior happens when using the AUR package, as well as the AppImage from [the release](https://github.com/hoppscotch/releases/releases/download/v25.2.2-0/Hoppscotch_linux_x64.AppImage).

Downgrading to **2025.2.1** resolved the issue, but setting the variable recommended in #3516 did not.

### Steps to reproduce

1. In a Linux environment using Wayland, download [The AppImage file](https://github.com/hoppscotch/releases/releases/download/v25.2.2-0/Hoppscotch_linux_x64.AppImage)
2. Make the AppImage executable with `chmod +x /path/to/file.AppImage`
3. Execute the AppImage

### Logs and Screenshots

```shell
## System Information
OS: Manjaro Linux (64-bit)

Desktop Environment: KDE Plasma 6.2.4

Kernel: `6.6.65-1-MANJARO (64-bit)`

`echo $WAYLAND_DISPLAY` = `wayland-0`

`inxi -Gxx`:


Graphics:
  Device-1: Intel Meteor Lake-P [Intel Arc Graphics] vendor: Dell driver: i915
[...]

Issue #3543, opened 2023-11-13
[bug]: Cannot Open AppImage File in Linux


erbaafidotama:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When i open Hoppscotch AppImage file,  just showing white blank. Using Manjaro I3 and i tried in Fedora 39 Gnome

### Steps to reproduce

1. Download file AppImage (Hoppscotch_linux_x64.AppImage)
2. Right click - click properties
3. In tab permissions check "Alloq this file to run as a program"
4. Open the AppImage

### Environment

Production

### Version

Local

chriswk:
Probably the same issue encountered with downloading the deb, related to #3542, I'm assuming both Manjaro and Fedora 39 have also made the security upgrade to openssl 3.

RiemaruKarurosu:
Same problem here, nobara 38 (fedora based distro)

rondymesquita:
I can reproduce the issue on opensuse leap-15.5.

```bash
❯ ./Hoppscotch_linux_x64.AppImage 
hoppscotch: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by hoppscotch)
hoppscotch: /lib64/libc.so.6: version `GLIBC_2.33' not found (required by hoppscotch)
hoppscotch: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by hoppscotch)
hoppscotch: /lib64/libm.so.6: version `GLIBC_2.35' not found (required by /tmp/.mount_Hoppscc5YDub/usr/lib/libwebkit2gtk-4.0.so.37)
hoppscotch: /lib64/libc.so.6: version `GLIBC_2.32' not found (required by /tmp/.mount_Hoppscc5YDub/usr/lib/libwebkit2gtk-4.0.so.37)
hoppscotch: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by /tmp/.mount_Hoppscc5YDub/usr/lib/libwebkit2gtk-4.0.so.37)
hoppscotch: /lib64/libc.so.6: version `GLIBC_2.33' not found (required by /tmp/.mount_Hoppscc5YDub/usr/lib/libwebkit2gtk-4.0.so.37)
hoppscotch: /lib64/libc.so.6: version `GLIBC_2.33' not found (required by /tmp/.mount_Hoppscc5YDub/usr/lib/libgtk-3.so.0)
[...]

PR #5850, opened 2026-02-09
Fix Linux desktop failing to start on Ubuntu 22.04 (GLIBC mismatch)

harihran-del:
The Linux desktop binary is currently built on ubuntu-24.04, which links
against GLIBC 2.38/2.39. Ubuntu 22.04 LTS ships with GLIBC 2.35, causing
the application to fail at startup with:

GLIBC_2.38 not found
GLIBC_2.39 not found

Linux binaries should be built on the oldest supported distribution to ensure
forward compatibility across newer distros.

This PR changes the build environment to ubuntu-22.04 so the produced binary
works on Ubuntu 22.04 and newer Linux distributions.

No runtime code changes were made — only CI build environment adjustment.

---
## Summary by cubic
Switches the Linux desktop CI build from ubuntu-24.04 to ubuntu-22.04 so the binary links against GLIBC 2.35 and starts correctly on Ubuntu 22.04 and newer. Also removes version pinning for WebKitGTK packages to match 22.04 repos; no runtime code changes.

- **Bug Fixes**
  - Build runs on ubuntu-22.04 and drops WebKitGTK version pinning for 22.04 compatibility.
  - Resolves “GLIBC_2.38/2.39 not found” startup errors on Ubuntu 22.04.

<sup>Written for commit 3fb83861c9c224f05a5b5b81e5f7421c760afa31. Summary will update on new commits.</sup>

### DIFF
--- a/.github/workflows/build-hoppscotch-desktop.yml
+++ b/.github/workflows/build-hoppscotch-desktop.yml
@@ -92,12 +92,12 @@ jobs:
             librsvg2-dev;
 
           sudo apt install -y \
-            libwebkit2gtk-4.1-0=2.44.0-2 \
-            libwebkit2gtk-4.1-dev=2.44.0-2 \
-            libjavascriptcoregtk-4.1-0=2.44.0-2 \
-            libjavascriptcoregtk-4.1-dev=2.44.0-2 \
-            gir1.2-javascriptcoregtk-4.1=2.44.0-2 \
-            gir1.2-webkit2-4.1=2.44.0-2;
+             libwebkit2gtk-4.1-0 \
+             libwebkit2gtk-4.1-dev \
+             libjavascriptcoregtk-4.1-0 \
+             libjavascriptcoregtk-4.1-dev \
+             gir1.2-javascriptcoregtk-4.1 \
+             gir1.2-webkit2-4.1;
       - uses: actions/cache@v4
         with:
           path: |


harihran-del:
Good catch — those WebKitGTK versions are not available on Ubuntu 22.04.

I removed strict version pinning so the workflow installs the correct
packages from Jammy repositories while still keeping compatibility.

This keeps the GLIBC fix and ensures the build step won't fail due to
missing package versions.

CuriousCorrelation:
Hi @harihran-del, thanks for the PR but at the moment this doesn't solve the underlying issue discussed in this thread https://github.com/hoppscotch/hoppscotch/issues/3543#issuecomment-2869628299. 

Plus I am a bit unclear on what this accomplishes exactly since a builder downgrade is a breaking process, specifically for AppImage users https://github.com/hoppscotch/hoppscotch/issues/4880.

If you have better workarounds that can be perhaps merged upstream first, that could be a perfect fix for this, since right now just downgrading will not work. There's also I believe a news about [WebKit technically dropping support for Ubuntu 22.04](https://docs.webkit.org/Ports/WebKitGTK%20and%20WPE%20WebKit/DependenciesPolicy.html).

harihran-del:
Thanks for the clarification!

You're right — changing the runner OS only shifts the compatibility window and doesn't truly solve the portability issue long-term, especially with WebKitGTK and AppImage expectations.

My intention was to restore compatibility for Ubuntu 22.04 users in the short term, but a more correct solution would be to build the Linux artifacts inside a controlled container/runtime so the produced binaries don't depend on the GitHub runner distribution.

I can explore using a containerized build (for example a Docker image matching the expected runtime or the recommended Tauri/AppImage build environment) so the generated binaries remain portable without downgrading the host runner.

Would you prefer:
• a containerized Linux build job in CI, or
• adopting an upstream recommended AppImage/Tauri build environment?

Happy to adjust the approach based on what aligns with the project's release strategy.

CuriousCorrelation:
Hi @harihran-del, thanks for the interest but I'm going to close this PR.

The GLIBC compatibility issue has context documented in #3543 (comment) that this change doesn't account for, including WebKitGTK version pinning requirements and AppImage compatibility constraints. Removing version pinning would reintroduce Wayland/EGL regressions (#4880).

This is an upstream issue in Tauri's dependency on libwebkit2gtk-4.1 and we're tracking it internally. If anyone finds an approach that addresses the constraints documented in #3543, PRs are welcome.

For future contributions, please engage with existing discussion on an issue before opening a PR, particularly when maintainers have already posted technical analysis.

harihran-del:
Ok 👍🏻

PR #5853, opened 2026-02-09
fix: auto-store Set-Cookie headers and fix cookie retrieval

### LINKED ISSUE #3532: [bug]: multiple set-cookie did not get fold correctly and not sending as well

FerreiraAdrien:
Closes #3532

When an API response includes Set-Cookie headers, Hoppscotch does not automatically store those cookies in the cookie jar or send them with subsequent requests. Users have to manually add cookies for each domain, which breaks common authentication flows relying on httpOnly cookies.

### What's changed

[x] Added applySetCookieHeaders method to CookieJarService that parses raw Set-Cookie header strings and upserts them into the cookie jar (merging by cookie name per domain)
[x] Hooked into the Agent interceptor response flow to automatically capture Set-Cookie headers and persist them via the new method
[x] Hooked into the Native interceptor response flow with the same auto-capture logic, handling both the structured response.cookies array and the raw header fallback
[x] Fixed getCookiesForURL to return stored Cookie objects directly instead of re-parsing cookieString.value, and fixed the expiry comparison to handle ISO string dates

### Notes to reviewers

- The Set-Cookie handling follows standard domain/path defaulting (falls back to the request hostname and / when not specified by the server)
- Cookie upsert is done by name within a domain — if a response updates an existing cookie, it is replaced rather than duplicated


CuriousCorrelation:
Hey @FerreiraAdrien, thanks for putting this together.

The interceptor and relay codepaths that this PR touches are currently going through some changes on our end (deprecation of the legacy interceptor system, kernel-level relay rework, and some related cookie/auth flow consolidation). Because of that, it's a bit tricky for us to test this reliably across the different deployment targets we support right now.

This will be considered for an upcoming release once the dust settles on those changes and we can give it a proper review and QA pass. Appreciate your patience and the effort here.

jamesgeorge007:
Closing in favour of #6416, which went with a different approach. Ships in the upcoming release; thanks for the contribution and the work that went into this one.

PR #5897, opened 2026-02-22
fix(backend): resolve security advisories for IDOR and onboarding bypass

mirarifhasan:
Closes BE-717 BE-718

Fixes GHSA-jwv8-867r-q9fg
Fixes GHSA-72rv-vc3j-5vqr

### What's changed

- This PR addresses multiple security advisories by enforcing proper authorization checks.

  1. https://github.com/hoppscotch/hoppscotch/security/advisories/GHSA-jwv8-867r-q9fg
  2. https://github.com/hoppscotch/hoppscotch/security/advisories/GHSA-72rv-vc3j-5vqr

  The `canReRunOnboarding` flag being tied to `usersCount === 0` means the endpoint is only accessible when no users exist, so both the initial setup window and the re-run scenario are covered by the same guarantee. There's nothing meaningful to compromise in either case, and once the first user signs up it locks permanently.

  On the other points from the report — the missing JWT/admin guard is intentional, since auth providers don't exist during initial setup and requiring auth to configure auth isn't feasible. The ask around guarding all `/v1/onboarding/*` endpoints also doesn't apply — `GET /status` has to stay public for the frontend to know whether to show the onboarding screen, and `GET /config` is already token-gated.

- Additionally, `getOnboardingStatus()` previously always returned a success value, making the downstream `E.isLeft` check dead code. Wrapped in a `try-catch` so fetch failures now return a proper error and surface as HTTP 422.

### Notes to reviewers

Nil

---
## Summary by cubic
Fixes two security issues: blocks unauthorized onboarding re-runs and removes an IDOR in user environment mutations. Addresses GHSA-jwv8-867r-q9fg, GHSA-72rv-vc3j-5vqr, GHSA-57c6-hq5p-rgjq; completes BE-717 and BE-718.

- **Bug Fixes**
  - Onboarding: blocks re-run after completion; 400 when blocked, 422 on status fetch failure; uses ONBOARDING_CANNOT_BE_RERUN.
  - User environments: scopes update/delete/clear to the authenticated user; resolver passes user to service; tests updated; adds missing JSDoc param.

<sup>Written for commit 843da99fbce13e11ebadc661fa5adf25ac6f686e. Summary will update on new commits.</sup>

### DIFF
--- a/packages/hoppscotch-backend/src/infra-config/onboarding.controller.ts
+++ b/packages/hoppscotch-backend/src/infra-config/onboarding.controller.ts
@@ -11,6 +11,7 @@ import { InfraConfigService } from './infra-config.service';
 import { RESTError } from 'src/types/RESTError';
 import { throwHTTPErr } from 'src/utils';
 import * as E from 'fp-ts/Either';
+import { ONBOARDING_CANNOT_BE_RERUN } from 'src/errors';
 import {
   GetOnboardingConfigResponse,
   GetOnboardingStatusResponse,
@@ -74,7 +75,7 @@ export class OnboardingController {
       !onboardingStatus.right.canReRunOnboarding
     )
       throwHTTPErr(<RESTError>{
-        message: 'Onboarding cannot be re-run',
+        message: ONBOARDING_CANNOT_BE_RERUN,
         statusCode: HttpStatus.BAD_REQUEST,
       });
 
--- a/packages/hoppscotch-backend/src/user-environment/user-environments.service.ts
+++ b/packages/hoppscotch-backend/src/user-environment/user-environments.service.ts
@@ -129,6 +129,7 @@ export class UserEnvironmentsService {
    * @param id environment id
    * @param name environments name
    * @param variables environment variables
+   * @param user User object for authorization
    * @returns an Either of `UserEnvironment` or error
    */
   async updateUserEnvironment(
--- a/packages/hoppscotch-backend/src/errors.ts
+++ b/packages/hoppscotch-backend/src/errors.ts
@@ -791,6 +791,13 @@ export const INFRA_CONFIG_FETCH_FAILED =
   'infra_config/fetch_failed' as const;
 
 /**
+ * Onboarding has already been completed and cannot be re-run
+ * (OnboardingController)
+ */
+export const ONBOARDING_CANNOT_BE_RERUN =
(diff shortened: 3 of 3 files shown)


packages/hoppscotch-backend/src/infra-config/onboarding.controller.ts
```diff
@@ -60,6 +60,20 @@ export class OnboardingController {
...
+    if (
+      E.isLeft(onboardingStatus) ||
+      (onboardingStatus.right.onboardingCompleted &&
+        !onboardingStatus.right.canReRunOnboarding)
+    ) {
+      throwHTTPErr(<RESTError>{
+        message: 'Onboarding cannot be re-run',
+        statusCode: HttpStatus.BAD_REQUEST,
+      });
+    }
```
jamesgeorge007: Minor nit. The `E.isLeft` and re-run checks are combined, so a `getOnboardingStatus()` failure surfaces `Onboarding cannot be re-run` instead of the actual error. Worth splitting to match how the `GET /status` handler above handles it.

https://github.com/hoppscotch/hoppscotch/blob/ff906b7c969c84dd42f913fed594fbd96d16463d/packages/hoppscotch-backend/src/infra-config/onboarding.controller.ts#L35-L42

Not a blocker though.

jamesgeorge007: Addressed in aaa52edb642835c9a8a2c7b643906cb78f0785f8. On closer inspection, `getOnboardingStatus()` was indeed always returning `E.right`, making the `E.isLeft` check dead code. Fixed by wrapping the method body in a `try-catch` that returns `E.left(INFRA_CONFIG_FETCH_FAILED)` on failure, so the defensive check is now meaningful.
Also, split the two separate conditions (fetch error vs. re-run guard) into distinct blocks to surface each correctly.

packages/hoppscotch-backend/src/user-environment/user-environments.service.ts
```diff
@@ -130,12 +131,17 @@ export class UserEnvironmentsService {
    * @param variables environment variables
```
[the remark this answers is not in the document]
jamesgeorge007: Addressed in 843da99fbce13e11ebadc661fa5adf25ac6f686e.

PR #5913, opened 2026-02-27
fix(backend): prevent request payload from overriding id and name

mirarifhasan:
Closes BE-729.

### What's changed
Reorder the spread in the request mapping so `id` and `name` are assigned after spreading `x.request`, preventing the stored request JSON payload from overriding the authoritative DB row values and applied to both export paths in `exportUserCollectionToJSONObject` and `exportUserCollectionsToJSON`.

Also fixes constructor brace formatting and adds regression tests.

### Notes to reviewers
`exportUserCollectionToJSONObject` is called on every personal workspace page load and not only on explicit collection exports. Users who duplicated requests before the #5781 fix (v2026.1.1) have rows where the stored `request` JSON contains the original request's `id`. Without this fix, the wrong `id` is served to the frontend on reload, causing delete/rename/edit operations to target the wrong row. This resolves the stuck-duplicate-after-upgrade regression reported post `v2026.1.1`.

---
## Summary by cubic
Fixes request mapping so a stored request payload can’t override id and name, ensuring correct identifiers in collection exports. Addresses BE-729.

- **Bug Fixes**
  - Spread cast request before setting id/name in both mappings; add tests for id/name precedence and USER_COLL_NOT_FOUND on invalid collectionID.
  - Cast Prisma.JSONValue to Record<string, unknown> to safely spread.

<sup>Written for commit 18b05213e3898ffd4853fb5469cdefcb2ff6696d. Summary will update on new commits.</sup>

### DIFF
--- a/packages/hoppscotch-backend/src/user-collection/user-collection.service.spec.ts
+++ b/packages/hoppscotch-backend/src/user-collection/user-collection.service.spec.ts
@@ -2314,3 +2314,69 @@ describe('updateUserCollection', () => {
     );
   });
 });
+
+describe('exportUserCollectionToJSONObject', () => {
+  test('should use DB row id and title over conflicting values in stored request payload', async () => {
+    const dbRowId = 'db-row-cuid-001';
+    const dbRowTitle = 'My Request';
+    const payloadId = 'stale-payload-id-from-original';
+    const payloadName = 'stale-payload-name-from-original';
+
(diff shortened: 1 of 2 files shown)


nahidhasan94 (approved):
LG

nahidhasan94 (dismissed):

jamesgeorge007 (approved):

PR #5914, opened 2026-02-28
fix(curl): handle output-control flags in curl import

### LINKED ISSUE #5910: [bug]: Unable to import cURL with -sS flags

### Platform

Web App

### Browser

Chrome

### Operating System

macOS

### Bug Description

When importing cURL commands that contain output control flags (such as `-s`, `-S`, `-v`, `-sS`, etc.), the import fails with "cURL is not formatted properly" error. These flags control cURL's output behavior but don't affect the actual HTTP request.

### Expected Behavior
The cURL command should be successfully imported.
The `-sS` flags should be automatically stripped during parsing since they only control cURL's output behavior.

### Actual Behavior
Import fails with error: **"cURL is not formatted properly"**

## Steps to reproduce
1. Open Hoppscotch (web or desktop app)
2. Click the dropdown next to the "Send" button → Select "Import cURL"
3. Paste the following cURL command that contains -sS flag
4. Click "Import"
5. **Error appears**: "cURL is not formatted properly"

### Current Workaround
Manually remove the `-sS` flags before importing.

## Additional context
**Interceptor used**: N/A (issue occurs during import, before any request is sent)

### Deployment Type

Hoppscotch Cloud

### Version

2026.2.0

A386official:
Fixes #5910

## What's going on

When you paste a curl command with flags like `-sS`, `-v`, or `-k`, the import fails with "cURL is not formatted properly". This happens because `yargs-parser` doesn't know these are boolean flags, so it treats the next argument as their value. If that next argument is the URL, it gets swallowed and the parser can't find it.

For example, `curl -sS https://example.com` gets parsed as:
```json
{ "_": ["curl"], "s": true, "S": "https://example.com" }
```

The URL ends up as the value of `-S` instead of in the positional args where `getURLObject` looks for it.

## Fix

Pass a `boolean` config to `yargs-parser` listing the curl flags that don't take arguments:

`-s`, `-S`, `-v`, `-k`, `-L`, `-N`, `-f`, `-q`, `-n`, `-#`, `-0`, `-4`, `-6`

These are all output/behavior flags that don't affect the HTTP request itself. After the fix:
```json
{ "_": ["curl", "https://example.com"], "s": true, "S": true }
```

## Tests

Added 3 test cases:
- `curl -sS <url>` (the exact case from the issue)
- `curl -sSL -k -H '...' <url>` (combined flags with headers)
- `curl -v <url>` (verbose flag)

All 32 tests pass (29 existing + 3 new).

---
## Summary by cubic
Fixes curl import failures by treating non-argument flags (short and long) as booleans so the URL isn’t swallowed. This resolves “cURL is not formatted properly” for commands using combined flags like -sS, -v, -k, -G, -I, and their long forms.

- **Bug Fixes**
  - Configure yargs-parser boolean flags: short (-s, -S, -v, -k, -L, -I, -G, -N, -f, -q, -n, -#, -0, -4, -6) and long (--silent, --show-error, --verbose, --insecure, --location, --head, --get, --no-buffer, --fail, --disable, --netrc, --progress-bar, --ipv4, --ipv6, --compressed).
  - Keeps the URL in positional args, enabling imports with combined flags.
  - Adds tests for -sS <url>, -sSL -k with headers, -v <url>, -G <url>, and --silent --verbose <url>.

<sup>Written for commit f4c54eeffb4f8d980789b2254feb8b7b7aa6da07. Summary will update on new commits.</sup>

### DIFF
--- a/packages/hoppscotch-common/src/helpers/curl/__tests__/curlparser.spec.js
+++ b/packages/hoppscotch-common/src/helpers/curl/__tests__/curlparser.spec.js
@@ -1092,6 +1092,44 @@ data2: {"type":"test2","typeId":"123"}`,
       responses: {},
     }),
   },
+  {
+    command: `curl -G https://example.com`,
+    response: makeRESTRequest({
+      method: "GET",
+      name: "Untitled",
+      endpoint: "https://example.com/",
+      auth: { authType: "inherit", authActive: true },
+      body: {
+        contentType: null,
+        body: null,
+      },
+      headers: [],
+      params: [],
+      preRequestScript: "",
+      testScript: "",
+      requestVariables: [],
+      responses: {},
+    }),
+  },
+  {
+    command: `curl --silent --verbose https://example.com`,
+    response: makeRESTRequest({
+      method: "GET",
+      name: "Untitled",
+      endpoint: "https://example.com/",
+      auth: { authType: "inherit", authActive: true },
+      body: {
+        contentType: null,
+        body: null,
+      },
+      headers: [],
+      params: [],
+      preRequestScript: "",
+      testScript: "",
+      requestVariables: [],
+      responses: {},
+    }),
+  },
 ]
 
 describe("Parse curl command to Hopp REST Request", () => {
--- a/packages/hoppscotch-common/src/helpers/curl/curlparser.ts
+++ b/packages/hoppscotch-common/src/helpers/curl/curlparser.ts
@@ -52,6 +52,8 @@ export const parseCurlCommand = (curlCommand: string) => {
       "v", // --verbose
       "k", // --insecure
       "L", // --location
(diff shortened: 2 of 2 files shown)


A386official:
Both issues from the cubic review have been addressed in f4c54ee:

- **P1**: Added `-G` (--get) and `-I` (--head) to the yargs-parser boolean config. These are now correctly treated as boolean flags, preventing URL consumption.
- **P2**: Added all long-form equivalents (--silent, --verbose, --location, --compressed, etc.) to the boolean array, ensuring both short and long flag forms are handled correctly.

Also added 5 test cases covering short flags, combined flags, -G specifically, and long-form flags. Ready for maintainer review.

A386official:
Closing in favor of #5911 which addresses the same issue.

PR #5917, opened 2026-02-28
fix(common): handle File objects in HAR postData text resolution

### LINKED ISSUE #5808: [bug]: error in generated code when there is an attachment in the body

### Platform

Desktop App

### Browser

Chrome

### Operating System

Windows

### Bug Description

hey fellas,

there is always a message 'Something went wrong' in the generated code within the hoppscotch desktop app.
anytime you attach a file to the body (content-type: application/octet-stream), the code is not generated.
(clearly visible in the attach screen shot).

expected behavior:
either generate the code or put there some reasonable message why it can't be generated.. :)

Image

### Deployment Type

Hoppscotch Cloud

### Version

2025.12.1

sh3xu:
Closes #5808

Code snippet generation (cURL, etc.) crashed with a "Something went wrong" toast when the request body was `application/octet-stream`. The `File` object was being cast to string, producing `"[object File]"` which HTTPSnippet couldn't parse.

### What's changed

- Handle `application/octet-stream` in `buildHarPostData` — return `text: ""` when no file is attached, and `@<path|name>` for file uploads
- Use `"path" in file` check for desktop-specific file path, falling back to `file.name`

### Notes to reviewers

The `application/octet-stream` body is `File | null` per the Zod schema in `body.ts`. The null case (no file selected) now produces an empty snippet instead of a misleading `@<binary-file>` placeholder. The `path` property is Electron/Node-specific and not part of the standard DOM File API.

### DIFF
--- a/packages/hoppscotch-common/src/helpers/new-codegen/har.ts
+++ b/packages/hoppscotch-common/src/helpers/new-codegen/har.ts
@@ -119,26 +119,26 @@ const buildHarPostData = (req: HoppRESTRequest): Har.PostData | undefined => {
   // "[object File]" which causes HTTPSnippet to throw and code generation to
   // fail entirely.
   if (req.body.contentType === "application/octet-stream") {
-  const body = req.body.body
+    const body = req.body.body
 
-  // Preserve string if it ever exists
-  if (typeof body === "string") {
+    // Preserve string if it ever exists
+    if (typeof body === "string") {
+      return {
+        mimeType: req.body.contentType,
+        text: body,
+      }
+    }
+
+    const pathOrName =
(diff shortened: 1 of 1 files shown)


anwarulislam (dismissed):
@sh3xu Thanks for the PR. LGTM 🚀

jamesgeorge007 (approved):

jamesgeorge007:
Thanks for the PR, @sh3xu; pushed a few [cleanup changes](https://github.com/hoppscotch/hoppscotch/pull/5917/commits/cf2b11cb1d1db667ef8b7d1d9db17a6c0aa8c80c). This is scheduled for the upcoming release.

PR #5918, opened 2026-02-28
feat: auto-save requests when enabled in settings (#3437)

### LINKED ISSUE #3437: [feature]: Auto-save requests

### Summary

When editing request in tabs it worth to make them auto-saved, at least as an option

Autosave is a standard option for a wide range of IDEs and even a default option

GaneshAdimalupu:
**Closes #3437**

Adds an optional auto-save feature for request tabs: when "Auto-save requests" is enabled in Settings → General, requests that are already saved to a collection (or team) are saved automatically after a configurable delay. Auto-save is silent (no toast); manual Save still shows "Request saved". Applies to REST and GraphQL, user and team collections.

### What's changed
- [x] Settings: added `AUTO_SAVE_REQUESTS` and `AUTO_SAVE_DELAY_MS` (store, persistence schema, Settings UI, i18n in en.json)
- [x] REST: debounced auto-save in `Request.vue` with silent toast, save-in-flight guard, and team save error handling
- [x] GraphQL: same pattern in `RequestOptions.vue` plus team-collection save path (was missing)
- [x] Only runs when tab has saveContext; no modal from auto-save; manual Save unchanged

### Notes to reviewers
@liyasthomas 
- Default is off; delay is 500–10000 ms (default 2000). Manual QA: enable in Settings, edit a saved request, wait for delay, confirm dot clears and no toast.

image

---
## Summary by cubic
Adds optional per-tab auto-save for saved REST and GraphQL requests. When enabled, edits save silently after you stop typing, with a configurable 500–10000 ms delay.

- New Features
  - Settings: `AUTO_SAVE_REQUESTS` toggle and `AUTO_SAVE_DELAY_MS` (500–10000 ms, default 2000) with clamped input and persisted validation.
  - Auto-save: Centralized `AutoSaveService` with debounced per-tab watchers that survive tab switches. Silent saves for REST & GraphQL; Manual Save unchanged. Supports user and team collections.

- Bug Fixes
  - Concurrency & stability: Per-tab save lock and per-tab manual-save polling; `saveInProgress` set before auth; `saveContext` re-verified after auth; request snapshotting with `isDirty` cleared only if unchanged.
  - Retry & debounce: Exponential backoff retries (2s → 4s → 8s, max 3) only for transient network errors and only when the request snapshot is unchanged; follow-up silent save if edits changed mid-flight; timers canceled on edits, unmount, or tab close using the tab’s ID.
  - Scoping: REST manual saves only when the invoking tab is active; result view scoped to the tab’s ID. GraphQL saves scoped to the component’s tab via `getTabRef(props.tabId)`; error modals/toasts gated behind silent saves; no analytics or toasts for auto-saves.

<sup>Written for commit f86ccd0236baade46100482cf170bb332766f9a0. Summary will update on new commits. <a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/5918?utm_source=github">Review in cubic</a></sup>


GaneshAdimalupu:
@nivedin @jamesgeorge007 @liyasthomas  any reviews ?

GaneshAdimalupu:
@nivedin @jamesgeorge007  made the changes based on the  bot review, can you guys have a look?

GaneshAdimalupu:
Thanks for all the automated reviews on this PR.
@nivedin @jamesgeorge007 
I’ve iterated through the bot feedback across the last dozens of commits:
- Timers (auto‑save, retries, manual‑save polling) are now tracked and cleared on unmount.
- Deferred saves check that the tab still exists before calling `saveRequest`.
- Team saves use request snapshots to avoid stale `isDirty` clears and out‑of‑order overwrites.
- Silent auto‑saves no longer show toasts or modals, and manual saves are queued behind in‑flight saves.

At this point the remaining bot comments are about very narrow retry/backoff semantics in extreme edge cases; the current behavior is stable and consistent between REST (`Request.vue`) and GraphQL (`RequestOptions.vue`).

Could a human maintainer please take a look and let me know if there are any *concrete, blocking* issues you’d like changed before merging, or if this is good enough to go in?

packages/hoppscotch-common/src/components/http/Request.vue
```diff
@@ -535,112 +523,233 @@ const fetchingShareLink = ref(false)
...
+      return
+    }
+
+    saveRequest({ silent: true })
+  }, delay)
+}
+
+const saveRequest = async (options?: { silent?: boolean }) => {
+  const silent = options?.silent === true
+
```
nivedin: Since the auto-save logic is the same for both REST and GQL tabs, let’s move it to a service-based architecture. This will make it easier to manage and maintain, while also giving us a single source of truth.

packages/hoppscotch-common/src/components/graphql/RequestOptions.vue
```diff
@@ -224,29 +229,222 @@ watch(
...
+  const silent = options?.silent ?? false
+
+  // Block ALL concurrent saves — both silent and manual — while a mutation
+  // is in-flight. This prevents out-of-order updates where a manual save
+  // races an auto-save and overwrites newer content with an older snapshot.
+  if (saveInProgress.value) {
+    if (!silent) {
+      // For a manual save blocked by an in-flight mutation, poll until
+      // the in-flight save completes then re-invoke so the user's intent
+      // is never silently dropped. We track the timer so repeated button
```
nivedin: let’s move it to a service-based architecture

nivedin (changes requested):

GaneshAdimalupu:
@nivedin I have moved the auto-save logic to a centralized `AutoSaveService` as requested, providing a single source of truth for both REST and GQL tabs. I've also addressed all the automated bot feedback regarding edge cases and concurrency locks. The latest commit has all these changes.

nivedin:
@GaneshAdimalupu Thank you for the changes, can you look into the AI reviews and close which have been fixed.

GaneshAdimalupu:
@nivedin done 
everything is resolved with the final commit

GaneshAdimalupu:
@cubic-dev-ai review

packages/hoppscotch-common/src/services/auto-save.service.ts
```diff
@@ -0,0 +1,338 @@
...
+    return this.gqlTabs
+  }
+
+  public saveRequest = async (
+    tab: HoppTab<TabDocument>,
+    options?: { silent?: boolean; onTriggerModal?: () => void }
+  ) => {
+    const silent = options?.silent ?? false
+    const t = getI18n()
+    const toast = useToast()
```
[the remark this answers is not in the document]
GaneshAdimalupu: // getI18n() bypasses the Vue setup-context requirement (see modules/i18n)
    const t = getI18n()
    // useToast() is safe outside setup: it returns a module-level singleton
    // created at plugin install time, not a composable that relies on inject()
    const toast = useToast()
    
    
useToast() → useToasted() is implemented as () => Ze, where Ze is a module-scoped variable initialized once when the @hoppscotch/vue-toasted plugin is installed via app.use(). No inject(), no getCurrentInstance() — just a plain closure.

packages/hoppscotch-common/src/services/auto-save.service.ts
nivedin, line 1: The service does the right things in the right shape but a few correctness gaps need closing.

**Lifecycle — watchers must outlive components.** Today the per-tab watcher is set up in component `onMounted` and stopped in `onUnmounted`, which means a tab switch drops a pending save. Move the entire watcher into this service, set it up inside `onServiceInit`, and wrap the setup in `effectScope(true)` so dioc's component-scoped binding doesn't dispose them:

```ts
import { effectScope, watch } from "vue"

private scope = effectScope(true) // detached — survives component unmounts

override onServiceInit() {
  this.registerAdapters()
  this.scope.run(() => {
    this.observeTabs("rest")
    this.observeTabs("gql")
  })
}
```

When the observer callback fires later (async, outside the original scope.run), call this.scope.run(() => watch(...)) again for per-tab watchers — without that, the inner watch is captured by whatever scope is active at callback time.

packages/hoppscotch-common/src/components/graphql/RequestOptions.vue
nivedin, line 1: **`toast` is undefined at runtime.**

The diff at the top of this file removes `import { useToast } from "@composables/toast"` and `const toast = useToast()` but leaves these calls in place:

- L165: `toast.success(\`${t("state.finished_in", { duration })}\`)`
- L167: `toast.success(t("authorization.graphql_headers"))`

Please re-add:
```ts
import { useToast } from "@composables/toast"
// ...
const toast = useToast()

nivedin (changes requested):
Thanks for the work on this, the feature surface is right (settings, debounce, retries, REST + GQL) but there are a few structural and correctness issues we should clear before merging

- **CI will fail**: `packages/hoppscotch-js-sandbox/package.json` adds `@types/acorn` but `pnpm-lock.yaml` is not updated. `pnpm install --frozen-lockfile` (CI) will reject this. The dep is also unrelated to autosave and `acorn@8` already ships its own types — recommend reverting entirely.
- **Silent regression in `RequestOptions.vue`**: `useToast` import was removed but `toast.success(...)` is still called at lines 165 and 167 inside `runQuery()`. The calls sit inside a `try/catch`, so the `ReferenceError` is silently swallowed by `console.error(e)` — every successful GraphQL run now loses its "Finished in Xms" toast, and any future error in that block is hidden.

The autosave state currently lives inside `Request.vue` and `RequestOptions.vue` (`watchDebounced` + `onUnmounted(stopAutoSave)`). Two consequences:

1. **Tab-switch loses pending saves.** `pages/index.vue` and `pages/graphql.vue` use `HoppSmartWindows` without `:render-inactive-tabs`, so inactive tabs unmount. Edit → switch tabs before the debounce fires → the watcher is disposed, pending save is dropped silently. Combined with the team-collection backend round-trip, this is the most common silent-failure mode.
2. **Opening a new tab won't autosave.** The first `useService(AutoSaveService)` call binds the service inside that component's effect scope. When that component unmounts on tab switch, all watchers created in `onServiceInit` are disposed too. Subsequent `useService` calls return the existing instance but never re-run `onServiceInit`.

Both are fixed by moving the watcher into `AutoSaveService` and wrapping it in `effectScope(true)` so the watchers survive any component unmount. Components then have zero autosave logic — they just pass `kind: "rest" | "gql"` to `saveRequest`.

### Revert these
- `packages/hoppscotch-js-sandbox/package.json` — unrelated `@types/acorn`.
- `packages/hoppscotch-common/tsconfig.json` — `"noEmit": true`

packages/hoppscotch-common/src/services/auto-save.service.ts
GaneshAdimalupu: Reverted unrelated changes: Removed the stray @types/acorn dependency from hoppscotch-js-sandbox and the noEmit flag from the hoppscotch-common TS config. CI (specifically pnpm install --frozen-lockfile) should pass cleanly now.

Fixed useToast regression in GraphQL: Good catch. I've re-added the useToast import and declaration in graphql/RequestOptions.vue so that the success and error toasts now fire correctly without throwing silent ReferenceErrors.

packages/hoppscotch-common/src/components/graphql/RequestOptions.vue
GaneshAdimalupu: Fixed lifecycle issue (Tab switching): You were absolutely right about watchDebounced getting torn down when switching tabs due to HoppSmartWindows unmounting inactive content.
I've moved the auto-save watchers out of the components entirely and into AutoSaveService.
The service now uses effectScope(true) in onServiceInit() to manage its own independent reactive watchers.
It reactively watches getActiveTabs() to attach/detach a debounced watcher per tab. Since these watchers live in a detached scope on the singleton service, they easily survive component unmounts during tab switches.
I also used useSettingStatic to safely access the auto-save settings outside the component setup context.

nivedin:
@GaneshAdimalupu please look into the AI reviews

PR #5948, opened 2026-03-05
fix(auth): allow refresh token extraction from Authorization header f…

### LINKED ISSUE #5635: [bug]: Session expiring after some idle time

### Platform

Desktop App

### Browser

Firefox

### Operating System

Windows

### Bug Description

When the desktop app is idle for some time the session expires even when the server is configured as
```env
# Default validity is 7 days (604800000 ms) in ms
REFRESH_TOKEN_VALIDITY=604800000
# Default validity is 1 day (86400000 ms) in ms
ACCESS_TOKEN_VALIDITY=86400000
```

I'm assuming this not only has to do with being idle because the webapp also expires the session (I always have to re-login in web when desktop app asks to login) so I have to request for the mail, click the link in mail, then re-load the dekstop login page, every single day, even sometimes twice or three times on the same day.

### Deployment Type

### Version

2025.11.0

sahilkhan09k:
Closes #5635 

This PR fixes an issue where the desktop client cannot refresh authentication tokens after the access token expires. The backend currently only extracts the refresh token from cookies, while the desktop client sends it through the Authorization Bearer header. As a result, the refresh request fails with COOKIES_NOT_FOUND, forcing users to repeatedly log in.

What's changed

Updated the refresh token extraction logic in RTJwtStrategy.

Added support for extracting the refresh token from the Authorization: Bearer <token> header.

Kept the existing cookie-based extraction to ensure the web client continues to work as before.

Added a fallback mechanism so the backend now accepts refresh tokens from:

refresh_token cookie (web client)

Authorization header (desktop client)

Notes to reviewers

The desktop client already sends the refresh token using the Authorization header during refresh requests. However, the backend strategy only checked for the token in cookies, causing refresh requests to fail with COOKIES_NOT_FOUND.

This change preserves the current cookie-based flow used by the web application while adding compatibility with the desktop client by supporting header-based token extraction.

---
## Summary by cubic
Allow refresh tokens to be read from the Authorization: Bearer header, fixing desktop token refresh failures. Keeps the cookie flow for web and resolves #5635.

- **Bug Fixes**
  - RTJwtStrategy checks the refresh_token cookie first, then parses Authorization: Bearer from string or array headers; standardized ASCII error message.
  - Prevents COOKIES_NOT_FOUND on desktop and avoids forced re-logins after access token expiry.

<sup>Written for commit 3bd64b5066003ab9630b70407558dbf1c05a5868. Summary will update on new commits.</sup>

### DIFF
--- a/packages/hoppscotch-backend/src/auth/strategies/rt-jwt.strategy.ts
+++ b/packages/hoppscotch-backend/src/auth/strategies/rt-jwt.strategy.ts
@@ -34,7 +34,10 @@ export class RTJwtStrategy extends PassportStrategy(Strategy, 'jwt-refresh') {
     // 2️. Fallback to Authorization header
     const authHeader = request.headers?.authorization;
     if (authHeader?.startsWith('Bearer ')) {
-      return authHeader.split(' ')[1];
+      const token = authHeader.slice(7).trim();
+      if (token) {
+        return token;
+      }
     }
 
     console.error('Refresh token not found in cookie or Authorization header');
--- a/packages/hoppscotch-backend/src/auth/strategies/rt-jwt.strategy.ts
+++ b/packages/hoppscotch-backend/src/auth/strategies/rt-jwt.strategy.ts
@@ -31,7 +31,7 @@ export class RTJwtStrategy extends PassportStrategy(Strategy, 'jwt-refresh') {
       return RTCookie;
     }
 
-    // 2️. Fallback to Authorization header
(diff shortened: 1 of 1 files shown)


sahilkhan09k:
@jamesgeorge007 Please review the PR and let me know if its right or not

parasiteoflife:
Any update on this?

alschenk:
this bug is really annoying when using the desktop client. any update on this?

PR #5971, opened 2026-03-10
feat: add option to clear local workspace data on logout

### LINKED ISSUE #5957: [feature]: Optional clearing of local workspace data on explicit logout

### Summary

Add an optional setting to clear locally persisted workspace data on explicit logout only.

The default behavior should remain unchanged, so local data continues to be preserved unless the user enables this option.

The current behavior makes sense for offline/disconnected continuity, and this should be preserved.

However, explicit logout is a different scenario. When a user intentionally clicks "Logout", there may be an expectation that locally cached workspace data should no longer remain accessible in that browser, especially on shared or public machines.

A good compromise would be to add a user-controlled option to clear local workspace data only on explicit logout, while keeping the current behavior as the default.

This cleanup should not happen when:
- the network connection is lost
- the backend is temporarily unavailable
- session validation fails unexpectedly
- a token expires during normal use
- the app is operating offline

This would support both use cases:
- preserving offline/disconnected continuity
- improving privacy on shared devices for users who explicitly want local data removed on logout

This would also help in support scenarios on customer machines.

In practice, it is easy to forget opening Hoppscotch in a private/incognito window before starting a support session. When that happens, manually clearing browser storage afterward is inconvenient and not always practical, especially in a customer environment.

An optional setting to clear local data on explicit logout would provide a safer and more convenient fallback, while still preserving the current default behavior for offline/disconnected use.

statxc:
Closes #5957

### What's changed

This PR adds a new privacy setting that gives users the option to clear their local workspace data when they log out.

**New setting: "Clear local workspace data on logout"**

A toggle has been added under **Settings → Privacy**. When enabled, logging out will remove the following data from the browser's localStorage:

- REST and GraphQL collections
- REST and GraphQL history
- Personal and global environments
- Secret environment variables
- Open tabs (REST and GraphQL)
- Realtime session data (WebSocket, Socket.IO, SSE, MQTT)
- Local state and sort preferences

Settings and schema version are intentionally preserved so user preferences (theme, language, etc.) remain intact across sessions.

**Implementation details**

- [x] Added `CLEAR_LOCAL_DATA_ON_LOGOUT` to the settings store with a default value of `false`
- [x] Added the toggle UI component in the profile settings under a new "Privacy" section
- [x] Implemented `clearWorkspaceData()` in `PersistenceService` to handle storage removal and in-memory state reset
- [x] Introduced a `_isPersistencePaused` flag in `PersistenceService` to prevent RxJS store subscribers from re-writing cleared data back to localStorage during logout
- [x] Integrated the clear logic into `signOutUser()` for both selfhost-web and desktop auth flows
- [x] Added i18n keys for the new UI strings
- [x] Added validation schema entry for the new setting
- [x] Added unit tests for `clearWorkspaceData` (success path and partial failure handling)

### Notes to reviewers

- The toggle defaults to **off**, so there is no change in behavior for existing users.
- The `_isPersistencePaused` flag is necessary because resetting in-memory stores (e.g., `setRESTCollections([])`) triggers persistence subscribers that would immediately re-save empty state back to localStorage. Without this flag, the cleared keys get re-created within milliseconds of removal.
- The flag remains `true` after logout - the `PersistenceService` reinitializes on the next login, so there is no need to unpause it.
- This feature only controls what stays in the browser's localStorage between sessions. Server-synced data (collections, environments, settings) will restore normally on next login through the existing sync mechanism.

---
## Summary by cubic
Adds a Privacy toggle to clear local workspace data on logout (Linear #5957). When enabled, logout wipes browser-stored workspace data while preserving user settings.

- **New Features**
  - Toggle under Settings → Privacy (default off).
  - On logout, clears localStorage workspace data (collections, history, environments incl. secrets, tabs, realtime, local state/sort) while preserving preferences; cloud-synced data restores on next login.
  - Works for both web and desktop sign-out flows.
  - Added i18n, validation, and unit tests; persistence is paused during clearing to avoid re-saving, with a toast on partial failures.

<sup>Written for commit 9a60e78761e1d1da9f2d40958c8f0575e4327a39. Summary will update on new commits.</sup>


statxc:
@nahidhasan94 @jamesgeorge007 Could you please review my PR? I'd really appreciate any feedback.

statxc:
@jamesgeorge007 would you review my PR? I didn't get any msg after submitting this PR

jamesgeorge007:
Hi @statxc, apologies for the delayed response. This change touches logout behaviour and local data handling/retention, so it needs a more careful review before we decide whether this specific approach is something we want to take forward. We’ll keep this on the backlog for now and revisit it when we’re able to do a deeper pass on this area. Thanks for the contribution and patience.

statxc:
Hi, @jamesgeorge007 I hope you are doing well. I know this is not priority. But I really appreciate you take a look when you have a chance. It took a long time since I submitted this PR.

PR #5977, opened 2026-03-12
fix: prevent modal content cutoff in tauri by removing fixed viewport…

### LINKED ISSUE #5938: Modal content with h-[80vh] breaks in tauri
HoppSmartModal has max-h-[60vh] (link https://github.com/hoppscotch/ui/blob/a77a9d920a8c74088326f483e58b82dce249d82c/src/components/modal/index.vue#L46C1-L47C1)

In here, there is height defined as h-[80vh]. This works okey in chrome on linux; but tauri on linux (and maybe elsewhere) has the modal cut off as shown in the image - scrolled all the way down, but cannot see the whole modal.

https://github.com/hoppscotch/hoppscotch/blob/1119a2209cc31f207138c424ff74a0cfc5951d1b/packages/hoppscotch-common/src/components/collections/documentation/index.vue#L11

Image

VishwaVikas2006:
This fixes an issue where the documentation modal content was cut off in Tauri.

Previously the modal body used a fixed height:

<div class="w-full h-[80vh] overflow-hidden">

This could conflict with the modal container’s own height limits in Tauri, causing the bottom portion of the modal to become inaccessible even when scrolling.

The change replaces the fixed height with:

<div class="w-full h-full overflow-auto">

This allows the content to inherit the modal height and enables proper scrolling without clipping.

Fixes #5938

---
## Summary by cubic
Prevents documentation modal content from being cut off in Tauri by removing the fixed viewport height and enabling scroll. Fixes #5938.

- **Bug Fixes**
  - Documentation modal body now uses h-full overflow-auto instead of h-[80vh] overflow-hidden, allowing content to inherit modal height and scroll in Tauri.

<sup>Written for commit 694f5a5bfe0132e8c23ea1c35ebcc8732ce76ad6. Summary will update on new commits.</sup>

### DIFF
--- a/packages/hoppscotch-common/src/components/collections/documentation/index.vue
+++ b/packages/hoppscotch-common/src/components/collections/documentation/index.vue
@@ -8,7 +8,7 @@
     @close="hideModal"
   >
     <template #body>
-      <div class="w-full h-[80vh] overflow-hidden">
+      <div class="w-full h-full overflow-auto">
         <div class="flex h-full">
           <div class="flex-1 flex">
             <CollectionsDocumentationPreview


VishwaVikas2006:
Thanks for the review.
I’ve removed the unrelated changes and updated the PR so it now only contains the modal height fix for #5938.

VishwaVikas2006:
@nivedin @abraham @silentmatt @kant @jamesgeorge007 Kindly review the PR

PR #5985, opened 2026-03-13
fix(common): parse JSON message body before emitting via Socket.IO

### LINKED ISSUE #4570: [bug]: json Socket.io messages are sent as string instead of object

### Current behavior

When I have connected to any (Socket.io) server and try to send an object, the message sent includes a serialized message containing the serialized object.

For example: `420["test-message","{\n  \"key\": \"value\"\n}"]`

### Steps to reproduce

1. Connect to wss://echo-socketio.hoppscotch.io with Client v4
2. Send a message `'test-message'` with the following body:
   ```
   {
     "key": "value"
   }
   ```
Observe through the network tab of your browser that the message sent is `420["test-message","{\n \"key\": \"value\"\n}"]`.

When the same message is sent directly through a socket.io client, (using `emit('test-message', {key: "value"});`) the message sent is `42["test-message",{"key":"value"}]`
Note the omission of all white-space characters, since the socket.io default parser is used instead.

### Environment

Production

### Version

Cloud

mahmoodhamdi:
## What does this PR do?

Fixes the Socket.IO client to parse JSON message bodies into actual objects before emitting them. Previously, JSON strings typed in the message body editor were emitted as-is, causing double-serialization on the wire.

## Why?

I was testing a Socket.IO server and noticed the messages being sent had double-serialized JSON payloads. For example, sending `{"key": "value"}` through the Socket.IO tab would produce:

```
420["test-message","{\n  \"key\": \"value\"\n}"]
```

instead of the correct format:

```
42["test-message",{"key":"value"}]
```

This happens because `socket.emit()` receives a string and serializes it again. Real socket.io clients (like the npm package) accept objects directly, so `emit(event, {key: value})` produces the correct wire format.

## Changes

- In `SIOConnection.sendMessage()`, the message body is now parsed via `JSON.parse()` before being passed to `socket.emit()`
- If the message is not valid JSON (plain text), it falls back to sending the raw string as before
- This is a minimal, safe change — only the emit payload is affected

## Testing

- All existing tests pass (680 tests in hoppscotch-common)
- TypeScript type checking passes
- ESLint passes with no new warnings
- Verified behavior:
  - Sending `{"key": "value"}` now correctly emits as a JSON object
  - Sending plain text like `hello` still works as a string
  - Sending arrays `[1, 2, 3]` correctly emits as an array
  - Sending numbers `42` correctly emits as a number

## Related Issues

Closes #4570

---
## Summary by cubic
Parse JSON message bodies before emitting via Socket.IO so objects and arrays are sent as native values instead of double-serialized strings. Only parse payloads starting with { or [ to avoid coercing primitives, and log the actual emitted payload in the sent-event history; fall back to the raw string for invalid JSON.

<sup>Written for commit 3b5f71cd0a6aea786aed2e3ce97ae0b887b7a2a6. Summary will update on new commits.</sup>

### DIFF
--- a/packages/hoppscotch-common/src/helpers/realtime/SIOConnection.ts
+++ b/packages/hoppscotch-common/src/helpers/realtime/SIOConnection.ts
@@ -135,13 +135,18 @@ export class SIOConnection {
     if (this.connectionState$.value === "DISCONNECTED") return
     const { message, eventName } = event
 
-    // Parse JSON strings into objects so socket.io sends them as
-    // native JSON rather than double-serialized string payloads
+    // Parse JSON object/array strings into native values so socket.io
+    // sends them as structured JSON rather than double-serialized strings.
+    // Only attempt parsing for object/array shapes to avoid silently
+    // coercing primitives like "null", "true", or "42".
     let payload: unknown = message
-    try {
-      payload = JSON.parse(message)
-    } catch {
-      // Not valid JSON, send as plain string
+    const trimmed = message.trim()
+    if (trimmed.startsWith("{") || trimmed.startsWith("[")) {
+      try {
+        payload = JSON.parse(message)
+      } catch {
+        // Not valid JSON, send as plain string
+      }
     }
 
     this.socket?.emit(eventName, payload, (data) => {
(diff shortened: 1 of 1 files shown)


anwarulislam (approved):
LGTM 🚀

PR #6001, opened 2026-03-16
fix: increase width of data structure drawer

### LINKED ISSUE #5999: [feature]: Adjust the width of the data structure drawer

### Summary

The function of generating data structures is excellent. However, the display drawer generated is too narrow, which is not very convenient. Can we make this width wider? Or I can manually drag to adjust the width.

Path: /documentation/getting-started/rest/response-handling

The function of generating data structures is excellent. However, the display drawer generated is too narrow, which is not very convenient. Can we make this width wider? Or I can manually drag to adjust the width.
thank you!

vikaskmalipatil:
Fixes #5999

The data structure / response interface drawer was too narrow,
which made generated code harder to read.

This PR increases the max width of the drawer to improve readability
while keeping the layout behavior unchanged.

Tested locally on the development server.

---
## Summary by cubic
Widened the data structure/response interface drawer and removed inner width caps to improve generated code readability without changing layout behavior.

- **Bug Fixes**
  - Applied a scoped style on `HoppSmartSlideOver` in `ResponseInterface.vue` via `.data-schema-drawer` to widen the slide-over (36rem, capped at viewport width).
  - Added `max-w-none` to content and code containers so they use the full drawer width.

<sup>Written for commit 9761c6ba9bd1875c2fce81bad2eb4d7f953bc9db. Summary will update on new commits.</sup>

### DIFF
--- a/packages/hoppscotch-common/src/components/http/ResponseInterface.vue
+++ b/packages/hoppscotch-common/src/components/http/ResponseInterface.vue
@@ -2,7 +2,7 @@
   <HoppSmartSlideOver
     :show="show"
     :title="t('response.data_schema')"
-    class="!max-w-[1000px]"
+    class="data-schema-drawer"
     @close="close()"
   >
     <template #content>
@@ -281,3 +281,9 @@ const { downloadIcon, downloadResponse } = useDownloadResponse(
   t("filename.response_interface")
 )
 </script>
+
+<style scoped>
+.data-schema-drawer :deep(aside) {
+  @apply !w-[36rem] !max-w-[100vw];
+}
+</style>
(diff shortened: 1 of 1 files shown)


anwarulislam (commented):
@vikaskmalipatil I’ve tested this locally, and I didn’t notice any changes – the behavior remains the same. Could you please share a screenshot and your testing method so I can verify it on my end?

vikaskmalipatil:
Hi @anwarulislam,

Thanks for pointing that out!

You're right — earlier it wasn’t clearly visible. I’ve rechecked and confirmed the change locally.

🧪 Steps to verify:

1. Open the data structure viewer
2. Expand the drawer panel
3. The width is now increased (~36rem)

📸 Before vs After:

Before:
image

After:
image


Please let me know if this looks correct or if I should adjust further.

PR #6024, opened 2026-03-21
fix(openapi): map custom json content types to supported types

### LINKED ISSUE #6018: [feature]: Support custom content type in request body

### Summary

For example take this spec:

    openapi: 3.1.0
    info:
      title: Minimal API
      version: 1.0.0
      license:
        name: Copyright 2026
        url: https://github.com

    security: []

    servers:
      - url: /api/v1

    paths:
      /example:
        post:
          summary: Example POST endpoint
          operationId: createExample
          requestBody:
            required: true
            content:
              application/custom+json:
                schema:
                  type: object
                  required:
                    - name
                  properties:
                    name:
                      type: string
                    age:
                      type: integer
          responses:
            '200':
              description: OK

https://hoppscotch.io says:

> This request does not have a body

Raw Request Body would be shown only when:

    application/custom+json:

is changed:

    application/json:

Request body template should be shown for specs that use custom content type.

sivasai9849:
Closes #6018

This PR updates the OpenAPI importer to properly recognize schemas and generate request body templates for endpoints with custom JSON (e.g. `application/custom+json`, `application/vnd.api+json`) and custom XML media types.

### What's changed

- Added a [getSupportedContentType](cci:1://file:///Users/sivasai/Dev/hoppscotch/packages/hoppscotch-common/src/helpers/import-export/import/openapi/index.ts:398:0-414:1) helper in [packages/hoppscotch-common/src/helpers/import-export/import/openapi/index.ts](https://github.com/sivasai9849/hoppscotch/blob/fix/issue-6018/packages/hoppscotch-common/src/helpers/import-export/import/openapi/index.ts) to dynamically identify custom `application/*+json` and `application/*+xml` media types.
- Updated [parseOpenAPIV2Body](cci:1://file:///Users/sivasai/Dev/hoppscotch/packages/hoppscotch-common/src/helpers/import-export/import/openapi/index.ts:416:0-473:1) and [parseOpenAPIV3Body](cci:1://file:///Users/sivasai/Dev/hoppscotch/packages/hoppscotch-common/src/helpers/import-export/import/openapi/index.ts:506:0-607:1) to map these dynamically recognized formats accurately into the strict [HoppRESTReqBody](cci:2://file:///Users/sivasai/Dev/hoppscotch/packages/hoppscotch-data/src/rest/v/10/body.ts:39:0-39:61) types supported by the UI (falling back to `application/json` or `application/xml` for syntax highlighting purposes).
- The raw `Content-Type` header parsed from the OpenAPI spec remains unmodified and accurate to the original schema.

### Notes to reviewers

Previously, because custom content types were not found within the strict array of `knownContentTypes`, the importer would silently discard the entire request body object. This enhancement ensures that Hoppscotch handles real-world specs utilizing custom content-type API semantics gracefully.

---
## Summary by cubic
OpenAPI importer now recognizes custom JSON/XML media types and generates request body templates instead of dropping them. It also keeps the original `Content-Type` header when the mapped type differs.

- **Bug Fixes**
  - Added `getSupportedContentType` to normalize and map `application/*+json` → `application/json` and `application/*+xml` → `application/xml` (trim/lowercase/strip params; safe match with `Object.hasOwn`).
  - Updated OpenAPI v2/v3 body parsers to use mapped types for examples and form data.
  - When mapped type differs, add or activate the original `Content-Type` header from the spec without overwriting an active non-empty value.
  - Use `cloneDeep(body)` for `originalRequest` to avoid shared mutable state.

<sup>Written for commit bce7c546007513ecc17f64fdc449ec8e732b57aa. Summary will update on new commits.</sup>

### DIFF
--- a/packages/hoppscotch-common/src/helpers/import-export/import/openapi/index.ts
+++ b/packages/hoppscotch-common/src/helpers/import-export/import/openapi/index.ts
@@ -396,17 +396,36 @@ const parseOpenAPIHeaders = (params: OpenAPIParamsType[]): HoppRESTHeader[] =>
     )
   )
 
+const getSupportedContentType = (
+  contentType: string
+): keyof typeof knownContentTypes | null => {
+  if (contentType in knownContentTypes) {
+    return contentType as keyof typeof knownContentTypes
+  }
+
+  if (contentType.match(/application\/.*\+json/)) {
+    return "application/json"
+  }
+
+  if (contentType.match(/application\/.*\+xml/)) {
+    return "application/xml"
+  }
+
+  return null
+}
+
 const parseOpenAPIV2Body = (op: OpenAPIV2.OperationObject): HoppRESTReqBody => {
   const obj = (op.consumes ?? [])[0] as string | undefined
 
-  // Not a content-type Hoppscotch supports
-  if (!obj || !(obj in knownContentTypes))
-    return { contentType: null, body: null }
+  if (!obj) return { contentType: null, body: null }
+
+  const supportedContentType = getSupportedContentType(obj)
+  if (!supportedContentType) return { contentType: null, body: null }
 
   // For form data types, extract form fields
   if (
-    obj === "multipart/form-data" ||
-    obj === "application/x-www-form-urlencoded"
+    supportedContentType === "multipart/form-data" ||
+    supportedContentType === "application/x-www-form-urlencoded"
   ) {
     const formDataValues = pipe(
       (op.parameters ?? []) as OpenAPIV2.Parameter[],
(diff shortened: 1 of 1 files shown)


howwohmm (commented):
solid approach — the `getSupportedContentType()` function mapping vendor MIME types via structured syntax suffixes aligns with RFC 6838.

two things worth considering:

1. the regex `/^application\/.+\+json$/` only matches `application/*+json`, but RFC 6838 structured suffixes can technically appear on any top-level type (e.g. `text/vnd.example+json`). if narrowing to `application/` is intentional that's fine, just worth a comment.

2. when the original content type is preserved as a header, the logic checks `!existingContentTypeHeader.active || !existingContentTypeHeader.value` to decide whether to overwrite. but if a spec already has an active Content-Type header with a different value (e.g. `application/vnd.api+json`), the current code drops it. the spec author likely set it deliberately — might want to preserve spec-defined headers over inferred ones.

otherwise the `cloneDeep(body)` fix for the response original request is a good catch.

PR #6025, opened 2026-03-21
feat(i18n): add Hindi (hi) translation

nikhil-shukl:
Closes # 

This PR adds a complete Hindi (`hi`) language translation to Hoppscotch. Hindi is the 3rd most spoken language in the world with 600M+ speakers. Currently Hoppscotch has no Hindi translation, so this PR fills that gap for a massive developer community.

### What's changed
- [x] Added `packages/hoppscotch-common/locales/hi.json` with 500+ manually translated strings
- [x] Covers all core UI sections: Actions, Auth, Collection, Environment, Error, Request, Response, Settings, Shortcuts, Workspace, GraphQL, Navigation and more
- [x] JSON structure exactly follows existing translation files
- [x] All keys match `en.json` structure

### Notes to reviewers
- All strings are **manually translated** — not machine translated
- Tested JSON validity — no syntax errors
- Hindi (`hi`) is one of the most requested missing languages

---
## Summary by cubic
Adds full Hindi (`hi`) translation, registers it in the language list, and syncs with the latest `en.json` keys so Hindi is complete and selectable. Incorporates locale review feedback for clearer, consistent wording.

- **New Features**
  - Added `packages/hoppscotch-common/locales/hi.json` (600+ strings) and registered `hi` (`hi-IN`) in `packages/hoppscotch-common/languages.json`.
  - Synced with the latest `en.json`; filled previously missing keys (e.g., downloads).

- **Bug Fixes**
  - Cleaned up `hi.json`: removed a duplicate key, fixed formatting, and applied review feedback (terminology, punctuation).
  - Fixed indentation in `languages.json`.

<sup>Written for commit c691271c9f65249419049cd2e3e6eb0e58d6a903. Summary will update on new commits. <a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6025?utm_source=github">Review in cubic</a></sup>

PR #6026, opened 2026-03-22
fix(common): track manual disconnect flag in realtime connections

### LINKED ISSUE #5987: [bug]: Realtime connections always report disconnects as manual, even when server-initiated
### Software and environment

- Hoppscotch version: latest (main branch)
- Browser: All
- Platform: Web, Desktop

### Steps to reproduce

1. Open a Socket.IO / WebSocket / SSE connection to a server
2. Have the server close the connection (e.g., server restart, timeout, or explicit server-side disconnect)
3. Observe the disconnect event in the log

### Expected behavior

The disconnect event should indicate that it was server-initiated (`manual: false`), so the user can tell the difference between intentionally closing a connection and an unexpected server-side disconnect.

### Actual behavior

All disconnect events are logged with `manual: true`, even when the server initiated the disconnect. This makes it impossible to distinguish between user-initiated disconnects and unexpected server disconnections.

### Root cause

In `SIOConnection.ts`, the `disconnect` event handler unconditionally sets `manual: true`:

```typescript
this.socket.on("disconnect", () => {
  this.connectionState$.next("DISCONNECTED")
  this.addEvent({
    type: "DISCONNECTED",
    time: Date.now(),
    manual: true, // Always true, even for server-initiated disconnects
  })
})
```

The same pattern exists in:
- `WSConnection.ts` — `onclose` always emits `manual: true`
- `SSEConnection.ts` — `stop()` always emits `manual: true`

**MQTT does it correctly** — `MQTTConnection.ts` uses a `manualDisconnect` boolean flag that is only set to `true` when the user explicitly calls `disconnect()`, and defaults to `false` for server-initiated disconnects.

### Suggested fix

Follow the MQTT pattern: introduce a `manualDisconnect` flag in each connection class. Set it to `true` only when `disconnect()` is explicitly called by the user, and use this flag in the disconnect/close event handlers.

### Priority

Low — this is primarily a UX/debugging issue. Users cannot currently tell from the event log whether a connection drop was intentional or unexpected.

howwohmm:
Closes #5987

Socket.IO, WebSocket, and SSE connections all hardcode `manual: true` on disconnect/stop events, making it impossible for users to distinguish between intentional disconnects and unexpected server-initiated ones. MQTT already handles this correctly with a `manualDisconnect` flag.

### What's changed

- [x] Added `manualDisconnect` flag to `SIOConnection.ts` — set to `true` only when `disconnect()` is called by the user, used in the `disconnect` event handler, and reset after use
- [x] Added `manualDisconnect` flag to `WSConnection.ts` — same pattern, set in `disconnect()`, read in `onclose`, reset after use
- [x] Added `manualStop` flag to `SSEConnection.ts` — extracted shared logic into private `closeConnection()` method so `onerror` emits `manual: false` while user-called `stop()` emits `manual: true`

All three files now follow the same pattern already used by `MQTTConnection.ts` (lines 53, 135–150, 283).

### Notes to reviewers

- Zero API changes — the `manual` field type (`boolean`) in each event type is unchanged
- The SSE fix required a small refactor: `stop()` now delegates to a private `closeConnection()` so the error handler path doesn't incorrectly mark disconnects as manual

---
## Summary by cubic
Fixes manual disconnect/stop flags for `Socket.IO`, `WebSocket`, and `SSE`, and guards redundant `WebSocket` closes so clients can tell user- vs server-initiated closures and connection state doesn’t get stuck. Matches `MQTTConnection` behavior; addresses #5987 and #6025.

- **Bug Fixes**
  - SIOConnection and WSConnection: add manualDisconnect; set only in disconnect(); errors use closeConnection(); reset after DISCONNECTED and when no socket.
  - WSConnection: guard closeConnection() when the socket is CLOSING/CLOSED so the flag resets even if onclose won’t fire, preventing stale connectionState$.
  - SSEConnection: add manualStop; stop() sets it; onerror uses closeConnection() so STOPPED emits manual: false; reset after emit.

<sup>Written for commit 7d119d41db0157ce4f0e0ceecb2822f4e440213d. Summary will update on new commits.</sup>

### DIFF
--- a/packages/hoppscotch-common/src/helpers/realtime/SIOConnection.ts
+++ b/packages/hoppscotch-common/src/helpers/realtime/SIOConnection.ts
@@ -44,6 +44,8 @@ export class SIOConnection {
   connectionState$: BehaviorSubject<ConnectionState>
   event$: Subject<SIOEvent> = new Subject()
   socket: SIOClient | undefined
+  private manualDisconnect = false
+
   constructor() {
     this.connectionState$ = new BehaviorSubject<ConnectionState>("DISCONNECTED")
   }
@@ -106,8 +108,9 @@ export class SIOConnection {
         this.addEvent({
           type: "DISCONNECTED",
           time: Date.now(),
-          manual: true,
+          manual: this.manualDisconnect,
         })
+        this.manualDisconnect = false
       })
     } catch (error) {
       this.handleError(error, "CONNECTION")
@@ -120,7 +123,7 @@ export class SIOConnection {
   }
 
   private handleError(error: unknown, type: SIOErrorType) {
-    this.disconnect()
+    this.closeConnection()
     this.addEvent({
       time: Date.now(),
       type: "ERROR",
@@ -157,8 +160,18 @@ export class SIOConnection {
     })
   }
 
-  disconnect() {
-    this.socket?.close()
+  private closeConnection() {
+    if (!this.socket) {
+      this.manualDisconnect = false
+      this.connectionState$.next("DISCONNECTED")
+      return
+    }
+    this.socket.close()
     this.connectionState$.next("DISCONNECTED")
+  }
+
+  disconnect() {
+    this.manualDisconnect = true
+    this.closeConnection()
   }
 }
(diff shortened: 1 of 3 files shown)


howwohmm:
bump — I've reviewed the automated feedback on the manualDisconnect edge cases and believe the current approach handles them correctly (the flag is reset on reconnect). happy to address any concerns from the team.

PR #6027, opened 2026-03-23
test: add timeout handling case to getResourceContents in CLI

cyphercodes:
This PR addresses the "Revisit existing unit tests for the CLI" issue by adding a critical edge case for the `getResourceContents` utility. 

**Added:**
- A test case for handling `ETIMEDOUT` errors in the network layer. This ensures that the CLI tool correctly identifies and handles timeout scenarios, reporting them as `UNKNOWN_ERROR` with the appropriate error context, rather than failing unexpectedly. 

This improves the reliability of the CLI when fetching remote collections or environments.

---
## Summary by cubic
Adds a timeout (ETIMEDOUT) test for `getResourceContents` in `hoppscotch-cli`, fixes async test assertions to await rejections, and ensures unhandled Axios errors are consistently rethrown as `UNKNOWN_ERROR` with the original error after specific checks. Aligns with the “Revisit existing unit tests for the CLI” task.

<sup>Written for commit e9399e775afa76cfd1bb6021f766f7fa96f2b17e. Summary will update on new commits.</sup>

PR #6028, opened 2026-03-23
test(cli): add coverage for collection variable priority in getResolvedVariables

### LINKED ISSUE #4136: Revisit existing unit tests for the CLI
> [!NOTE]
This is intended for [FOSSHACK](https://github.com/hoppscotch/hoppscotch/discussions/4131) participants.

### Description

Various [unit tests](https://github.com/hoppscotch/hoppscotch/tree/main/packages/hoppscotch-cli/src/__tests__/functions)s exist for the CLI which are disabled at the moment since they were found flaky. They have to be revisited and necessary updates are to be performed (migrating to the Vitest set of APIs, resolving failed assertions, etc) to be made part of the CI pipeline. This will depend on #4095 which introduces unit tests for the newly added functionality alongside migrating the test runner from Jest to Vitest bringing in better ESM support.

cyphercodes:
This PR adds missing unit test coverage for collection variables in the Hoppscotch CLI's `getResolvedVariables` utility. It ensures that variables correctly follow the priority order: Request Variables > Collection Variables > Environment Variables. Fixes #4136.

---
## Summary by cubic
Adds tests for `getResolvedVariables` to enforce variable priority (Request > Collection > Environment), and hardens `getResourceContents` error handling for timeouts and unhandled `AxiosError`s. Fixes #4136.

- **Bug Fixes**
  - `getResourceContents`: keep `INVALID_SERVER_URL` for non‑JSON responses; otherwise always throw `UNKNOWN_ERROR` with the original error (e.g., `ETIMEDOUT`).
  - Tests: add collection-variable priority test; include `ETIMEDOUT` in parameterized cases; make `UNKNOWN_ERROR` assertions robust with partial matching; use async `await expect(...).rejects`.

<sup>Written for commit 28c36ca0a475cfe92b019d7b97c5377a1dc6eb63. Summary will update on new commits.</sup>

### DIFF
--- a/packages/hoppscotch-cli/src/__tests__/unit/getters.spec.ts
+++ b/packages/hoppscotch-cli/src/__tests__/unit/getters.spec.ts
@@ -275,14 +275,14 @@ describe("getters", () => {
       test("Promise rejects with the code `UNKNOWN_ERROR` while encountering an error that is not an instance of `AxiosError`", async () => {
         const expected = {
           code: "UNKNOWN_ERROR",
-          data: new Error("UNKNOWN_ERROR"),
+          data: expect.objectContaining({ message: "UNKNOWN_ERROR" }),
         };
 
         vi.spyOn(axios, "get").mockImplementation(() =>
           Promise.reject(new Error("UNKNOWN_ERROR"))
         );
 
-        await expect(getResourceContents(args)).rejects.toEqual(expected);
+        await expect(getResourceContents(args)).rejects.toMatchObject(expected);
       });
     });
 
--- a/packages/hoppscotch-cli/src/__tests__/unit/getters.spec.ts
+++ b/packages/hoppscotch-cli/src/__tests__/unit/getters.spec.ts
@@ -237,7 +237,7 @@ describe("getters", () => {
         },
       ];
 
-      test.each(cases)("$description", ({ args, axiosMock, expected }) => {
+      test.each(cases)("$description", async ({ args, axiosMock, expected }) => {
         const { code, response } = axiosMock;
         const axiosErrMessage = code ?? response?.data?.reason;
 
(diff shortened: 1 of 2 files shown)


howwohmm (commented):
nice additions — the collection variable priority test is correct (traced through getResolvedVariables: request > collection > environment), and the \`await\` additions on \`rejects\` fix a real bug in the existing tests (without await, failed assertions are fire-and-forget promises that don't fail the test).

one issue: there are two identical standalone ETIMEDOUT tests (same body, same mock, same assertion) placed outside the \`describe("Network call failure")\` block. the first one also has a broken description — missing \`\`UNKNOWN_ERROR\`\` between "code" and "if":

\`\`\`
"Promise rejects with the code  if the network call fails..."
\`\`\`

vs the second one which correctly reads:

\`\`\`
"Promise rejects with the code \`UNKNOWN_ERROR\` if the network call fails..."
\`\`\`

looks like a copy-paste — one should be removed, and the surviving one should live inside \`describe("Network call failure")\` alongside the other network error tests. the table-driven \`test.each\` already covers ETIMEDOUT too, so the standalone test is arguably redundant.

cyphercodes:
Fixed! I've removed the redundant standalone `ETIMEDOUT` test and corrected the description. I also fixed the expected order in the variable priority test to correctly reflect the `Request > Collection > Environment` precedence. Thanks for the thorough review! 🦞

jamesgeorge007 (commented):
Since this PR already includes the timeout/error-handling work from #6027 in the same files, it makes sense to consolidate review here. For any further changes in this CLI getters/test surface, please keep them on this PR rather than opening a parallel one.

cyphercodes:
Updated this PR to address the remaining CLI getters test-surface review note:

- Changed the non-`AxiosError` `UNKNOWN_ERROR` rejection assertion to use `toMatchObject` with `expect.objectContaining({ message: "UNKNOWN_ERROR" })`, avoiding a brittle raw `Error` instance equality check.

Local verification:
- `pnpm install --frozen-lockfile` ✅
- `pnpm --dir packages/hoppscotch-cli exec vitest run src/__tests__/unit/getters.spec.ts --reporter=verbose --testTimeout=10000 --hookTimeout=10000` ✅ (23 tests)
- `pnpm --filter @hoppscotch/cli do-typecheck` ✅
- `git diff --check` ✅

I kept the update on this PR per the consolidation note for CLI getters/test changes. I did not rebase because the PR is currently mergeable; the only remaining change was this focused test assertion cleanup.

PR #6030, opened 2026-03-23
feat: Support Configuration for Client Authentication in OAuth 2.0 Password Grant

### LINKED ISSUE #6029: [feature]: Support Configuration for Client Authentication in OAuth 2.0 Password Grant

### Summary

Currently, when using the OAuth 2.0 Password Grant type, it defaults to POST body for sending the client_id and client_secret. However, many Identity Providers strictly require client credentials to be sent via the Authorization Header (Basic Auth), while others only support them in the Request Body.

To improve compatibility with various OAuth 2.0 implementations, I think we should allow users to toggle between these two authentication methods.

MartinPham:
Closes #6029

### Descriptions
OAuth 2.0's Password grant now supports the same client authentication choice as client credentials: send credentials in the body (IN_BODY) or as Basic auth (AS_BASIC_AUTH_HEADERS). 

### What's changed
In the Authorization > OAuth 2.0 option, specifically when the Grant Type is set to Password Payloads, there should be a configuration option to choose how client credentials are transmitted: Send as Basic Auth header (Recommended by RFC 6749) OR Send client credentials in body

* `packages/hoppscotch-data/src/rest/v/15/auth.ts`: PasswordGrantTypeParams was extended with clientAuthentication, an optional enum AS_BASIC_AUTH_HEADERS | IN_BODY, defaulting to "IN_BODY".

* `packages/hoppscotch-common/src/types/post-request.d.ts`: added optional `clientAuthentication`

* `packages/hoppscotch-common/src/types/pre-request.d.ts`: optional clientAuthentication on the PASSWORD variant

* `packages/hoppscotch-common/src/services/oauth/flows/password.ts`: Default params include clientAuthentication: "IN_BODY".

* `packages/hoppscotch-common/src/composables/oauth2/useOAuth2GrantTypes.ts`: exposes the new option in the Authorization UI next to other password fields (after scopes, consistent with client-credentials ordering for this control).

* `packages/hoppscotch-common/src/helpers/import-export/import/openapi/index.ts`:  2 OpenAPI-derived password flows (OAuth2 flows object and legacy flow === "password") now set clientAuthentication: "IN_BODY" on the imported PASSWORD grantTypeInfo.

* `packages/hoppscotch-cli/src/utils/workspace-access.ts`: added clientAuthentication: "IN_BODY" alongside empty tokenRequestParams / refreshRequestParams.

### Notes to reviewers

Thank you for all your hard work on Hoppscotch! This is my first contribution to the project, so I apologize if I’ve missed any specific formatting rules or technical details. 

image

---
## Summary by cubic
Adds a choice for OAuth 2.0 Password grant client authentication. You can now send client ID/secret in the body or as a Basic Auth header, with “In body” as the default.

- **New Features**
  - Added “Send as” option to Password grant in Authorization > OAuth 2.0: “In body” (default) or “Basic Auth header”.
  - Request builder applies the selected mode; creates a Basic Auth header and omits client credentials from the body when chosen.
  - OpenAPI v2/v3 imports set Password flow to “IN_BODY” by default.
  - CLI workspace access sets Password flows to “IN_BODY” by default.

- **Migration**
  - No action needed. Existing configs keep “IN_BODY” and behave the same.

<sup>Written for commit f3fd863d6864ed6685cb761b3f4da164c24d12dd. Summary will update on new commits.</sup>


howwohmm (commented):
clean implementation — correctly mirrors the existing Client Credentials pattern for Password grant. two things:

1. **`encodeBasicAuthComponent` is duplicated** between `password.ts` and `clientCredentials.ts`. if the encoding logic needs to change for RFC 7617 UTF-8 handling, it needs updating in two places. worth extracting to a shared utility in the OAuth helpers.

2. **the `.stop.prevent` on Enter keydown in TagsInput (if applicable here) — actually, more importantly:** the guard in the Password callback uses `auth.value.grantTypeInfo.grantType !== "PASSWORD"` (hardcoded string), while the Client Credentials implementation uses `!("clientAuthentication" in auth.value.grantTypeInfo)` (property check). the property check is more resilient to refactoring. consider aligning them.

minor: `clientSecret` is spread conditionally via `...(clientSecret && { client_secret: clientSecret })`, which treats empty string as falsy. some OAuth servers require `client_secret` to be present even if empty. pre-existing behavior but since you're restructuring this code, good time to switch to `clientSecret !== undefined`.

no tests for the Basic Auth encoding path — given that encoding has subtle edge cases with special characters, at least one test for `encodeBasicAuthComponent` with a password containing `%` or `:` would be valuable.

Issue #4488, opened 2024-10-28
[bug]: Inconsistant tab location REST Page Tabs Above URL vs. GraphQL Tabs Below URL


barrettluke:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

On the REST page, tabs are positioned above the URL input field, while on the GraphQL page, they are positioned below the URL input field. This inconsistency may affect user experience, as it changes the visual flow and interaction between the pages.

REST page with tabs above URL 
![Screenshot 2024-10-27 at 8 50 19 PM](https://github.com/user-attachments/assets/f41f3f12-dbde-48b5-873d-9c8d35c7a269)

GraphQL page with tabs below URL
![Screenshot 2024-10-27 at 9 07 40 PM](https://github.com/user-attachments/assets/edbd9a68-15ca-427f-898b-1868453c93f8)

### Steps to reproduce

	1.	Go to the REST page.
	2.	Observe that the tabs are positioned above the URL input field.
	3.	Switch to the GraphQL page.
	4.	Observe that the tabs are positioned below the URL input field.

### Environment

Production

### Version

Cloud

nivedin:
Closing since this is now addressed in the [v2026.8.0 release](https://github.com/hoppscotch/hoppscotch/releases/tag/2026.8.0)

Issue #4383, opened 2024-09-29
[bug]: GraphQL collections are not shared in a workspace


tvvignesh:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

We created a workspace, added team members and provided them editor and admin access in hoppscotch. The issue is that while REST APIs we create are visible across members in the team, GraphQL APIs are not (irrespective of who created it)

We are unable to use hoppscotch to collaborate with GraphQL because of this. Kindly help.

### Steps to reproduce

1. Create a workspace and add team members into it
1. Create a collection, save GQL requests into the collection as team member 1
2. Team member 2 tries to see the collection and he/she sees nothing in GQL but sees collections in REST

### Environment

Production

### Version

Self-hosted

MVelez94:
Hey, I'd like to work on this one.

tvvignesh:
@jamesgeorge007 @AndrewBastin @CuriousCorrelation Tagging you all for visibility. Thanks a lot.

Another issue is lack of environment support in GraphQL which is being tracked here: https://github.com/hoppscotch/hoppscotch/issues/2172

If both are done, hoppscotch would be complete for us. We are using it comprehensively and would love to give feedback and help in the process. Thanks in advance.

toddgower-hiatus:
Unfortunately, this issue prevents us from using Hoppscotch. Is there any movement towards getting this working? Seems like it's not that much of a stretch to share GQL collections since it already works for REST/

ivank:
This is affecting us as well! Neither Environment variables nor the collections themselves are stored in the workspace.

sanirasheed:
was this resloved?

ThiloHettmer:
We would like to have this too. Can we expect this feature to be implemented in the near future?

EverWinter23:
We are also facing the same issue.

ItaiPendler:
also need this! 
happy to see there's a PR for this. hope it will be released soon!

nivedin:
Hoppscotch now support unified REST-GQL workspace where you can have both gql and rest requests inside the same collection, both in personal and shared workspace
Closing since this is now addressed in the [v2026.8.0 release](https://github.com/hoppscotch/hoppscotch/releases/tag/2026.8.0)

Issue #2172, opened 2022-03-17
[feature]: GraphQL environment variables support and should be shared with REST


xpol:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Summary

Add environment variables for GraphQL:
- in the GraphQL endpoint
- In the GraphQL variables editor
- In the GraphQL Headers editor
- In the GraphQL Authorization editor

And the environment variables should be shared with REST tab.

So that I can first request a REST api to get authorization information, and store tokens and other values in the environment variables, the the GraphQL requests can use those variables.

### Why should this be worked on?

Get the GraphQL editor in hoppscotch more user friendly.

jopijuco:
I would also add that environment variables should be shared with the team like collections. Otherwise, your team collections won't work because the requests are using variables that only you have.

xpol:
Yes, we must allow environments to extend other environments.

AndrewBastin:
We are working on revamping the GraphQL page completely. Should be done by next month hopefully. We will consider this feature at that point.

As for extending, we are brainstorming a UX for request based variables.

Tobychidi:
Yes. Environments should be tied to collections or teams...

hiroingk:
Are there any updates to this feature? It would be so cool if this feature was added.

Mautriz:
I'd say that environments should be global (same environments available to REST/graphql/realtime) as:
- There are usually some web services which have at least 2 of those 3 things
- Sharing a single config file per environment per team is easier/creates less friction than sharing a config file per environment per service per protocol per team   
As for me personally it's an important feature, I'd gladly switch from insomnia to hoppscotch is it was not for the absence of configs
I'm up to help in case

camflan:
Same here, was excited to transition our teams to Hoppscotch from Insomnia and hit this road block soon after setting up our second collection

Any updates?

wzrdtales:
@AndrewBastin I guess the original plans have been thrown over?

wzrdtales:
Alternatively, if this takes too much time, maybe a good golden path for a fast solution is to do it how postman integrated it, by offering a graphql option on the REST tab for body format.

azzamsa:
This is the only feature that stops me from fully migrating to hopscotch 😢

p-kuen:
What is the current status of this issue? Are you open for PR's or is there something already in progress?

Timeraa:
[...]

PR #6041, opened 2026-03-25
feat: REST-GQL unified workspace

nivedin:
Closes BE-757 FE-1162 FE-1173 #4383 #2172 #4488 

https://github.com/user-attachments/assets/ca529824-92dc-4f78-8822-ff97d2225995

Merges REST and GraphQL into a unified tab experience, allowing both protocol types to coexist within the same tabbed interface. Users can create, switch between, and manage REST and GraphQL requests from a single workspace.
We will move forward as phase by phase adding features from REST to GQL workspace.

## Core Changes

Unified tab system: Both HoppRequestDocument (REST) and HoppGQLRequestDocument (GQL) are now managed by RESTTabService via the HoppTabDocument union type. The main index.vue page conditionally renders HttpRequestTab or GqlRequestTab based on the active tab's document type.

Protocol switching: New ProtocolSwitcher.vue component lets users toggle between REST and GraphQL within an open tab. A bidirectional type converter (helpers/rest/type-converter.ts) preserves compatible fields (name, headers, auth) during conversion.

New GQL components: Full set of GraphQL-specific components — gql/Request.vue (URL bar + query editor), gql/RequestOptions.vue (query/variables/headers/auth tabs), gql/TabHead.vue, gql/DocExplorer.vue, and gql/SchemaSearch.vue.

GQL tab connection service: New gql-tab-connection.service.ts manages per-tab WebSocket connections, schema introspection, and query execution for GraphQL tabs.

Context-aware sidebar: http/Sidebar.vue conditionally shows GQL-specific tabs (Documentation, Schema) when a GraphQL tab is active, while shared tabs (Collections, Environments) remain always visible.

## Phases
- [x] Phase 1: REST-GQL unification: merges the GraphQL request interface into the existing tab system
- [x] Phase 2: GQL collection management (save/organize GQL requests in collections)
- [x] Phase 3: Unified environment handling, collection runner support for GQL and Documentation of GQL API
- [ ] Phase 4: Mock server, Scripting support

---
## Summary by cubic
Unifies REST and GraphQL into one tabbed workspace with protocol switching, per‑tab GQL execution, unified history, shareable embeds, example responses, and a collection runner. Adds backend `ReqType` (REST/GQL) with strict type enforcement, hardens auth and GQL connection UX, and improves docs/imports to meet FE-1162 and FE-1173.

- **New Features**
  - Backend: add `ReqType` to team collections/requests, register the enum in GraphQL, default root listing to REST when omitted, inherit type for child collections, validate on create/move/reorder, and set mock‑server defaults to REST.
  - Workspace/GQL: switch to `WorkspaceTabsService`; add protocol switcher; docs/schema explorer with search; extracted schema sidebar; shareable GQL embeds; sticky URL/options; per‑tab WS negotiation for `graphql-ws`/`graphql-transport-ws` with ACK/timeouts; subscription log; settings toggle for unified GQL.
  - Collections/Runner/CLI: “Add GQL request”; GQL example responses; unified history with a merged GQL card; runner supports unified collections; CLI remains REST‑only and warns/skips GQL.
  - Documentation: render GQL Query/Variables with copy actions; simpler schema downloads; new English labels for GQL states.

- **Bug Fixes**
  - Backend security/typing: scope user data to requester; enforce team ownership and collection/request type matches; default root‑collection type filter; inherit parent type; new errors `TEAM_COLL_TYPE_MISMATCH`, `TEAM_COLL_NOT_SAME_PARENT`, `TEAM_REQ_TYPE_MISMATCH`; minor model decorator fixes.
  - GQL connections: read‑only URL during connect; handle disabled introspection and invalid JSON; shape‑check introspection payloads; append auth params; robust stop/disconnect/cleanup; freeze request and deep‑clone inherited props at run; guard URL parsing; route errors by operation type; mark subscription run boundaries; prevent schema fetch clobbering.
  - Collections/docs/history/UX: resolve and cascade inherited headers; guard mixed imports; preserve JSON bodies in cURL; success toast after saving GQL; unify request naming; block sends when auth is unresolved; generate missing request ref IDs; capture cancel handles earlier in embeds; inspection results scoped per tab across REST/GQL; stronger error handling in GQL ResponseMeta/SubscriptionLog; remove redundant role attribute in GQL Field; prevent incorrect success toast on subscription operations.

<sup>Written for commit 9428a568a56f9e9fe416ae3d5d0f630f1e568549. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6041?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


bt-88:
🎉!

jamesgeorge007:
@greptileai, please review again.

nivedin:
@greptileai, please review again.

nivedin:
@copilot, please review again.

nivedin:
@greptileai, please review again.

nivedin:
@copilot, please review again.

nivedin:
@cubic-dev-ai review

nivedin:
@greptileai, please review again.

nivedin:
@greptileai, please review again.

nivedin:
@copilot please review

nivedin:
@greptileai

- mode="rest" isn't GQL-specific breakage: the manual GQL save modal uses the same mode="rest" (components/gql/Request.vue). Both flows are identical; "rest" here means the REST-page workspace, not the request type.
- SaveRequest.vue is GQL-aware: requestContext branches on doc.type === "gql-request", and after save it checks isGQLRequest() and builds a proper gql-request document + save context (personal and team paths).
- GQL requests living in REST-typed collections is the unified model — sync stores them as opaque JSON via the REST mutations.
- Loads don't corrupt: normalizeCollectionRequest (lib/sync/collections/index.ts) has explicit GQL-in-REST / REST-in-GQL branches, and team loads discriminate by shape (normalizeTeamRequestBody).
- The confirm-close handler is already GQL-aware — it runs gqlTabConn.cleanupTab() for gql-request docs before closing (pages/index.vue).

nivedin:
@greptileai please review again

jamesgeorge007 (changes requested):
1. Export a collection holding both REST and GQL requests, re-import it, and every GQL request comes back as a blank `Untitled` GET.

2. Switching a request between REST/GQL drops the REST-only fields (method, endpoint, params, body, scripts), and they aren't recovered. Saving after a switch overwrites the saved entry with no confirmation, and after a reload, switching back re-seeds defaults instead of restoring (only headers/auth survive).

3. Porting a collection that contains GQL requests currently errors. It should filter/skip GQL rather than fail the whole export.

packages/hoppscotch-backend/src/team-request/team-request.service.ts
```diff
@@ -35,13 +36,14 @@ export class TeamRequestService {
...
    */
-  private cast(tr: DbTeamRequest) {
+  private cast(tr: DbTeamRequest): TeamRequest {
     return {
       id: tr.id,
       collectionID: tr.collectionID,
       teamID: tr.teamID,
       title: tr.title,
       request: JSON.stringify(tr.request),
+      type: tr.type as ReqType,
```
nahidhasan94, line 47: The PR adds a type column to TeamRequest and exposes it in the GraphQL model, but createTeamRequest never sets it. Every request including a GQL request saved into a GQL collection will be persisted with the DB default REST.

packages/hoppscotch-backend/src/team-request/team-request.service.ts
```diff
@@ -35,13 +36,14 @@ export class TeamRequestService {
    * A helper function to cast the Prisma TeamRequest model to the TeamRequest model
    * @param tr TeamRequest model from Prisma
    */
-  private cast(tr: DbTeamRequest) {
+  private cast(tr: DbTeamRequest): TeamRequest {
```
nahidhasan94, line 40: Collections got a type-mismatch guard on move/reorder (moveCollection, updateCollectionOrder), but requests did not. A REST request can be moved into a GQL collection and vice versa.

packages/hoppscotch-backend/src/team-collection/team-collection.service.ts
```diff
@@ -210,6 +222,14 @@ export class TeamCollectionService {
     if (!Array.isArray(collectionsList.right))
       return E.left(TEAM_COLL_INVALID_JSON);
 
+    // When importing into an existing parent, ensure the type matches the parent
+    if (parentID) {
```
nahidhasan94, line 227: The new parent lookup only compares type. It does not verify the parent belongs to teamID. The resolver guard validates team membership for the teamID argument, but nothing checks that parentCollectionID is inside that team before parent: { connect: { id: parentID } } runs. Since you're already fetching the parent here, close that gap.

suggested change:

```
if (parentID) {
  const parentCollection = await this.getCollection(parentID);
  if (E.isLeft(parentCollection)) return E.left(TEAM_COLL_NOT_FOUND);
  if (parentCollection.right.teamID !== teamID)
    return E.left(TEAM_NOT_OWNER);
  if (parentCollection.right.type !== type)
    return E.left(TEAM_COLL_TYPE_MISMATCH);
}
```

packages/hoppscotch-backend/src/team-collection/team-collection.service.ts
```diff
@@ -462,6 +488,12 @@ export class TeamCollectionService {
     if (parentID !== null) {
```
nahidhasan94, line 479: Two sequential DB round-trips for the same row: isOwnerCheck(parentID, teamID) fetches the parent, then getCollection(parentID) fetches it again just to read type.

suggested change:

```
if (parentID !== null) {
  const parentCollection = await this.prisma.teamCollection.findFirst({
    where: { id: parentID, teamID },
    select: { type: true },
  });
  if (!parentCollection) return E.left(TEAM_NOT_OWNER);
  if (parentCollection.type !== type)
    return E.left(TEAM_COLL_TYPE_MISMATCH);
}
```

nahidhasan94 (changes requested):
@mirarifhasan please look into the backend feedbacks.

packages/hoppscotch-common/src/components/http/ProtocolSwitcher.vue
```diff
@@ -0,0 +1,226 @@
...
+  },
+  set: (value: string) => {
+    const doc = currentDoc.value
+    if (!doc) return
+    if (doc.type === "request") {
+      doc.request.name = value
+    } else if (doc.type === "gql-request") {
+      doc.request.name = value
+    }
+  },
```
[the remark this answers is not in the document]
nivedin: Not a bug — RequestTab's deep dirty watcher already marks the tab dirty on rename (its equality check includes name), so the close prompt fires. updateTab() is unnecessary for the same reason — it's the same reactive object. Added a comment in the setter pointing at the watcher.

packages/hoppscotch-common/src/services/gql-tab-connection.service.ts
```diff
@@ -0,0 +1,1464 @@
...
+
+    const { authHeaders, authParams } = await this.generateAuthHeader(
+      effective.effectiveFinalURL,
+      effective.effectiveFinalAuth,
+      envVars
+    )
+
+    let finalUrl = effective.effectiveFinalURL
+    if (Object.keys(authParams).length > 0) {
+      try {
```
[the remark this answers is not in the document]
nivedin: Not reachable as described, a URL that new URL() rejects also can't route to the real endpoint (browser fetch resolves it against the app origin; native interceptors reject it), so what surfaces is the invalid-URL error itself, and the env inspector already flags unresolved <<VAR>>s on the URL field. The fall-through matches fetchSchema and testRunner by design. Keeping as-is.

nivedin:
@greptileai, please review again

nivedin:
@greptileai it's an intended behavior rather than a gap: the retry loop exists specifically to protect an active subscription from a transient introspection blip. Once the socket closes, the armed retry still fires once, success promotes back to CONNECTED and resumes polling; failure ends the chain in ERROR, which is the same terminal state as any non-subscription tab whose poll fails (those don't retry either). Retrying a failing endpoint indefinitely with no live stream to protect is the unbounded-polling behavior we deliberately avoid; recovery is the Connect button. Happy to add a bounded post-close retry (e.g. 2 attempts) if product prefers, but correctness-wise the current terminal is consistent by design.

@greptileai Please review again

packages/hoppscotch-common/src/components/gql/RequestOptions.vue
```diff
@@ -0,0 +1,318 @@
...
+        // key off `response[0]`) and replace otherwise — errors stay
+        // visible instead of hiding behind a stale result, and a stream
+        // whose log was displaced (e.g. by a connection error) restores
+        // itself on the next frame.
+        const current = props.response ?? []
+        const head = current[0]
+        const logOnScreen =
+          head?.type === "response" && head.operationType === "subscription"
+        const appendsToLog =
+          event.operationType === "subscription" && logOnScreen
```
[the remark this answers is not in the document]
nivedin: Fixed with the earlier-reset approach: subscription runs now emit reset at operation start, so the stale-run bail replaces the panel instead of joining the previous log. Run-level IDs weren't needed. One subtlety: the reset must sit a flush tick ahead of the bail (Vue batches ref writes — same-tick reset+error collapses to just the error), which is why it's at function entry rather than beside the bail; verified live with a run/disconnect race.

nivedin:
> Switching a request between REST/GQL drops the REST-only fields (method, endpoint, params, body, scripts), and they aren't recovered. Saving after a switch overwrites the saved entry with no confirmation, and after a reload, switching back re-seeds defaults instead of restoring (only headers/auth survive).

Its intended behaviour. Switching snapshots the outgoing request into per-tab drafts, and switching back restores it verbatim, nothing is lost within a session. Drafts are intentionally session-only (scratch state), so a reload re-seeds from the converter, durable data belongs in a save. Save-after-switch is an explicit action, and the old-protocol draft stays recoverable until then. A confirm dialog on save-after-switch can be a follow-up if product wants it. Let me know this works @jamesgeorge007 , I have patched the rest of the issues.

packages/hoppscotch-backend/src/team-request/team-request.service.ts
mirarifhasan: In phase 1, we'll keep as it is (intentionally) to maintain current behaviour. 
In next phase, we'll take input from the FE weather this is GQL or REST.

nivedin:
@greptileai, errors render once via ResponseMeta's placeholder, and SubscriptionLog renders server error frames as rows with entry.error.message. The failure-after-success case was specifically fixed in f77160319/e7e4fbe37 and verified live.

@greptileai, please review again

packages/hoppscotch-common/src/pages/index.vue
```diff
@@ -288,12 +377,25 @@ const duplicateTab = (tabID: string) => {
...
+    const newTab = tabs.createNewTab({
+      type: "gql-request",
+      request: {
+        ...cloneDeep(doc.request),
+        _ref_id: generateUniqueRefId("req"),
+      },
+      isDirty: true,
+      cursorPosition: doc.cursorPosition ?? 0,
+    })
+    tabs.setActiveTab(newTab.id)
```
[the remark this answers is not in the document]
nivedin: Declined by design, duplicates intentionally mirror REST tab duplication: no inheritedProperties (the copy is detached from the source collection, so inherit resolves to none until it's saved somewhere, at which point saving cascades fresh properties from the real destination), and no response/sub-tab preference carry-over, which REST doesn't persist either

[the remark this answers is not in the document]
nivedin: Please review again

mirarifhasan:
Re: `createRequestInCollection` hardcoding `ReqType.REST` — this is intentional for this PR's scope, not a leftover:

1. **The `type` column tracks the collection's type, not the request body's protocol.** Team collections are REST-typed only in this PR (the migration defaults everything to `'REST'`, and there is no path to create a GQL-typed team collection yet). Stamping `REST` keeps every request row consistent with its parent collection and with the `TEAM_REQ_TYPE_MISMATCH` guard on cross-collection moves.
2. **The request's actual protocol is derived from the request JSON itself**, not the DB column — the client discriminates via `isGQLRequest` in `normalizeTeamRequestBody`, so GQL requests saved to a team collection load and execute as GQL correctly today.
3. **`CreateTeamRequestInput` deliberately does not expose `reqType`.** Accepting it from clients would allow creating GQL-typed rows inside REST collections — exactly the mismatch the new guards prevent. When first-class GQL team collections land in a follow-up, the create path will derive `type` from the parent collection rather than trusting client input.

So there's no data-correctness issue in the current scope; the column becomes meaningful once GQL team collections are introduced, and rows created until then are correctly typed relative to their (REST) collections. If we later key any query off `TeamRequest.type`, we'll backfill from the request body at that point.

mirarifhasan:
@greptileai, please review again

mirarifhasan:
@greptileai, please review again

nivedin:
@greptileai Error rendering: Present at HEAD and previously re-verified — errors render once via ResponseMeta's placeholder (Response.vue defers by design, a second branch there double-rendered and was removed). The blank-panel-on-retry bug had a different root cause, error events appending behind stale responses, fixed in f77160319/e7e4fbe37 and verified live.

nivedin:
@greptileai, please review again

packages/hoppscotch-backend/src/team-collection/team-collection.service.ts
```diff
@@ -453,15 +459,22 @@ export class TeamCollectionService {
...
+    // collection has the same type as its parent
     if (parentID !== null) {
-      const isOwner = await this.isOwnerCheck(parentID, teamID);
-      if (O.isNone(isOwner)) return E.left(TEAM_NOT_OWNER);
+      const parentCollection = await this.prisma.teamCollection.findFirst({
+        where: { id: parentID, teamID },
+        select: { type: true },
+      });
+      if (!parentCollection) return E.left(TEAM_NOT_OWNER);
+      if (parentCollection.type !== type)
```
nahidhasan94: Any client that does not explicitly send type (all existing/older clients, and any caller creating a child under a GQL parent without setting type) gets TEAM_COLL_TYPE_MISMATCH and cannot create sub-collections under GQL collections. The resolver comment on duplicateTeamCollection even states the type "is derived from the source collection," but child creation does not derive it .it trusts the REST-defaulted arg.


**Suggested fix**: derive the child's type from the parent rather than defaulting to REST. Either drop the type arg for child creation, or in the service compute type ?? parentCollection.type and only reject when an explicitly-provided type conflicts

packages/hoppscotch-backend/src/team-collection/input-type.args.ts
```diff
@@ -55,6 +73,15 @@ export class CreateChildTeamCollectionArgs {
   @IsString()
   @IsOptional()
   data: string;
+
+  @Field(() => ReqType, {
+    nullable: true,
+    defaultValue: ReqType.REST,
```
nahidhasan94: The ReqType.REST default here is the root cause of the child-creation bug.
Because this defaults to REST, any client that omits type sends REST to createCollection, which then rejects it against a GQL parent with TEAM_COLL_TYPE_MISMATCH (service.ts:476). This breaks all existing/older clients and blocks creating sub-collections under any GQL collection.

packages/hoppscotch-backend/src/team-request/team-request.service.ts
```diff
@@ -370,6 +388,29 @@ export class TeamRequestService {
...
+      select: { teamID: true, type: true },
+    });
+    if (!destCollection) return E.left(TEAM_INVALID_COLL_ID);
+    if (destCollection.teamID !== request.teamID) {
+      return E.left(TEAM_REQ_INVALID_TARGET_COLL_ID);
+    }
+
+    // When moving across collections, the source and destination collections
+    // must hold the same request type (REST/GQL)
+    if (request.collectionID !== destCollID) {
```
nahidhasan94, line 404: This fails open. When srcCollection is null the srcCollection && guard skips the type check entirely and the move proceeds unvalidated. It should fail closed. if the source collection can't be found, return an error instead of silently allowing the move.

suggested changes:
```
const srcCollection = await this.prisma.teamCollection.findUnique({
        where: { id: request.collectionID },
        select: { type: true },
      });
      if (!srcCollection) return E.left(TEAM_INVALID_COLL_ID);
      if (srcCollection.type !== destCollection.type) {
        return E.left(TEAM_REQ_TYPE_MISMATCH);
      }
```

nahidhasan94 (changes requested):
@mirarifhasan please look into the feedbacks.

packages/hoppscotch-backend/src/team-collection/team-collection.service.ts
mirarifhasan: Intentional, currently we're not saving GQL collections/requests.
Just added support ReqType in this PR for `phase-2` implementation of unified playground.

So i think we can skip for now.

mirarifhasan:
### Backend feedback — status

@nahidhasan94 I've gone through the remaining backend items. Where each one stands:

#### Fixed in this push

**1. Child collections now inherit their parent's type** — `team-collection.service.ts`, `input-type.args.ts`

`CreateChildTeamCollectionArgs.type` carried `defaultValue: ReqType.REST`, so an omitted argument arrived at the service as `REST` rather than `undefined`. `createCollection` then applied a strict `parentCollection.type !== type` check, which meant **every child collection created under a GQL parent was rejected** with `TEAM_COLL_TYPE_MISMATCH` — and the service had no way to tell "caller omitted the type" apart from "caller explicitly said REST".

`type` is now genuinely optional. `null` or omitted means *inherit*: a child takes its parent's type, a root collection still falls back to REST. Only an explicitly supplied type that disagrees with the parent is rejected. This matches the pattern `createTeamRequest` already uses (`type ?? collection.type`), so the two services are now consistent. Because the field was already `nullable: true`, this is not a schema-breaking change.

**2. `moveRequest` now fails closed on the source collection** — `team-request.service.ts`

The guard read `if (srcCollection && srcCollection.type !== destCollection.type)`. A null `srcCollection` short-circuited the whole check and let the move proceed unvalidated. It now returns `TEAM_INVALID_COLL_ID` when the source collection can't be resolved, then compares types unconditionally — as requested in the earlier review.

#### Already addressed earlier in the branch

These came up in review but the code has since moved on, so flagging them as no longer outstanding:

- **`createTeamRequest` does set `type`.** It derives from the destination collection (`type ?? collection.type`) and rejects explicit mismatches with `TEAM_REQ_TYPE_MISMATCH`. My earlier "phase 1, keep as-is" reply is out of date.
- **Parent-ownership gap on import is closed.** `importCollectionsFromJSON` verifies `parentCollection.teamID === teamID` before connecting the parent.
- **The duplicate round-trip in `createCollection` is gone.** `isOwnerCheck` was replaced by a single `findFirst({ id: parentID, teamID }, select: { type })`.
- **Cross-team move validation runs on every move**, not only when `nextRequestID` is null.

#### On the "GQL requests persist as REST" note (greptile, 3/5)

This is intended phase-1 behaviour rather than a guard being defeated. A request's type follows the collection it lives in — `createTeamRequest` derives it — and in phase 1 the unified workspace stores GQL requests inside REST collections as opaque JSON, which the frontend reconstitutes via `isGQLRequest()`. Greptile reached the same conclusion in a later thread on this PR ("GQL-in-REST collection storage is intentional").

Real GQL collections arrive in phase 2. The inheritance added in (1) is precisely what lets that happen without a breaking schema change later.

#### Verification

- `nest build` passes.
- `team-collection` + `team-request` suites green — 132 tests, 0 failures.
- No new lint errors; the 15 `no-unused-vars` reported in these files are pre-existing unused catch-block parameters, identical before and after this change.

mirarifhasan:
### Re: "GQL team-request type never wired from frontend"

I don't think this one belongs on the frontend — the suggested change would break phase-1 saves.

The proposal is for `createRequestInCollection` to send `type: isGQLRequest(parsedRequest) ? ReqType.GQL : ReqType.REST`. In phase 1 there are no GQL team collections, so GQL requests are saved into **REST** collections. `createTeamRequest` guards on exactly that:

```ts
if (!collection) return E.left(TEAM_INVALID_COLL_ID);
if (type && type !== collection.type) {
  return E.left(TEAM_REQ_TYPE_MISMATCH);
}
const requestType = type ?? (collection.type as ReqType);
```

With the collection at `REST` and the frontend sending `GQL`, that guard fires and the save fails with `TEAM_REQ_TYPE_MISMATCH`. The suggested fix contradicts a guard added in this same PR.

So `type = REST` on those rows isn't a gap — it's the direct consequence of the invariant this PR enforces: a collection and the requests inside it carry one type. A `GQL` request stored inside a `REST` collection is precisely the mixed-protocol state that `TEAM_COLL_TYPE_MISMATCH` / `TEAM_REQ_TYPE_MISMATCH` exist to prevent.

On the phase-2 concern about rows carrying the wrong type: no frontend wiring is needed then either. `type ?? collection.type` already resolves correctly — once GQL collections exist, a GQL request saved into one inherits `GQL` automatically. That inheritance is the reason the field is optional rather than required; the frontend never has to compute it.

The column also isn't load-bearing for rendering — the frontend reconstitutes the document type from the payload via `isGQLRequest()`, as established in an earlier thread on this PR.

Keeping the current behaviour.

mirarifhasan:
@greptileai, please review again

nivedin:
@tembo please review

nivedin:
@greptileai please review

nivedin:
@tembo please review

nivedin:
@tembo please review

nahidhasan94 (approved):

jamesgeorge007 (approved):
Really nice to see this land where it is 🙌

It's come a long way with the earlier mixed-JSON import, OpenAPI GQL export, team create/move type-guard, and unsaved-draft reload issues are all resolved.



Agree with carrying the remaining AI-review comments into phase 2. The ReqType team-collection BE work is intentional for this phase, so those are fine to skip here. A few findings to keep track of, subject to evaluation, would be:



- `graphql/FieldLink.vue` — `clickable ? handleClick : undefined` returns the function without invoking it, so the field-link click is a no-op; wants `handleClick($event)`.

- `graphql/Response.vue` — `JSON.parse(response[0].data)` in a computed with no `try/catch`, so malformed/non-JSON response data can take the response panel down.

- `graphql/Request.vue` — `interceptorService.current.value!.id` assumes an interceptor exists; a null during setup will throw.

Let's get it in and knock out the rest in phase 2 🚀

PR #6043, opened 2026-03-26
feat(desktop): cloud for orgs support - phase 3

CuriousCorrelation:
Adds the kernel log module, store org-scoping with diagnostic
 logging, a Vue Router guard to preserve the ?org= query param
 across navigations, and the ScopedStore/extendStore surface in
 the kernel barrel. Also moves prisma packages to Linux-only in
 devenv to fix a nightly Rust build failure.

 Closes FE-795
 Closes FE-813
 Closes FE-820
 Closes FE-818
 Closes FE-827
 Closes FE-804
 Closes FE-814
 Closes FE-819
 Closes FE-815
 Closes FE-816
 Closes FE-822
 Closes FE-823
 Closes FE-1080
 Closes FE-1081
 Closes FE-1099
 Closes FE-1082
 Closes FE-1120
 Closes FE-1089
 Closes FE-1090
 Closes FE-1091
 Closes FE-828
 Closes FE-1083
 Closes FE-1084
 Closes FE-1085
 Closes FE-1086
 Closes FE-1092
 Closes FE-817
 Closes FE-826
 Closes FE-1093
 Closes FE-1094
 Closes FE-825
 Closes FE-1141
 Closes FE-1142
 Closes FE-1143
 Closes FE-1144
 Closes FE-1145
 Closes FE-1146
 Closes FE-1147
 Closes FE-1148
 Closes FE-1149
 Closes FE-1150
 Closes FE-1151
 Closes FE-1152
 Closes FE-1153
 Closes FE-1154
 Closes FE-1155
 Closes FE-1156
 Closes FE-1157
 Closes FE-1158
 Closes FE-1159
 Closes FE-1160
 Closes FE-1161
 Closes FE-1166
 Closes FE-794

 ## Kernel log module

 New `log` module following the existing versioned API pattern
 (v1 interface, web impl, desktop impl, registered in initKernel).

 `packages/hoppscotch-kernel/src/log/v/1.ts` defines `LogV1` with
 `init(logPath)` and `log(logPath, level, tag, message, data?)`.
 Two error kinds (`init`, `write`), three capability flags
 (`console`, `file`, `buffer`), and a default stub that no-ops.

 `impl/web/v/1.ts` has `BrowserLogManager` (singleton). Routes to
 the appropriate `console[level]` method and pushes to an in-memory
 buffer capped at 5000 entries. Exposes `window.__DIAG_LOGS__` and
 `window.__dumpDiagLogs__()` for devtools inspection.

 `impl/desktop/v/1.ts` has `TauriLogManager` with per-path
 singletons (same pattern as `TauriStoreManager` in the store
 module). Does everything the web impl does plus batched file writes
 every 500ms via `invoke("append_log", { filename, content })`.
 On `init()` writes a session header with org context, href, host,
 and kernel presence. Each webview's log block is identifiable by
 its org context. The Tauri `invoke` for file operations is
 lazy-loaded to avoid import-time failures in non-desktop builds.

 `packages/hoppscotch-common/src/kernel/log.ts` follows the same
 delegation pattern as `store.ts`. Provides `Log.init()`,
 `Log.debug()`, `Log.info()`, `Log.warn()`, `Log.error()`. Also
 exports `diag()` as a convenience alias for `Log.info()` so call
 sites can import directly from `~/kernel/log`.

 The log file path resolves to `{get_logs_dir()}/diag-logs.txt`
 using `path.rs`'s `get_logs_dir` command. This puts diag logs next
 to the main app log (`io.hoppscotch.desktop.log`).

 Circular dependency avoidance: `log.ts` does NOT import from
 `store.ts` even though `store.ts` has a `getLogsDir` function.
 This is because `store.ts` imports `diag` from `log.ts` and calls
 it at module evaluation time (the top-level
 `diag("store", "--- COMMON store.ts module evaluated ---")`). If
 `log.ts` imported from `store.ts`, the ESM circular resolution
 would leave `logAtLevel` (a const in `log.ts`) uninitialized when
 `store.ts` tries to call `diag()`. Instead, `log.ts` inlines its
 own copy of the Tauri invoke lazy-load pattern to call
 `get_logs_dir` directly.

 `kernel.js` (the pre-module init script that runs before any ES
 modules load) also writes to the log file. It uses the raw
 `__TAURI_INTERNALS__.invoke` directly since `@tauri-apps/api` isn't
 available yet. The webview init block logs the tag, `?org=` param,
 href, hostname, and origin.

 ## Store org-scoping

 With shared origin, `window.location.host` is `hoppscotch` for
 every webview, which means `STORE_PATH` in
 `hoppscotch-common/src/kernel/store.ts` would resolve to the same
 `hoppscotch.hoppscotch.store` file for every webview. Auth tokens,
 settings, collections, and other per-org data would all collide.

 The fix reads the `?org=` param and uses it as the file prefix:

 ```typescript
 const orgParam =
   new URLSearchParams(window.location.search).get("org")
 const STORE_PATH = orgParam
   ? `${orgParam.replace(/[^a-zA-Z0-9]/g, "_")}.hoppscotch.store`
   : `${window.location.host}.hoppscotch.store`
 ```

 The sanitization (`replace(/[^a-zA-Z0-9]/g, "_")`) produces the
 same filenames as the old per-hostname approach.
 `test-org.hoppscotch.io` becomes `test_org_hoppscotch_io`, so the
 store file is `test_org_hoppscotch_io.hoppscotch.store`. Existing
 store files from the per-hostname era stay compatible. The vendored
 webview (no `?org=` param) falls back to
 `hoppscotch.hoppscotch.store` as before.

 `getStorePath()` resolves the full disk path by joining `STORE_PATH`
 with the Rust-provided `get_store_dir()`. Every `Store.get()`,
 `Store.set()`, and `Store.init()` call goes through `getStorePath()`
 which caches the result after first resolution. Diagnostic logging
 at module evaluation time prints the computed `STORE_PATH` and
 `orgParam` so disk logs show exactly which file each webview is
 using. Additional diag calls in `Store.init()`, `Store.set()`, and
 `Store.get()` (with value preview) trace the full store lifecycle.

 A new `getLogsDir()` export wraps the Rust `get_logs_dir` command
 for use by the log module's path resolution.

 ## ScopedStore and extendStore

 `ScopedStore` interface in `kernel/store/v/1.ts` (added in phase 2)
 provides a namespace-bound store surface (get, set, remove, clear,
 has, listKeys, watch) so consumers don't need to pass namespace on
 every call. `extend()` function constructs one from any StoreV1
 instance plus a store path and namespace.

 This phase adds the barrel export from `kernel/index.ts`
 (`extendStore`) and the `Store.extend(namespace)` method in
 `common/kernel/store.ts` that wires it up.

 ## Router guard for ?org= preservation

 `Login.vue` calls `window.location.reload()` after
 `signInWithCustomToken` succeeds. A full-page reload strips query
 params from the URL, so on the reloaded page `desktopOrgHost()`
 returns null and the app thinks it's vendored.

 The fix is a Vue Router `beforeEach` guard in
 `hoppscotch-common/src/modules/router.ts`:

 ```typescript
 const initialOrgParam =
   new URLSearchParams(window.location.search).get("org")

 router.beforeEach(async (to, from) => {
   if (initialOrgParam && !to.query.org) {
     return { ...to, query: { ...to.query, org: initialOrgParam } }
   }
   _isLoadingInitialRoute.value = isInitialRoute(from)
   // ... rest of existing guard
 })
 ```

 `initialOrgParam` is captured at router init time from
 `window.location.search`. At that point the query param is always
 present because Tauri set it on the initial webview URL. Every
 subsequent route transition goes through `beforeEach`. If the
 target route doesn't have `?org=`, the guard returns a modified
 route with it injected. For full-page reloads, the URL in the
 address bar already has `?org=` because the guard ensured the
 previous navigation included it.

 The guard runs as the first check in `beforeEach`, before the
 existing initial-route loading state logic. If it returns a
 redirect, the guard re-triggers (Vue Router calls `beforeEach`
 again for the redirected route), but this time `to.query.org` is
 present so the guard passes through. No infinite loop.

 On web this is a no-op since there's no `?org=` param.

 ## devenv.nix

 Moved `nodePackages.prisma` and `prisma-engines` to `linuxPackages`.
 `prisma-engines` fails to compile on recent Rust nightlies due to
 the `metrics` crate v0.23.0 using a raw pointer lifetime pattern
 rejected by rust-lang/rust#141402. Since the `PRISMA_*` env vars
 are already Linux-only (the database runs in Docker on macOS),
 neither package needs to be in the cross-platform set.

 Also bumps Go from `go_1_24` to `go_1_25`.

### DIFF
--- a/devenv.lock
+++ b/devenv.lock
@@ -3,10 +3,11 @@
     "devenv": {
       "locked": {
         "dir": "src/modules",
-        "lastModified": 1764669403,
+        "lastModified": 1774428097,
+        "narHash": "sha256-yQAutPgbsVHsN/SygZDyzMRxQn6Im53PJkrI377N8Sg=",
         "owner": "cachix",
         "repo": "devenv",
-        "rev": "3f2d25e7af748127da0571266054575dd8fec5ab",
+        "rev": "957d63f663f230dc8ac3b85f950690e56fe8b1e0",
         "type": "github"
       },
       "original": {
@@ -24,80 +25,26 @@
         "rust-analyzer-src": "rust-analyzer-src"
       },
       "locked": {
-        "lastModified": 1764658058,
+        "lastModified": 1774423251,
+        "narHash": "sha256-g/PP8G9WcP4vtZVOBNYwfGxLnwLQoTERHnef8irAMeQ=",
         "owner": "nix-community",
         "repo": "fenix",
-        "rev": "12bd9c7bcbeb949741b3ad0ca2b3506d0718cf4d",
+        "rev": "b70d7535088cd8a9e4322c372a475f66ffa18adf",
         "type": "github"
       },
       "original": {
         "owner": "nix-community",
         "repo": "fenix",
-        "type": "github"
-      }
-    },
-    "flake-compat": {
-      "flake": false,
-      "locked": {
-        "lastModified": 1761588595,
-        "owner": "edolstra",
-        "repo": "flake-compat",
-        "rev": "f387cd2afec9419c8ee37694406ca490c3f34ee5",
-        "type": "github"
-      },
-      "original": {
-        "owner": "edolstra",
-        "repo": "flake-compat",
-        "type": "github"
-      }
-    },
-    "git-hooks": {
-      "inputs": {
-        "flake-compat": "flake-compat",
-        "gitignore": "gitignore",
-        "nixpkgs": [
-          "nixpkgs"
-        ]
-      },
-      "locked": {
-        "lastModified": 1763988335,
-        "owner": "cachix",
-        "repo": "git-hooks.nix",
-        "rev": "50b9238891e388c9fdc6a5c49e49c42533a1b5ce",
-        "type": "github"
-      },
-      "original": {
-        "owner": "cachix",
-        "repo": "git-hooks.nix",
-        "type": "github"
-      }
-    },
-    "gitignore": {
-      "inputs": {
-        "nixpkgs": [
-          "git-hooks",
-          "nixpkgs"
-        ]
-      },
-      "locked": {
-        "lastModified": 1762808025,
-        "owner": "hercules-ci",
-        "repo": "gitignore.nix",
-        "rev": "cb5e3fdca1de58ccbc3ef53de65bd372b48f567c",
-        "type": "github"
-      },
-      "original": {
-        "owner": "hercules-ci",
-        "repo": "gitignore.nix",
         "type": "github"
       }
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1764611609,
+        "lastModified": 1774273680,
+        "narHash": "sha256-a++tZ1RQsDb1I0NHrFwdGuRlR5TORvCEUksM459wKUA=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "8c29968b3a942f2903f90797f9623737c215737c",
+        "rev": "fdc7b8f7b30fdbedec91b71ed82f36e1637483ed",
         "type": "github"
       },
       "original": {
@@ -111,21 +58,18 @@
       "inputs": {
         "devenv": "devenv",
         "fenix": "fenix",
-        "git-hooks": "git-hooks",
         "nixpkgs": "nixpkgs",
-        "pre-commit-hooks": [
-          "git-hooks"
-        ],
         "rust-overlay": "rust-overlay"
       }
     },
     "rust-analyzer-src": {
       "flake": false,
       "locked": {
-        "lastModified": 1764603480,
+        "lastModified": 1774376228,
+        "narHash": "sha256-7oA0u4aghFjjIcIDKZ26NUpXH7hVXGPC0sI1OfK7NUk=",
         "owner": "rust-lang",
         "repo": "rust-analyzer",
-        "rev": "f25db5500baa047106d74962fe361ea59ce6f91e",
+        "rev": "eabb84b771420b8396ab4bb4747694302d9be277",
         "type": "github"
       },
       "original": {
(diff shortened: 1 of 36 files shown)


CuriousCorrelation:
Pushed fixes for AI reviews:

The CI failure was the `@log` path alias missing from `tsconfig.base.json` and `vite.config.ts`. Added it.

`Object.keys(null)` crash in the Store.get() preview is fixed, `typeof null === "object"` in JS so null values would throw. Also reworked the preview to log type and key count only (`Right(object(3 keys))`, `Right(string)`) instead of actual values, since `String(val).slice(0, 100)` could persist auth tokens into the diag log.

The flush data loss was a good catch. `pendingWrites` was getting cleared before the `invoke` call, so if the write failed those lines were just gone. Now it saves a snapshot, clears the array for new writes during the await, and prepends the snapshot back on failure.

`earlyBuffer` not cleared after drain is fixed, `console[method]` indexing replaced with explicit if/else chains in all three files, added type annotations to the `log()` params, removed the dead try/catch in web init, added a window guard in `tryModule()`, and `.catch(() => {})` on the fire-and-forget promise in `diag()`.

A few things I'm leaving as-is:

Desktop log `init()` swallowing errors is intentional. File logging is supplementary to console, so if `append_log` isn't available or the disk is full, the app shouldn't crash. Console logging still works and the outer API wraps in Either so callers that care can check the result.

ScopedStore surface is intentionally minimal for v1 (`isAvailable`, `get`, `set`, `remove`). `clear`/`has`/`listKeys`/`watch` will come when consumers need them. The interface is in the kernel package so extending it is non-breaking. Updated the PR description to match the actual surface.

`isAvailable()` calling `store.init()` is fine because both impls are idempotent. `TauriStoreManager` checks an `initialized` flag and returns early, web init is a no-op.

`get()` returning null for both errors and missing keys is a v1 simplicity choice. Consumers that need to distinguish can use `Store.get()` directly, which returns `Either<StoreError, T | undefined>`

CuriousCorrelation:
Pushed the router guard fix, returning an explicit { path, hash, params, query } instead of spreading to. Good catch on that one, the spread was pulling in matched, meta, and other internal RouteLocationNormalized fields that Vue Router doesn't expect in the return value.

The appload URL construction comments (unwrap and encoding) are valid observations but the inputs are controlled on both sides: host comes from the instance registry which stores clean domain names, and sanitize_window_label already validates bundle_name. Going to leave those as-is for now and revisit if the input surface ever widens.
The ScopedStore.get() returning null for errors was covered in the previous round, intentional v1 choice

jamesgeorge007 (approved):

PR #6097, opened 2026-04-07
fix(workspace): paginate all teams in spotlight search

### LINKED ISSUE #6096: [bug]: Workspaces missing from the search in web and desktop clietn

### Platform

Desktop App

### Browser

Firefox

### Operating System

Windows

### Bug Description

We have around 55 workspaces for our colleagues to work with.
Currently it is quite tedious to use the dropdown in the upper right to switch workspaces, since it is a long list which is not even sorted alphabetically. That makes searching for the right workspace quite hard. I tried the search function on the top, but that does not show all of my workspaces. I would expect it to allow me to search for every workspace, but just some of them show up.

We are self-hosting so I am wondering whether I set something up incorrectly or if it is just not working well.

Steps to reproduct:
1. Add a lot of workspaces
2. Search for the namespaces (some show up and some do not)

### Deployment Type

### Version

v2026.2.1

sandiprv9898:
Closes #6096

## Introduction

The spotlight workspace search only fetches the first page of teams (max 10) because `fetchMyTeams()` makes a single API call without looping through cursor-based pagination. Users with more than 10 workspaces cannot find all their workspaces in the search.

## What's changed

- Added a `while(true)` pagination loop to `SwitchWorkspaceSpotlightSearcherService.fetchMyTeams()` in `workspace.searcher.ts`
- The loop fetches pages of teams using cursor-based pagination until fewer than `BACKEND_PAGE_SIZE` (10) results are returned
- Imported `BACKEND_PAGE_SIZE` from `~/helpers/backend/helpers` (already exported and used elsewhere in the codebase)
- Error handling: if an API call fails mid-pagination, the loop breaks gracefully and returns the teams fetched so far

**Root cause:** The `cursor` variable was computed once from an empty `results` array (always `undefined`), and no loop existed to fetch subsequent pages. This is the same pagination pattern already used correctly in `TeamListAdapter.fetchList()`.

**Before:**
```typescript
const results = []
const cursor = results.length > 0 ? results[results.length - 1].id : undefined // always undefined
const result = await platform.backend.getUserTeams(cursor) // one call, max 10 teams
```

**After:**
```typescript
const results = []
while (true) {
  const cursor = results.length > 0 ? results[results.length - 1].id : undefined
  const result = await platform.backend.getUserTeams(cursor)
  if (E.isLeft(result)) break
  results.push(...result.right.myTeams)
  if (result.right.myTeams.length !== BACKEND_PAGE_SIZE) break
}
```

## Notes to reviewers

- This fix only addresses the spotlight search pagination. The workspace dropdown (`Selector.vue`) already paginates correctly via `TeamListAdapter`.
- The issue also mentions workspaces not being sorted alphabetically — that is a separate concern in `team.service.ts` (`getTeamsOfUser` has no `orderBy`) and will be addressed in a separate PR to keep changes small.
- The `BACKEND_PAGE_SIZE` constant is imported from `~/helpers/backend/helpers` rather than redefined locally, matching the pattern in `ShortcodeListAdapter.ts`.

---
## Summary by cubic
Fixes Spotlight workspace search to return all teams and alphabetizes the workspace dropdown for a stable, consistent list.

- **Bug Fixes**
  - Paginate in `SwitchWorkspaceSpotlightSearcherService.fetchMyTeams()` by looping `platform.backend.getUserTeams(cursor)` until fewer than `BACKEND_PAGE_SIZE`; import `BACKEND_PAGE_SIZE` from `~/helpers/backend/helpers`.
  - Sort teams by name in `components/workspace/Selector.vue` via a computed list; use `team.id` as the `v-for` key for stable rendering.

<sup>Written for commit a7961445af57b255ddb5664f2ae9d978ccc0e857. Summary will update on new commits. <a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6097?utm_source=github">Review in cubic</a></sup>


sandiprv9898:
Good point  I'll address this in a follow-up

sandiprv9898:
As requested in #6099, I've folded the alphabetical sorting fix into this PR.

**New commit added:**
- Sort workspaces alphabetically by name (client-side `computed` with `localeCompare`)
- Use `team.id` as `v-for` key instead of array index for stable rendering after sort

This PR now addresses both parts of #6096:
1. ✅ Spotlight search pagination (all workspaces searchable)
2. ✅ Workspace dropdown sorted alphabetically

PR #6098, opened 2026-04-07
feat(backend): use stateless OAuth2 state store

mirarifhasan:
Closes BE-743

#### The Problem
The OAuth flow relies on `express-session` with the default `MemoryStore` — an in-process, in-memory session store. Here's what breaks:

1. User hits GET /v1/auth/google → routed to Instance A
  - Passport generates an OAuth state parameter
  - `store: true` serializes `{ redirect_uri }` into Instance A's in-memory session store
  - User is redirected to Google
2. Google calls back `GET /v1/auth/google/callback` → load balancer routes to Instance B
  - Passport tries to look up the state from Instance B's memory store
  - State not found → authentication fails

This affects Google, GitHub, Microsoft (all use store: true).

#### The Solution
**Go stateless (no session store at all):**

Instead of storing state server-side, encode the OAuth state as a signed JWT or encrypted token:

1. Remove `store: true` from all strategies and set state: false
2. In each SSO guard's getAuthenticateOptions(), generate a signed state token containing `{ redirect_uri, nonce, exp }` and pass it as the state parameter
3. On callback, verify the signature and extract the `redirect_uri` from the state parameter directly (no session lookup needed)
4. Remove express-session entirely

### Notes to reviewers

Test authentication flows across all providers and verify correct redirection behavior for both the application and admin panel during login.

### DIFF
--- a/packages/hoppscotch-backend/src/auth/stateless-state-store.ts
+++ b/packages/hoppscotch-backend/src/auth/stateless-state-store.ts
@@ -52,6 +52,16 @@ export class StatelessStateStore {
     maxAgeMs: number = 600_000,
     cookieName?: string,
   ) {
+    if (!secret) {
+      if (process.env.NODE_ENV === 'production') {
+        throw new Error(
+          'StatelessStateStore: INFRA.SESSION_SECRET must be set in production. Without a shared secret, OAuth callbacks will fail across load-balanced instances.',
+        );
+      }
+      console.warn(
+        '[StatelessStateStore] INFRA.SESSION_SECRET not set; using ephemeral fallback. OAuth will fail in any multi-instance deployment.',
+      );
+    }
     this.secret = secret || FALLBACK_SECRET;
     this.maxAgeMs = maxAgeMs;
     this.cookieName = cookieName || DEFAULT_COOKIE_NAME;
--- a/packages/hoppscotch-backend/src/auth/stateless-state-store.ts
+++ b/packages/hoppscotch-backend/src/auth/stateless-state-store.ts
@@ -46,11 +46,13 @@ export class StatelessStateStore {
   private readonly secret: string;
   private readonly maxAgeMs: number;
   private readonly cookieName: string;
+  private readonly secureCookies: boolean;
 
   constructor(
     secret?: string,
     maxAgeMs: number = 600_000,
     cookieName?: string,
+    secureCookies?: boolean,
   ) {
     if (!secret) {
       if (process.env.NODE_ENV === 'production') {
@@ -65,6 +67,7 @@ export class StatelessStateStore {
     this.secret = secret || FALLBACK_SECRET;
     this.maxAgeMs = maxAgeMs;
     this.cookieName = cookieName || DEFAULT_COOKIE_NAME;
+    this.secureCookies = secureCookies ?? false;
   }
 
   /**
@@ -144,7 +147,7 @@ export class StatelessStateStore {
         req.res.cookie(this.cookieName, nonce, {
           httpOnly: true,
           sameSite: 'lax',
-          secure: req.secure || req.protocol === 'https',
+          secure: this.secureCookies,
           maxAge: this.maxAgeMs,
           path: '/',
         });
--- a/packages/hoppscotch-backend/src/auth/stateless-state-store.ts
+++ b/packages/hoppscotch-backend/src/auth/stateless-state-store.ts
@@ -1,0 +1,185 @@
+import * as crypto from 'crypto';
+
+/**
+ * Module-level fallback secret for single-instance deployments
+ * where INFRA.SESSION_SECRET is not configured.
+ */
+const FALLBACK_SECRET = crypto.randomBytes(32).toString('hex');
+
+/**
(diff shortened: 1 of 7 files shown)


packages/hoppscotch-backend/src/auth/stateless-state-store.ts
```diff
@@ -0,0 +1,252 @@
...
+ *
+ * IMPORTANT: For multi-instance deployments, `INFRA.SESSION_SECRET` must be
+ * set to the same value across all instances.
+ */
+export class StatelessStateStore {
+  private readonly secret: string;
+  private readonly maxAgeMs: number;
+  private readonly cookieName: string;
+
+  constructor(
```
nahidhasan94, line 51: Silent failure on missing SESSION_SECRET in production


suggested block:
```
constructor(
  secret?: string,
  maxAgeMs: number = 600_000,
  cookieName?: string,
) {
  if (!secret) {
    if (process.env.NODE_ENV === 'production') {
      throw new Error(
        'StatelessStateStore: INFRA.SESSION_SECRET must be set in production. ' +
        'Without a shared secret, OAuth callbacks will fail across load-balanced instances.',
      );
    }
    console.warn(
      '[StatelessStateStore] INFRA.SESSION_SECRET not set; using ephemeral fallback. ' +
      'OAuth WILL fail in any multi-instance deployment.',
    );
  }
  this.secret = secret || FALLBACK_SECRET;
  this.maxAgeMs = maxAgeMs;
  this.cookieName = cookieName || DEFAULT_COOKIE_NAME;
}
```

packages/hoppscotch-backend/src/auth/strategies/google.strategy.ts
```diff
@@ -23,7 +24,11 @@ export class GoogleStrategy extends PassportStrategy(Strategy) {
       callbackURL: configService.get<string>('INFRA.GOOGLE_CALLBACK_URL'),
       scope: configService.get<string>('INFRA.GOOGLE_SCOPE').split(','),
       passReqToCallback: true,
-      store: true,
+      store: new StatelessStateStore(
```
nahidhasan94, line 27: all three strategies pass the same cookie name . If a user opens GitHub login in one tab, then Google login in another tab before completing the first, the second Set-Cookie overwrites the first nonce. The first callback then fails with "Invalid state signature" and the user has no idea why. This is a real, reproducible end-user bug.

suggested fix:

```
store: new StatelessStateStore(
  configService.get<string>('INFRA.SESSION_SECRET'),
  undefined,
  (configService.get<string>('INFRA.SESSION_COOKIE_NAME') || '__oauth_nonce') + '_google',
),
```

packages/hoppscotch-backend/src/auth/strategies/github.strategy.ts
```diff
@@ -21,7 +22,11 @@ export class GithubStrategy extends PassportStrategy(Strategy) {
       clientSecret: configService.get<string>('INFRA.GITHUB_CLIENT_SECRET'),
       callbackURL: configService.get<string>('INFRA.GITHUB_CALLBACK_URL'),
       scope: [configService.get<string>('INFRA.GITHUB_SCOPE')],
-      store: true,
+      store: new StatelessStateStore(
```
nahidhasan94, line 25: all three strategies pass the same cookie name . If a user opens GitHub login in one tab, then Google login in another tab before completing the first, the second Set-Cookie overwrites the first nonce. The first callback then fails with "Invalid state signature" and the user has no idea why. This is a real, reproducible end-user bug.

suggested fix:

```
store: new StatelessStateStore(
  configService.get<string>('INFRA.SESSION_SECRET'),
  undefined,
  (configService.get<string>('INFRA.SESSION_COOKIE_NAME') || '__oauth_nonce') + '_github',
),
```

packages/hoppscotch-backend/src/auth/strategies/microsoft.strategy.ts
```diff
@@ -22,7 +23,11 @@ export class MicrosoftStrategy extends PassportStrategy(Strategy) {
       callbackURL: configService.get<string>('INFRA.MICROSOFT_CALLBACK_URL'),
       scope: configService.get<string>('INFRA.MICROSOFT_SCOPE').split(','),
       tenant: configService.get<string>('INFRA.MICROSOFT_TENANT'),
-      store: true,
+      store: new StatelessStateStore(
```
nahidhasan94, line 26: all three strategies pass the same cookie name . If a user opens GitHub login in one tab, then Google login in another tab before completing the first, the second Set-Cookie overwrites the first nonce. The first callback then fails with "Invalid state signature" and the user has no idea why. This is a real, reproducible end-user bug.

suggested fix:

```
store: new StatelessStateStore(
  configService.get<string>('INFRA.SESSION_SECRET'),
  undefined,
  (configService.get<string>('INFRA.SESSION_COOKIE_NAME') || '__oauth_nonce') + '_microsoft',
),
```

nahidhasan94 (changes requested):
please look into the feedbacks once

nahidhasan94 (dismissed):

nahidhasan94 (approved):

Issue #312, opened 2019-11-19
Graphql support 


arangates:
with GraphQL support , this can be future proof

liyasthomas:
#311 fixes this.

PR #6121, opened 2026-04-11
feat: add checkbox to disable environment entries

FarahThoufeek:
Corresponding to issue  #312 
Added a checkbox to each environment variable row to enable/disable entries. 
Disabled entries are excluded when saving the environment.
Changes:
1. Added enabled toggle checkbox to each variable row in Details.vue
2. New entries default to enabled: true
3. Disabled entries are filtered out on save

---
## Summary by cubic
Add a checkbox to enable/disable environment variables without deleting them, so you can keep entries but exclude them when using an environment. Addresses #312.

- **New Features**
  - Added an enable/disable checkbox to each variable row in `packages/hoppscotch-common/src/components/environments/my/Details.vue`.
  - New variables default to enabled.
  - Disabled variables are filtered out on save.

- **Refactors**
  - Updated `packages/hoppscotch-selfhost-web/vite.config.ts` to disable sitemap generation/copy and turn off HMR overlay.

<sup>Written for commit 3f6f050ee7d2cb1e5e01fbf28d0fb4bb98fb9bb8. Summary will update on new commits.</sup>

### DIFF
--- a/packages/hoppscotch-common/src/components/environments/my/Details.vue
+++ b/packages/hoppscotch-common/src/components/environments/my/Details.vue
@@ -130,7 +130,15 @@
                     v-for="({ id, env }, index) in tab.variables"
                     :key="`${tab.id}-${id}-${index}`"
                     class="flex divide-x divide-dividerLight"
-                  >
+>
+                    <div class="flex items-center px-2">
+                      <input
+                        type="checkbox"
+                        :checked="env.enabled !== false"
+                        class="cursor-pointer"
+                        @change="env.enabled = ($event.target as HTMLInputElement).checked"
+                      />
+                    </div>
                     <input
                       v-model="env.key"
                       v-focus
@@ -510,6 +518,7 @@ const addEnvironmentVariable = () => {
       currentValue: "",
       initialValue: "",
       secret: selectedEnvOption.value === "secret",
+      enabled: true,
     },
   })
 }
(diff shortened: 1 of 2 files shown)


nivedin:
@FarahThoufeek Thank you for the PR. Could you clarify the use case for this feature? Also, the issue you’ve linked doesn’t seem to match the feature, could you please check?

Instead of clicking the checkbox and filtering out environment variables, users can click the Remove button to remove individual variables. Please let me know if I’ve understood the feature correctly.

FarahThoufeek:
Hi @nivedin. Thank you for reviewing. The use case is different from the remove button, the idea is to temporarily disable a variable without losing it. For example, a user might have multiple variables with the same key pointing to different values, and wants to quickly switch between them by enabling one and disabling the others, without having to delete and re add them each time.Could you let me know if this use case makes sense and if the implementation approach is on the right track? Happy to make any adjustments!

nivedin:
> Hi @nivedin. Thank you for reviewing. The use case is different from the remove button, the idea is to temporarily disable a variable without losing it. For example, a user might have multiple variables with the same key pointing to different values, and wants to quickly switch between them by enabling one and disabling the others, without having to delete and re add them each time.Could you let me know if this use case makes sense and if the implementation approach is on the right track? Happy to make any adjustments!

Understood. In that case, we might need to tweak the code a bit. Currently, we filter out the disabled variables while saving and remove them, but we don’t want to do it that way. Instead, we should exclude the disabled variables during runtime or while parsing the environment.

PR #6122, opened 2026-04-11
refactor(cli): match test-result helper name to documented contract

### LINKED ISSUE #6118: `hasFailedTestCases` returns true when all tests pass, false when tests fail
`hasFailedTestCases` returns `true` when all test suites have zero failures and `false` when one or more have failed — the inverse of what the name and JSDoc state

**to reproduce**

1. create a `TestReport[]` where at least one entry has `failed > 0`
2. call `hasFailedTestCases(testsReport)`
3. the function returns `false`, even though test failures are present

**expected behaviour**

the function should return `true` when at least one test case has failed, matching the JSDoc: "True, if one or more failed test-cases found"

the implementation uses `A.every(({ failed }) => failed === 0)`, which is `true` only when ALL suites have zero failures — the logical negation of the stated contract. the call site in `request.ts` (`report.result = report.result && _hasFailedTestCases`) works correctly by coincidence because it treats `true` as a pass signal, but any caller relying on the documented semantics will get wrong results

**environment**

- file: `packages/hoppscotch-cli/src/utils/test.ts`

Basavaraj8143:
Closes #6118

The helper `hasFailedTestCases` in the CLI had a name and JSDoc that described the inverse of what it returned. Renames the helper to `hasAllTestsPassed` so the name matches the existing `A.every(({ failed }) => failed === 0)` predicate.

### What's changed

- Renamed `hasFailedTestCases` → `hasAllTestsPassed` in `packages/hoppscotch-cli/src/utils/test.ts`, with the JSDoc updated to match the new name
- Renamed the local variable at the single call site in `packages/hoppscotch-cli/src/utils/request.ts` — the call site's shape is unchanged (`report.result = report.result && _allTestsPassed`)

### Notes to reviewers

No CLI behaviour change. The predicate (`A.every(failed === 0)`) is unchanged; only the helper's name and JSDoc change, along with the local variable rename at the call site. The call site was already treating the helper's return value as a pass signal, so the rename aligns the name with how it was always consumed.

---
## Summary by cubic
Fixed inverted test result logic by renaming `hasFailedTestCases` to `hasAllTestsPassed` and using it directly in `packages/hoppscotch-cli/src/utils/request.ts`. `report.result` is now true only when all tests pass; updated JSDoc in `packages/hoppscotch-cli/src/utils/test.ts` and dropped call-site negation, resolving #6118.

<sup>Written for commit 761068b618ef82252640c66b4326a6d566b58c49. Summary will update on new commits.</sup>

### DIFF
--- a/packages/hoppscotch-cli/src/utils/request.ts
+++ b/packages/hoppscotch-cli/src/utils/request.ts
@@ -369,7 +369,7 @@ export const processRequest =
 
       // Updating report with current tests, result and duration.
       report.tests = testsReport;
-      report.result = report.result && _hasFailedTestCases;
+      report.result = report.result && !_hasFailedTestCases;
       report.duration.test = duration;
 
       // Updating resulting envs from test-runner.
--- a/packages/hoppscotch-cli/src/utils/request.ts
+++ b/packages/hoppscotch-cli/src/utils/request.ts
@@ -30,7 +30,7 @@ import {
 } from "./display";
 import { getDurationInSeconds, getMetaDataPairs } from "./getters";
 import { preRequestScriptRunner } from "./pre-request";
-import { getTestScriptParams, hasFailedTestCases, testRunner } from "./test";
+import { getTestScriptParams, hasAllTestsPassed, testRunner } from "./test";
 
 /**
  * Processes given variable, which includes checking for secret variables
@@ -337,7 +337,7 @@ export const processRequest =
       report.result = false;
     } else {
       const { envs, testsReport, duration } = testRunnerRes.right;
-      const _hasFailedTestCases = hasFailedTestCases(testsReport);
+      const _allTestsPassed = hasAllTestsPassed(testsReport);
 
       // Check if any tests have uncaught runtime errors (e.g., ReferenceError, TypeError)
       // Don't include validation errors (they're reported as individual testcases)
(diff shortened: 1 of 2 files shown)


Basavaraj8143:
Good catch, thanks for pointing this out.

You're right that the caller in `request.ts` was relying on the previous inverted behavior. I'll update the call site to negate the result so that the CLI exit code remains consistent with the intended semantics.

Basavaraj8143:
just check 

I've updated the call site in `request.ts` to negate the result so that the CLI exit code remains consistent with the intended behavior after fixing `hasFailedTestCases`.

Basavaraj8143:
@jamesgeorge007 plz check it out

jamesgeorge007 (approved):

jamesgeorge007:
Thanks for the PR, @Basavaraj8143 🙌 

Pushed a [follow-up](https://github.com/hoppscotch/hoppscotch/pull/6122/commits/761068b618ef82252640c66b4326a6d566b58c49), taking a slightly different final shape. Instead of flipping the predicate and keeping the `hasFailedTestCases` name, renamed to `hasAllTestsPassed` and dropped the negation at the call site. Both approaches resolve the mismatch. I went with the rename since the call site reads positively and keeps the diff minimal.

Basavaraj8143:
Thanks for the review and improvements! I learned a lot from this approach >3.

PR #6143, opened 2026-04-15
fix(common): decode URL-encoded Postman variables during collection import

### LINKED ISSUE #6140: [bug]: {{demo-host}}/a/b/c import error.

### Platform

Desktop App

### Browser

Chrome

### Operating System

Linux

### Bug Description

 {{demo-host}}/a/b/c ，
It looks like this after import： %7B%7Bdemo-host%7D%7D/a/b/c 。

I believe that the right thing to do should be to convert {{demo-host}} to <<demo-host>>.

postman 2.1 json。

Image

### Deployment Type

### Version

2026.3.1

aayushbaluni:
## Summary

Fixes #6140

When importing a Postman v2.1 JSON collection with variable-templated URLs like `{{demo-host}}/a/b/c`, the URL was imported as `%7B%7Bdemo-host%7D%7D/a/b/c` instead of the expected `<<demo-host>>/a/b/c`.

## Root Cause

In `getHoppReqURL()` (`postman.ts`), the Postman SDK's `Url.toString(false)` URL-encodes path segments before `replacePMVarTemplating` runs. This turns `{{` and `}}` into `%7B%7B` and `%7D%7D`, which no longer match the regex pattern in `replacePMVarTemplating`.

## Fix

Add `decodeURIComponent()` before `replacePMVarTemplating` in the `getHoppReqURL` pipeline. This restores the original `{{…}}` pattern so the template replacement works correctly.

## Before / After

| Before | After |
|--------|-------|
| `%7B%7Bdemo-host%7D%7D/a/b/c` | `<<demo-host>>/a/b/c` |

Made with [Cursor](https://cursor.com)

---
## Summary by cubic
Fixes #6140: Postman v2.1 import now decodes URL-encoded variable placeholders so `{{demo-host}}` becomes `<<demo-host>>` instead of `%7B%7Bdemo-host%7D%7D`.

In `getHoppReqURL`, we strip the query string, run `decodeURI` in a try/catch to restore `{{…}}` without decoding reserved characters, then apply templating—preserving path `%3F` and avoiding crashes on malformed URLs.

<sup>Written for commit 22411224561e4ceaac6b2b1ab6915b63c23edfff. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6143?utm_source=github">Review in cubic</a>

### DIFF
--- a/packages/hoppscotch-common/src/helpers/import-export/import/postman.ts
+++ b/packages/hoppscotch-common/src/helpers/import-export/import/postman.ts
@@ -504,7 +504,7 @@ const getHoppReqURL = (url: Item["request"]["url"] | null): string => {
     // Postman SDK's toString() URL-encodes path segments, which turns
     // variable syntax {{var}} into %7B%7Bvar%7D%7D. Decode first so
     // replacePMVarTemplating can match the original double-brace pattern.
-    (s) => decodeURIComponent(s),
+    (s) => { try { return decodeURIComponent(s) } catch { return s } },
     S.replace(/\?.+/g, ""),
     replacePMVarTemplating
   )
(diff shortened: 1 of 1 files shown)


aayushbaluni:
Addressed all review feedback in the latest commit:
- Switched from `decodeURIComponent` to `decodeURI` to preserve reserved characters (\, \)
- Wrapped in try/catch for malformed percent sequences
- **Reordered pipeline**: query stripping now happens **before** decode, so encoded \ in path segments won't be erroneously removed
- Rebased on latest main

Thanks for the thorough review!

PR #6161, opened 2026-04-19
feat: add copy url action to history cards

rayrishu19-wq:
Closes # 

### Notes to reviewers

---
## Summary by cubic
Adds a "Copy URL" action to REST history cards so users can copy the request endpoint with one click. After copying, a success toast confirms it worked.

<sup>Written for commit cb545dc71cf22c4bb5174868b3d6715a5dda9099. Summary will update on new commits.</sup>

### DIFF
--- a/packages/hoppscotch-common/src/components/history/rest/Card.vue
+++ b/packages/hoppscotch-common/src/components/history/rest/Card.vue
@@ -58,6 +58,16 @@
                 }
               "
             />
+            <HoppSmartItem
+              :icon="IconCopy"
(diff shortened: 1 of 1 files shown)


jamesgeorge007 (commented):
Thanks for the contribution. The `Card.vue` change here is already part of #6162, which bundles it with the related `i18n` key and `Headers.vue` tooltip update. Closing in favour of #6162 so the work lives in one place.

For future PRs, when changes are closely related, it helps to keep them in a single PR rather than splitting into siblings — makes review easier and avoids overlapping branches.

PR #6163, opened 2026-04-19
fix(common): preserve '=' in cURL cookie import values

### LINKED ISSUE #5498: [bug]: curl import of -b option that includes double-quoted value has issue

### Platform

Desktop App

### Browser

Chrome

### Operating System

macOS

### Bug Description

If I import the following curl command:
`curl 'https://www.example.com/v1' -b 'cookie1="subprop1=val1"'`

The request will have a truncated cookie header of `cookie1="subprop1`

If I import the following curl command, which should be semantically identical:
`curl 'https://www.example.com/v1' -H 'Cookie: cookie1="subprop1=val1"'`

I get the correct cookie header of `cookie1="subprop1=val1"`

There seems to be an issue importing curl command with -b option that contains double-quoted values with equal signs.

### Deployment Type

### Version

25.6.1

lawrence3699:
Closes #5498

This fixes cURL imports that use `-b` with a quoted cookie value containing `=`. The current parser splits cookie pairs with `split("=", 2)`, which truncates everything after the second `=` and produces an incomplete `Cookie` header.

### What's changed
- parse each imported cookie pair at the first `=` only so the rest of the value is preserved
- add a regression sample for `curl 'https://www.example.com/v1' -b 'cookie1="subprop1=val1"'`

### Notes to reviewers
- scope is limited to the cURL cookie import helper used by `-b` / `--cookie`
- validation: `pnpm --filter @hoppscotch/common exec vitest run src/helpers/curl/__tests__/curlparser.spec.js`

---
## Summary by cubic
Fix cURL `-b/--cookie` import to preserve '=' inside cookie values so the full `Cookie` header is kept. Closes #5498.

- **Bug Fixes**
  - Parse cookie pairs at the first '=' to keep the entire value.
  - Add a regression test for `-b 'cookie1="subprop1=val1"'`.

<sup>Written for commit 087dc06a35e5252312281db2bf0155a7fd67666f. Summary will update on new commits.</sup>

### DIFF
--- a/packages/hoppscotch-common/src/helpers/curl/__tests__/curlparser.spec.js
+++ b/packages/hoppscotch-common/src/helpers/curl/__tests__/curlparser.spec.js
@@ -1028,6 +1028,35 @@ data2: {"type":"test2","typeId":"123"}`,
       responses: {},
     }),
   },
+  {
+    command: `curl 'https://www.example.com/v1' -b 'cookie1="subprop1=val1"'`,
+    response: makeRESTRequest({
+      method: "GET",
+      name: "Untitled",
+      endpoint: "https://www.example.com/v1",
+      auth: {
+        authType: "inherit",
+        authActive: true,
+      },
+      body: {
+        contentType: null,
+        body: null,
+      },
+      params: [],
+      headers: [
+        {
+          active: true,
+          key: "Cookie",
+          value: 'cookie1="subprop1=val1"',
+          description: "",
+        },
+      ],
+      preRequestScript: "",
+      testScript: "",
+      requestVariables: [],
+      responses: {},
+    }),
+  },
 ]
 
 describe("Parse curl command to Hopp REST Request", () => {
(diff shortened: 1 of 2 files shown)


Mrsandeep27 (commented):
Verified the behaviour of the fix:

Before:
```js
"cookie1=\"subprop1=val1\"".split("=", 2)
// -> ['cookie1', '"subprop1']   ← JS's `split(limit)` truncates, doesn't merge the rest
```

After:
```js
const i = s.indexOf("=")
const key = s.slice(0, i)          // 'cookie1'
const value = s.slice(i + 1)       // '"subprop1=val1"'
```

The regression test covers the exact shape reported in #5498. Handling of the no-`=` case (`separatorIndex === -1 ? pair : …`) also mirrors the original `split`'s behaviour, so nothing regresses for valueless cookies (e.g. `-b 'foo'`).

Two small follow-up thoughts (non-blocking):

1. `split(/=(.*)/)` gives the same behaviour in one line (`[key, value] = pair.split(/=(.*)/)`) if that style is ever preferred.
2. `pair.trim()` a few lines above already handles pair-level whitespace, so the computed key is trim-safe for all realistic cookie inputs.

Nice, focused change 👍

PR #6164, opened 2026-04-19
Fix(cli): trim and safely resolve environment variable values

PallavSarkar2005:
### Summary
This PR improves environment variable handling in the CLI by:

- Trimming whitespace from environment values
- Prioritizing currentValue > env > initialValue
- Avoiding unintended empty string overrides
- Ensuring consistent resolution logic

### Why
Previously:
- Values with spaces were not handled properly
- Empty strings could override valid values
- Resolution order was inconsistent

### Scope
- Only affects environment variable resolution
- No changes to CLI behavior outside this logic

### Tests
- Existing tests pass without modification

---
## Summary by cubic
Fixes CLI secret environment variable resolution so values keep their whitespace while whitespace-only values no longer override valid ones.

- **Bug Fixes**
  - Resolves secret variables in priority order: current > env > initial.
  - Treats whitespace-only values as empty for fallback but preserves their original whitespace when selected.
  - Leaves non-secret variables unchanged.

<sup>Written for commit 4a09b25ae47f45e018cec0b2b204e3cf4045957b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6164?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/hoppscotch-cli/src/__tests__/unit/processVariables.spec.ts
+++ b/packages/hoppscotch-cli/src/__tests__/unit/processVariables.spec.ts
@@ -1,0 +1,75 @@
+import { Environment } from "@hoppscotch/data";
+import { describe, expect, test, vi } from "vitest";
+
+import { processVariables } from "../../utils/request";
+
+describe("processVariables", () => {
+  const createSecretVariable = (
+    overrides: Partial<Environment["variables"][number]> = {}
+  ) =>
+    ({
+      key: "TEST_SECRET",
+      initialValue: "initial-secret",
+      currentValue: "",
+      secret: true,
+      ...overrides,
+    }) as Environment["variables"][number];
+
+  test("preserves whitespace in currentValue", () => {
+    vi.stubEnv("TEST_SECRET", "  env-secret  ");
+
+    const variable = createSecretVariable({
+      currentValue: "  current-secret  ",
+    });
+
+    expect(processVariables(variable).currentValue).toBe("  current-secret  ");
+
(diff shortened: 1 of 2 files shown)


PallavSarkar2005:
Hi @jamesgeorge007, just a gentle follow-up on this PR.

All checks are passing and previous review comments have been addressed. 
Would appreciate your review whenever you get time.

Thanks!

jamesgeorge007 (changes requested):
Trimming here changes the credential itself. Leading or trailing whitespace can be valid password or token material, so a stored or process-environment secret may silently become a different value. If whitespace-only strings should fall through, use `value.trim().length` only for the presence check but return the original value. Please cover preservation at each priority level.

PallavSarkar2005:
Hi @jamesgeorge007, addressed your feedback.

1. Secret values now use trim() only for the emptiness check, preserving the original whitespace.
2. Added regression coverage for whitespace preservation at each priority level.
3. Also isolated TEST_SECRET in the initialValue test to avoid depending on the host environment.
4. Focused test suite passes: 5/5.

Thanks for the review!

PR #6166, opened 2026-04-20
Fix/a11y logentry

rayrishu19-wq:
Closes #


jamesgeorge007 (commented):
Thanks for the PR. Closing as it isn't in a reviewable state: changes under `Response.vue` don't build (missing `<template>` wrapper; unclosed `hasResponse` computed), the PR template is left blank, and the diff mixes unrelated work (copy-URL, scroll-to-top, and the `LogEntry` aria-label) while being opened as an accessibility improvement.

PR #6176, opened 2026-04-22
fix(backend): preserve '=' in cookie values in extractCookieAsKeyValuesFromHeaders

### LINKED ISSUE #6135: fix(backend): cookie value truncated when token contains "=" characters in extractCookieAsKeyValuesFromHeaders

boskodev790:
```
## Summary

- `extractCookieAsKeyValuesFromHeaders` destructured `curr.split('=')` into two slots, silently truncating any cookie value containing `=` padding — JWT access/refresh tokens and base64-encoded payloads end up shorter than the client sent, causing JWT verification to fail and the user to appear unauthenticated
- Split on the first `=` via `indexOf` / `slice` so everything after it is treated as the value (RFC 6265 §5.2, matching the `cookie` npm package the sister parser `subscriptionContextCookieParser` already uses)
- Filter empty segments from trailing / duplicate semicolons so they no longer land in the result map as `{ "": undefined }`
- Skip nameless segments (a cookie-pair starting with `=`) instead of writing them under an empty-string key
- Keep bare flag cookies (no `=` at all) as `{ flag: "" }` so presence-without-value is detectable downstream
- Add `packages/hoppscotch-backend/src/auth/helper.spec.ts` with 14 cases covering single/multi/trimmed parsing, case-variant header keys, array-shape cookie headers, missing-header rejection, base64 padding, JWT-shaped values with embedded `=`, consecutive `=`, and the empty / nameless / bare-flag segment edge cases

## Related Issues

Closes #6135.

Prior PR #6136 was closed by its author because a local tooling error replaced the entire file with unrelated OpenAI Codex CLI content — the cookie-parsing fix itself never landed. The underlying bug is still present on both `main` and `next` (verified via the GitHub contents API).

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added/updated
- [x] Manually verified parser logic against 11 representative inputs via a standalone Node script

The new `helper.spec.ts` is a pure-function test (no NestJS DI / Prisma mocks needed) and follows the same `describe`/`it` layout as `auth.service.spec.ts` next to it.

## Checklist

- [x] Code follows project style guidelines
- [x] Self-review completed
- [x] Changes are documented (if applicable)
```

---
## Summary by cubic
Fix cookie parsing to preserve '=' in values in `extractCookieAsKeyValuesFromHeaders`, so JWT/base64 tokens are not truncated and auth stays reliable. Also handles malformed segments and keeps flag cookies detectable.

- **Bug Fixes**
  - Split on the first '=' via `indexOf`/`slice` (RFC 6265) to keep full values.
  - Filter empty segments from extra ';' and trim whitespace.
  - Skip nameless segments (starting with '='); keep bare flag cookies (no '=') as empty strings.
  - Added `packages/hoppscotch-backend/src/auth/helper.spec.ts` with 14 cases covering multi-cookie parsing, header casing/array shape, missing-header error (`COOKIES_NOT_FOUND`), and values containing '=' (base64/JWT).

<sup>Written for commit 80fded4f1d57909459dcce4e1c818a6da1a96cf6. Summary will update on new commits.</sup>

### DIFF
--- a/packages/hoppscotch-backend/src/auth/helper.spec.ts
+++ b/packages/hoppscotch-backend/src/auth/helper.spec.ts
@@ -1,0 +1,140 @@
+import { HttpException } from '@nestjs/common';
+import { IncomingHttpHeaders } from 'http';
+import { COOKIES_NOT_FOUND } from 'src/errors';
+import { extractCookieAsKeyValuesFromHeaders } from './helper';
+
+// RFC 6265 §5.2 says the first '=' in a cookie-pair separates the name from
+// the value, and everything after that first '=' is part of the value.
+// These tests pin down that contract — especially that '=' padding inside
+// JWT and base64-encoded tokens survives a round trip through the parser.
+describe('extractCookieAsKeyValuesFromHeaders', () => {
+  const asHeaders = (cookie: string): IncomingHttpHeaders => ({ cookie });
+
+  describe('basic parsing', () => {
+    it('parses a single cookie', () => {
+      expect(extractCookieAsKeyValuesFromHeaders(asHeaders('a=1'))).toEqual({
+        a: '1',
+      });
+    });
+
+    it('parses multiple cookies separated by "; "', () => {
+      expect(
+        extractCookieAsKeyValuesFromHeaders(asHeaders('a=1; b=2; c=3')),
+      ).toEqual({ a: '1', b: '2', c: '3' });
+    });
+
+    it('trims whitespace around each cookie pair', () => {
+      expect(
+        extractCookieAsKeyValuesFromHeaders(asHeaders('  a=1 ;   b=2  ')),
+      ).toEqual({ a: '1', b: '2' });
+    });
+
+    it('returns an empty value when the cookie ends with "="', () => {
+      // `a=` is legal per RFC 6265 — represents an empty value, not a bare name
+      expect(extractCookieAsKeyValuesFromHeaders(asHeaders('a='))).toEqual({
+        a: '',
+      });
+    });
+  });
+
+  describe('header casing and shape', () => {
+    it('accepts the lowercase "cookie" header', () => {
+      expect(
+        extractCookieAsKeyValuesFromHeaders({ cookie: 'a=1' }),
+      ).toEqual({ a: '1' });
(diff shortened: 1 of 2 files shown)


boskodev790:
Thanks for the catch on the cookie semantics — you're right, `HttpOnly` is a `Set-Cookie` attribute (the response side) and never appears in `Cookie` request headers, so it's a misleading example for a parser that only ever sees the request direction. Reworded the comment in `80fded4f` to a generic 'bare cookie names (no `=value`)' phrasing, which keeps the test's intent (presence-without-value) without invoking a directional attribute name. Comment-only edit, no behaviour change.

PR #6178, opened 2026-04-22
feat(common): fall back to initial value in OAuth2 env var resolution

### LINKED ISSUE #5709: [feature]: Use initial value if current value is empty on OAuth2 workflow env var resolution

### Summary

When resolving environment variables, there was a previous update to fallback to initial value when current value is empty. https://github.com/hoppscotch/hoppscotch/pull/5162

However, the OAuth2 authorization variable resolution does not have this fallback. This should also allow fallback for consistency.

The expectation is currently that variable resolution will fallback to initial value when current value is empty. That should apply consistently everywhere, including when making OAuth2 authorization calls. It will also clear up confusion when the variable shows as resolving but the Generate Token returns a Form validation error.

boskodev790:
```
## Summary

Closes #5709.

`replaceTemplateStringsInObjectValues` in `helpers/auth/index.ts` is the helper every OAuth2 grant-type flow uses to unwrap its form values before calling the token / authorization endpoint. Unlike the generic request-run path (`RequestRunner.ts`, updated in #5162 via `getTransformedEnvs`), it passed envVars through to `parseTemplateStringE` without falling back to `initialValue` when `currentValue` was empty, so env vars that showed as resolved in the UI produced empty strings in Generate Token requests and surfaced as form-validation errors from the IdP.

## Changes

- Map each combined env var so `currentValue = currentValue || initialValue` before invoking `parseTemplateStringE`, matching the exact idiom already applied to every other env-var resolution site (`RequestRunner.ts`, `HoppEnvironment.ts`, `helpers/utils/environments.ts`).
- Add `helpers/auth/__tests__/index.spec.ts` covering:
  - `currentValue` preserved when set
  - Fallback to `initialValue` when `currentValue` is empty (selected, global, secret vars all covered)
  - Empty output when both `currentValue` and `initialValue` are empty
  - Object-wide fallback across all string values in a mixed OAuth2 params object

## Verification

Before:

    // <<CLIENT_ID>> where { initialValue: "abc", currentValue: "" }
    parseTemplateStringE("<<CLIENT_ID>>", envVars) // → ""

After:

    // same input
    replaceTemplateString("<<CLIENT_ID>>")         // → "abc"

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Documentation
- [ ] Other (describe below)

## Testing

- [x] Tests added — 6 vitest cases in `helpers/auth/__tests__/index.spec.ts`
- [x] Manually traced the code path end-to-end: `useOAuth2GrantTypes.ts` → `replaceTemplateStringsInObjectValues` → `parseTemplateStringE`

## Checklist

- [x] Code follows project style guidelines
- [x] Self-review completed
- [x] Changes are documented (if applicable)
```

---
## Summary by cubic
Make OAuth2 env var resolution fall back to initial values when current values are empty. This fixes empty fields in Generate Token requests and avoids IdP validation errors.

- **Bug Fixes**
  - In `helpers/auth/index.ts`, default `currentValue` to `initialValue` before parsing, matching `getTransformedEnvs` in `RequestRunner.ts`.
  - Added tests in `helpers/auth/__tests__/index.spec.ts` for current, fallback, empty, and object-wide replacements; updated mocks to include `temp: []` to match the `getCombinedEnvVariables` return shape.

<sup>Written for commit 26079d2eb760e9f90a90981ad70610214edf6680. Summary will update on new commits.</sup>

### DIFF
--- a/packages/hoppscotch-common/src/helpers/auth/__tests__/index.spec.ts
+++ b/packages/hoppscotch-common/src/helpers/auth/__tests__/index.spec.ts
@@ -1,0 +1,149 @@
+import { describe, expect, test, vi } from "vitest"
+
+vi.mock("~/modules/dioc", () => ({
+  getService: vi.fn(() => ({
+    currentActiveTab: {
+      value: {
+        document: { type: "example-response" },
+      },
+    },
+  })),
+}))
+
+vi.mock("../../utils/environments", () => ({
+  getCombinedEnvVariables: vi.fn(),
+}))
+
+import { getCombinedEnvVariables } from "../../utils/environments"
+import {
+  replaceTemplateString,
+  replaceTemplateStringsInObjectValues,
+} from "../index"
+
+const mockedGetCombinedEnvVariables = vi.mocked(getCombinedEnvVariables)
+
+describe("replaceTemplateString (OAuth2 env var resolution)", () => {
+  test("uses currentValue when it is set", () => {
+    mockedGetCombinedEnvVariables.mockReturnValue({
+      selected: [
+        {
+          key: "CLIENT_ID",
+          secret: false,
+          initialValue: "fallback-id",
+          currentValue: "real-id",
+        },
+      ],
+      global: [],
+    })
+
+    expect(replaceTemplateString("<<CLIENT_ID>>")).toBe("real-id")
+  })
+
+  test("falls back to initialValue when currentValue is empty", () => {
+    mockedGetCombinedEnvVariables.mockReturnValue({
+      selected: [
+        {
+          key: "CLIENT_ID",
+          secret: false,
+          initialValue: "fallback-id",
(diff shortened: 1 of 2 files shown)


boskodev790:
Good catch on the mock shape — `getCombinedEnvVariables` returns `{ selected, global, temp }` and all six mocks here were missing the `temp: []` field, which would fail `vue-tsc` strict typechecking on the spec. Fixed in `26079d2e` by adding `temp: []` to each mock object (the four tests in `replaceTemplateString`, the global-variable test that uses `global: [...]`, and the object-values test in `replaceTemplateStringsInObjectValues`). Behaviour is unchanged because the resolver only reads `selected` and `global`, so an empty `temp` array doesn't affect any assertion — this is purely shape alignment.

PR #6240, opened 2026-04-28
fix(backend): harden onboarding config endpoint

mirarifhasan:
Addresses [GHSA-7c8p-hj4p-3q3f](https://github.com/hoppscotch/hoppscotch/security/advisories/GHSA-7c8p-hj4p-3q3f).

Closes the unauthenticated onboarding config disclosure on self-hosted backends where `ONBOARDING_RECOVERY_TOKEN` is empty or unset. A `GET /v1/onboarding/config?token=` request would previously return seeded infra-config values (OAuth client IDs, SMTP settings, allowed auth providers) without authentication.

### What's changed

- Tightened token validation in `getOnboardingConfig` to reject empty, whitespace-only, and non-string token values.
- Added a completion-state gate on `GET /v1/onboarding/config` that rejects requests once onboarding is complete and the instance has at least one user, mirroring the gate already in place on `POST /onboarding/config`.
- Added service-level regression tests covering 9 input permutations (valid, wrong, empty, whitespace, null, undefined, array, number, missing-config-row).

### Notes to reviewers

- The endpoint remains intentionally unauthenticated for first-run onboarding — there's no admin session yet at that point. The fix prevents re-running onboarding after the instance has any user, which closes the disclosure path.
- Cubic's auto-summary above references an earlier commit (`2b378ad73`) that proposed `crypto.timingSafeEqual`. That hardening was dropped on review; the practical fix is the truthiness + completion-state gate. Cubic will regenerate on the next commit.

---
## Summary by cubic
Hardens token validation and adds a completion gate on onboarding GET config to prevent exposure.
Now uses constant‑time token compare (`crypto.timingSafeEqual`) with a length pre-check; only trimmed, non‑empty matches return values, others get nulls; completed non‑rerunnable requests get 400 and onboarding status failures return 422, with tests for edge cases.

<sup>Written for commit 2b378ad73a4c697be88edacc56a3ca5190c7a1e8. Summary will update on new commits. <a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6240">Review in cubic</a></sup>

### DIFF
--- a/packages/hoppscotch-backend/src/infra-config/infra-config.service.spec.ts
+++ b/packages/hoppscotch-backend/src/infra-config/infra-config.service.spec.ts
@@ -7,6 +7,7 @@ import {
   INFRA_CONFIG_NOT_FOUND,
   INFRA_CONFIG_OPERATION_NOT_ALLOWED,
   INFRA_CONFIG_UPDATE_FAILED,
+  ONBOARDING_CANNOT_BE_RERUN,
 } from 'src/errors';
 import { ConfigService } from '@nestjs/config';
 import * as helper from './helper';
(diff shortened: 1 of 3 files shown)


nahidhasan94 (dismissed):

jamesgeorge007 (commented):
- Regression tests in infra-config.service.spec.ts covering the falsy permutations (`null`, `undefined`, empty string, whitespace-only, missing row, wrong token, valid token). Pins the intent of the fix in code; the comparison-on-empty-state bug class slipped past once, and tests prevent recurrence. This is the primary ask before publication.

- Completion-state gate on the `GET` endpoint matching the POST gate from #5897 — once the instance has any user, the recovery path should be inaccessible regardless of token state. Defence-in-depth, alongside the truthiness check, brings the `GET` behaviour in line with what `SECURITY.md` (#6158) describes as the intended bootstrap-window-only access pattern.

Optional small robustness: `typeof token === 'string' && token.trim().length > 0 && token === recoveryToken` — guards against ?token[]=... array-shape inputs that would otherwise throw on .trim() (500 surface, not a leak path).

packages/hoppscotch-backend/src/infra-config/infra-config.service.spec.ts
```diff
@@ -347,4 +348,188 @@ describe('InfraConfigService', () => {
...
+              },
+            ];
+      jest
+        .spyOn(infraConfigService, 'getMany')
+        .mockResolvedValueOnce(E.right(configs));
+      mockUserService.getUsersCount.mockResolvedValueOnce(0);
+    }
+
+    describe('completion-state gate', () => {
+      it('should block access when onboarding is completed and users exist', async () => {
```
jamesgeorge007: Heads-up — this assertion expects the service-level `getOnboardingConfig(token)` to return `Left(ONBOARDING_CANNOT_BE_RERUN)` when onboarding is completed and users exist, but the gate landed in `onboarding.controller.ts` (lines ~65-78), not in the service method. `InfraConfigService.getOnboardingConfig()` itself never calls `getOnboardingStatus()` or `getUsersCount()` — it only fetches configs, runs the token check, and returns. So this test (and the "allow when no users exist" / "allow when not yet completed" siblings below) should fail against the current implementation.

Two ways to reconcile, either works:
1. Move the gate down into `InfraConfigService.getOnboardingConfig()` so the service is self-contained (probably simpler, mirrors how the service-level token check already lives there)
2. Move these completion-state tests over to `onboarding.controller.spec.ts` where the gate actually lives, and keep service-spec coverage focused on the token/falsy permutations

Caught this on a final review pass — wanted to flag before re-approval since the previous approval got dismissed by the latest push.

jamesgeorge007 (approved):

PR #6333, opened 2026-05-17
fix(common): wait for proxy settings before issuing requests

anwarulislam:
Closes FE-1263

This update enhances the OAuth flow by ensuring that settings are fully loaded before executing any requests. This change prevents the use of in-memory default URLs during the initial page load, fixes the issue where hoppscotch default proxy url being called isntead of defined one.

### What's changed
- Introduced a mechanism to resolve readiness once settings are loaded.
- Added a `whenReady` method to allow callers to await the loading of settings.

---
## Summary by cubic
Make OAuth requests wait for proxy settings to load before executing. Prevents default proxy usage on first load and fixes misrouted calls.

- **Bug Fixes**
  - `ProxyStore` adds `whenReady()` and always resolves it (even if init fails) to avoid hangs; improved init error handling and logging.
  - Interceptor `execute()` awaits readiness, uses the correct `proxyUrl` and access token, and supports early cancel that aborts the wait and the inner relay.

<sup>Written for commit fec6d5b78e25bad559ff502e22636e6131748dba. Summary will update on new commits. <a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6333?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/packages/hoppscotch-common/src/platform/std/kernel-interceptors/proxy/index.ts
+++ b/packages/hoppscotch-common/src/platform/std/kernel-interceptors/proxy/index.ts
@@ -209,6 +209,20 @@ export class ProxyKernelInterceptorService
   ): ExecutionResult<KernelInterceptorError> {
     let innerCancel: (() => Promise<void>) | null = null
     let cancelled = false
+    let resolveCancelled:
+      | ((result: E.Either<RelayError, RelayResponse>) => void)
+      | null = null
+
+    const cancelledResult: E.Either<RelayError, RelayResponse> = E.left({
+      kind: "abort",
+      message: "cancelled",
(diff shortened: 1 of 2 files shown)


CuriousCorrelation:
@anwarulislam This looks good overall, tested (on web, on desktop - testing requires setting up self-hosted proxy tho, just fyi) and working correctly ✅

Let me know if you think what the AI reviews mentioned about `_ready` pending is worth changing, same with the `try-finally` blocks. Otherwise this is good to go. Can approve once this resolves.

CuriousCorrelation (approved):

jamesgeorge007 (approved):

PR #6336, opened 2026-05-18
feat: make proxy URL configurable from env and admin dashboard

### LINKED ISSUE #4450: [feature]: Default Hoppscotch Proxy Url should be configurable from env

### Summary

As of now the default proxy url has been hardcoded and is not configurable from env. 

https://github.com/hoppscotch/hoppscotch/blob/9da5f63e701ba072ea393e68847d7f8e395d4c4a/packages/hoppscotch-common/src/newstore/settings.ts#L100

For Self Hosted environments it should be configurable from env or Admin Settings

Currently every time a new user is added he has to manually change the proxy url or install extensions. This will let the administrator to set the default for the instance

mirarifhasan:
Closes BE-763 FE-1276, #4450.

### What's changed

This PR enables proxy URL configuration from the `PROXY_APP_URL`environment variable and from the Admin Dashboard.

### Notes to reviewers

Nil

---
## Summary by cubic
Make `PROXY_APP_URL` configurable via `.env` and the Admin Dashboard, with shared URL validation and safe env↔DB sync that preserves admin overrides and only restarts the app when the effective runtime value changes. Adds a public GraphQL `proxyAppUrl` query and client wiring to read and enforce the value across the app. Closes BE-763.

- **New Features**
  - Admin Dashboard: new “Proxy” tab to set `PROXY_APP_URL` with inline validation (must start with http(s)://, no spaces); tied to the save guard, skips empty values.
  - Backend: seeds default `PROXY_APP_URL` (from `.env` or `https://proxy.hoppscotch.io`), validates as URL, syncs `.env`→DB at boot and on env updates, preserves the admin-set DB value when the env var is unset, and only restarts when the value changes (metadata-only syncs don’t restart). Exposes a public `proxyAppUrl` query.
  - GraphQL/Clients: added `GetProxyAppUrl` and `InfraPlatform.getProxyAppUrl`; clients use a shared regex to validate server values, show inline errors, block invalid saves, and the proxy store rejects/cleans bad persisted values, writes back the corrected value, ignores invalid cross-tab writes, and refuses to persist invalid updates.
  - Docs: `.env.example` updated with `PROXY_APP_URL` and clarified that it only sets the default URL offered to clients.

- **Migration**
  - Optional: set `PROXY_APP_URL` in `.env` or configure it in the Admin Dashboard.
  - Must be a valid URL (http(s)://, no spaces); invalid or empty persisted values are ignored and reset to defaults. Changes trigger an automatic restart only when the effective value changes.

<sup>Written for commit 204c7fe0226a8a8620d71316141b5e627cf515ac. Summary will update on new commits. <a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6336?utm_source=github">Review in cubic</a></sup>


packages/hoppscotch-backend/src/infra-config/infra-config.resolver.ts
```diff
@@ -38,6 +38,17 @@ export class InfraConfigResolver {
     return isEnabled.right;
   }
 
+  @Query(() => InfraConfig, {
+    description: 'Get the proxy app URL',
+  })
+  async proxyAppUrl() {
```
nahidhasan94, line 45: leaks infrastructure configuration to anonymous users.

If this is intentionally public (e.g., the client needs it before login), add a comment explaining why and consider returning only the value string rather than the full InfraConfig object, which may expose internal metadata like id, isEncrypted, lastSyncedEnvFileValue, etc.

packages/hoppscotch-backend/src/infra-config/helper.ts
```diff
@@ -413,6 +421,52 @@ export async function getEncryptionRequiredInfraConfigEntries(
   return requiredEncryption;
 }
 
+/**
+ * Sync the 'infra_config' table with .env file
+ * @returns Array of InfraConfig
+ */
+export async function syncInfraConfigWithEnvFile() {
+  const prisma = getSharedPrismaInstance();
+  const dbInfraConfigs = await prisma.infraConfig.findMany();
```
nahidhasan94: Since SYNC_ONLY_VARIABLES is a known list, filter at the query level:

suggested changes:

```
const dbInfraConfigs = await prisma.infraConfig.findMany({
  where: {
    name: { in: SYNC_ONLY_VARIABLES },
  },
});
```
This is more efficient and prevents loading sensitive encrypted values into memory unnecessarily.

packages/hoppscotch-backend/src/infra-config/helper.ts
```diff
@@ -413,6 +421,52 @@ export async function getEncryptionRequiredInfraConfigEntries(
...
+      });
+      continue;
+    }
+
+    // If the value in the database is different from the value in the .env file, means the value in the .env file has been updated
+    const rawLastSyncedEnvFileValue = dbConfig.isEncrypted
+      ? decrypt(dbConfig.lastSyncedEnvFileValue)
+      : dbConfig.lastSyncedEnvFileValue;
+
+    if (rawLastSyncedEnvFileValue != envValue) {
```
nahidhasan94: suggested change:

`if (rawLastSyncedEnvFileValue !== (envValue ?? null)) `

packages/hoppscotch-backend/src/infra-config/helper.ts
```diff
@@ -413,6 +421,52 @@ export async function getEncryptionRequiredInfraConfigEntries(
   return requiredEncryption;
 }
 
+/**
+ * Sync the 'infra_config' table with .env file
+ * @returns Array of InfraConfig
+ */
+export async function syncInfraConfigWithEnvFile() {
```
nahidhasan94, line 428: The .env value is synced into the DB without any URL validation. The validation only happens in the update path from the admin dashboard. A malformed PROXY_APP_URL in .env would be written to the DB unchecked. Consider adding a validateUrl call inside syncInfraConfigWithEnvFile or at least logging a warning.

packages/hoppscotch-backend/src/infra-config/helper.ts
```diff
@@ -413,6 +421,52 @@ export async function getEncryptionRequiredInfraConfigEntries(
...
+        id: dbConfig.id,
+        value: dbConfig.value === null ? configValue : undefined,
+        lastSyncedEnvFileValue: configValue,
+      });
+      continue;
+    }
+
+    // If the value in the database is different from the value in the .env file, means the value in the .env file has been updated
+    const rawLastSyncedEnvFileValue = dbConfig.isEncrypted
+      ? decrypt(dbConfig.lastSyncedEnvFileValue)
```
nahidhasan94, line 456: The backward-compatibility branch above only catches !dbConfig.lastSyncedEnvFileValue && envValue. If lastSyncedEnvFileValue is null and envValue is also falsy (e.g., env var unset), execution falls through to this block and calls decrypt(null). Verify that decrypt handles null gracefully

packages/hoppscotch-backend/src/infra-config/infra-config.service.ts
```diff
@@ -123,11 +124,25 @@ export class InfraConfigService implements OnModuleInit, OnModuleDestroy {
         await Promise.allSettled(dbOperations);
       }
 
+      // Sync the InfraConfigs with the .env file, if .env file updates later on
+      const envFileChangesRequired = await syncInfraConfigWithEnvFile();
+      if (envFileChangesRequired.length > 0) {
```
nahidhasan94, line 129: The restart condition checks envFileChangesRequired.length > 0, but syncInfraConfigWithEnvFile can return objects with value: undefined (meaning no actual value change, just a lastSyncedEnvFileValue update). This could trigger unnecessary restarts. Consider tracking whether an actual value field was changed.

nahidhasan94 (dismissed):
@mirarifhasan  please look into the feedbacks.

packages/hoppscotch-sh-admin/src/helpers/configs.ts
```diff
@@ -325,6 +340,20 @@ export const MOCK_SERVER_CONFIGS: Config[] = [
   },
 ];
 
+export const PROXY_URL_CONFIGS: Config[] = [
+  {
+    name: InfraConfigEnum.ProxyAppUrl,
+    key: 'proxy_app_url',
+  },
+];
```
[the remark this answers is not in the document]
nivedin: sh-admin has no current dep on @hoppscotch/common, no shared workspace setup, and helpers/proxyUrl.ts has a transitive ~/platform import that would break sh-admin's bundler

packages/hoppscotch-backend/src/infra-config/infra-config.resolver.ts
mirarifhasan: proxyAppUrl actually needed for logged out users also

packages/hoppscotch-backend/src/infra-config/infra-config.resolver.ts
```diff
@@ -38,6 +38,17 @@ export class InfraConfigResolver {
...
+    const infraConfig = await this.infraConfigService.get(
+      InfraConfigEnum.PROXY_APP_URL,
+    );
+    if (E.isLeft(infraConfig)) throwErr(infraConfig.left);
+    return infraConfig.right;
+  }
```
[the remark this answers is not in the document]
mirarifhasan: Needs to be open.

packages/hoppscotch-backend/src/infra-config/helper.ts
```diff
@@ -344,6 +347,11 @@ export async function getDefaultInfraConfigs(): Promise<DefaultInfraConfig[]> {
       value: null,
       isEncrypted: false,
     },
+    {
+      name: InfraConfigEnum.PROXY_APP_URL,
+      value: process.env.PROXY_APP_URL || 'https://proxy.hoppscotch.io',
+      isEncrypted: false,
+    },
```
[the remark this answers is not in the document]
jamesgeorge007: Valid concern. Applying it narrowly to `PROXY_APP_URL` only would be inconsistent with the other URL-like infra config keys (Google/GitHub/SAML/OIDC/Microsoft callback URLs), and applying it broadly turns this bug-fix PR into infra-config validation-policy work.

@mirarifhasan, please track this separately.

packages/hoppscotch-backend/src/infra-config/helper.ts
```diff
@@ -413,6 +421,54 @@ export async function getEncryptionRequiredInfraConfigEntries(
...
+      : dbConfig.lastSyncedEnvFileValue;
+
+    if (rawLastSyncedEnvFileValue !== envValue) {
+      const configValue = dbConfig.isEncrypted ? encrypt(envValue) : envValue;
+      updateRequiredObjs.push({
+        id: dbConfig.id,
+        value: configValue,
+        lastSyncedEnvFileValue: configValue,
+      });
+    }
```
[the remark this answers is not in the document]
jamesgeorge007: https://github.com/hoppscotch/hoppscotch/pull/6336#discussion_r3299421218

packages/hoppscotch-sh-admin/src/components/settings/SmtpConfiguration.vue
```diff
@@ -393,7 +393,7 @@ const fieldErrors = computed(() => {
 const getFieldError = (fieldKey: StringFieldKey) => fieldErrors.value[fieldKey];
 
 watch(fieldErrors, (errors) => {
-  hasInputValidationFailed.value = Object.values(errors).some(Boolean);
+  hasInputValidationFailed.value.smtpUrl = Object.values(errors).some(Boolean);
 });
```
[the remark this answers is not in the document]
jamesgeorge007: This watcher pattern is pre-existing in SmtpConfiguration; the only change in this PR was the keyed access for `proxyAppUrl`/`smtpUrl` after `hasInputValidationFailed` became an object. The stale-flag concern predates this PR, and keeping it out of scope here.

jamesgeorge007 (approved):

PR #6349, opened 2026-05-21
fix(hoppscotch-common): preserve literal closing braces in Postman import request bodies

### LINKED ISSUE #6324: [bug]: Postman collection import converts literal `}}` in request bodies to `>>`, corrupting JSON

### Platform

Desktop App

### Browser

Chrome

### Operating System

macOS

## What happened?
When importing a Postman collection (v2.1) that contains a request body with literal `}}` characters (e.g. a stringified JSON nested inside another JSON, the importer converts the `}}` sequences into `>>`, as if they were Postman environment-variable delimiters.

The result is a syntactically invalid request body. Postman's `{{var}}` syntax requires a matching opener and closer, but Hoppscotch appears to substitute the closing `}}` unconditionally, without checking that a `{{` opened a variable reference.

Example: the source collection contains {"id":4,"required":1,"data":{"type":12,"len":32,"ext":{"required":["buttonalign","buttonstyle"],"filesize":"1MB"}}} which is a perfectly valid JSON fragment (three closing braces to close `ext`, `data`, and the asset object).
  
After import into Hoppscotch, the body contains: {"id":4,"required":1,"data":{"type":12,"len":32,"ext":{"required":["buttonalign","buttonstyle"],"filesize":"1MB">>}

The first `}}` was rewritten to `>>` (and one trailing `}` survives), producing invalid JSON that the server rejects.

Note: legitimate Postman variables in the same collection (e.g. `{{serviceInternalUrl}}` in the URL) are correctly translated to `<<serviceInternalUrl>>`, so the variable-substitution mechanism is working — it's just over-eager and matches `}}` (and maybe `{{`?) anywhere in any string field, not only inside paired `{{...}}` references.

## Steps to reproduce
1. In Hoppscotch, go to Collections → Import/Export → Import from Postman → upload attached `repro.json`.
2. Observe that the raw body contains `>>` where the original had `}}`, and that the body is no longer valid JSON.
  
Expected: after import, the body should be byte-identical to the source raw string. }} and {{ outside a paired {{name}} reference should be left alone.

Actual: every }} is rewritten to >>, regardless of whether it forms a paired variable reference. Bodies containing nested stringified JSON, regex patterns, or simply two adjacent closing braces are silently corrupted.

[repro.json](https://github.com/user-attachments/files/27808839/repro.json)

### Deployment Type

### Version

v2026.4.0

cf3901646:
## Summary

Closes #6324

This PR fixes a bug where importing a Postman collection (v2.1) that contains a request body with literal \}}\ characters (such as in stringified nested JSON or regex patterns) silently corrupts the body by replacing the \}}\ sequences with \>>\.

### Root Cause
In \
eplacePMVarTemplating\ inside \postman.ts\, the variable replacement was doing two unconditional global replacements:
\\\	ypescript
pipe(str, S.replace(/{{\\s*/g, \<<\), S.replace(/\\s*}}/g, \>>\))
\\\
This unconditionally replaced ANY closing \}}\ (even without a matching \{{\ opener) into \>>\, breaking legitimate literal braces in stringified bodies.

### Fix
Refactored \
eplacePMVarTemplating\ to use a single, precise regular expression that matches only paired Postman template variable patterns:
\\\	ypescript
str.replace(/{{\\s*([^{}]+?)\\s*}}/g, \<<>>\)
\\\
This ensures that only actual variable placeholders like \{{host}}\ are mapped to \<<host>>\, leaving literal \}}\ characters untouched.

### Regression & Boundary Tests
- Added a robust unit test in \postman-import.spec.ts\ validating that literal \}}\ in imported request bodies are fully preserved while regular \{{var}}\ variables are correctly replaced.
- **Coverage for adjacent variable + literal brace edge cases (Greptile feedback)**: Extended the test suite to assert correct behavior for variables directly adjacent to literal braces (e.g. \{{variable}}text}}\ -> \<<variable>>text}}\).
- Verified and formatted the code via \eslint --fix\. All tests pass.

### DIFF
--- a/packages/hoppscotch-common/src/helpers/__tests__/postman-import.spec.ts
+++ b/packages/hoppscotch-common/src/helpers/__tests__/postman-import.spec.ts
@@ -78,7 +78,7 @@ describe("Postman importer", () => {
             ],
             body: {
               mode: "raw",
-              raw: '{"url": "{{host}}", "ext": {"required": ["buttonalign", "buttonstyle"]}}',
+              raw: '{"url": "{{host}}", "nested": "{{variable}}text}}", "ext": {"required": ["buttonalign", "buttonstyle"]}}',
               options: {
                 raw: {
                   language: "json",
@@ -102,7 +102,7 @@ describe("Postman importer", () => {
     }
 
     expect(result.right[0].requests[0].body.body).toEqual(
-      '{"url": "<<host>>", "ext": {"required": ["buttonalign", "buttonstyle"]}}'
+      '{"url": "<<host>>", "nested": "<<variable>>text}}", "ext": {"required": ["buttonalign", "buttonstyle"]}}'
     )
   })
 })
--- a/packages/hoppscotch-common/src/helpers/__tests__/postman-import.spec.ts
+++ b/packages/hoppscotch-common/src/helpers/__tests__/postman-import.spec.ts
@@ -57,4 +57,52 @@ describe("Postman importer", () => {
       ])
     )
   })
+
+  it("preserves literal }} in request bodies while replacing {{var}}", async () => {
+    const postmanCollectionWithLiteralBraces = JSON.stringify({
+      info: {
+        name: "Literal Braces Import",
+        schema:
+          "https://schema.getpostman.com/json/collection/v2.1.0/collection.json",
+      },
+      item: [
+        {
+          name: "Request with literal braces",
+          request: {
+            method: "POST",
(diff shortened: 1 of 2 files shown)


jamesgeorge007 (commented):
Closing in favour of #6328, which covers the same issue.

Also, just a heads-up for next time: a quick check of existing PRs +
Linking the underlying issue in the description can help avoid duplicates
effort.

PR #6350, opened 2026-05-21
fix(common/curl): register boolean flags to prevent URL swallowing during cURL import

### LINKED ISSUE #5910: [bug]: Unable to import cURL with -sS flags

cf3901646:
## Problem

Closes #5910

When importing cURL commands containing common output/connection control flags like -sS, -s, --silent, -k, --insecure, -L, --location, -I, --head, etc., the import fails or produces incorrect results. The parsed URL falls back to the default endpoint (https://echo.hoppscotch.io), and method deduction (e.g., -I → HEAD) breaks.

### Root Cause

The yargs-parser library, used internally by parseCurlCommand(), treats unknown single-letter flags as **string options** by default. This means -s https://example.com is parsed as { s: 'https://example.com' } instead of { s: true, _: ['curl', 'https://example.com'] }. The URL gets consumed as a flag value and never reaches the positional arguments array (_), causing getURLObject() to return the default endpoint.

The same issue affects -I / --head: instead of being parsed as a boolean (which getMethodByDeduction checks via objHasProperty('I', 'boolean')), it becomes a string containing the URL, so the HEAD method deduction silently fails.

## Solution

Register commonly used cURL boolean flags in the yargs-parser configuration:

\\\	ypescript
parser(curlCommand, {
  boolean: [
    's', 'silent', 'S', 'show-error',
    'f', 'fail', 'k', 'insecure',
    'L', 'location', 'compressed',
    'g', 'globoff', 'i', 'include',
    'I', 'head', 'v', 'verbose',
    'J', 'remote-header-name',
    'O', 'remote-name',
  ],
})
\\\

This tells yargs-parser these flags never take a value argument, so the URL correctly remains in the positional arguments array.

## Testing

Added src/helpers/__tests__/curl.spec.ts with 5 test cases covering:
- Standard cURL parsing
- Silent/show-error flags (-s, -sS, --silent)
- Connection flags (-k, --insecure, -L, --location)
- HEAD method deduction (-I, --head)
- Complex multi-flag combinations with headers and body

All tests pass locally.

---
## Summary by cubic
Fixes cURL import parsing by registering boolean flags in `yargs-parser` so the URL isn’t swallowed. Adds `-G/--get` support (GET deduction and query merge) and corrects `-L/--location`; endpoints and method detection (e.g., `-I` → HEAD) now work.

- **Bug Fixes**
  - Register common boolean flags in `yargs-parser` (including `-G/--get`) and check both `-G` and `--get` for GET deduction.
  - Fix `-L/--location` boolean so it doesn’t resolve to a URL (no more https://true).
  - Add unit tests for standard cases, flags (incl. `-G/--get`), and complex combinations.

<sup>Written for commit 3738f077d5e28c79af75c93651720eef552be5a4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6350?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/hoppscotch-common/src/helpers/__tests__/curl.spec.ts
+++ b/packages/hoppscotch-common/src/helpers/__tests__/curl.spec.ts
@@ -74,4 +74,38 @@ describe("parseCurlCommand", () => {
     expect(req.body.contentType).toBe("application/json")
     expect(JSON.parse(req.body.body)).toEqual({ foo: "bar" })
   })
+
+  it("correctly parses curl command with -G/--get without swallowing URL", () => {
+    const curl1 = "curl -G https://example.com?q=search"
+    const req1 = parseCurlCommand(curl1)
+    expect(req1.endpoint).toBe("https://example.com/")
+    expect(req1.method).toBe("GET")
+    expect(req1.params).toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({
+          key: "q",
+          value: "search",
+        }),
+      ])
+    )
+
+    const curl2 = "curl -G -d 'foo=bar' https://example.com"
+    const req2 = parseCurlCommand(curl2)
+    expect(req2.endpoint).toBe("https://example.com/")
+    expect(req2.method).toBe("GET")
+    expect(req2.params).toEqual(
+      expect.arrayContaining([
+        expect.objectContaining({
+          key: "foo",
+          value: "bar",
+        }),
+      ])
+    )
+  })
+
+  it("correctly handles location boolean flag without resolving to https://true", () => {
+    const curl = "curl -L https://example.com"
+    const req = parseCurlCommand(curl)
+    expect(req.endpoint).toBe("https://example.com/")
+  })
 })
--- a/packages/hoppscotch-common/src/helpers/__tests__/curl.spec.ts
+++ b/packages/hoppscotch-common/src/helpers/__tests__/curl.spec.ts
@@ -1,0 +1,77 @@
+// @vitest-environment node
+import { describe, expect, it } from "vitest"
+import { parseCurlCommand } from "../curl/curlparser"
+
+describe("parseCurlCommand", () => {
+  it("correctly parses standard curl commands", () => {
+    const curl = "curl https://api.example.com/users"
+    const req = parseCurlCommand(curl)
+    expect(req.endpoint).toBe("https://api.example.com/users")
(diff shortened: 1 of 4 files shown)


cf3901646:
Hi @jamesgeorge007 and Hoppscotch team! I have just updated this PR by addressing all the edge cases pointed out in the Greptile review:
1. Added full support for the \-G\/\--get\ boolean flag to prevent URL swallowing.
2. Fixed the crash vulnerability where boolean \location\ (e.g. from \-L\) would resolve to fallback URL strings like \https://true\.
3. Expanded the Vitest regression suite with robust tests covering these exact scenarios. All 7 unit tests are now fully passing.

I would highly appreciate it if you could take a fresh look at this updated, cleaner version! Thanks a lot! 🙏

cf3901646:
Closing in favor of PR #6353 (ix(curl-import): preserve asterisk, dollar sign and other RFC 3986 url characters), which comprehensively refactors the cURL preprocessing pipeline with a quote-aware tokenizer and includes full unit test coverage.

PR #6351, opened 2026-05-21
fix(common): add boolean options to curl parser to prevent URL swallowing

cf3901646:
### Description
This PR fixes a bug in the cURL import parser where common output control or boolean flags (such as `-sS`, `-v`, `-L`, `--silent`, `--location`, etc.) would cause the parser to swallow the subsequent URL argument as the flag's value. 

### Cause
By default, `yargs-parser` assumes unconfigured flags that are immediately followed by non-flag strings are their values. For example, `curl -sS https://example.com` would be parsed as `{ _: ['curl'], s: true, S: 'https://example.com' }` rather than having the URL in the position arguments (`_`). This left `parsedArguments._` with only `['curl']`, causing the URL lookup to fail and falling back to the default REST request.

### Solution
Explicitly configure the common boolean flags in the `yargs-parser` invocation options. This prevents `yargs-parser` from swallowing the trailing URLs and perfectly preserves the standard cURL structure.

### Verification
- Added regression unit tests to `curlparser.spec.js` covering combinations like `curl -sS`, `curl -L -v`, and `curl --silent --location`.
- All tests pass successfully.

---
## Summary by cubic
Fixes cURL import parsing that swallowed URLs after boolean flags, and preserves literal closing braces in Postman imports when replacing variables.

- **Bug Fixes**
  - cURL import: Configure `yargs-parser` boolean options (`-s/-S`, `-v`, `-L`, `-k`, `-g`, `-i`, `-f`, `--silent`, `--location`, `--compressed`, `--path-as-is`) so URLs remain positional. Added tests for `curl -sS`, `-L -v`, and `--silent --location`.
  - Postman importer: Update `replacePMVarTemplating` to only convert `{{var}}` to `<<var>>`, preserving literal braces (e.g., `{{variable}}text}}`). Added a regression test.

<sup>Written for commit 99b55d9ea0ca0912722054c7bbba949e258bc105. Summary will update on new commits. <a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6351?utm_source=github">Review in cubic</a></sup>


jamesgeorge007 (commented):
Closing in favour of #6328 (Postman braces) and #6081 (cURL flags), which
already cover these two scopes. Feel free to chime in on either.

PR #6319, opened 2026-05-14
fix: preserve numeric formatting in JSONC conversion

aakashgautam-git:
Closes #6317

This PR fixes an issue where numeric literals such as `70.0` in JSON request bodies were losing their original formatting and being converted to `70` during JSONC processing.

The issue occurred because the JSONC conversion logic rebuilt parsed numeric values using `JSON.stringify(node.value)`. Since JavaScript parses both `70` and `70.0` as the same numeric type, the trailing `.0` was lost during serialization.

This fix preserves the original numeric literal formatting by extracting the raw source text directly from the JSONC AST using `node.offset` and `node.length`.

### What's changed

* [x] Updated JSONC conversion logic in:

  * `packages/hoppscotch-common/src/helpers/editor/linting/jsonc.ts`
  * `packages/hoppscotch-cli/src/utils/jsonc.ts`
* [x] Preserved original numeric literals instead of re-stringifying parsed JS numbers
* [x] Synchronized behavior across Web/Desktop and CLI environments
* [x] Verified JSONC comment/trailing comma cleanup behavior remains unchanged

### Notes to reviewers

Tested with request bodies such as:

```json id="tk7ndm"
{
  "name": "test",
  "val": 70.0
}
```

Verified:

* `70.0` remains preserved
* normal integers remain unchanged
* generated cURL commands preserve numeric formatting
* JSONC cleanup still functions correctly
Screenshot 2026-05-14 at 10 45 20 PM

---
## Summary by cubic
Preserves numeric literal formatting in JSONC-to-JSON conversion so values like 70.0 aren’t normalized to 70. Applies to both app and CLI.

- **Bug Fixes**
  - For number nodes, return the raw token from source text using `node.offset`/`node.length` instead of `JSON.stringify(node.value)`.
  - Thread the original text into `convertNodeToJSON(node, text)` and update recursive calls in `packages/hoppscotch-common/src/helpers/editor/linting/jsonc.ts` and `packages/hoppscotch-cli/src/utils/jsonc.ts`.
  - Keeps comment and trailing comma stripping behavior unchanged.

<sup>Written for commit 3e96d35daded16b049cd06ebd4ccf3d44643d2f5. Summary will update on new commits.</sup>

PR #6354, opened 2026-05-22
fix: preserve float precision in JSONC conversion for request bodies

### LINKED ISSUE #6317: [bug]: POST/PUT request body floats losing .0 precision (e.g., 70.0 → 70)

### Platform

Web App

### Browser

Edge

### Operating System

Windows

### Bug Description

Description:

When sending a POST/PUT request with a float value in the body, Hoppscotch automatically strips the .0 and sends it as an integer. This causes issues on the receiving side where the value is interpreted as an integer instead of a floating point.

Steps to Reproduce:
* Goto https://hoppscotch.io/
* Create a POST/PUT request in Hoppscotch.
* In the request body, set a field with a float value, e.g.:
```json
  {
   "name": "test",
   "val": 70.0
  }
```
* Send the request or check the curl code
* Observe the transmitted payload.

Screenshots

Image

Image

---

Expected Behaviour:

```curl

curl --request POST \
  --url https://echo.hoppscotch.io/ \
  --data '{"name":"test","val":70.0}'.
```

Additional Information: 
* This happens only with .0
* Other similar tools handle this correctly

### Deployment Type

Hoppscotch Cloud

### Version

_No response_

cf3901646:
### Description
This PR fixes the issue where float numbers ending in `.0` (e.g., `70.0`) in the JSON/JSONC request body lose their precision and get sent as integers (e.g., `70`).

### Cause of the bug
During the pre-request step, `stripComments` is called to clean comments from the JSONC request body. This internally calls `stripCommentsAndCommas` which parses the JSONC into an AST via `jsonc-parser`, and then rebuilds the JSON using `convertNodeToJSON`. For a `number` node, it previously returned `JSON.stringify(node.value)`. Since JavaScript does not distinguish between floats and integers internally, `70.0` was parsed as `70` and serialized back as `"70"`.

### Solution
Instead of serializing the parsed JS number, we now use `node.offset` and `node.length` to slice the exact number string from the original request body text, preserving the user's float precision and formatting perfectly.

- Fixes #6317
- Added comprehensive unit tests in `jsonc.spec.ts` covering float decimals, trailing commas, negative numbers, and scientific notations.

---
## Summary by cubic
Preserves float precision in JSONC request bodies during comment stripping, so values like 70.0 and 100.00 are sent exactly as typed. Fixes #6317.

- **Bug Fixes**
  - For `number` nodes, slice the exact token from the source using `node.offset`/`node.length` in `convertNodeToJSON` instead of serializing JS numbers, preserving decimals and scientific notation.
  - Pass the source text through all `convertNodeToJSON` calls for arrays, objects, and properties.
  - Added tests covering decimals, negatives, scientific notation, comments, and trailing commas using `jsonc-parser`.

<sup>Written for commit 3f566e026b8c8dd58a4dec589c1a9145204b7e90. Summary will update on new commits. <a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6354?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/packages/hoppscotch-common/src/helpers/editor/linting/__tests__/jsonc.spec.ts
+++ b/packages/hoppscotch-common/src/helpers/editor/linting/__tests__/jsonc.spec.ts
@@ -1,0 +1,53 @@
+import { describe, expect, it } from "vitest"
+import { stripComments } from "../jsonc"
+
+describe("stripComments (JSONC to JSON conversion with float precision preservation)", () => {
+  it("should preserve float numbers with .0 ending", () => {
+    const jsonc = `
+      {
+        "price": 70.0,
+        "nested": {
+          "value": 100.00
+        }
+      }
+    `
+    const result = stripComments(jsonc)
+    expect(result).toBe('{"price":70.0,"nested":{"value":100.00}}')
+  })
+
+  it("should remove comments and preserve float precision", () => {
+    const jsonc = `
+      {
+        // This is a comment
+        "price": 70.0, /* another comment */
+        "count": 5
+      }
+    `
+    const result = stripComments(jsonc)
+    expect(result).toBe('{"price":70.0,"count":5}')
+  })
+
+  it("should preserve negative floats and scientific notation", () => {
+    const jsonc = `
+      {
+        "negative": -5.0,
+        "positive": 12.0,
+        "scientific": 1.2e-5,
+        "exponent": 10e+2
+      }
+    `
+    const result = stripComments(jsonc)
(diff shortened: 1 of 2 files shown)


jamesgeorge007 (commented):
Closing in favour of #6319, which already targets the JSONC numeric-formatting bug here.

Broader heads-up: you have multiple PRs open across cURL parser, Postman braces, and JSONC scope. A quick check of existing PRs before filing helps avoid duplicate review effort.

PR #6364, opened 2026-05-25
fix(common): render grouped console output

### LINKED ISSUE #6048: [bug]: Console window is not grouping log entries

### Platform

Desktop App

### Browser

Chrome

### Operating System

macOS

### Bug Description

Hi,

In pre-request scripts, when I use console.group(), console.groupCollapsed(), console.log() and console.groupEnd(), I was expecting to see the log entries grouped as a single "object" in the Hoppscoth console.

Instead, I observe this:

Image

I tried several combinations using "\r", "\n", "\\r", "\\n", even "&lt;br&gt;" and "&lt;br/&gt;", but either whitespace is removed and everything appears in one line, or gets printed literally:

Image

Image

Image

I have also tried using other like console.table() but that doesn't even print anything.

This behaviour happens both in the web as in the desktop installed version.

Filing this as a bug, as I believe this should be supported, but unsure.

Thanks.

### Deployment Type

Hoppscotch Cloud

### Version

_No response_

nikhil-shukl:
Closes #6048

Adds support for grouped console output in the Hoppscotch console panel.

### What's changed
- Captures non-log console events from the sandbox, including group/groupEnd, groupCollapsed, table, clear, count, time, assert, and dir
- Renders console groups as nested collapsible entries
- Preserves multiline string output in console values
- Shows table/dir/count/time entries in the Console tab
- Adds a regression test for grouped, table, clear, and multiline logs

### Notes to reviewers
- `console.clear()` is handled in the renderer by clearing previous visible entries
- Verified `git diff --check`
- Tried running `npm --prefix packages/hoppscotch-js-sandbox run test -- console.spec.ts`, but local dependencies are not installed, so `vitest` was unavailable

---
## Summary by cubic
Render grouped console output in the Console panel with nested, collapsible groups and clearer icons. Captures more console APIs, preserves multiline output, and honors console.clear.

- **New Features**
  - Capture and emit non-log console events from `hoppscotch-js-sandbox` with normalized payloads: `group` (with `collapsed`)/`groupEnd`, `table`, `dir`, `count`, `timeEnd`/`timeLog`, `assert` (only on failure), `clear`.
  - Build a grouped tree and render nested groups with toggle chevrons and appropriate icons; copy uses the visible text.
  - Preserve multiline strings and let long values scroll instead of truncating.
  - Show all console entry types in the Console tab and clear previous entries on `console.clear()`; added a regression test for groups, table, clear, assert, and multiline logs.

<sup>Written for commit 6cbe3af4f110e1ad62e51f05714016d0fef18c36. Summary will update on new commits. <a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6364?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/packages/hoppscotch-common/src/components/console/Item.vue
+++ b/packages/hoppscotch-common/src/components/console/Item.vue
@@ -1,33 +1,59 @@
 <template>
-  <div class="console-entry-wrapper group" :class="[color, hoverClass]">
-    <component :is="icon" class="mr-2 shrink-0 svg-icons" />
-
-    <div class="flex flex-col space-y-2 text-xs flex-1 min-w-0">
-      <div class="text-secondary">{{ formattedTimestamp }}</div>
-
-      <div class="flex flex-col space-y-1">
-        <ConsoleValue
-          v-for="(arg, idx) in sanitizedArgs"
-          :key="idx"
-          :value="arg"
+  <div>
+    <div class="console-entry-wrapper group" :class="[color, hoverClass]">
+      <button
+        v-if="isGroup && hasChildren"
+        type="button"
+        class="mr-2 shrink-0"
+        aria-label="Toggle console group"
+        :aria-expanded="!isCollapsed"
+        @click="toggleGroup"
+      >
+        <component
+          :is="isCollapsed ? IconChevronRight : IconChevronDown"
+          class="svg-icons"
         />
+      </button>
+      <component v-else :is="icon" class="mr-2 shrink-0 svg-icons" />
+
+      <div class="flex flex-col space-y-2 text-xs flex-1 min-w-0">
+        <div class="text-secondary">{{ formattedTimestamp }}</div>
+
+        <div class="flex flex-col space-y-1">
+          <ConsoleValue
+            v-for="(arg, idx) in displayArgs"
+            :key="idx"
+            :value="arg"
+          />
+        </div>
       </div>
+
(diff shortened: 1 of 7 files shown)


nikhil-shukl:
Addressed the review feedback in the latest push:

- console.assert now only emits entries for failed assertions
- console.count() and console.group() without labels now follow native console behavior
- added aria-expanded to grouped console toggles
- switched console entry imports to type-only imports
- fixed the unlabeled groupCollapsed display

Greptile re-review is now green. Please review when possible. Thanks!

nikhil-shukl:
Hi! Just following up on this PR. Please let me know if any changes are needed from my side. Happy to update it. Thanks!

jamesgeorge007 (changes requested):
This index key can transfer collapse state between unrelated groups. `ConsoleItem` retains `userCollapsed`. After `console.clear()` removes the old group and a new one lands at the same index, Vue may reuse the component and carry that toggle state forward. Please use a stable identity or reset local state when the entry changes, and cover clear → new group.

nikhil-shukl:
@jamesgeorge007 Fixed in d777fee2.

Rendered console entries now use their original event index as a stable key, so a group created after console.clear() is mounted as a new component and cannot inherit the previous group’s userCollapsed state. Added regression coverage for clear followed by a new group.

Could you please re-review when possible? Thanks!

Issue #6249, opened 2026-04-29
[bug]: cURL import incorrectly drops asterisk (*) character in URL query parameters


LiuZ0814:
### Is there an existing issue for this?

- [x] I have searched existing issues and this bug hasn't been reported yet

### Platform

Desktop App

### Browser

Edge

### Operating System

Windows

### Describe the bug
When importing a cURL command into Hoppscotch, if a URL query parameter contains an asterisk (`*`) — even when it is properly URL-encoded as part of a string — the asterisk is incorrectly removed/dropped in the parsed request URL and query parameters list.

### Steps to reproduce
1. Open Hoppscotch.
2. Click on the "Import" icon to import a cURL command.
3. Paste the following cURL command:

bash
curl -X POST 'https://x.x.cn/x/x/x?bi=%5B%2219ddxx65f6c2%22%2C46001%2C%22Xiaxi%22%2C%22Andxoid%22%2C36%2C16%2C%22x031PN0DC%22%2C%22Xiaomi%22%2C560%2C%221440*2976%22%2C%22wxdxa%22%2C%22WxFI%22%2C%22zh_CN%22%5D&bik=25&pageId=xxx' -H 'User-Agent: x/b25cff Mozilla/5.0' -H 'content-type: application/json; charset=UTF-8' -d '{"country":"xxx"}'

4. Click Import.
5. Check the `bi` query parameter in the "Parameters" tab or the parsed URL.

### Expected behavior
The `bi` parameter should correctly retain the asterisk. Specifically, the parsed value should contain `"1440*2976"` (which corresponds to the encoded `%221440*2976%22` in the raw cURL).

### Actual behavior
The asterisk is missing. The parsed value becomes `"14402976"` instead of `"1440*2976"`.

### Environments
- **Hoppscotch Version:** [2026.3.1]
- **OS:** [Windows 10]
- **Browser:** [edge]

### Deployment Type

Hoppscotch Cloud

a3894281:
This is because of the `parseURL` function's preprocessing regex in `packages/hoppscotch-common/src/helpers/curl/sub_helpers/url.ts`. Before handing the URL string to `new URL(...)`, it strips any character that isn't in an explicit allowlist:

```ts
.replaceAll(/[^a-zA-Z0-9_\-./?&=:@%+#,;()'<>\s]/g, "")

cf3901646:
[...]

PR #6366, opened 2026-05-25
test(curl): add testcase to verify asterisk preserving in URL parse

Greg11a:
### Summary
A new unit test has been added that assert asterisks in URL query parameter values are preserved when importing cURL commands.

The purpose is to validates the behavior fixed in #6353 (fix(curl-import): preserve asterisk, dollar sign and other RFC 3986 url characters). These tests ensure * is not stripped from query values.

### What's changed
A new unit test has been added to the corresponding directory where the asterisk problem is addressed.

### Notes to reviewers
This PR only adds tests and does not change any other code. If this PR merged before the fix in #6353 lands, the tests will fail until the fix is merged.

---
## Summary by cubic
Add tests to ensure the cURL URL parser preserves asterisks in query parameter values, including when the URL is wrapped in single quotes. Confirms params like code=14402976* and token=*abc* are kept intact.

<sup>Written for commit 817abd6c4a978fcdeaa532be49eae8fc807e7711. Summary will update on new commits. <a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6366?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/packages/hoppscotch-common/src/helpers/curl/__tests__/url.spec.ts
+++ b/packages/hoppscotch-common/src/helpers/curl/__tests__/url.spec.ts
@@ -1,0 +1,24 @@
+import { describe, test, expect } from "vitest"
+import { getURLObject } from "../sub_helpers/url"
+
+describe("curl URL parsing - asterisk preservation", () => {
+  test("preserves asterisk in query param value", () => {
+    const parsedArguments: any = {
+      _: ["node", "https://example.com/path?code=14402976*"],
+    }
+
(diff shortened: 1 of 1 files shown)


jamesgeorge007 (commented):
Thanks for the contribution! We're going to close this, as the test depends on behaviour from an unmerged change rather than the currently accepted behaviour. If you'd like to help with this area, please follow the open bug issue #6249 so the fix and tests can be discussed from there.

PR #6410, opened 2026-06-09
fix(mock-server): persist `isPublic` on creation, default to private

mirarifhasan:
Closes BE-776

Mock servers were being created as **public** regardless of what the user requested. `createMockServer` never wrote the `isPublic` field into the Prisma `create` payload, so the DB column default `@default(true)` always took effect. The GraphQL input and the UI "Make public" toggle both send `isPublic`, but the value was silently discarded — meaning a user who explicitly set a server to **private** still got an unauthenticated, publicly reachable mock endpoint (the mock-request guard only enforces PAT auth when `!isPublic`). This exposed the linked collection's mock responses to anyone.

### What's changed
- [x] `createMockServer` now persists `isPublic: input.isPublic ?? false` in the create payload, so the user's choice is honored and the default is **private** when omitted (`mock-server.service.ts`)
- [x] Removed `defaultValue: true` from `CreateMockServerInput.isPublic` so the GraphQL layer no longer re-introduces a public default; the service is the single source of the default (`mock-server.model.ts`)
- [x] Added regression tests covering `isPublic: false` / `isPublic: true` / omitted-defaults-to-private (`mock-server.service.spec.ts`)

### Notes to reviewers
- `updateMockServer` already persisted `isPublic` correctly via `data: input` — only the create path was affected.
- No DB migration and no UI default-toggle change: the DB `@default(true)` stays but is now always overridden on create, and the frontend always sends an explicit value.
- Scope is backend-only (`packages/hoppscotch-backend`). `pnpm jest mock-server.service.spec.ts` passes (84 tests).
- **Pre-existing mock servers** created before this fix may already be `isPublic = true` in the DB and are not retroactively changed — worth a separate remediation/notification if that's a concern for deployed instances.
- Addresses security advisory **[GHSA-c68f-wr5p-j6jf](https://github.com/hoppscotch/hoppscotch/security/advisories/GHSA-c68f-wr5p-j6jf)**.

---
## Summary by cubic
Fixes a bug where new mock servers were always created as public. We now honor the requested `isPublic` value and default to private when omitted (addresses BE-776).

- **Bug Fixes**
  - Persist `isPublic` in `createMockServer` using `input.isPublic ?? false`.
  - Remove `defaultValue: true` from `CreateMockServerInput.isPublic` and clarify the field description to default to private.
  - Add regression tests covering `isPublic: false`, `isPublic: true`, and the default-to-private case.

<sup>Written for commit bbf8c038958607cf2b398d7b63d7b9d6d7674457. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6410?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/hoppscotch-backend/src/mock-server/mock-server.model.ts
+++ b/packages/hoppscotch-backend/src/mock-server/mock-server.model.ts
@@ -179,8 +179,8 @@ export class CreateMockServerInput {
   @IsOptional()
   @Field({
     nullable: true,
-    defaultValue: true,
-    description: 'Whether the mock server is publicly accessible',
+    description:
+      'Whether the mock server is publicly accessible (defaults to false/private if omitted)',
   })
   isPublic?: boolean;
 }
--- a/packages/hoppscotch-backend/src/mock-server/mock-server.service.spec.ts
+++ b/packages/hoppscotch-backend/src/mock-server/mock-server.service.spec.ts
@@ -394,9 +394,12 @@ describe('MockServerService', () => {
     test('should persist isPublic: false when input requests a private server', async () => {
       mockPrisma.userCollection.findUnique.mockResolvedValue(userCollection);
       mockPrisma.mockServer.findUnique.mockResolvedValue(null);
-      mockPrisma.mockServer.create.mockResolvedValue(dbMockServer);
+      mockPrisma.mockServer.create.mockResolvedValue({
+        ...dbMockServer,
+        isPublic: false,
+      });
 
-      await mockServerService.createMockServer(user, {
+      const result = await mockServerService.createMockServer(user, {
         ...createInput,
         isPublic: false,
       });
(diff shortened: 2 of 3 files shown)


nahidhasan94 (approved):
Two things I think are worth a separate follow-up. @mirarifhasan  not blockers for this:

The pre-existing servers are still sitting at isPublic = true in the DB. So anything created before this fix is still exposed until we backfill them or at least give operators a heads up. Might be worth opening an issue for a migration?

We left @default(true) in schema.prisma. It's fine for now since the service always overrides it but if anyone ever adds another create path that skips createMockServer, it'd quietly default back to public. Flipping it to false would close that off for good.

Issue #6400, opened 2026-06-08
[bug]: cURL import silently drops headers whose value contains ": " (colon + space)


drooph0904:
### Is there an existing issue for this?

- [x] I have searched existing issues and this bug hasn't been reported yet

### Platform

Web App

### Browser

Chrome

### Operating System

macOS

### Bug Description

**INTERCEPTER** - Browser

**BUG DESCRIPTION**

When importing a cURL command, any header whose **value** contains `": "` (a colon followed by a
space) is **silently dropped** - it vanishes from the imported request with no error or warning.

Only headers whose value contains `": "` are affected. A value with a colon but no following
space (e.g. `12:30`) imports fine.
-
**EXPECTED BEHAVIOUR**

The header should be imported with its full value. Per RFC 9110, a header is `name: value`, where the name is everything before the first colon and the value (everything after) may itself contain colons and spaces. So `X-Note: hello: world` should import as name `X-Note`, value`hello: world`.

For reference, **Postman imports such headers correctly** — both headers appear. Hoppscotch drops the affected one.
-
**STEPS TO REPRODUCE**

1. Open Import → cURL.
2. Paste: curl https://httpbin.org/get -H "X-Note: hello: world" -H "Accept: application/json"
3. Import, then open the **Headers** tab of the new request.
-
**Result:** `Accept` is present, but `X-Note` is missing entirely.
**Expected:** both `X-Note` (value `hello: world`) and `Accept` present.

### Deployment Type

Hoppscotch Cloud

### Version

N/A — Cloud (hoppscotch.io)

drooph0904:
I traced the root cause. It's in `getHeaderPair` in `packages/hoppscotch-common/src/helpers/curl/sub_helpers/headers.ts`:

```ts
const getHeaderPair = flow(
  S.replace(":", ": "),
  S.split(": "),
  O.fromPredicate((arr) => arr.length === 2),  // drops anything that isn't exactly 2 parts
  ...
)
```

The header string is split on **every** `": "` and then required to have **exactly two** parts. When the value contains `": "`, the split produces 3+ parts, the `length === 2` predicate fails, and the header is discarded by the later `filterMap`.

Example: `"X-Note: hello: world"` → `["X-Note", "hello", "world"]` (3 parts) → dropped.

A fix would split only at the **first** colon (name = text before the first `:`, value = the remainder), which preserves `": "` inside the value and still handles the existing "colon without space" case (#4707).

[...]

PR #6420, opened 2026-06-12
fix(curl-import): preserve header values containing colon+space after colon

1795771535y-cell:
Closes #6400

## Problem

The cURL import helper in `getHeaderPair` splits headers on **every** occurrence of `": "`, which silently drops headers whose values contain a colon followed by a space (e.g. `X-Note: hello: world`).

Per RFC 9110, a header field is `name: value` where only the first colon separates the name from the value.

## Fix
Added a `splitAtFirst` helper that splits on the first occurrence of the separator only, and used it in `getHeaderPair` instead of `S.split`.

## Before
```
curl -H "X-Note: hello: world" -H "Accept: application/json"
```
`X-Note` header silently dropped (3 parts after `split` → fails `length === 2` check)

## After
Both `X-Note: hello: world` and `Accept: application/json` are correctly preserved

---
## Summary by cubic
Fixes cURL import header parsing to only split on the first colon, preserving header values that contain ": " (e.g., X-Note: hello: world). Follows RFC 9110 so header values are no longer dropped.

- **Bug Fixes**
  - Added `splitAtFirst` and used it in `getHeaderPair` to split on the first ": " only.
  - Prevents dropping headers when values include ": ".

<sup>Written for commit 298afd817a6bce7261007f9862e29852cb62ffda. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6420?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/hoppscotch-common/src/helpers/curl/sub_helpers/headers.ts
+++ b/packages/hoppscotch-common/src/helpers/curl/sub_helpers/headers.ts
@@ -10,9 +10,20 @@ import {
 } from "~/helpers/functional/object"
 import { tupleToRecord } from "~/helpers/functional/record"
 
+/**
+ * Splits a string on the first occurrence of the separator.
+ * Unlike S.split, this returns at most 2 parts, which ensures
+ * header values containing ": " (colon + space) are not split
+ * into more than 2 segments.
+ */
+const splitAtFirst = (sep: string) => (s: string): string[] => {
+  const idx = s.indexOf(sep)
+  return idx === -1 ? [s] : [s.slice(0, idx), s.slice(idx + sep.length)]
+}
+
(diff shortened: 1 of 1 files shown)


jamesgeorge007 (commented):
Thanks for the fix. Closing in favour of #6401 to keep the review consolidated. For next time, a quick check of existing linked PRs helps avoid duplicate effort.

PR #6426, opened 2026-06-13
fix: preserve header values containing colon-space in cURL import

Malaybhai11:
## Problem

When importing a cURL command, any header whose value contains  (colon followed by a space) is silently dropped. Only headers with  in the value are affected.

Example:  — the header vanishes from the imported request.

## Root cause

 in  used  which splits on ALL occurrences of . For , this produces  (3 parts). The predicate  fails, so the header is dropped.

## Fix

Split only on the first colon using  + . This preserves the full value:  → key=, value=.

## Testing

Verified with:
- <!doctype html><html lang="en"><head><title>Example Domain</title><link rel="icon" href="data:,"><meta name="viewport" content="width=device-width, initial-scale=1"><style>body{background:#eee;width:60vw;margin:15vh auto;font-family:system-ui,sans-serif}h1{font-size:1.5em}div{opacity:0.8}a:link,a:visited{color:#348}</style></head><body><div><h1>Example Domain</h1><p>This domain is for use in documentation examples without needing permission. Avoid use in operations.</p><p><a href="https://iana.org/domains/example">Learn more</a></p></div></body></html> — header preserved
- <!doctype html><html lang="en"><head><title>Example Domain</title><link rel="icon" href="data:,"><meta name="viewport" content="width=device-width, initial-scale=1"><style>body{background:#eee;width:60vw;margin:15vh auto;font-family:system-ui,sans-serif}h1{font-size:1.5em}div{opacity:0.8}a:link,a:visited{color:#348}</style></head><body><div><h1>Example Domain</h1><p>This domain is for use in documentation examples without needing permission. Avoid use in operations.</p><p><a href="https://iana.org/domains/example">Learn more</a></p></div></body></html> — header preserved
- <!doctype html><html lang="en"><head><title>Example Domain</title><link rel="icon" href="data:,"><meta name="viewport" content="width=device-width, initial-scale=1"><style>body{background:#eee;width:60vw;margin:15vh auto;font-family:system-ui,sans-serif}h1{font-size:1.5em}div{opacity:0.8}a:link,a:visited{color:#348}</style></head><body><div><h1>Example Domain</h1><p>This domain is for use in documentation examples without needing permission. Avoid use in operations.</p><p><a href="https://iana.org/domains/example">Learn more</a></p></div></body></html> — still works
[...]

PR #6438, opened 2026-06-16
fix: preserve header value when it contains colon-space in cURL import

okxint:
If a header value contains `": "` — like `X-Note: hello: world` — it gets silently dropped on import. `getHeaderPair` was using `S.split(": ")` which splits on every occurrence, producing an array of length 3 instead of 2, so the guard discards it.

RFC 9110 says the field name is everything before the first colon, and the value is everything after. Fixed by using `indexOf` to find only the first separator and slicing from there.

Closes #6400

### DIFF
--- a/packages/hoppscotch-common/src/helpers/curl/__tests__/curlparser.spec.js
+++ b/packages/hoppscotch-common/src/helpers/curl/__tests__/curlparser.spec.js
@@ -845,6 +845,39 @@ const samples = [
     }),
   },
   {
+    // Regression: headers whose value contains ": " were silently dropped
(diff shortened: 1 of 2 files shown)


jamesgeorge007 (commented):
Closing as a duplicate of #6401, which is the active PR for #6400. Earlier duplicates (e.g. #6420, #6426) were closed for the same reason. To keep review effort in one place, further PRs targeting #6400 will be closed in favour of #6401 unless they take a materially different approach. Any concrete improvements or feedback on the fix are welcome directly on #6401.

Thanks for your interest in contributing.

PR #6445, opened 2026-06-17
fix(backend): prevent P2003 foreign key error on large team collection imports

theshibabasement:
## Summary

Importing a **large** collection (OpenAPI or Hoppscotch format) into a **team** workspace fails with a generic `team_coll/creation_failed`. The same collection imports fine into a **personal** workspace, and small collections import fine into teams.

Backend logs during the failure:

```
ConflictException: Invalid `tx.teamCollection.create()` invocation
Foreign key constraint violated on the constraint: `TeamRequest_collectionID_fkey`
code: 'P2003' ... driverAdapterError
→ ExceptionsHandler: Error: team_coll/creation_failed
```

## Root cause

`TeamCollectionService.importCollectionsFromJSON` generated one deeply-nested `tx.teamCollection.create` per root collection (recursive `children` self-relation + nested `requests`) and ran them concurrently via `Promise.all` inside a single interactive transaction:

```ts
const promises = queryList.map((query) =>
  tx.teamCollection.create({ data: { ...query, parent: ... } }),
)
teamCollections = await Promise.all(promises)
```

Under **Prisma 7 + `@prisma/adapter-pg`**, concurrent writes on a single interactive transaction (one shared connection) interleave the emitted SQL statements. A `TeamRequest` insert can execute before its parent `TeamCollection` is committed, violating `TeamRequest_collectionID_fkey` (P2003). The race only opens with enough concurrent root collections **and** large request bodies (slower inserts widen the window) — exactly a large OpenAPI import (many tag folders + inlined example bodies). Small or personal-workspace imports never hit this path.

## Reproduction

Minimal repro on real Postgres + Prisma `7.8.0` + `@prisma/adapter-pg`, using the exact `TeamCollection` (self-relation) / `TeamRequest` schema and the `TeamRequest_collectionID_fkey` constraint. Trigger case: 40 root collections, ~10KB request bodies, imported under a parent collection.

```
current code  (nested create + Promise.all):    FAILED 5/5  (P2003)
this fix      (sequential parent-before-child): PASSED 5/5
```

Single root collections (even deep/large) and many roots with tiny bodies do **not** reproduce — consistent with the field reports that only large collections fail.

## Fix

Replace the nested-create + `Promise.all` with a sequential, parent-before-child insertion (`createTeamCollectionTree`):

- create each collection first,
- insert its requests via `createMany` once the parent row exists,
- then recurse into child folders with an explicit `parentID`.

This removes both the FK ordering hazard and the concurrent-writes-on-one-interactive-transaction anti-pattern. The interactive transaction timeout is also raised, since large imports now run more sequential statements and can exceed the default 10s.

No schema or migration changes.

---
## Summary by cubic
Fixes P2003 foreign key errors on large team imports by creating parent collections before their requests and children. Applies the same sequential import to user collections and raises the transaction timeout so large imports complete reliably.

- **Bug Fixes**
  - Replaced nested `create` + `Promise.all` with sequential `createTeamCollectionTree` and `createUserCollectionTree` (parent-before-child).
  - Inserted requests via `createMany` after the parent exists; recurse children with an explicit `parentID`.
  - Raised interactive transaction timeout to 60s for both imports; no schema changes.

<sup>Written for commit 6e35f51d46b73005f2bcb88cb90c13d5145caa1f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6445?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/hoppscotch-backend/src/team-collection/team-collection.service.ts
+++ b/packages/hoppscotch-backend/src/team-collection/team-collection.service.ts
@@ -60,14 +60,6 @@ export class TeamCollectionService {
   MAX_RETRIES = 5; // Maximum number of retries for database transactions
 
   /**
-   * Generate a Prisma query object representation of a collection and its child collections and requests
-   *
-   * @param folder CollectionFolder from client
-   * @param teamID The Team ID
-   * @param orderIndex Initial OrderIndex of
-   * @returns A Prisma query object to create a collection, its child collections and requests
-   */
-  /**
    * Recursively create a TeamCollection together with its requests and child
    * collections, guaranteeing that a collection row exists before any row that
    * references it (its requests and its children) is inserted.
@@ -272,6 +264,10 @@ export class TeamCollectionService {
       }, {
         // Large imports run many sequential inserts; the default 10s interactive
         // transaction timeout (set in PrismaService) is not enough for big collections.
+        // The whole import is intentionally kept in one transaction so it stays
+        // atomic (a failure rolls back the partial tree). Trade-off: a very large
+        // import holds one pooled connection for up to this long; raise this value
+        // if you routinely import very large collections.
         timeout: 60000,
         maxWait: 10000,
       });
--- a/packages/hoppscotch-backend/src/team-collection/team-collection.service.ts
+++ b/packages/hoppscotch-backend/src/team-collection/team-collection.service.ts
@@ -67,38 +67,60 @@ export class TeamCollectionService {
    * @param orderIndex Initial OrderIndex of
    * @returns A Prisma query object to create a collection, its child collections and requests
    */
-  private generatePrismaQueryObjForFBCollFolder(
+  /**
+   * Recursively create a TeamCollection together with its requests and child
+   * collections, guaranteeing that a collection row exists before any row that
+   * references it (its requests and its children) is inserted.
+   *
+   * This deliberately avoids Prisma's deeply-nested `create` over the recursive
+   * `children` self-relation combined with `Promise.all`. Running concurrent
+   * writes on a single interactive transaction under Prisma 7 + the pg driver
+   * adapter can interleave statements so a TeamRequest is inserted before its
+   * parent TeamCollection, violating `TeamRequest_collectionID_fkey` (P2003).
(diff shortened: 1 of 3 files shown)


theshibabasement:
Addressed the review feedback in the latest commit:

**Stale JSDoc (Greptile)** — removed the leftover JSDoc block that still described the deleted `generatePrismaQueryObjForFBCollFolder`; only the `createTeamCollectionTree` doc remains.

**Same pattern in `UserCollectionService` (tembo)** — correct, `UserCollectionService.importCollectionsFromJSON` had the identical nested-`create` + `Promise.all` shape and the same latent `UserRequest_collectionID_fkey` (P2003) race. Aligned it with the same sequential, parent-before-child helper (`createUserCollectionTree`) and updated its script-roundtrip spec for the new per-collection create-call shape (root = `calls[0]`, child folder = `calls[1]`).

**Hardcoded 60s timeout (Greptile, P2)** — the import is intentionally kept in a single interactive transaction so it stays atomic: on failure the partial tree is rolled back instead of leaving a half-imported collection, so chunking outside the transaction would trade that away. I documented the value in code as a deliberate trade-off (it only bounds how long one pooled connection is held). Imports are infrequent and the pool default is 20, so concurrent-import pressure is low in practice — happy to expose it as an env-var if you'd prefer it configurable.

**Pubsub only for root collections (Greptile)** — this matches the pre-existing behavior: the previous nested `create` also only returned (and published `coll_added` for) the root collections; nested children were never individually published. I left it unchanged here to avoid altering subscription semantics in a bugfix PR — happy to make it a separate change if real-time events for nested folders are wanted.

The reproduction (real Postgres + Prisma 7.8.0 + `@prisma/adapter-pg`) still holds for the team path: the previous code fails the trigger case 5/5 with P2003 and the sequential fix passes 5/5.

PR #6447, opened 2026-06-18
fix(common): resolve environment variables in inherited collection headers

### LINKED ISSUE #6446: [bug]: Environment variables are not applied on collection level auth

### Platform

Desktop App

### Browser

Firefox

### Operating System

Windows

## Description
I've set basic auth for my whole collection using variables defined within an environment.

Image

Firstly: the variable preview shows that something is off.

Image

But the variable itself is recognized (using the wrong variable name instantly marks it in red, green is a recognized environment variable according to the docs).

Within my request it shows correctly that basic auth is inherited from the parent collection.
But the authorization header value shows that it is empty and leads to 401 upon execution.

Image

When I set the autorization to basic auth directly in the request, everything works as expected.

## Steps to reproduce
- create environment
- define variables for username and password
- create collection
- set authorization in collection to basic auth and insert environment variables
- create request in collection
- set authorization to inherit
- check created request headers

### Deployment Type

### Version

v2026.5.0

sahil29roy:
Closes #6446

Had a look at the issue — when a request inherits Basic Auth from a 
collection and the username/password fields use `<<variable>>` syntax, 
the Authorization header comes out empty and the request 401s.

The problem is in `Headers.vue` — it was pulling from `aggregateEnvs$` 
which only has the saved initial values, not whatever the user has 
actually set in their active environment session. So the variables never 
resolve.

Fix threads `aggregateEnvsWithCurrentValue$` through instead, which 
includes current session values and secrets. Also wired the `envs` prop 
down through `RequestTab`, `ResponseTab`, and the collection 
`Properties` modal so the preview tooltips in the auth fields resolve 
correctly too.

## Files changed

- `http/Headers.vue` — switched to `aggregateEnvsWithCurrentValue$`, added `envs` prop with fallback, updated the watch on `inheritedProperties`
- `http/RequestTab.vue` — pass resolved envs down to `HttpRequestOptions`  
- `http/ResponseTab.vue` — same
- `collections/Properties.vue` — combine inherited collection vars + active env and pass to `HttpHeaders` and `HttpAuthorization`

## Testing

1. Create an environment, add `username` and `password` variables
2. Create a collection, set Basic Auth using `<<username>>` / `<<password>>`
3. Add a child request set to Inherit auth
4. Check the Headers tab — Authorization should show the resolved Base64 value
5. Switch environment and confirm it updates
6. Open collection Properties → Authorization tab, hover the variable tokens — tooltip should show the resolved value not the raw placeholder
7. Send the request — should get 200 not 401

---
## Summary by cubic
Fixes #6446. Resolves env vars in inherited Basic Auth and headers, enforces request → collection → environment precedence in previews and execution, prevents stale previews/401s, and keeps collection secrets resolving in tooltips (incl. after OAuth redirects); also normalizes legacy env rows so embeds and previews resolve.

- **Bug Fixes**
  - Switch to `aggregateEnvsWithCurrentValue$`, add `:envs` prop, compute headers/auth with stale-update cancellation and non-empty filtering; normalize legacy `{ key, value }` rows via `normalizeAggregateEnvs`; clear inherited rows when the next request has no inherited properties.
  - Add `getEffectiveVariablesForRequest` (request > collection > environment) and use it in `RequestTab`, example `ResponseTab`, `Codegen`, `EnvInput`, the editor env tooltip extension, embeds, and the environment inspector; expand tests for precedence and empty-value fallthrough, and pass auth as a request-like object for correct header context.
  - In collection `Properties`/`Variables`, merge collection + inherited vars with the active env and pass to `HttpHeaders`/`HttpAuthorization`; introduce `collectionStoreKey`, set `sourceEnvID` on collection variables so tooltips resolve secret/current values, and preserve it across OAuth redirect rehydration; exclude request-level vars.

<sup>Written for commit 6625895a368af1297a2f28b5217b4a5bc2dca046. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6447?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/hoppscotch-common/src/components/http/example/ResponseTab.vue
+++ b/packages/hoppscotch-common/src/components/http/example/ResponseTab.vue
@@ -5,6 +5,8 @@
       <HttpRequestOptions
         v-model="tab.document.response.originalRequest"
         v-model:option-tab="optionTabPreference"
+        v-model:inherited-properties="tab.document.inheritedProperties"
+        :envs="envs"
       />
     </template>
     <template #secondary>
@@ -21,6 +23,11 @@ import { HoppTab } from "~/services/tab"
 import { HoppSavedExampleDocument } from "~/helpers/rest/document"
 import { RESTOptionTabs } from "../RequestOptions.vue"
 import { isEqual } from "lodash-es"
+import { useReadonlyStream } from "@composables/stream"
+import {
+  aggregateEnvsWithCurrentValue$,
+  getAggregateEnvsWithCurrentValue,
+} from "~/newstore/environments"
 
 const props = defineProps<{ modelValue: HoppTab<HoppSavedExampleDocument> }>()
 
@@ -31,6 +38,11 @@ const emit = defineEmits<{
 const tab = useVModel(props, "modelValue", emit)
 
 const optionTabPreference = ref<RESTOptionTabs>("params")
+
+const envs = useReadonlyStream(
+  aggregateEnvsWithCurrentValue$,
+  getAggregateEnvsWithCurrentValue()
+)
 
 // TODO: Come up with a better dirty check
 let oldResponse = cloneDeep(tab.value.document.response)
--- a/packages/hoppscotch-common/src/components/http/example/ResponseTab.vue
+++ b/packages/hoppscotch-common/src/components/http/example/ResponseTab.vue
@@ -6,7 +6,7 @@
         v-model="tab.document.response.originalRequest"
         v-model:option-tab="optionTabPreference"
         v-model:inherited-properties="tab.document.inheritedProperties"
-        :envs="envs"
+        :envs="resolvedEnvs"
       />
     </template>
     <template #secondary>
@@ -16,7 +16,7 @@
 </template>
 
 <script setup lang="ts">
-import { watch, ref } from "vue"
+import { watch, ref, computed } from "vue"
 import { useVModel } from "@vueuse/core"
 import { cloneDeep, isEqual } from "lodash-es"
 import { HoppTab } from "~/services/tab"
@@ -27,6 +27,7 @@ import {
   aggregateEnvsWithCurrentValue$,
   getAggregateEnvsWithCurrentValue,
 } from "~/newstore/environments"
+import { getEffectiveVariablesForRequest } from "~/helpers/utils/environments"
 
 const props = defineProps<{ modelValue: HoppTab<HoppSavedExampleDocument> }>()
 
@@ -42,6 +43,14 @@ const envs = useReadonlyStream(
   aggregateEnvsWithCurrentValue$,
   getAggregateEnvsWithCurrentValue()
 )
+
+const resolvedEnvs = computed(() => {
+  return getEffectiveVariablesForRequest(
+    tab.value.document.response.originalRequest.requestVariables,
+    tab.value.document.inheritedProperties?.variables,
+    envs.value
+  )
+})
 
 // TODO: Come up with a better dirty check
 let oldResponse = cloneDeep(tab.value.document.response)
--- a/packages/hoppscotch-common/src/components/http/example/ResponseTab.vue
+++ b/packages/hoppscotch-common/src/components/http/example/ResponseTab.vue
@@ -18,11 +18,10 @@
 <script setup lang="ts">
 import { watch, ref } from "vue"
 import { useVModel } from "@vueuse/core"
-import { cloneDeep } from "lodash-es"
+import { cloneDeep, isEqual } from "lodash-es"
 import { HoppTab } from "~/services/tab"
 import { HoppSavedExampleDocument } from "~/helpers/rest/document"
 import { RESTOptionTabs } from "../RequestOptions.vue"
-import { isEqual } from "lodash-es"
 import { useReadonlyStream } from "@composables/stream"
 import {
   aggregateEnvsWithCurrentValue$,
--- a/packages/hoppscotch-common/src/components/collections/Properties.vue
+++ b/packages/hoppscotch-common/src/components/collections/Properties.vue
@@ -264,6 +264,9 @@ export type EditingProperties = {
   isRootCollection: boolean
   path: string
   inheritedProperties?: HoppInheritedProperty
+  // Key the collection's secret/current values are stored under, computed
+  // authoritatively by `editProperties` (team → `id`, personal → `_ref_id`).
+  collectionStoreKey?: string
 }
 type HoppCollectionAuth = HoppRESTAuth | HoppGQLAuth
 type HoppCollectionHeaders = HoppRESTHeaders | GQLHeader[]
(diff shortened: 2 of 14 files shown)


Kinetic-Labs-GT (commented):
### Async Race Condition Detected

This PR correctly addresses the root cause of the inherited auth issue (#6446) by switching the stream to `aggregateEnvsWithCurrentValue$`. The variable resolution logic is sound and correctly handles live session values.

However, structural analysis indicates that the implementation introduces a **concurrency bug (race condition)** inside the newly updated watchers in `Headers.vue`.

####  The Introduced Bug
In `Headers.vue`, two watchers (`computedHeaders` and `inheritedProperties`) are configured to react to deep state changes (e.g., typing in an environment variable, toggling environments rapidly). 

Because these watchers execute asynchronous functions (`getComputedHeaders` and `getComputedAuthHeaders`) *without* any cancellation mechanism, the following failure mode occurs under rapid user interaction:

1. **Tick 1:** The user enters State A. The watcher fires -> Promise A begins.
2. **Tick 2:** The user quickly enters State B. The watcher fires -> Promise B begins.
3. Promise B resolves quickly over the network/compute thread. The UI updates to State B (Correct).
4. Promise A resolves slowly. The UI overwrites the current DOM with State A (Incorrect/Stale).

This lack of promise invalidation will permanently corrupt the UI state if an older network or compute execution resolves after a newer one. Additionally, the `inheritedProperties` watcher pushes to `inheritedProperty.value` asynchronously after evaluating a synchronous filter, which can lead to duplicate entries or array corruption if multiple promises resolve out of order.

#### The Required Fix: Vue `onCleanup`
To safely run asynchronous logic inside Vue watchers, we must invalidate stale executions using the `onCleanup` callback. This guarantees that only the results of the *most recent* state change are applied to the DOM.

Please apply the following patches to `Headers.vue`:

**1. `computedHeaders` Watcher (Lines ~595):**
```typescript
watch(
  [() => props.modelValue, resolvedEnvs],
  async (newVals, oldVals, onCleanup) => {
    let isStale = false;
    
    // Vue calls this automatically if the watcher re-triggers before completion
    onCleanup(() => {
      isStale = true;
    });

    const headers = await getComputedHeaders(props.modelValue, resolvedEnvs.value);
    
    // Discard the result if a newer watcher execution has already begun
    if (isStale) return;

    computedHeaders.value = headers.map((header, index) => ({
      id: `header-${index}`,
      ...header,
    }));
  },
  { immediate: true, deep: true }
);

```

**2. `inheritedProperties` Watcher (Lines ~633):**

```typescript
watch(
  () => [props.inheritedProperties, request.value, resolvedEnvs.value],
  async (newVals, oldVals, onCleanup) => {
    let isStale = false;
    
    // Invalidate this execution run if dependencies change again
    onCleanup(() => {
      isStale = true;
    });

    if (!props.inheritedProperties) return;

    // ... [keep the existing filter logic for inheritedHeaders here] ...

    if (/* ... existing Authorization header check ... */) {
      const [computedAuthHeader] = await getComputedAuthHeaders(
        aggregateEnvs.value,
        resolvedEnvs.value,
        request.value,
        props.inheritedProperties.auth.inheritedAuth,
        false
      );

      // Discard stale async results before pushing to the reactive array
      if (isStale) return; 

      if (computedAuthHeader) {
        // ... [keep the existing inheritedProperty.value.push logic here] ...
      }
    }
  },
  { immediate: true, deep: true }
);

```

Once these async safety checks are in place, the state management will be rock solid. Let me know if you have any questions on the concurrency flow!

nivedin:
@sahil29roy please look into the AI reviews

packages/hoppscotch-common/src/components/http/example/ResponseTab.vue
```diff
@@ -32,6 +38,11 @@ const tab = useVModel(props, "modelValue", emit)
 
 const optionTabPreference = ref<RESTOptionTabs>("params")
 
+const envs = useReadonlyStream(
```
sviat-barbutsa, line 42: Nice fix however I think this still misses request variables in this preview path becayse `envs` here is only built from `aggregateEnvsWithCurrentValue$`, but execution/codegen resolve variables with request variables first, then collection variables and then environment variables. Since `parseTemplateStringE` uses the first matching key, inherited auth can preview a different value than the one actually sent.

Here is the case - if inherited auth uses `<<token>>`, the selected env has `token=env-token`, and the request has an active `token=request-token` variable, the actual request/codegen should use `request-token`, but this preview path can still resolve `env-token`. 

Could you, please, build the list passed to `HttpRequestOptions` with the same precedence as execution/codegen, and add a regression test for a request-variable override in inherited auth?

sahil29roy: Good catch, I only used environment variables in the preview path, 
so request-level variables can end up resolving differently from 
execution/codegen. I'll update the variable list passed to 
`HttpRequestOptions` to follow the same precedence 
(request > collection > environment) and add a regression test 
covering the inherited auth override case.

nivedin (dismissed):

nivedin:
@sahil29roy Thank you for the PR, pushed a few follow-up commits on top of your branch. The main thing was pulling the request/collection/env variable merging into shared helpers in environments.ts (getEffectiveVariablesForRequest, transformRequestVariablesToAggregateEnv, normalizeAggregateEnvs) and pointing Codegen, EnvInput, the env inspector, and HoppEnvironment at them, they were all doing the same merge inline.

Along the way I fixed a handful of things that came up:

- the preview wasn't matching what actually gets sent, so it now runs through filterNonEmptyEnvironmentVariables
- inherited headers/auth were sticking around after switching to a request with no inherited props
- inherited-header parentName was using the wrong index after the list got filtered
- collection secrets showed up empty in the env tooltip because sourceEnvID was getting dropped
- the collection store key now comes from editProperties (team uses id, personal _ref_id) so it can't drift from the save side
- embeds were passing { key, value }, which resolved to empty — switched them to the proper shape

Also added more cases to variablePrecedence.spec.ts. Lint's clean and the common test suite passes.

jamesgeorge007 (approved):

jamesgeorge007:
Thanks, @sahil29roy. This is scheduled for the upcoming release.

sahil29roy:
@nivedin  Thanks a lot for the follow-up commits and for cleaning up the 
edge cases I missed — the sourceEnvID drop and the embed shape 
issues especially weren't on my radar. Really appreciate you 
taking the time to get it across the finish line properly.

sahil29roy:
@jamesgeorge007 Thanks for merging Looking forward to seeing it in the release. 
Will keep an eye out for more issues to contribute to.

PR #6448, opened 2026-06-19
feat(mock-server): append /backend to domain-based URL when subpath access is enabled

mirarifhasan:
Closes BE-782

When a deployment uses subpath-based access (`ENABLE_SUBPATH_BASED_ACCESS=true`), the backend is served under the `/backend` subpath. The generated domain-based mock server URL (`serverUrlDomainBased`) previously omitted this suffix, so the URL did not route correctly in those deployments. This PR appends `/backend` to that URL when subpath-based access is enabled.

### What's changed

- `mock-server.service.ts`: in the `cast` helper, read `ENABLE_SUBPATH_BASED_ACCESS` via `ConfigService` and conditionally append `/backend` to `serverUrlDomainBased`.
  - Only the domain-based URL is affected. `serverUrlPathBased` is unchanged since it derives from `VITE_BACKEND_API_URL`, which already reflects the subpath.
  - The existing `null` guard (when no wildcard domain is configured) is preserved — no suffix is added in that case.
  - Example: `https://myapp.mock.hopp.io` → `https://myapp.mock.hopp.io/backend`
- `mock-server.service.spec.ts`: add a `cast - server URL generation` test block covering the unset, `"false"`, `"true"`, secure-cookies (`https`), and null-wildcard cases.

### Notes to reviewers

- `ENABLE_SUBPATH_BASED_ACCESS` is a plain process env var (not under the `INFRA.` namespace), read the same way as `WHITELISTED_ORIGINS`.
- The new field is only consumed by the frontend display layer; there are no other side effects.

### DIFF
--- a/packages/hoppscotch-backend/src/mock-server/mock-server.service.ts
+++ b/packages/hoppscotch-backend/src/mock-server/mock-server.service.ts
@@ -64,15 +64,20 @@ export class MockServerService {
     const isSecure =
       this.configService.get<string>('INFRA.ALLOW_SECURE_COOKIES') === 'true';
     const protocol = isSecure ? 'https://' : 'http://';
+
+    // ENABLE_SUBPATH_BASED_ACCESS is a flat config key (no INFRA. prefix) to support flexible deployment strategies
+    const SUBPATH_BACKEND_SUFFIX = '/backend';
     const subpathSuffix =
       this.configService.get<string>('ENABLE_SUBPATH_BASED_ACCESS') === 'true'
-        ? '/backend'
+        ? SUBPATH_BACKEND_SUFFIX
         : '';
-    const serverUrlDomainBased = wildcardDomain
-      ? protocol +
-        dbMockServer.subdomain +
-        wildcardDomain.substring(1) +
-        subpathSuffix
+
+    const domainPart = wildcardDomain
+      ? dbMockServer.subdomain + wildcardDomain.substring(1)
+      : null;
+
+    const serverUrlDomainBased = domainPart
+      ? `${protocol}${domainPart.replace(/\/+$/, '')}${subpathSuffix}`
       : null;
 
     return {
--- a/packages/hoppscotch-backend/src/mock-server/mock-server.service.ts
+++ b/packages/hoppscotch-backend/src/mock-server/mock-server.service.ts
@@ -64,8 +64,15 @@ export class MockServerService {
     const isSecure =
       this.configService.get<string>('INFRA.ALLOW_SECURE_COOKIES') === 'true';
     const protocol = isSecure ? 'https://' : 'http://';
+    const subpathSuffix =
+      this.configService.get<string>('ENABLE_SUBPATH_BASED_ACCESS') === 'true'
+        ? '/backend'
+        : '';
     const serverUrlDomainBased = wildcardDomain
-      ? protocol + dbMockServer.subdomain + wildcardDomain.substring(1)
+      ? protocol +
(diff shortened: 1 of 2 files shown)


mirarifhasan:
Addressed the review feedback from Greptile and cubic (both flagged the same gap: missing test coverage for the new `ENABLE_SUBPATH_BASED_ACCESS` branch).

Added a `cast - server URL generation` test block in `mock-server.service.spec.ts` (commit 6e9c3b786) covering:

- `ENABLE_SUBPATH_BASED_ACCESS` unset → no `/backend` suffix
- `ENABLE_SUBPATH_BASED_ACCESS="false"` → no `/backend` suffix
- `ENABLE_SUBPATH_BASED_ACCESS="true"` → `/backend` appended
- secure cookies + subpath access → `https://` protocol with `/backend` suffix
- wildcard domain unset → `serverUrlDomainBased` stays `null` even with subpath access enabled

This guards the config-key lookup against regressions (wrong namespace / env var not injected). All 5 tests pass.

packages/hoppscotch-backend/src/mock-server/mock-server.service.ts
```diff
@@ -64,8 +64,15 @@ export class MockServerService {
     const isSecure =
```
nahidhasan94, line 64: Add inline comment explaining the flat config key + extract constant

suggestion:
```
const isSecure =
      this.configService.get<string>('INFRA.ALLOW_SECURE_COOKIES') === 'true';
    const protocol = isSecure ? 'https://' : 'http://';

    const SUBPATH_BACKEND_SUFFIX = '/backend';
    const subpathSuffix =
      this.configService.get<string>('ENABLE_SUBPATH_BASED_ACCESS') === 'true'
        ? SUBPATH_BACKEND_SUFFIX
        : '';

    const serverUrlDomainBased = wildcardDomain
      ? `${protocol}${dbMockServer.subdomain}${wildcardDomain.substring(1)}${subpathSuffix}`
      : null;
```

why: The inline comment prevents future devs from "fixing" the key by adding an INFRA. prefix. The constant makes the magic string discoverable. The template literal is easier to read than 4-line string concatenation inside a ternary.

packages/hoppscotch-backend/src/mock-server/mock-server.service.ts
```diff
@@ -64,8 +64,15 @@ export class MockServerService {
...
       this.configService.get<string>('INFRA.ALLOW_SECURE_COOKIES') === 'true';
     const protocol = isSecure ? 'https://' : 'http://';
+    const subpathSuffix =
+      this.configService.get<string>('ENABLE_SUBPATH_BASED_ACCESS') === 'true'
+        ? '/backend'
+        : '';
     const serverUrlDomainBased = wildcardDomain
```
nahidhasan94: suggested code:

```
const domainPart = wildcardDomain
      ? dbMockServer.subdomain + wildcardDomain.substring(1)
      : null;

    const serverUrlDomainBased = domainPart
      ? `${protocol}${domainPart.replace(/\/+$/, '')}${subpathSuffix}`
      : null;
```
Why: Guards against a wildcard domain value like *.mock.hopp.io/ producing https://x.mock.hopp.io//backend. Low probability, but zero-cost defensiveness.

nahidhasan94 (changes requested):

nahidhasan94 (approved):
Team mock server test coverage: a nice-to-have, can be a follow-up @mirarifhasan

PR #6452, opened 2026-06-20
fix(common): preserve headers containing colon-space in cURL import

linze0721:
## Summary
Fix cURL import silently dropping headers whose value contains ": " (colon+space).

Closes #6400

## Root Cause
The header parser splits on all occurrences of ": " instead of only the first one. Per RFC 9110, a header is name: value where only the first colon separates name from value.

## Changes
Updated the cURL header parser to split only on the first colon, preserving colon-space sequences in header values.

---
## Summary by cubic
Fix cURL import dropping headers when the value contains ": ". We now split on only the first colon to keep the full header value.

- **Bug Fixes**
  - Updated `hoppscotch-common` cURL header parser to split on the first `:` only (per RFC 9110).
  - Added a test for `X-Signature: sha256: abc:def` to ensure values with colons are preserved.

<sup>Written for commit 1fbc0eb8aae3f310989f398d50b949acf4430e35. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6452?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/hoppscotch-common/src/helpers/curl/__tests__/curlparser.spec.js
+++ b/packages/hoppscotch-common/src/helpers/curl/__tests__/curlparser.spec.js
@@ -870,6 +870,36 @@ const samples = [
     }),
   },
   {
+    command: `curl google.com -H "X-Signature: sha256: abc:def"`,
+    response: makeRESTRequest({
+      method: "GET",
+      name: "Untitled",
+      endpoint: "https://google.com/",
+      auth: {
+        authType: "inherit",
+        authActive: true,
+      },
+      body: {
+        contentType: null,
+        body: null,
+      },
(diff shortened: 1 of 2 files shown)


jamesgeorge007 (commented):
Closing in favour of #6401, which covers the same `headers.ts` fix and was opened earlier.

PR #6454, opened 2026-06-20
fix(common): preserve float .0 precision in request body

### LINKED ISSUE #6317: [bug]: POST/PUT request body floats losing .0 precision (e.g., 70.0 → 70)

linze0721:
## Summary
Fix request body float values like 70.0 being sent as 70 (integer).

Closes #6317

## Root Cause
JSON.parse/JSON.stringify round-trip drops .0 from float values. The body should be sent as-is when it is already valid JSON text.

## Changes
Avoid re-serializing JSON body through parse/stringify when the raw text is already valid JSON, preserving the original numeric representation.

---
## Summary by cubic
Preserves the exact JSON request body text so numbers like 70.0 aren’t normalized to 70 when sending requests. Prevents precision loss for JSON bodies.

- **Bug Fixes**
  - For JSON content types, send the body exactly as written if it’s already valid JSON; skip parse/stringify.
  - Only strip comments when the body isn’t valid JSON, keeping JSONC support.
  - Added a unit test to ensure numeric tokens (e.g., 70.0) are preserved.

<sup>Written for commit 6019a862d5adccf4881762fa64db7e61705d0ae2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6454?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/hoppscotch-common/src/helpers/kernel/__tests__/kernel.spec.ts
+++ b/packages/hoppscotch-common/src/helpers/kernel/__tests__/kernel.spec.ts
@@ -331,6 +331,32 @@ describe("REST Request Transformation", () => {
     })
   })
 
+  it("preserves JSON request body text without re-serializing numbers", async () => {
+    const jsonBody = '{"amount":70.0,"nested":{"value":1.0}}'
+    const request: EffectiveHoppRESTRequest = {
+      ...baseEffectiveRequest,
+      method: "POST",
+      body: { contentType: "application/json", body: jsonBody },
+      effectiveFinalBody: jsonBody,
+      effectiveFinalHeaders: [
+        {
+          active: true,
+          key: "content-type",
+          value: "application/json",
+          description: "",
+        },
+      ],
+    }
+
+    const result = await RESTRequest.toRequest(request)
+
+    expect(result.content).toEqual({
+      kind: "text",
+      content: jsonBody,
+      mediaType: "application/json",
+    })
+  })
+
   it("includes auth parameters when basic auth is active", async () => {
     const request: EffectiveHoppRESTRequest = {
       ...baseEffectiveRequest,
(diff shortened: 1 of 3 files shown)


jamesgeorge007 (commented):
Closing since this is part of a rapid batch of generated PRs, and #6455 in the same batch still has its "# Agent Report" header intact.

A quick check of existing issues and PRs, plus a read-through of the diff before filing, helps avoid duplicate review effort next time. Thanks for your interest in contributing!

PR #6391, opened 2026-06-04
Fix/json rendering performance

hardikkaurani:
## Overview

This PR improves the user experience of the monthly credit limit notification by making the message clearer, more informative, and easier to act upon. When users exhaust their available monthly credits, it is important that they immediately understand why requests are failing and what options are available to continue using the service.

The update focuses on improving the presentation and clarity of the notification while preserving all existing functionality and behavior.

---

## The Problem

Users who reach their monthly credit limit are presented with a notification explaining that their credits have been exhausted. While the required information is available, the overall experience can be improved by making the message more readable and by providing clearer guidance on the available next steps.

A notification shown during a blocked workflow should quickly answer the following questions:

- Why did my request fail?
- Have I reached a usage limit?
- What can I do next?
- When will my credits become available again?

Improving the clarity of this message helps reduce confusion and creates a smoother user experience.

---

### Improved Notification Experience

- Refined the monthly credit limit notification flow.
- Improved the presentation of the exhausted-credit message.
- Enhanced readability and overall message structure.
- Made important information easier to identify at a glance.

### Improved User Guidance

- Clearly surfaces the actions available after credits are exhausted.
- Helps users understand the available recovery options without additional investigation.
- Provides better guidance for continuing usage after reaching the limit.

### Improved Information Visibility

- Ensured credit reset information remains visible and easy to understand.
- Improved the hierarchy of information shown within the notification.
- Reduced ambiguity around account status and usage limits.

### User Experience Enhancements

- Improved overall notification clarity.
- Reduced potential confusion during failed request scenarios.
- Created a more actionable and user-friendly experience.
- Maintained consistency with the existing application workflow.

---

## Benefits

This change provides several user-facing improvements:

- Better understanding of why requests cannot be processed.
- Faster discovery of available recovery options.
- Clear visibility into credit reset timing.
- Reduced friction when usage limits are encountered.
- Improved overall usability of the notification experience.

---

## Testing

The following checks were performed:

- Verified notification rendering behavior.
- Confirmed exhausted-credit messaging displays correctly.
- Verified action options remain accessible and visible.
- Confirmed reset timing information is displayed correctly.
- Tested overall readability and user experience.
- Ensured no existing functionality was affected.

---

## Impact

- UI/UX improvement only.
- No backend changes.
- No API changes.
- No business logic changes.
- No impact on existing workflows.

---

## Notes to Reviewers

[...]

PR #6463, opened 2026-06-23
fix(common): lazy-load `jq-wasm` in the JSON lens renderer

### LINKED ISSUE #6291: [bug]: UI freezes/becomes unresponsive after every request while JSON response renders
## Workaround 27-05-2026
It appears that using the `.AppImage`  may work for some users.

### Platform

Desktop App

### Browser

Firefox

### Operating System

Linux

### Bug Description

**Operating System**
Linux — Ubuntu 24.04 LTS

**Version**
v2026.4.0, installed via .deb

**Bug Description**

After sending any request, the entire UI becomes unresponsive while the response renders. This happens on every request, not just the first. It looks the first request is triggering a full batch of asset loads on the main thread which blocks the UI (mainly the 1.3MB JSONLensRenderer). Following requests have no asset loads, and a 'loading' spinner appears, but the UI is still completely unresponsive.

- First request after launch: app is completely frozen and unresponsive for ~1 minute before the response appears.
- Subsequent requests: a loading spinner appears (so the renderer is mounted), but the rest of the UI is still fully blocked - it is not possible to interact with any other part of the app (e.g. navigate to scripts, switch tabs) while waiting for the JSON response to load.

**Steps to Reproduce**
1. Open Hoppscotch Desktop on Ubuntu 24.04
2. Send any POST request that returns a JSON response body
3. The UI is completely unresponsive until the response panel finishes rendering (~1-2 minutes)
4. Send a second request to the same endpoint
5. A 'loading' spinner appears but the UI is still blocked (~30-60s)

**Expected Behavior**
The rest of the UI should remain interactive while the response is rendering.

**Logs**
The request completes successfully (200 OK) and then the main thread (`ThreadId(01)`) immediately begins serving a large number of assets for the response renderer including `JSONLensRenderer-tSbXhImA.js` at 1,301,785 bytes for the first request. Not sure in the second request. Both log files are attached as .txt files.

[first_request_logs.txt](https://github.com/user-attachments/files/27476286/first_request_logs.txt)
[second_request_logs.txt](https://github.com/user-attachments/files/27476285/second_request_logs.txt)

### Deployment Type

Hoppscotch Cloud

### Version

v2026.4.0

hassams:
Closes #6291

### What's changed

`JSONLensRenderer.vue` statically imported `jq-wasm` at the top of the file, even though it's only ever used inside the (debounced) jq filter computation, gated behind the user actually typing a filter query. `jq-wasm` bundles a sizeable Emscripten/WASM payload (~1.3MB on disk locally), and a static import pulls that into the same chunk that has to be fetched and parsed every time a JSON response is rendered — not just when the filter feature is used.

This matches the symptoms reported in the issue: a large `JSONLensRenderer-*.js` chunk (~1.3MB) being loaded on every JSON response render, with the main thread blocked while it's parsed — most noticeable on Linux/WebKitGTK desktop builds, but not specific to any one platform.

- [x] Replaced the static `import * as jq from "jq-wasm"` with a type-only import (`import type * as JqWasm from "jq-wasm"`) plus a memoized `getJqWasm()` helper that does a dynamic `import("jq-wasm")`, only called from inside the filter computation when a query is actually run.
- [x] No behavior change for the filter feature itself — same `jq.raw(...)` call, just resolved lazily and cached after first use.

### Notes to reviewers

- Verified via the dev server that `jq-wasm` no longer appears in the component's static imports and is only pulled in through the dynamic `import()` inside `getJqWasm()` (confirmed by inspecting Vite's transformed module output).
- A previous attempt at this exact fix (#6391) was closed unmerged — per review comments there, it bundled the one real fix together with ~36 unrelated/unwired files. This PR isolates just that one change.
- Ran the full `hoppscotch-common` test suite (no regressions) and `pnpm -r do-lint && pnpm -r do-typecheck` across all workspace packages via the repo's pre-commit hook — all passing.
- No interface/env var changes, so no README updates were needed.

---
## Summary by cubic
Lazy-loads `jq-wasm` in the JSON lens so its ~1.3MB WASM is fetched only when a jq filter runs, avoiding main-thread stalls during JSON renders. Replaces the static import in `JSONLensRenderer.vue` with a memoized dynamic import that retries on failure and uses `typeof import("jq-wasm")` for typing; behavior unchanged, resolving #6291.

<sup>Written for commit 96581d8137f4d60249b433c4314affb7fc8f0b1b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6463?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/hoppscotch-common/src/components/lenses/renderers/JSONLensRenderer.vue
+++ b/packages/hoppscotch-common/src/components/lenses/renderers/JSONLensRenderer.vue
@@ -303,7 +303,15 @@ const t = useI18n()
 // JSON response render has to load and parse.
 let jqWasmPromise: Promise<typeof JqWasm> | null = null
 const getJqWasm = () => {
-  if (!jqWasmPromise) jqWasmPromise = import("jq-wasm")
+  if (!jqWasmPromise) {
+    // if the dynamic import fails (e.g. a transient network error), drop the
+    // cached promise so the next filter attempt can retry instead of being
+    // stuck failing until the page is reloaded
+    jqWasmPromise = import("jq-wasm").catch((err) => {
+      jqWasmPromise = null
+      throw err
+    })
+  }
   return jqWasmPromise
 }
 
--- a/packages/hoppscotch-common/src/components/lenses/renderers/JSONLensRenderer.vue
+++ b/packages/hoppscotch-common/src/components/lenses/renderers/JSONLensRenderer.vue
@@ -273,7 +273,6 @@ import * as E from "fp-ts/Either"
 import { pipe } from "fp-ts/function"
 import { computed, ref, reactive } from "vue"
 import { computedAsync, refDebounced } from "@vueuse/core"
-import type * as JqWasm from "jq-wasm"
 import { useCodemirror } from "@composables/codemirror"
 import { HoppRESTResponse } from "~/helpers/types/HoppRESTResponse"
 import jsonParse, { JSONObjectMember, JSONValue } from "~/helpers/jsonParse"
@@ -301,7 +300,7 @@ const t = useI18n()
 // ever needed when the user actively filters a JSON response with a jq
 // query. Importing it lazily keeps that payload out of the chunk every
 // JSON response render has to load and parse.
-let jqWasmPromise: Promise<typeof JqWasm> | null = null
+let jqWasmPromise: Promise<typeof import("jq-wasm")> | null = null
 const getJqWasm = () => {
   if (!jqWasmPromise) {
     // if the dynamic import fails (e.g. a transient network error), drop the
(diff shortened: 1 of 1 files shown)


hassams:
Good catch — fixed in b42665b64: `jqWasmPromise` now resets to `null` on rejection so a failed import (e.g. transient network error) doesn't permanently break the filter feature until a page reload.

hassams:
Addressed both Copilot comments in 952c8c7b6: dropped the `import type * as JqWasm` namespace import and typed the cache as `Promise<typeof import("jq-wasm")>` directly, which sidesteps any ambiguity around `typeof` on a type-only namespace import and removes the now-unused import in one change.

(Verified the original `typeof JqWasm` form actually type-checked fine under `vue-tsc@2.2.0`/`typescript@5.9.3` in an isolated repro — this repo's pinned `vue-tsc@1.8.8` for this package is itself broken against the locked TS version, unrelated to this PR — but the `typeof import("jq-wasm")` form is unambiguously correct and removes the dependency on that question entirely.)

nivedin (approved):

jamesgeorge007 (approved):

jamesgeorge007:
Thanks for the PR, @hassams 🙌 
This is scheduled for the upcoming release.

PR #6486, opened 2026-07-04
fix(curl): preserve header values containing ": " on import

okxint:
## Summary

Fixes a bug where importing a cURL command silently drops any header whose **value** contains `": "` (colon + space).

**Steps to reproduce:**
```bash
curl https://httpbin.org/get -H "X-Note: hello: world" -H "Accept: application/json"
```
Import this in Hoppscotch → only `Accept` appears; `X-Note` is gone.

## Root cause

`getHeaderPair` in `headers.ts` had a three-step pipeline:

```ts
const getHeaderPair = flow(
  S.replace(":", ": "),   // normalize "Header:value"
  S.split(": "),           // ← splits on ALL ": " occurrences
  O.fromPredicate((arr) => arr.length === 2),  // ← drops if > 2 segments
  O.map(([k, v]) => [k.trim(), v?.trim() ?? ""] as [string, string])
)
```

A header like `"X-Note: hello: world"` produces `["X-Note", "hello", "world"]` (three segments) and is discarded by the `length === 2` predicate.

## Fix

Per RFC 9110, the header name is everything before the **first** colon; the value (which may itself contain `": "`) is everything after it. Replace the multi-step flow with `indexOf(":")` + `slice`:

```ts
const getHeaderPair = (raw: string): O.Option<[string, string]> => {
  const idx = raw.indexOf(":")
  if (idx === -1) return O.none
  const key = raw.slice(0, idx).trim()
  const value = raw.slice(idx + 1).trim()
  if (!key) return O.none
  return O.some([key, value])
}
```

This also correctly handles compact form (`Authorization:Bearer token`) by trimming the value.

## Tests

Added `packages/hoppscotch-common/src/helpers/curl/__tests__/headers.spec.ts` with 4 cases:
- Value with `": "` → preserved in full (`hello: world`)
- Value with `:` but no space → preserved (`12:30`)
- Compact form without space after colon → key/value both trimmed
- Multiple headers with `": "` values → all preserved

Closes #6400

---
## Summary by cubic
Fix cURL import dropping headers when the value contains ": ". We now split on the first colon only (RFC 9110) so values like "hello: world" are preserved.

- **Bug Fixes**
  - Parse header lines using first-colon `indexOf` + `slice`, trimming key/value and supporting compact form (e.g., `Authorization:Bearer token`).
  - Add regression tests covering colon-space values, colon without space, compact form, and multiple headers with colons.

<sup>Written for commit 5397ee78b30c51f38ffd82c5e4282d5f1a328ecc. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6486?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/hoppscotch-common/src/helpers/curl/__tests__/headers.spec.ts
+++ b/packages/hoppscotch-common/src/helpers/curl/__tests__/headers.spec.ts
@@ -1,0 +1,56 @@
+import { describe, expect, it } from "vitest"
+
+import { parseCurlCommand } from "../curlparser"
+
+// Regression tests for https://github.com/hoppscotch/hoppscotch/issues/6400
+// A header value that itself contains ": " was silently dropped because the old
+// parser split on every occurrence of ": " and then rejected arrays with > 2 elements.
+describe("cURL header parsing", () => {
+  it("imports a header whose value contains ': ' (colon + space)", () => {
+    const result = parseCurlCommand(
+      `curl https://example.com -H "X-Note: hello: world" -H "Accept: application/json"`
+    )
+
+    const xNote = result.headers.find((h) => h.key === "X-Note")
+    const accept = result.headers.find((h) => h.key === "Accept")
+
+    expect(xNote).toBeDefined()
+    expect(xNote?.value).toBe("hello: world")
+    expect(accept).toBeDefined()
+    expect(accept?.value).toBe("application/json")
+  })
+
+  it("imports a header whose value has a colon without trailing space", () => {
+    const result = parseCurlCommand(
+      `curl https://example.com -H "X-Time: 12:30"`
+    )
+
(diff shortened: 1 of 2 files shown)


jamesgeorge007 (commented):
Closing as a duplicate of #6401, which fixes #6400 with the same first-colon approach.

Separately, this is your second PR for this issue. Please stop re-filing and follow up on the existing PRs instead.

Issue #6491, opened 2026-07-07
[feature]: Enable editing of Response Headers and Response Body in saved Examples / Mocks


CG-Ziv:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Summary

Currently, when a user saves a response as an **Example** in a collection, the response metadata (Status Code, Response Headers, and Response Body) becomes completely read-only in the UI. 
If a developer wants to manually design a mock response (e.g., simulating success and failure cases with specific custom headers and JSON structures) without hitting a live server, they are unable to do so. Clicking the `...` menu next to the saved Example only allows renaming the example name, and the response headers/body tabs at the bottom have no inputs or edit options.

### Why should this be worked on?

We would like to have the ability to edit the saved example's response directly in the Hoppscotch UI. Specifically:
1. **Response Headers Editor**: Under the Response > Headers tab, provide an input table (similar to the Request Headers editor) where users can add, edit, or delete custom key-value pairs.
2. **Response Body Editor**: Under the Response > Raw (or Body) tab, make the text area editable so users can paste or write custom mock payloads.
3. A **Save** button to persist these changes back to the Example.

shyam-rix:
Took a closer look — good news, most of this already landed on `main`: status code, body (Raw/JSON editors), and header edit/delete are editable in saved Examples, and there's a **Save** button that persists to user & team collections. If you're on the stable release, it should reach you in an upcoming version.

The one piece that was genuinely missing was **adding a brand-new response header** (you could edit/delete captured ones but not add your own) — which is exactly what you'd need to author a mock from scratch. Opened #6494 to add that. Could you confirm which version you're on so we can verify the rest covers your use case?

CG-Ziv:
Hi, thanks for checking. I'm currently on version v2026.6.0 for both the Server and Desktop.

shyam-rix:
Thanks for confirming! `v2026.6.0` is indeed the current stable release, but this issue came in just after that build was cut. All the editing work — editable status code, Raw/JSON body, header edit/delete, and the **Save** button — landed on `main` afterward, so it isn't in `2026.6.0` yet. That's why the response still shows as read-only on your end.

The last remaining piece, adding a brand-new response header from scratch, is in #6494 and on its way in. All of this is queued for the next release (`2026.7.0`), so you'll get the full mock-authoring flow in that update.

shyam-rix:
Thanks for assigning this to me, @jamesgeorge007! Picking it up.

[...]

PR #6494, opened 2026-07-07
feat(response): allow adding new headers when editing a saved example

shyam-rix:
### What

Adds an **Add** button to the response Headers editor so users can add brand-new header rows while editing a saved Example, not just edit or delete the ones captured from the original response.

Refs #6491

### Why

#6491 asks for editable response headers/body in saved Examples so developers can design mock responses. Most of that already works on `main` — status code, body (via the Raw/JSON CodeMirror lenses), and header **edit/delete** are all editable, and `saveExample()` already persists changes back to user and team collections.

The one missing primitive is **adding a header that wasn't in the original response** — the exact thing you need when authoring a mock from scratch. `LensesHeadersRenderer` only iterated over existing headers, so there was no way to introduce, say, an `X-RateLimit-Remaining` header on a saved Example.

### Changes

`packages/hoppscotch-common/src/components/lenses/HeadersRenderer.vue`:
- Add an **Add** button (`lucide/plus`, `add.new` i18n key), gated on `isEditable` so it only shows in the editable saved-Example context and stays hidden on live responses.
- `addHeader()` pushes an empty `{ key: "", value: "" }` row, mirroring the existing `deleteHeader()`/`splice()` pattern. The new row uses the same editable `SmartEnvInput` entries already rendered by `HeadersRendererEntry`.

No schema, store, or new-component changes. Persistence rides the existing dirty-tracking (`ResponseTab.vue`) → `saveExample()` flow.

### Scope / call sites

`LensesHeadersRenderer` is used in three places; behavior verified for each:
- `http/example/LenseBodyRenderer.vue` — `is-editable="true"` → **Add button shows** (target flow).
- `lenses/ResponseBodyRenderer.vue` (live response headers + request headers) — `is-editable="false"` → **Add button hidden** (unchanged).

### Testing

- [ ] Save a response as an Example, open it, go to the Headers tab → click **Add** → new editable row appears.
- [ ] Fill key/value, click **Save** → reopen the Example → the added header persists.
- [ ] Open a live response's Headers tab → **no** Add button (unchanged behavior).

### Notes for reviewers

Content-type headers remain non-editable by design (`HeadersRendererEntry` guards this, since changing content-type can switch the active lens). Adding *other* headers is unaffected. Happy to revisit if you'd like new rows to allow a content-type override.

---
## Summary by cubic
Allow adding response headers when editing a saved Example (refs #6491) to support mock response design. Previously you could only edit/delete existing headers; now you can add new rows. Live response views remain unchanged.

- Show an Add button only when `isEditable` and `headers` are present; hidden on live responses.
- Reassign the `headers` array on add/delete so `v-model` emits and persists; `update:modelValue` now emits `HoppRESTResponseHeader[]`.
- Content-Type rows remain non-editable.

<sup>Written for commit 729663b3c3f9dc49312ef03275a276551e5fae9c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6494?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/hoppscotch-common/src/components/lenses/HeadersRenderer.vue
+++ b/packages/hoppscotch-common/src/components/lenses/HeadersRenderer.vue
@@ -55,7 +55,7 @@ const props = defineProps<{
 }>()
 
 const emit = defineEmits<{
-  (e: "update:modelValue"): void
+  (e: "update:modelValue", value: HoppRESTResponseHeader[]): void
 }>()
 
 const headers = useVModel(props, "modelValue", emit)
--- a/packages/hoppscotch-common/src/components/lenses/HeadersRenderer.vue
+++ b/packages/hoppscotch-common/src/components/lenses/HeadersRenderer.vue
@@ -72,10 +72,10 @@ const copyHeaders = () => {
 }
 
 const addHeader = () => {
-  headers.value.push({ key: "", value: "" })
+  headers.value = [...headers.value, { key: "", value: "" }]
 }
 
 const deleteHeader = (index: number) => {
-  headers.value.splice(index, 1)
+  headers.value = headers.value.filter((_, i) => i !== index)
 }
 </script>
--- a/packages/hoppscotch-common/src/components/lenses/HeadersRenderer.vue
+++ b/packages/hoppscotch-common/src/components/lenses/HeadersRenderer.vue
@@ -8,6 +8,13 @@
       </label>
       <div class="flex">
         <HoppButtonSecondary
+          v-if="isEditable"
+          v-tippy="{ theme: 'tooltip' }"
+          :title="t('add.new')"
+          :icon="IconPlus"
+          @click="addHeader"
+        />
+        <HoppButtonSecondary
           v-if="headers"
           v-tippy="{ theme: 'tooltip' }"
           :title="t('action.copy')"
@@ -30,6 +37,7 @@
 <script setup lang="ts">
 import IconCopy from "~icons/lucide/copy"
 import IconCheck from "~icons/lucide/check"
+import IconPlus from "~icons/lucide/plus"
 import { refAutoReset } from "@vueuse/core"
 import { copyToClipboard } from "~/helpers/utils/clipboard"
 import { useI18n } from "@composables/i18n"
@@ -61,6 +69,10 @@ const copyHeaders = () => {
   copyToClipboard(JSON.stringify(props.modelValue))
   copyIcon.value = IconCheck
   toast.success(`${t("state.copied_to_clipboard")}`)
+}
+
+const addHeader = () => {
+  headers.value.push({ key: "", value: "" })
 }
 
 const deleteHeader = (index: number) => {
--- a/packages/hoppscotch-common/src/components/lenses/HeadersRenderer.vue
+++ b/packages/hoppscotch-common/src/components/lenses/HeadersRenderer.vue
@@ -8,7 +8,7 @@
       </label>
       <div class="flex">
         <HoppButtonSecondary
-          v-if="isEditable"
+          v-if="isEditable && headers"
           v-tippy="{ theme: 'tooltip' }"
           :title="t('add.new')"
           :icon="IconPlus"


shyam-rix:
Thanks for the review! Good catch on the null-guard — applied in 7ccdc74. The Add button now uses `v-if="isEditable && headers"`, matching the sibling Copy button so it only renders when the headers model is present.

shyam-rix:
Good catch on the content-type edge case (thanks @cubic-dev-ai / @tembo). Fixed in cd2d868: the delete action is now gated on `isEditable` instead of `isEntryEditable`, so a row whose key is `content-type` (still read-only for editing, to avoid switching the active lens) can always be removed. That resolves the stuck-row case where a user adds a row and types `content-type`.

shyam-rix:
Reverted the content-type delete change (previously cd2d868) to keep this PR tightly scoped.

@greptile-apps raised a fair P1: gating delete on `isEditable` also made the *captured* `content-type` header — present in almost every saved example — deletable, which changes lens selection on reopen. That's a behavior change beyond "add a header row," so I've dropped it. The branch is back to two focused commits (add button + null guard).

On the @cubic-dev-ai / @tembo edge case (typing `content-type` into a newly added row makes it read-only with no delete): that's a narrow facet of the existing `content-type` lock in `HeadersRendererEntry`, not something introduced by the core feature. Happy to address it in a follow-up if you'd like — e.g. keeping user-added rows removable, or blocking reserved keys — but I'd rather not change captured-header behavior implicitly here. Deferring to maintainer preference.

Re: @cubic-dev-ai's note that `addHeader` uses `.push()` and doesn't emit `update:modelValue` — this mirrors the existing `deleteHeader`/`splice` pattern, and dirty-tracking still fires via the deep watcher on `document.response` in `ResponseTab.vue`, so the save flow works. Kept it consistent with the sibling method.

shyam-rix:
@jamesgeorge007 this one's ready for review whenever you get a chance. It adds the ability to add a brand-new response header when editing a saved Example (the last missing piece from #6491) — status/body/header edit + Save already landed on `main`.

Also addressed the reviewer note on `addHeader`: both add and delete now reassign the `headers` array instead of mutating in place, so `update:modelValue` fires correctly. Would be great to get this into `2026.7.0` alongside the rest.

shyam-rix:
Good call, @Copilot — fixed in 0ab253c. `update:modelValue` is now typed with its `HoppRESTResponseHeader[]` payload instead of `void`, matching what `useVModel`'s setter actually emits when `addHeader`/`deleteHeader` reassign the array. Both call sites bind an array via `v-model`, so the tighter signature is source-compatible, and the package typecheck passes.

CG-Ziv:
Hi @shyam-rix / @jamesgeorge007 ,

Just checking on the status of #6494. I noticed the PR is now out-of-date with the base branch. Is there anything blocking the review or merge?

This PR would complete the original feature request in #6491. Thanks!

shyam-rix:
Thanks for the nudge, @CG-Ziv — nothing blocking on my end.

I've just rebased onto the latest `main`, so the branch is up to date again and checks are green.

**Status:** this is a single-file change to `HeadersRenderer.vue` adding the one missing primitive from #6491 — adding a *new* response header when editing a saved Example (status/body/header edit + Save already landed on `main`). All reviewer feedback is addressed: null guard on the Add button, `update:modelValue` typed with its `HoppRESTResponseHeader[]` payload, and both add/delete reassign the array so v-model emits correctly.

It's just waiting on a maintainer pass — @jamesgeorge007, happy to adjust anything you'd like here.

One item I deliberately left out of scope and deferred to maintainer preference: typing `content-type` into a newly added row makes it read-only with no delete. That's a facet of the existing content-type lock in `HeadersRendererEntry`, not something this PR introduces — happy to take it as a follow-up.

Ziv81:
@jamesgeorge007 
@nivedin 

The branch is up to date and all review comments appear to be resolved.
Could someone please help review/approve this PR when available?
Thanks!

PR #6515, opened 2026-07-18
fix: preserve header values containing colons in cURL import

ekanshsaxena:
The previous implementation split on all ': ' occurrences, then rejected headers where the split produced more than 2 parts. This silently dropped any header whose value contained ': ' (e.g., 'X-Note: hello: world').

Split on the first colon only per RFC 9110: the header name is everything before the first colon, and the value is everything after.

Closes #6400

Closes # 

### Notes to reviewers

---
## Summary by cubic
Fix cURL import to preserve headers when values contain colons by splitting only on the first ":" per RFC 9110. This prevents dropping headers like "X-Note: hello: world".

<sup>Written for commit f03ad03e8b6659be75bb7ebcf20937d7eeab8fd3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6515?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/hoppscotch-common/src/helpers/curl/sub_helpers/headers.ts
+++ b/packages/hoppscotch-common/src/helpers/curl/sub_helpers/headers.ts
@@ -10,13 +10,14 @@ import {
 } from "~/helpers/functional/object"
 import { tupleToRecord } from "~/helpers/functional/record"
 
-const getHeaderPair = flow(
-  S.replace(":", ": "),
-  S.split(": "),
-  // must have a key and a value
-  O.fromPredicate((arr) => arr.length === 2),
-  O.map(([k, v]) => [k.trim(), v?.trim() ?? ""] as [string, string])
-)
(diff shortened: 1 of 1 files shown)


jamesgeorge007 (commented):
Closing in favour of #6401, which is the canonical fix for #6400, is already up for review and uses the same first-colon-split approach. For context, the [status note on the issue](https://github.com/hoppscotch/hoppscotch/issues/6400#issuecomment-5005895067) asks contributors to hold off on new PRs because the issue is already covered. If you spot any refinements, please propose them as comments or review suggestions on #6401 so the discussion stays with the canonical fix.

PR #6554, opened 2026-07-30
feat: add option to disable cookies globally and per-request

### LINKED ISSUE #6526: [feature]: Option to disable cookies

### Summary

Add an option (per request, collection or globally) to disable cookies.

Currently, in Hoppscotch, cookies are sent with the request without being visible. It is impossible to see them, except in the desktop application, where they can be deleted.
Even in the code generator, the cookies are not visible. This can be confusing.
I wasted an hour before realizing that a cookie was taking precedence over a header.

An option to disable cookies can be useful for getting rid of them and avoiding unnecessary request pollution.

For example in Postman :
In request :
Image
In settings :
Image

VigneshrajNadar:
Closes #6526

This PR adds the ability to disable the cookie jar both globally and on a per-request basis. This resolves the problem of cookies polluting requests unexpectedly without visibility, and provides an intuitive UX patterned after Postman's cookie jar settings.

### What's changed

- **Global Setting:** Added a new "Disable cookies" toggle in the main Settings page (`pages/settings.vue`) under a new Network section. When toggled, the cookie jar is disabled for all requests.
- **Per-Request Setting:** Added a new `Settings` tab in the request options area (`RequestOptions.vue`). It includes a "Disable cookie jar" toggle that applies only to the current request.
- **Request Kernel:** Updated `RESTRequest.toRequest` to dynamically calculate whether to send cookies by evaluating an `OR` condition between the global setting and the local request override.
- **Data Migration:** Safely migrated the `HoppRESTRequest` schema to support the new `requestOptions` object without breaking existing request history.

### Notes to reviewers

- The UI for the per-request settings tab intentionally mirrors the language and layout used by Postman to make the migration experience familiar for users. 
- I recommend viewing the visual changes in the screenshots below!

### Screenshots
Screenshot 2026-07-30 at 10 56 48 PM

---
## Summary by cubic
Add global and per-request controls to disable the cookie jar and enforce request-level isolation. Previously all requests sent and mutated cookies; now you can disable them globally or per request. Scripts cannot change this during execution. Note: due to browser limits, same-origin requests may still include cookies.

- Settings and schema: Adds Settings > Network “Disable cookies” (`DISABLE_COOKIES`) and per-request Options > Settings “Disable cookie jar” (`requestOptions.disableCookies`, defaults to false). Migrates REST schema to v18 in `@hoppscotch/data` with persisted/validated `requestOptions` (no manual action).
- Enforcement: Centralizes policy in `cookiePolicy.ts` and applies it in runners and fetch hooks. `RESTRequest.toRequest` sets `meta.options.cookies`; the web relay maps this to `withCredentials`. Runners skip cookie-jar injection when disabled and drop post-script cookie mutations. Cookie APIs throw when disabled and `getUpdatedCookies` returns null.
- Scripts and fetch: Exposes `hopp.request.requestOptions` read-only in scripts; attempting to change `disableCookies` throws. `hopp.fetch` honors the effective policy (global OR request) captured at start of execution; scripts cannot relax or enable cookies mid-run.
- UI/tests: Adds a Request Settings tab and i18n strings; updates unit tests to assert restrictive policy and the same-origin caveat.

<sup>Written for commit 219f1f98cd1f4f629d59f151b8e1e0ce587ff3c1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6554?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/codemirror-lang-graphql/package.json
+++ b/packages/codemirror-lang-graphql/package.json
@@ -11,8 +11,10 @@
   "main": "dist/index.cjs",
   "module": "dist/index.js",
   "exports": {
-    "import": "./dist/index.js",
-    "require": "./dist/index.cjs"
+    ".": {
+      "import": "./dist/index.js",
+      "require": "./dist/index.cjs"
+    }
   },
   "types": "dist/index.d.ts",
   "sideEffects": false,
--- a/packages/hoppscotch-common/locales/en.json
+++ b/packages/hoppscotch-common/locales/en.json
@@ -1241,7 +1241,7 @@
     "generate_name_error": "Failed to generate request name.",
     "disable_cookies": "Disable cookie jar",
     "disable_cookies_default": "(Default: Settings)",
-    "disable_cookies_description": "Prevent cookies used in this request from being stored in the cookie jar. Existing cookies in the cookie jar will not be added as headers for this request."
+    "disable_cookies_description": "Prevent cookies used in this request from being stored in the cookie jar. Existing cookies in the cookie jar will not be added as headers for this request. Note: Due to browser limitations, same-origin requests will still include cookies."
   },
   "response": {
     "audio": "Audio",
@@ -1430,7 +1430,7 @@
     "proxy_auth": "You can also include username and password in the URL.",
     "network": "Network",
     "disable_cookies": "Disable cookies",
-    "disable_cookies_description": "Disable cookie jar for all requests."
+    "disable_cookies_description": "Disable cookie jar for all requests. Note: Due to browser limitations, same-origin requests will still include cookies."
   },
   "shared_requests": {
     "button": "Button",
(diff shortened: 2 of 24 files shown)


jamesgeorge007 (changes requested):
`getSharedCookieMethods` now takes a second `request` argument, but `shared.spec.ts` still calls it with one at around 10 sites, so 9 of the 10 cookie tests throw before they assert. Could you thread `request` through those calls, or make the parameter optional and guard its use?

VigneshrajNadar:
Thanks for pointing that out! I've just pushed a fix to make the request parameter optional in getSharedCookieMethods and safely guarded its usage with optional chaining (currentRequest?.requestOptions?.disableCookies).

This restores the 10 cookie tests in shared.spec.ts without requiring a dummy request to be threaded through every test call. Let me know if everything looks good now!

PR #6560, opened 2026-08-07
fix(common): workspace environment selector not marking new selection as active

### LINKED ISSUE #6540: [bug]: Workspace (team) environment selector does not mark the newly selected environment as active

I searched open and closed issues and did not find one covering this.

### Platform

Desktop App

### Operating System

macOS

### Deployment Type

Hoppscotch Cloud

### Bug Description

Selecting a different environment under the **Workspace Environments** tab of the environment selector does not take effect: the check mark stays on the previously selected environment, and the selection does not register as active. Personal environments are unaffected.

### Steps to reproduce

1. Switch to a team workspace that has multiple team environments.
2. Open the environment selector and go to the **Workspace Environments** tab.
3. Click an environment other than the currently active one.
4. The check mark remains on the previous environment.

### Root cause

In `packages/hoppscotch-common/src/components/environments/Selector.vue`, `isEnvActive` resolves the active team environment through `selectedEnv.value.teamEnvID`. `selectedEnv` is a computed that, for the `TEAM_ENV` branch, looks the selection up via a `.find()` over `teamEnvironmentList`:

```js
const teamEnv = teamEnvironmentList.value.find(
  (env) => env.id === (... && selectedEnvironmentIndex.value.teamEnvID)
)
if (teamEnv) { return { type: "TEAM_ENV", teamEnvID: ..., ... } }
return { type: "NO_ENV_SELECTED" }
```

When that lookup does not resolve — for example while `teamEnvironmentList` is still being populated by its GraphQL query/subscription — the computed falls back to `NO_ENV_SELECTED`, which carries no `teamEnvID`. `isEnvActive` then matches nothing and the check mark does not move.

`handleEnvironmentChange` does write the selection to the store correctly, so `selectedEnvironmentIndex.value.teamEnvID` already holds the right value. The `MY_ENV` branch does not depend on any external lookup, which is why personal environments behave correctly.

### Proposed fix

Compare directly against `selectedEnvironmentIndex.value.teamEnvID`, mirroring how the `MY_ENV` branch resolves its active state. PR: #6535

### Note on verification

The behaviour was observed on the Hoppscotch Cloud desktop app. The root cause above comes from reading the code rather than from stepping through a running build, so an independent confirmation of the timing scenario would be welcome.

CipherCraftXIshant:
Closes #6540

### Problem
In the **Workspace Environments** tab of the environment selector, clicking a different environment doesn't update the checkmark — it stays on the previously selected one, and the selection doesn't visibly register as active. Personal environments work fine; only team/workspace environments are affected.

### Root cause
`isEnvActive()` in `Selector.vue` checks the active state by reading it off the `selectedEnv` computed property. For team environments, that computed re-derives `teamEnvID` by looking the selection up in a separately-fetched `teamEnvironmentList`. If that list is momentarily stale or hasn't finished loading (e.g. right after a workspace switch), the lookup misses and `selectedEnv` silently falls back to `NO_ENV_SELECTED` — so nothing ever shows as active, even though the real selection state (`selectedEnvironmentIndex`) was already updated correctly by the click handler.

### What's changed

- Fixed `isEnvActive()` in `packages/hoppscotch-common/src/components/environments/Selector.vue`, which decides whether an environment shows the active checkmark.

- **Fix:** compare directly against `selectedEnvironmentIndex` for both `MY_ENV` and `TEAM_ENV`, removing the dependency on the separate list lookup. This mirrors the pattern already used (and already working) in `environments/teams/index.vue`.
- [x] Root cause identified and fixed
- [x] Verified `eslint` passes clean on the changed file
- [ ] Not manually verified against a live backend (team workspace + auth required) — flagging for reviewer/CI confirmation

Testing

- eslint passes clean on the changed file.
- I wasn't able to reproduce this against a live backend on my end (team workspaces need the full Postgres + GraphQL backend running). If a maintainer or CI can confirm against a real team workspace, I'm happy to iterate quickly on any feedback.

### Notes to reviewers

This is a minimal, 2-line change — no behavior change for personal (`MY_ENV`) environments, since that branch already read from `selectedEnvironmentIndex` indirectly via `selectedEnv.value.index`, which was always in sync (both come from the same store). The `selectedEnv` computed itself is left untouched and still used elsewhere in the file (e.g. displaying the active environment's name/variables in the header), since fixing that would need a broader refactor beyond this issue's scope.

I wasn't able to reproduce this live against a running backend (team workspaces require the full Postgres + GraphQL backend stack), so I'd appreciate a reviewer or CI confirming the fix against a real team workspace if possible.

---
## Summary by cubic
Fixes the active checkmark in the Workspace Environments tab so a newly selected team/workspace environment is marked immediately. `isEnvActive()` now compares against `selectedEnvironmentIndex` for both `MY_ENV` and `TEAM_ENV`, removing the stale `selectedEnv` lookup.

<sup>Written for commit 23981764298913dc21cb9d0283938859d8544362. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6560?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


CipherCraftXIshant:
When will be the pr be approved and merged?

PR #6563, opened 2026-08-09
fix: avoid shared global regex in environment variable .test() checks

Arunendra21:
## Problem

`HOPP_ENVIRONMENT_REGEX` in `packages/hoppscotch-common/src/helpers/environment-regex.ts` is a module-level regex created with the global (`g`) flag. That is correct for the code that needs every occurrence:

- `environment.inspector.ts` uses `element.match(HOPP_ENVIRONMENT_REGEX)`
- `HoppEnvironment.ts` uses it as the `regexp` for CodeMirror's `MatchDecorator`

But three places use the same shared regex with `.test()`:

- `environment.inspector.ts` -> `isENVInString`
- `curl/curlparser.ts` -> `containsEnvVariables`
- `editor/extensions/HoppEnvironment.ts` -> the tooltip word check

A global regex keeps a `lastIndex` between calls, and `.test()` resumes searching from that position. Since the regex object is shared across all of these modules, the state leaks between them. After one successful match, the next `.test()` can start past the beginning of a shorter string and return `false` even when the string does contain a variable.

## Reproduction

```js
const re = /(<<[a-zA-Z0-9_.-]+>>)/g

re.test("<<token>>") // true
re.test("<<token>>") // false  <- same input, different result

re.test("https://<<host>>/api") // true, lastIndex now past a short string
re.test("<<url>>")              // false, but it clearly has a variable
```

In practice this means, for example, a curl import whose URL uses an environment variable can be treated as having no variables (in `containsEnvVariables`), depending on what was tested just before it.

## Fix

Add a non-global `HOPP_ENVIRONMENT_TEST_REGEX` and use it for the `.test()` containment checks. The global `HOPP_ENVIRONMENT_REGEX` is kept unchanged for `String.match` extraction and the `MatchDecorator`, which need it. Added a spec covering repeated calls, the cross-length state-leak case, and that the global regex still extracts all matches.

No behavior change for the match/highlight paths; only the boolean containment checks become order-independent and correct.

---
## Summary by cubic
Fixes false negatives in environment variable detection by using a non-global regex for `.test()` checks. Detection is now consistent across cURL import, the environment inspector, and editor tooltips; extraction/highlighting remain unchanged.

- **Bug Fixes**
  - Introduced non-global `HOPP_ENVIRONMENT_TEST_REGEX` for containment checks; kept global `HOPP_ENVIRONMENT_REGEX` for extraction and `MatchDecorator`.
  - Replaced `.test()` usages in `curlparser.ts`, `environment.inspector.ts`, and `editor/extensions/HoppEnvironment.ts`.
  - Added tests to ensure stable `.test()` results and that the global regex still extracts all matches.

<sup>Written for commit cd7b89035be2b72dd251313b0fbed2bdc0dedf04. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6563?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/hoppscotch-common/src/helpers/curl/curlparser.ts
+++ b/packages/hoppscotch-common/src/helpers/curl/curlparser.ts
@@ -29,7 +29,7 @@ import {
 } from "./sub_helpers/preproc"
 import { getQueries } from "./sub_helpers/queries"
 import { concatParams, getURLObject } from "./sub_helpers/url"
-import { HOPP_ENVIRONMENT_REGEX } from "../environment-regex"
+import { HOPP_ENVIRONMENT_TEST_REGEX } from "../environment-regex"
 
 const defaultRESTReq = getDefaultRESTRequest()
 
@@ -38,7 +38,8 @@ const defaultRESTReq = getDefaultRESTRequest()
  * @param str The string to test for environment variables.
  * @returns A boolean indicating whether the string contains environment variables.
  */
-const containsEnvVariables = (str: string) => HOPP_ENVIRONMENT_REGEX.test(str)
+const containsEnvVariables = (str: string) =>
+  HOPP_ENVIRONMENT_TEST_REGEX.test(str)
 
 export const parseCurlCommand = (curlCommand: string) => {
   // const isDataBinary = curlCommand.includes(" --data-binary")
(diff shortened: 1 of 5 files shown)


Arunendra21:
Good suggestion — trimmed the comment to a concise two lines while keeping the 'why' (lastIndex leaking across calls/modules). Pushed in cd7b890.

PR #6583, opened 2026-08-16
feat: show search match counts in editor search panel (#6561)

### LINKED ISSUE #6561: [feature]: Show search match counts (e.g., "1 of 10") in Response Body search

### Summary

Currently, when searching (Ctrl+F / Cmd+F) within the response body viewer, the internal editor highlights the matched text, but it does not display the total number of matches found or the current match index. For example, you cannot see if you are on match "1 of 10" or if only a single match exists. This makes it difficult to know how many times a specific key, ID, or value appears in a large JSON/XML response.

Τo find out how many times a term appears in a response, I have to click "Copy" on the response body and paste it into an external text editor  or download the response as a file and open it locally to use a fully-featured search.

Displaying search counts is a standard, expected feature in almost all modern IDEs, text editors, and web browsers. Adding this small UI/UX improvement to Hoppscotch will significantly enhance the Developer Experience and save time when debugging API responses.

VigneshrajNadar:
## Description

Implements the "Search match counts" feature requested in #6561.

### Changes

- Added a custom CodeMirror search panel implementation (`SearchMatchCount.ts`) that extends the default CodeMirror search panel.
- Integrates `getSearchQuery` and `SearchCursor` from `@codemirror/search` to calculate total matches and the currently active match index.
- Plugs the custom panel into the existing `basicSetup` extension in `baseTheme.ts` so that it applies to all editors globally (request body, pre-request scripts, etc.).
- Adds CSS styles to properly format the search panel with the new `X of Y` match counter alongside the search inputs.

### Expected Behavior

- When you press `Ctrl+F` / `Cmd+F` in any Hoppscotch editor and start typing, a live count (e.g., `1 of 5`) will appear next to the search input.
- If there are no matches, it displays `No results`.
- Pressing Enter or clicking Next/Prev will smoothly update the current index as you jump between matches.

Closes #6561

---
## Summary by cubic
Adds a custom CodeMirror search panel that shows live match counts in the editor. Previously the panel showed no counts; now it displays “X of Y” for the active match, “Y results” when none is active, or “No results,” updating on edits, selection moves, and query changes.

- Integrates via `search({ top: true, createPanel: createSearchPanelWithCounts })` in `basicSetup`; new panel lives in `@helpers/editor/extensions/SearchMatchCount`.
- Counts use `getSearchQuery().getCursor` to iterate matches and the main selection head to determine the active index; aria-live updates are polite.
- Updates `baseTheme`/`inputTheme` to stack the search UI vertically and adds styling for `.cm-search-match-count`.
- Replace actions and navigation shortcuts are unchanged; replace controls are omitted in read-only editors.
- Review focus: verify counts with regex/case/word options, next/previous navigation, replace/replace all, and performance in large files.

Addresses #6561.

<sup>Written for commit ec1ddc95021c53cce9e4520b650e26c873a125d2. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6583?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/hoppscotch-common/src/helpers/editor/extensions/SearchMatchCount.ts
+++ b/packages/hoppscotch-common/src/helpers/editor/extensions/SearchMatchCount.ts
@@ -1,0 +1,242 @@
+import { EditorView, Panel } from "@codemirror/view"
+import {
+  getSearchQuery,
+  setSearchQuery,
+  SearchQuery,
+  findNext,
+  findPrevious,
+  replaceNext,
+  replaceAll,
+  closeSearchPanel,
+  selectMatches,
+} from "@codemirror/search"
+
+/**
+ * Computes the total match count and current match index for a search query.
+ *
+ * Uses SearchQuery.getCursor to iterate all matches in the document.
+ * The current match index is determined by finding the match whose range
+ * contains or starts at the editor's main selection head.
+ *
+ * @returns `{ total, current }` where current is 1-indexed (0 if no match is active)
+ */
+function computeMatchCounts(view: EditorView): {
+  total: number
+  current: number
+} {
+  const state = view.state
+  const query = getSearchQuery(state)
+
+  if (!query.valid) {
+    return { total: 0, current: 0 }
+  }
+
+  const cursor = query.getCursor(state)
+  const mainHead = state.selection.main.head
+  let total = 0
+  let current = 0
+
+  let result = cursor.next()
+  while (!result.done) {
+    total++
+    const { from, to } = result.value
+    // The active match is the one whose range contains the cursor head
+    if (mainHead >= from && mainHead <= to) {
+      current = total
+    }
+    result = cursor.next()
+  }
+
+  return { total, current }
+}
+
+/**
+ * Creates a custom search panel that includes a live "X of Y" match count.
+ *
+ * This mirrors the default CodeMirror SearchPanel structure but appends a
+ * `<span class="cm-search-match-count">` element that updates on every
+ * editor state change.
+ */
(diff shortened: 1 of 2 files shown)


3wwwgr:
I tested this PR locally with the self-hosted Docker build.
The match counter works correctly, but I noticed a layout regression in the CodeMirror search panel.

The new .cm-search styles set flexDirection: "column", while SearchMatchCount.ts appends all controls directly to .cm-search. The .cm-search-row and .cm-search-options classes added in this PR are currently not used by the generated DOM, so all search controls end up stacked vertically.

I tested restoring the original horizontal .cm-search layout and keeping only the .cm-search-match-count styling, and the panel works correctly again while still showing the match count.

I can submit the CSS fix as a small PR against your feature branch if you'd like.

VigneshrajNadar:
Thanks for testing this and for catching the layout regression. That makes sense — the `.cm-search-row` and `.cm-search-options` classes aren't being applied to the generated DOM, so changing `.cm-search` to a column layout causes the controls to stack vertically.

Please go ahead with the small CSS fix against the feature branch. Keeping the original horizontal `.cm-search` layout and only adding the match-count styling sounds good to me.

Thanks!

PR #6586, opened 2026-08-16
fix(common): agent Set-Cookie header fallback

CuriousCorrelation:
The agent code path captures no cookies when the Hoppscotch Agent is a stable build. The agent interceptor reads structured cookies from `response.cookies`, which the agent binary populates only when it was built against a relay revision that has `parse_cookies`. An older agent leaves the field undefined, the capture path returns without doing anything, and the `Set-Cookie` headers it did return, joined with newlines in the response headers, are never read. The app and the agent update separately, so this applies to every user until they update the agent.

Closes FE-1286

### What's changed

- `captureResponseCookies` reads the `set-cookie` response header when `response.cookies` is undefined, splits on newlines, and parses each line with the parser the jar already uses.
- Parsed results go through `extractFromResponse`, so canonicalisation, domain defaulting and expiry handling are not duplicated.
- `normalizeSameSite` maps the parser's SameSite union, lowercase strings or a bare boolean, onto the casing `@hoppscotch/data` expects, leaving anything unrecognised undefined so the extractor applies its `Lax` default.
- `headers` added as an optional field on the existing response parameter.
- Four tests covering the fallback and its guard.

### Notes to reviewers

The fallback runs when `cookies` is undefined and not when it is an empty array. An empty array is the relay reporting no cookies in that response, and parsing the headers again would store each cookie twice. The third test covers that case.

`headers` is optional on the existing parameter, so the native and proxy interceptors need no change. Both already pass an object that includes `headers`.

---
## Summary by cubic
Preserves cookie capture with older Hoppscotch Agent builds and blocks cross-domain Domain attributes. Previously, older agents returned no response.cookies so we dropped cookies, and responses could set cookies for unrelated hosts. For IP hosts, only an exact Domain match is allowed. Addresses FE-1286.

- Fallback runs only when response.cookies is undefined; a defined empty array means "no cookies."
- Reads newline-joined Set-Cookie header values, parses with the existing parser (no value decoding), and passes results to extractFromResponse.
- Resolves Max-Age into expires with RFC precedence; Max-Age=0 expires one millisecond before capture so same-tick reads do not see it.
- Normalizes SameSite to `@hoppscotch/data` casing; unrecognized/boolean values become undefined so the extractor applies its Lax default.
- Rejects a Domain that does not domain-match the request host (RFC 6265 5.3); IP literals must match exactly; enforced during canonicalization for both structured and header-parsed cookies.
- Adds optional headers to the captureResponseCookies input; native and proxy interceptors already include headers, so no caller changes.
- Tests cover the header fallback, domain-match rejection and parent-domain acceptance, IP-literal equality and partial-suffix rejection, no re-parse when cookies is [], Max-Age handling (including 0), raw value preservation, header splitting, and the header-absent case.

<sup>Written for commit d2e0c8757f598da62ceee541ac1cfa11f0091b65. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6586?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/packages/hoppscotch-common/src/services/__tests__/cookie-jar.service.spec.ts
+++ b/packages/hoppscotch-common/src/services/__tests__/cookie-jar.service.spec.ts
@@ -410,7 +410,7 @@ describe("CookieJarService", () => {
   })
 
   describe("captureResponseCookies header fallback", () => {
-    it("parses the Set-Cookie header when structured cookies is undefined", async () => {
+    it("parses the Set-Cookie header when structured cookies are undefined", async () => {
       await service.captureResponseCookies(
         { headers: { "set-cookie": "sid=abc123; Path=/; HttpOnly" } },
         "https://example.com/"
@@ -419,6 +419,27 @@ describe("CookieJarService", () => {
       expect(stored).toHaveLength(1)
       expect(stored[0].name).toBe("sid")
       expect(stored[0].value).toBe("abc123")
+    })
+
+    it("resolves Max-Age into an expiry on the header fallback", async () => {
+      await service.captureResponseCookies(
+        { headers: { "set-cookie": "sid=abc; Max-Age=60; Path=/" } },
+        "https://example.com/"
+      )
+      const stored = service.getCookiesForURL(new URL("https://example.com/"))
+      expect(stored).toHaveLength(1)
+      expect(stored[0].expires).toBeDefined()
+      expect(new Date(stored[0].expires!).getTime()).toBeGreaterThan(Date.now())
+    })
+
+    it("keeps a percent-encoded value undecoded on the header fallback", async () => {
+      await service.captureResponseCookies(
+        { headers: { "set-cookie": "sid=abc%20123; Path=/" } },
+        "https://example.com/"
+      )
+      const stored = service.getCookiesForURL(new URL("https://example.com/"))
+      expect(stored).toHaveLength(1)
+      expect(stored[0].value).toBe("abc%20123")
     })
 
     it("splits newline-joined Set-Cookie headers the agent relay concatenates", async () => {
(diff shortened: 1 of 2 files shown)


jamesgeorge007 (dismissed):
Looks good. Sensible fallback for older agent builds that would otherwise drop the cookies. Also, please look into the pending AI review comments.

CuriousCorrelation:
The public-suffix case is the known gap, Domain=github.io passes because canonAttrDomain only rejects single-label domains. That is FE-1285 and FE-1324, which cover the list and its bundle-size cost.

@greptileai review

CuriousCorrelation:
@greptileai, please re-evaluate. A redirect chain resolves to one response here, and the URL this code receives is the original request URL, so the guard compares against the host the user typed rather than the host that answered. This PR narrows that case rather than introducing it. At the merge base 3aa5c8530, extractFromResponse stored an explicit Domain with no host comparison at packages/hoppscotch-common/src/services/cookie-jar.service.ts:499, so a redirect target could set a cookie for any domain it named; with the guard it can only set one that domain-matches the original request host, and an IP-literal host now needs an exact match. The responding host is never passed to this code. ResponseHandler in packages/hoppscotch-desktop/plugin-workspace/relay/src/response.rs:15 has fields for headers, body, status, sizes and timing and none for a URL, the interop Response has no URL field, and packages/hoppscotch-common/src/platform/std/kernel-interceptors/native/index.ts:228 passes effectiveRequest.url because that is the only URL available at the call site. Per-hop validation needs the relay to report the effective URL, which is tracked separately. The public-suffix case is FE-1285 and FE-1324.

jamesgeorge007 (approved):

PR #6648, opened 2026-09-13
fix(i18n): restore placeholders lost in the Afrikaans (af) translation

owgreen-dev:
Seven `af` strings translated or hardcoded the interpolation variable, so vue-i18n can't substitute it and users see the literal token (or a stale value). Caught while running an open-source i18n placeholder checker I maintain (shipi18n) across a few OSS locale trees; each change below was verified by hand against `en.json`.

### What's changed

- [x] `state.connected_to` / `state.connecting_to` — `{naam}` → `{name}` (the variable name was translated, so the connection name never renders)
- [x] `team.invited_to_team` / `team.join_team` / `team.joined_team` — `{team}` → `{workspace}` (matches the variable `en.json` passes; the last two were added in a second commit after re-running the check on the branch)
- [x] `import.file_size_limit_exceeded_warning_single_file` / `…_multiple_files` — hardcoded `10MB` → `{sizeLimit}MB`, so the limit follows the value from `en.json` again

### Notes to reviewers

Only `packages/hoppscotch-common/locales/af.json` is touched (+8/−7); no wording changes beyond the variable tokens. After this branch, `npx @shipi18n/cli check packages/hoppscotch-common/locales -s en --severity 'missing-key=off,untranslated=off,orphan-key=off'` reports 0 placeholder errors for `af`. Happy to split or adjust if you'd prefer a different phrasing for the restored strings.

Full scan for this repo: https://shipi18n.com/oss#hoppscotch


jamesgeorge007 (approved):

nivedin (approved):

owgreen-dev:
Thanks for merging, and for the rebase.

I re-ran the same check against `next` after the merge: `af` is clean now. The other 33 locales are not: **25 of them still drop the same placeholders, 123 strings across 12 keys**. It's the same three patterns the `af` fix covered:

- **`{sizeLimit}` hard-coded as a number — 23 locales.** `de`: `… das empfohlene Limit von 10 MB …`, `ja`: `… limit of 10MB …`. The component passes the real limit, so the message is wrong where the configured limit isn't 10.
- **`{team}` where en now says `{workspace}` — 21 locales × `team.invited_to_team` / `join_team` / `joined_team`.** `de`: `{team} beitreten`, `es`: `Entrar al espacio de trabajo {team}`. vue-i18n has nothing to substitute, so users see the literal `{team}`.
- **translated variable names — a handful.** `ja` `state.connected_to`: `{名前}に接続しました`.

Would be happy to send that as a follow-up PR, in the same shape as this one. It's mechanical, no wording changes past the token itself. One PR for all 25, or split by pattern if that's better to review; your call, and no problem if you'd rather leave it.

Either way, once the tree is clean this keeps it that way: no API key, no account, runs offline:

```yaml
# .github/workflows/i18n-check.yml
name: i18n check
on: [pull_request]
jobs:
  check:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - run: >-
          npx @shipi18n/cli@2 check packages/hoppscotch-common/locales -s en
          --severity 'untranslated=off,missing-key=off,orphan-key=off'
```

The severity flags are what make it useful rather than noisy on your tree, it then reports only placeholder and plural breakage. If you'd rather not fix the 123 first, `--baseline` freezes today's state and fails only on newly-introduced ones.

Full scan for this repo: https://shipi18n.com/oss#hoppscotch

PR #6653, opened 2026-09-17
fix: move pnpm overrides and build allowlist to pnpm-workspace.yaml

### LINKED ISSUE #6636: [bug]: pnpm 10 ignores root dependency overrides and build allowlist
### Platform

Development tooling / package installation

### Browser

Other

### Operating System

Windows

### Bug Description

At the current main commit (ac145e7f758151b41fd46d3e5f513886ce9068ba), the root package.json declares pnpm@10.34.5 but stores the dependency overrides and onlyBuiltDependencies allowlist under package.json's pnpm object. Running pnpm 10.34.5 in a clean checkout prints: The pnpm field in package.json is no longer read by pnpm. The following keys were ignored: pnpm.overrides, pnpm.onlyBuiltDependencies. pnpm config get overrides and pnpm config get onlyBuiltDependencies both return undefined.

Expected behavior: the committed dependency overrides and approved dependency build-script allowlist are active for pnpm 10 installs.

Actual behavior: pnpm does not load either setting. The existing lockfile still contains resolved overrides, but future dependency resolution and install-script policy reads do not receive the declarations from package.json.

The pnpm 10 settings documentation says overrides can only be set at the root of pnpm-workspace.yaml, and documents onlyBuiltDependencies there as well: https://pnpm.io/10.x/settings

A focused fix would move these settings to pnpm-workspace.yaml (or use the current allowBuilds equivalent where appropriate) and verify that pnpm no longer warns and reports the settings as active.

Tyagiquamar:
Fixes #6636

pnpm 10 no longer reads the pnpm field in package.json, so the committed overrides and the onlyBuiltDependencies allowlist were silently inactive (pnpm config get overrides / onlyBuiltDependencies both returned undefined on pnpm 10.34.5).

This moves both settings unchanged to the root of pnpm-workspace.yaml, where pnpm 10 reads them, and removes the dead pnpm object from package.json. Deliberately keeps the onlyBuiltDependencies key name (still valid in 10.x) instead of migrating to allowBuilds, to keep this a pure relocation with no behavior change.

Verified in Docker (node:22-bookworm, pnpm 10.34.5, container-local clones):
- Before (main): pnpm config get overrides -> undefined; onlyBuiltDependencies -> undefined.
- After (this branch): overrides returns all 25 entries; onlyBuiltDependencies returns all 14 entries.
- pnpm install --lockfile-only --ignore-scripts completes with no root-field warning.
- prettier --check passes on package.json; the added YAML lines are prettier-clean (the only findings are two pre-existing single-quoted lines that also fail on main, left untouched).

Note: install still warns about a nested pnpm.overrides in packages/hoppscotch-desktop/plugin-workspace/tauri-plugin-relay/package.json. That is pre-existing and out of scope here: non-root overrides are ignored by design, and hoisting that entry to the workspace root would broaden its resolution scope.

---
## Summary by cubic
Fixes #6636 by moving pnpm `overrides` and the `onlyBuiltDependencies` allowlist from `package.json` to `pnpm-workspace.yaml`. pnpm 10 no longer reads the `pnpm` field in `package.json`, so both settings were silently inactive; pnpm 10 reads them from the workspace file.

No dependency versions or allowlisted packages changed. The `onlyBuiltDependencies` key is kept instead of migrating to `allowBuilds`, since it is still valid in pnpm 10. The workspace file also now uses double quotes consistently for string scalars. A pre-existing warning about a nested `pnpm.overrides` in `packages/hoppscotch-desktop/plugin-workspace/tauri-plugin-relay/package.json` remains out of scope: non-root overrides are ignored by design, and hoisting it would broaden its resolution scope.

<sup>Written for commit 488dcdcf963adbb84550b832f6287a4b32d0084f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/hoppscotch/hoppscotch/pull/6653?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Tyagiquamar:
Hi, just following up on this when you get a chance. The branch is up to date and checks are green. If it looks good from your side, it should be ready to merge. Happy to make any changes if needed. Thanks!

